PR #33, opened 2015-08-12
Update spec to include requirements for authenticated clients

mattheworiordan:
Addresses https://github.com/ably/realtime/issues/131, https://github.com/ably/realtime/issues/245 and https://github.com/ably/realtime/issues/246

@paddybyers can you review, I believe this covers off what we discussed in https://github.com/ably/realtime/issues/131


content/client-lib-development-guide/features.textile
```diff
@@ -74,21 +74,32 @@ h3(#restclient). RestClient
...
 ** @(RSC15d)@ Automatic retries to fallback hosts will not occur for any other HTTP status codes. For example, if the credentials are invalid or the maxmimum message length has been exceeded, then resending the request to an alternative data center will not fix the problem and will only increase the load on other data centers. See the "complete list of Ably error codes":https://github.com/ably/ably-common/blob/master/protocol/errors.json#L11 and a "discussion in the Java library about this":https://github.com/ably/ably-java/issues/14
+* @(RSC17)@ When instancing the library, if a @clientId@ attribute is set in @ClientOptions@, then the @Auth#clientId@ attribute will contain the provided @clientId@
 
 h3(#rest-auth). Auth
 
 * @(RSA1)@ Basic Auth connects over HTTPS by default.  If HTTP and Basic Auth are used, an exception will be raised
-* @(RSA11)@ When using Basic Auth, the API key is sent as the username & password portion of an HTTP URL or in the @Authorization: Basic@ header with a Base64 encoded value
+* @(RSA11)@ When using Basic Auth, the API key is sent as the username & password portion of an HTTPS URL or in the @Authorization: Basic@ header with a Base64 encoded value
```
paddybyers: I don't know what it means to say "sent as the username and password portion of an HTTPS URL".

The API key is sent as a `key` query param of the URL (encoded as a URI component) or is Base64 encoded and included in an Authorization header as specified in RFC7235.

content/client-lib-development-guide/features.textile
```diff
@@ -74,21 +74,32 @@ h3(#restclient). RestClient
...
 * @(RSA2)@ Basic Auth is the default authentication scheme if an API key exists
 * @(RSA3)@ Token Auth:
 ** @(RSA3a)@ Can be used over HTTP or HTTPs
 ** @(RSA3b)@ For REST requests, the token string is Base64 encoded and used in the @Authorization: Bearer@ header
 ** @(RSA3c)@ For Realtime connections, the querystring param @access_token@ is appended to the URL endpoint
 * @(RSA4)@ Token Auth is the default if option @useTokenAuth@ is set to true, a @clientId@ is specified, @authUrl@ or @authCallback@ is configured, or a an explicit @token@ is provided
```
paddybyers: I think the use of the word "default" here is wrong.

It would be more accurate to say:

token auth is used if option @useTokenAuth@ is set to true, or if @useTokenAuth@ is unspecified and one of the following conditions causes token auth to be selected as the default:
-  @clientId@ is specified;
- @authUrl@ or @authCallback@ is configured;
- an explicit @token@ or @TokenDetails@ is provided.

content/client-lib-development-guide/features.textile
```diff
@@ -74,21 +74,32 @@ h3(#restclient). RestClient
...
-* @(RSA14)@ If Token Auth is the default authentication, an exception will be raised if a token is not provided or there is no means to generate a token. For example, if only the option @useTokenAuth@ is specified, and thus a @key@ is not provided, then the client library is unable to authenticate or issue a token
+* @(RSA14)@ If Token Auth is the default authentication scheme, an exception will be raised if a token is not provided or there is no means to generate a token. For example, if only the option @useTokenAuth@ is specified, and a @key@ is not provided, then the client library is unable to authenticate or issue a token
```
paddybyers: Again the use of "default" here is confused.

If Token Auth is selected, either because it was explicitly specified in @useTokenAuth@, or was selected by default (eg by clientId being specified), then an exception will be raised ....

content/client-lib-development-guide/features.textile
```diff
@@ -74,21 +74,32 @@ h3(#restclient). RestClient
...
+* @(RSA15)@ If Token Auth is being used and additionally @clientId@ has been set in the @ClientOptions@ when the library was instanced, the @clientId@ of the token must match the @clientId@ provided in @ClientOptions@. Following an auth request which uses a @TokenDetails@ or @TokenRequest@ object that contains an incompatible @clientId@, the library should change the connection state to @FAILED@ and emit an error
```
paddybyers: The wording here implies that a token always has a clientId, which is not true.

content/client-lib-development-guide/features.textile
```diff
@@ -74,21 +74,32 @@ h3(#restclient). RestClient
...
 * @(RSA5)@ TTL for new tokens is specified in milliseconds and defaults to the REST API default (1 hour)
 * @(RSA6)@ The @capability@, if not specified, defaults to allow all operations permitted for the key associated with the token
-* @(RSA7)@ If a @clientId@ is provided in the options, then the string will be encoded in UTF-8 format before it is sent to Ably
+* @(RSA7)@ @clientId@ and authenticated clients:
+** @(RAS7a)@ If a @clientId@ is present in the authentication token, or in the case of a realtime connection, the endpoint has an explicit @clientId@ param configured, then the client is considered to be authenticated i.e. it has an identity that is implicit in all operations. The following applies to authenticated clients:
```
paddybyers: I think this should be expressed in terms of the client API, not the params determined by the library. Do instead of referring to the endpoint having an explicit clientId, it should talk about a clientId being associated with the library, either by being specified in ClientOptions, or being inferred from a TokenDetails. This way of expressing things should then also apply when we have server-indicated clientId.

content/client-lib-development-guide/features.textile
```diff
@@ -74,21 +74,32 @@ h3(#restclient). RestClient
...
+*** @(RSA7a2)@ The server will reject the publishing of messages with an explicit @clientId@ that does not match the authenticated client's @clientId@. The connection and channel remains available for further operations
+*** @(RSA7a3)@ The client library raises an exception if the @clientId@ attribute of any @Message@ or @PresenceMessage@ published is not @null@ and does not match the non-@null@ @Auth#clientId@. The connection and channel remains available for further operations
+*** @(RSA7a4)@ @Auth#clientId@ attribute returns a string value for the authenticated client's @clientId@. If the client is not authenticated and a @clientId@ value was not provided when instanced with @ClientOptions@, then @Auth#clientId@ is @null@
+** @(RSA7e)@ @Auth#clientId@ is not @null@ when:
+*** @(RSA7e1)@ A @clientId@ is provided in the @ClientOptions@. @clientId@ should be a UTF-8 string
```
paddybyers: Sorry, this is confusing strings and encodings. clientId is a string.

content/client-lib-development-guide/features.textile
```diff
@@ -129,6 +142,16 @@ h3(#rest-channel). Channel
...
+** @(RSL1g)@ Authenticated clients with a @clientId@ (as a result of either an explicitly configured @clientId@ in @ClientOptions@, or implicitly through token auth):
+*** @(RSL1g1)@ When publishing a @Message@ with the @clientId@ attribute set to @null@:
+**** @(RSL1g1a)@ It is unnecessary for the client to set the @clientId@ of the @Message@ before publishing
+**** @(RSL1g1b)@ Ably is expected to automatically assign a @clientId@ upon receiving the @Message@. A test should assert via the history API that the @clientId@ value is populated for the @Message@ when received
+*** @(RSL1g2)@ When publishing a @Message@ with the @clientId@ attribute value set to the authenticated client's @clientId@, Ably will accept the message and publish it. A test should assert that the @clientId@ value is populated for the @Message@ when received
+*** @(RSL1g3)@ When publishing a @Message@ with a different @clientId@ attribute value to the authenticated @clientId@, the client library should reject publishing immediately. The message should not be sent to Ably and an exception should be raised. The connection and channel remains available for further operations
```
paddybyers: I think this should just say that the client library should reject the message. Saying "reject publishing immediately" sounds more like all publishing is then disabled.

content/client-lib-development-guide/features.textile
```diff
@@ -129,6 +142,16 @@ h3(#rest-channel). Channel
...
+*** @(RSL1g4)@ When using token auth with a literal token, an auth URL or an auth callback, until authentication has completed, the client library will not be constrained to a @clientId@. However, as authentication always occurs before the message is published in a synchronous REST library, an exception should be raised if publishing a message with an incompatible @clientId@ to the @clientId@ that is set following authentication. The connection and channel remains available for further operations
```
paddybyers: I think it is better to say "with a token string" instead of "with a literal token".

"However, as authentication always occurs before the message is published in a synchronous REST library": I don't know what this means.

content/client-lib-development-guide/features.textile
```diff
@@ -309,7 +336,16 @@ h3(#realtime-channel). Channel
...
+*** When using token auth with a literal token, an auth URL or an auth callback, until authentication has completed, the client library will not be constrained to a @clientId@. As such, the @Auth#clientId@ attribute will be @null@ until authentication is complete. Therefore, if a @Message@ with a @clientId@ value is published before authentication has completed, the client library should not reject the @clientId@ value and the message with the specified @clientId@ should be delivered to the Ably system. A test should publish a message before authentication with a valid @clientId@ to ensure it is received back with the @clientId@ intact, and then a further test should publish a message with a @clientId@ value that does not match the @clientId@ following auth, which should be rejected by Ably when delivered and the error should contain the server error message. The connection and channel remains available for further operations
```
paddybyers: "until authentication has completed": I don't know what this means.

content/client-lib-development-guide/features.textile
```diff
@@ -504,6 +543,16 @@ h4. Capability - *API not agreed yet*
 * @(TC1)@ This type represents a capability for a key or token
 * @(TC2)@ For now a string representation of the JSON will suffice wherever @capability@ is used
 
+h4. ConnectionDetails
+
+* @(CD1)@ Connection details are optionally passed to the client library in the @CONNECTED@ @ProtocolMessage#connectionDetails@ attribute to inform the client about any constraints it should adhere to, and provide additional meta data about the connection. For example, if a request is made to publish a message that exceeds the @maxMessageSize@, the client library can reject the message immediately before publishing it to Ably
```
paddybyers: meta data -> metadata

content/client-lib-development-guide/features.textile
```diff
@@ -504,6 +543,16 @@ h4. Capability - *API not agreed yet*
...
+* @(CD2)@ Attributes available in @ConnectionDetails@:
+** @(CD2a)@ @clientId@ contains the client ID assigned to the token, when present
+** @(CD2b)@ @connectionKey@ is the connection secret key string that is used to resume a connection and its state
+** @(CD2c)@ @maxMessageSize@ is the maximum individual message size in bytes
+** @(CD2d)@ @maxFrameSize@ is the maximum size for a single frame of data sent to Ably. This restriction applies to a @ProtocolMessage@ sent over a realtime connection, or the total body size for a REST request
+** @(CD2e)@ @maxInboundRate@ is the maximum allowable number of requests per second from a client or Ably. In the case of a realtime connection, this restriction applies to the number of @ProtocolMessage@ objects sent, whereas in the case of REST, it is the total number of REST requests per second
```
paddybyers: This is accurate but now I read this I'm wondering why we don't count the number of messages instead of the number of protocol messages / api requests.

content/client-lib-development-guide/features.textile
mattheworiordan: @paddybyers well I would have thought that is always the case, but it seems in the REST library Faraday I was using, whilst I explicitly configured the Base64 encoded Authorization header, it is in fact removing that header and in all debug messages showing the URL as `username:password@url.com`. Under the hood I expect it must be transforming that back into an Authorization header.  Either way, the change I made was to the protocol from HTTP to HTTPS, but I will correct this anyway.

content/client-lib-development-guide/features.textile
mattheworiordan: Ok, I will amend, again not necessarily related to this PR, but sure, it's clearer.

content/client-lib-development-guide/features.textile
mattheworiordan: I don't think explaining the reasons why it was selected is right by duplicating the spec unnecessarily.  I am happy to go with `If Token Auth is selected` as opposed to `default authentication scheme`.

content/client-lib-development-guide/features.textile
mattheworiordan: I am not sure how it implies that?

content/client-lib-development-guide/features.textile
mattheworiordan: ok

content/client-lib-development-guide/features.textile
mattheworiordan: Yes, we've discussed this before, whilst in Java this may not be an issue, I thought we had agreed that @clientId@, like @name@ or channels and events, is always expected to be UTF-8 string.  Therefore, if someone were to provide a string in another format, the client library should encode as UTF-8.  See https://github.com/ably/ably-ruby/blob/master/spec/unit/auth_spec.rb#L13-L67 for a set of unit tests in the Ruby lib

content/client-lib-development-guide/features.textile
mattheworiordan: Agreed, will fix

content/client-lib-development-guide/features.textile
mattheworiordan: Fine re: token string.

In Realtime, it is plausible to publish a message, and then later the client library will authenticate, at which point the @clientId@ could be inferred.

With the REST library, Auth will always occur before the client library publishes the message, so if a @clientId@ is inferred following Authentication, it can reject the message immediately.

content/client-lib-development-guide/features.textile
mattheworiordan: As above in my previous comments. Authentication may happen after the message is published (well at least queued for publishing)

content/client-lib-development-guide/features.textile
mattheworiordan: Yes, well I was wondering that too.  However, it is linked to my question re: rate limiting and thus publishing a portion of the messages within a `ProtocolMessage` and rejecting others because they exceed the limit.  I assume that would never happen, and as discussed, we reject the entire `ProtocolMessage` let the it succeed.

content/client-lib-development-guide/features.textile
paddybyers: sure

content/client-lib-development-guide/features.textile
paddybyers: > I am not sure how it implies that?

"the @clientId@ of the token ..."

perhaps:

"any clientId associated with the token must match ... "

content/client-lib-development-guide/features.textile
paddybyers: No, strings are strings, ie sequences of characters, no more and no less.

If the Ruby binding places additional constraints on how strings are presented by the caller, then that fact belongs in the Ruby library documentation.

content/client-lib-development-guide/features.textile
paddybyers: > With the REST library, Auth will always occur before the client library publishes the message, so if a @clientId@ is inferred following Authentication, it can reject the message immediately.

Ok, if "it" is Ably, not the library. It's just that with that wording it reads as though the client library could reject the message. I agree that, since the Ably service will perform authentication as part of the query, then the client library will return an error directly from the `publish()` call, instead of at some later time.

content/client-lib-development-guide/features.textile
paddybyers: But I still don't understand the comment. Are you saying that after having performed a single successful rest query the client library will then know the `clientId` ?

content/client-lib-development-guide/features.textile
paddybyers: I will update and clarify (you already have an issue open), then we can come back here and update.

content/client-lib-development-guide/features.textile
mattheworiordan: Ok, it is quite possible I am misunderstanding how languages tend to deal with encodings, but here is how I see it.

If you publish a message event name as a string, it's simply treated as an array of bytes.  So if one client publishes a message the event name with one encoding, and subscribes with another encoding, the bytes must be the same.

```ruby
4] pry(main)> "ハローワー".encode("SHIFT_JIS").unpack("C*")
=> [131, 110, 131, 141, 129, 91, 131, 143, 129, 91]
[5] pry(main)> "ハローワー".encode("UTF-8").unpack("C*")
=> [227, 131, 143, 227, 131, 173, 227, 131, 188, 227, 131, 175, 227, 131, 188]
```

The two encodings yield very different byte arrays, so I would not expect a subscribe with different encodings to be compatible.

content/client-lib-development-guide/features.textile
mattheworiordan: No, I meant the library. If you are using Token Auth because and have provided a `authCallback`, and are using the Realtime library, the Ably client library will let you publish a message with an incompatible `clientId` because at the time of publish there is no way for the client library to know what the clientId should be.  However, in the case of the REST client library, when publishing a message, it needs to call the `authCallback` and then execute the POST HTTP request to Ably.  However, before it even issues the POST request, it will know from the `TokenDetails` object that was returned from the `authCallback` that the `clientId` is incompatible, and can thus raise an exception without having tried to send it.

content/client-lib-development-guide/features.textile
mattheworiordan: As above, with realtime a message will be queued until authentication has happened, whereas with rest it will publish immediately after authentication has happened.

content/client-lib-development-guide/features.textile
mattheworiordan: Great, for reference, issue Realtime#258

content/client-lib-development-guide/features.textile
paddybyers: > Ok, it is quite possible I am misunderstanding how languages tend to deal with encodings, but here is how I see it.

I think so.

Abstractly, an event name is a string. A sequence of characters. That name is a member of a Message, say, which in turn is a member of ProtocolMessage. We pack the ProtocolMessage with msgpack (say) and the whole thing gets turned into bytes, and it's the msgpack spec says what happens when a string member is packed. (It does get turned into utf-8 in msgpack v5 although it is mooted that future versions will admit different encodings.) The client passes a string and how it is encoded at that point ought to be irrelevant.

When a client subscribes, they call the `subscribe()` API and pass a string. It is the job of the library to compare that string with the event name which (after having unpacked it) is also a string. How that string happened to be encoded in transit is irrelevant.

The problem is that a Ruby string isn't a string at all, but a byte array together with some metadata that says that it can be interpreted as a string with a given encoding. Comparing these "strings" in Ruby, by the sound of it, doesn't in fact do a string comparison but a comparison of the byte arrays. So that's an issue that needs to be accommodated in how we specify the client API in Ruby, perhaps, but an encoding is not an attribute of a string (abstractly, or in any language that makes a proper distinction between strings and binary data).

mattheworiordan: Ok, fine, but I fear that not all languages are equal, and being explicit about Strings being an array of Unicode characters is important so as to ensure compatibility between client libraries.  So if the language forces an encoding, we expect client libraries to use UTF-8 encoding.

Just to confirm, in Ruby:

```
[1] pry(main)> "ハローワー".encode("SHIFT_JIS") == "ハローワー".encode("UTF-8")
=> false
[2] pry(main)> "ハローワー".encode("SHIFT_JIS").encode("UTF-8") == "ハローワー".encode("UTF-8")
=> true
```

paddybyers: > being explicit about Strings being an array of Unicode characters is important

Absolutely, but really this goes without saying. For absolute clarity we should state that clientIds, channel names, event names are all Unicode strings and Unicode support is therefore required for those entities in all libraries. That's not in dispute.

> So if the language forces an encoding, we expect client libraries to use UTF-8 encoding.

In a language like Ruby in which string values are internally tagged with an encoding, then I think it's fine for the API to prefer (or insist), for efficiency reasons, that the caller supplies values tagged (ie encoded) with UTF-8. That is really an issue I think for the Ruby client library documentation.

Any sensible language hides from the developer how strings are internally encoded, except sometimes for artifacts of the encoding visible from the handling of surrogate pairs. It is not correct that we ask for a specific default if the language does expose this; for example Java and Javascript will internally use UCS-2 or UTF-16. If the language does offer the programmer a choice of internal representation for strings, then the library IMO should use the language default. Then it is up to the msgpack or other libraries to ensure compatibility on the wire.

mattheworiordan:
@paddybyers I have addressed the issues raised in the PR previously. I would like to get this merged in ASAP as Matus is working now on the PHP lib.

@SimonWoolf FYI

content/client-lib-development-guide/features.textile
```diff
@@ -128,7 +141,19 @@ h3(#rest-channel). Channel
...
+*** @(RSL1g5)@ When using the REST library, publishing a message with an explicit @clientId@ that is incompatible with the @clientId@ inferred following authentication will cause the library to reject the message immediately. Unlike the Realtime library, the message is never queued,and as such, authentication will be completed synchronously before attempting to publish the message over REST
+** Where the library language permits, the @Channel#publish(name, data)@ method should provide an optional argument that allows the @clientId@ value to be specified such as @Channel#publish('event', 'data', { clientId: 'John' })@
```
SimonWoolf: Line duplicates RSL1h below

mattheworiordan: Thanks, it was caused by a merge issue.  The branch when merged did not have the codes in it, so was a nightmare.  Thanks for spotting.

content/client-lib-development-guide/features.textile
```diff
@@ -349,24 +386,25 @@ h3(#realtime-presence). Presence
...
 ** @(RTP8b)@ Optionally a callback can be provided that is called for both success or failure to enter
-** @(RTP8c)@ A @PRESENCE ProtocolMessage@ with a @PresenceMessage@ with the action @ENTER@ is sent to the Ably service
+** @(RTP8c)@ A @PRESENCE ProtocolMessage@ with a @PresenceMessage@ with the action @ENTER@ is sent to the Ably service. The @clientId@ attribute of the @PresenceMessage@ must not be present. Entering without an explicit @PresenceMessage#clientId@, implicitly uses the @clientId@ for the current connection
 ** @(RTP8d)@ Implicitly attaches to the channel if not attached
 ** @(RTP8e)@ Optional data can be included when entering a channel that will be encoded / decoded as with normal messages. A test should exist to ensure data used with enter is encoded & decoded correctly. Also, when data is provided when entering, but no data is provided when leaving, the data attribute should be emitted in the @LEAVE@ event for this client
-** @(RTP8f)@ Raises an exception of the client library has not @clientId@ configured
+** @(RTP8f)@ Raises an exception if the client library does not have a @clientId@ configured and basic auth is being used
```
SimonWoolf: > Raises an exception if the client library does not have a @clientId@ configured and basic auth is being used

Why do we want different behaviour if client tries Presence#enter but fails because no clientId if using basic auth (exception) vs if using token auth without a clientId (error emitted but no exception)?

content/client-lib-development-guide/features.textile
```diff
@@ -508,6 +548,16 @@ h4. Capability - *API not agreed yet*
 * @(TC1)@ This type represents a capability for a key or token
 * @(TC2)@ For now a string representation of the JSON will suffice wherever @capability@ is used
 
+h4. ConnectionDetails
+
+* @(CD1)@ Connection details are optionally passed to the client library in the @CONNECTED@ @ProtocolMessage#connectionDetails@ attribute to inform the client about any constraints it should adhere to, and provide additional metadata about the connection. For example, if a request is made to publish a message that exceeds the @maxMessageSize@, the client library can reject the message immediately before publishing it to Ably
```
SimonWoolf: suggest comma after 'immediately' (currently misparsable as "the client library can reject the message, immediately before publishing it")

content/client-lib-development-guide/features.textile
mattheworiordan: The reason for this is that if you connect to Ably using an API key, and don't specify a `clientId` or any other argument that would make the client library use token authentication, then the client will be connected to Ably without having an implicit clientId.  If you then `enter`, the current client is not able to do that because `enter` requires the currently connected `client` to have an implicit or explicit `clientId`.  So the client library can raise an exception immediately as it will know there is no way this can work.  In the case of `RTP8i`, it is not the same as you can `enter` before the client is authenticated, and as such, we cannot raise an exception immediately.  We can only fail the `enter`, and emit an error explaining why this has happened.  I will try and reword this to make this all a bit clearer.

mattheworiordan: BTW. Following your comment, I was thinking that perhaps a client should raise an exception if using token auth, but no `clientId` is configured.  However, I don't like this because generally in the realtime library we raise an exception if the developer could have predicted the outcome in advance i.e. entering presence when using basic auth is not possible (unless on behalf of a client use `enterClient`), so we can raise an exception immediately.  In the case of using token auth but with no `clientId`, the behaviour will vary based on whether the authentication scheme that was used specifies a `clientId` or not, so I think it's better in these cases to emit an error and fail the operation because the operation was valid, just not at this time.

SimonWoolf: fair enough :+1:

SimonWoolf:
lgtm

paddybyers:
lgtm

PR #43, opened 2015-09-08
Tweak cipherparams description, add getDefaultParams

SimonWoolf:
discussion in https://github.com/ably/ably-js/pull/120


mattheworiordan:
:+1: merge away unless @paddybyers objects

paddybyers:
The default `keyLength` is not 128; each algorithm has its own range of possible values and default. I suggest:

    @(TZ2b)@ @keyLength@ integer - An integer, from an algorithm-dependent range of possible values. If unspecified, an algorithm-dependent default is used. For AES the allowed values at least include 128 and 256, with a default of 128.

paddybyers:
Other than that, lgtm

SimonWoolf:
Merged with Paddy's change

PR #48, opened 2015-09-30
Lower severity for invalid client id in realtime

mattheworiordan:
@paddybyers, please see e9d06bb specifically which removes some duplication of the spec and ensures REST covers only it's requirements, and the same for Realtime.  Also, I noticed that we were referring to exceptions in the Realtime spec or invalid messages, however given we mostly have a callback, I have removed all references to exceptions so that a failure to publish in most cases will simply call the error callback.

FYI, I've updated the [Realtime Base Spec](https://docs.google.com/spreadsheets/d/11fxeoCsaW1j5Q5VWtFNf04v6LXFYoHRhVTRgsEysZkU/edit#gid=0) and [Ruby Realtime Spec](https://docs.google.com/spreadsheets/d/1Zv2NQygt4CzakILWQ42N_Uv-g_GCTgqh_t1DGq82-20/edit#gid=0) to include all the recent changes merged in.  It's a right pain, so expect it will unfortunately cause grief for our client lib devs to meet the spec.

Are you happy with this change?


content/client-lib-development-guide/features.textile
```diff
@@ -93,9 +93,7 @@ h3(#rest-auth). Auth
...
 *** @(RAS7a1)@ All operations (such as message publishing or presence) will have an implicit @clientId@. The Ably service automatically updates the @clientId@ attribute (when empty) for all @Message@ and @PresenceMessage@ messages received from that authenticated client, and any messages then published from the Ably service, will have the @clientId@ attribute populated. It is therefore expected that Ably client libraries do not explicitly set the @clientId@ field on messages published when @clientId@ is implicit in the connection or authentication scheme
-*** @(RSA7a2)@ The server will reject the publishing of messages with an explicit @clientId@ that does not match the authenticated client's @clientId@. The connection and channel remains available for further operations
-*** @(RSA7a3)@ The client library raises an exception if the @clientId@ attribute of any @Message@ or @PresenceMessage@ published is not @null@ and does not match the non-@null@ @Auth#clientId@. The connection and channel remains available for further operations
```
paddybyers: So with this change are you saying it's never a requirement on the library to detect this disparity and indicate an error?

content/client-lib-development-guide/features.textile
```diff
@@ -152,7 +150,7 @@ h3(#rest-channel). Channel
 ** @(RSL1c)@ When an array of @Message@ objects is provided, a single request is made to Ably. When publishing multiple messages, this approach is more efficient. However, a yet to be implemented feature should limit the total number of messages bundled in a single POST based on the default max request size, and would raise an exception if any single message exceeds that limit.
 ** @(RSL1d)@ Indicates an error if the message was not successfully published to Ably
 ** @(RSL1e)@ Allows @name@ and or @data@ to be @null@.  If any of the values are null, then key is not sent to Ably i.e. a payload with a @null@ value for @data@ would be sent as follows @{ "name": "click" }@
-** @(RSL1f)@ Unauthenticated clients without a @clientId@ (any @clientId@ is permitted):
+** @(RSL1f)@ Unauthenticated clients using basic auth without a @clientId@ (i.e. any @clientId@ is permitted):
```
paddybyers: You can have a token that's not bound to a `clientId`.

content/client-lib-development-guide/features.textile
```diff
@@ -160,8 +158,7 @@ h3(#rest-channel). Channel
...
-*** @(RSL1g5)@ When using the REST library, publishing a message with an explicit @clientId@ that is incompatible with the @clientId@ inferred following authentication will cause the library to reject the message immediately. Unlike the Realtime library, the message is never queued,and as such, authentication will be completed synchronously before attempting to publish the message over REST
+*** @(RSL1g4)@ When publishing a message with an explicit @clientId@ that is incompatible with the @clientId@ (either inferred or explicitly configured in the token or @ClientOptions@), the library will reject the message immediately and raise an exception.
```
paddybyers: ... indicate an error.

content/client-lib-development-guide/features.textile
```diff
@@ -350,14 +347,14 @@ h3(#realtime-channel). Channel
 *** @(RTL6c2)@ If the connection is @CONNECTING@ or @DISCONNECTED@, and @ClientOptions#queueMessages@ has not been explicitly set to false, then the message will be queued and delivered as soon as the connection state returns to @CONNECTED@
 *** @(RTL6c3)@ Else an exception is raised preventing the message publishing
 ** @(RTL6d)@ Messages are delivered using a single @ProtocolMessage@ where possible by bundling in all messages for that channel into the @ProtocolMessage#messages@ array. However, a yet to be implemented feature should limit the total number of messages bundled per @ProtocolMessage@ based on the default max message size, and would raise an exception if any single message exceeds that limit
-** @(RTL6e)@ Unauthenticated clients without a @clientId@ (any @clientId@ is permitted):
+** @(RTL6e)@ Unauthenticated clients using basic auth without a @clientId@ (i.e. any @clientId@ is permitted):
```
paddybyers: See above. Token auth, with no bound clientId, also applies.

content/client-lib-development-guide/features.textile
```diff
@@ -350,14 +347,14 @@ h3(#realtime-channel). Channel
...
 **** @(RTL6g1b)@ Ably is expected to automatically assign a @clientId@ upon receiving the @Message@. A test should assert that the @clientId@ value is populated for the @Message@ when received
 *** @(RTL6g2)@ When publishing a @Message@ with the @clientId@ attribute value set to the authenticated client's @clientId@, Ably will accept the message and publish it. A test should assert that the @clientId@ value is populated for the @Message@ when received
-*** @(RTL6g3)@ When publishing a @Message@ with a different @clientId@ attribute value to the authenticated @clientId@, the client library should reject publishing immediately. The message should not be sent to Ably and an exception should be raised. The connection and channel remains available for further operations
+*** @(RTL6g3)@ When publishing a @Message@ with a different @clientId@ attribute value to the authenticated @clientId@, the client library should reject publishing immediately. The message should not be sent to Ably and the errr callback should be called. The connection and channel must remain available for further publishing and other operations
```
paddybyers: Again I think generically "... an error indicated."

content/client-lib-development-guide/features.textile
mattheworiordan: Nope, it's just that it is discovered and described elsewhere in the spec so I don't see the need to duplicate this requirement, see below

content/client-lib-development-guide/features.textile
mattheworiordan: But I thought we had agreed a token without a `clientId` meant that a client could not use any `clientId` i.e. it would always be anonymous / unauthenticated?  Is that not the case, either way we should a test in realtime for this.

content/client-lib-development-guide/features.textile
mattheworiordan: Well with all client libraries so far, an exception is raised in the case of REST.  Do you think that is not always going to be the case?  Perhaps not, your wording may be more flexible moving forwards.

content/client-lib-development-guide/features.textile
mattheworiordan: As above, not sure token auth without a `clientId` allows any `clientId` to be used, at least as far as I was concerned

content/client-lib-development-guide/features.textile
mattheworiordan: Yes, probably better

content/client-lib-development-guide/features.textile
paddybyers: > I thought we had agreed a token without a clientId meant that a client could not use any clientId

I don't think so. As a matter of principle, there shouldn't be anything that you can do with basic auth that you can't do with token auth. We originally had tokens with a `'*'` wildcard `clientId` but we agreed to remove the wildcard, so now a token with no `clientId` string can act on behalf of any clientId, as if using basic auth.

> either way we should a test in realtime for this.

yes

mattheworiordan: > As a matter of principle, there shouldn't be anything that you can do with basic auth that you can't do with token auth.

That I agree with

> We originally had tokens with a '*' wildcard clientId but we agreed to remove the wildcard, so now a token with no clientId string can act on behalf of any clientId, as if using basic auth.

My main concern with this approach then is that if a user (developer) does not specify a `clientId` in their token params, then they could very easily mistakenly assume the user is anonymous. From an API usability perspective, I am a little concerned by this actually.  Can you think of any way to avoid this problem easily within the API, or will we have to rely on documentation to get around this?

paddybyers: > Can you think of any way to avoid this problem easily within the API, or will we have to rely on documentation to get around this?

We could revert to expllcitly requiring a wildcard? And an empty `clientId` means anonymous?

mattheworiordan: > We could revert to explicitly requiring a wildcard? And an empty clientId means anonymous?

Well there is one other option, which is to not allow a `clientId` if `clientId` is null / undefined in the TokenParams, and if `clientId` is an empty string, then any `clientId` is allowed.  The reasoning behind that is because an empty string `clientId` is not allowed so there is no possibility of a `clientId` with an empty string ever colliding.  However, I don't think that is particularly user friendly either.  I would perhaps go with no `clientId` means the user is not authenticated, and `*` is any `clientId` is allowed.  It's the most intuitive, and we're only then making one `clientId` impossible to specify in TokenParams, that being `*`.  Hardly a big problem I would say.

mattheworiordan: Also, if we don't do this, then there is no way to issue a token for a user and not assign them a `clientId`, which is odd

mattheworiordan:
@paddybyers and @SimonWoolf, I have now updated this PR to primarily address the support for wildcard tokens that we need, and also address some comments. Do give me your thoughts. I am tempted now to add some more realtime tests to cover off this expected functionality.

mattheworiordan:
@paddybyers @SimonWoolf are you happy to merge this in?  If so, I will write some specs for Realtime as well so that we can support this functionality...

mattheworiordan:
@paddybyers @SimonWoolf are you happy to merge this in?

paddybyers:
> are you happy to merge this in?

Fine by me

Issue #50, opened 2015-10-01
Document ability to change the defaults and lower the defaults


mattheworiordan:
* Retry immediately if a connection disconnects from a connected state
* Retry every 15s whilst in the disconnected state
* Moves to suspended state within 2 minutes
* Retry every 30s whilst in the suspended state

* Review the fallback strategy when connection fails because of network or server issues

* Allow defaults to be reconfigured before instancing the library, with a `defaults` option.

* Add REST request timeout

mattheworiordan:
Add mention of TTL defaults for tokens, https://github.com/ably/ably-ruby/blob/master/lib/ably/auth.rb#L34-L38

Issue #38, opened 2015-08-22
Add a spec for OS network connection state detection


mattheworiordan:
See https://github.com/ably/ably-js/issues/99 for more information.

If we know the underlying internet connection has gone away or come back, we can act on that immediately.

FYI @paddybyers

Issue #37, opened 2015-08-21
Ensure spec loses recover / resume state if CLOSED explicitly or FAILED


mattheworiordan:
Ensure closing & reopening starts a new connection

PR #51, opened 2015-10-06
Configurable timing options

mattheworiordan:
@paddybyers this addresses the recent discussion we have had on configurable options for the client library, as well as fixing a few other small issues.

Fixes:
* https://github.com/ably/docs/issues/50
* https://github.com/ably/wiki/issues/13
* https://github.com/ably/docs/issues/38
* https://github.com/ably/docs/issues/37

BTW. Please feel free to suggest better naming for the options...

cc @SimonWoolf


content/client-lib-development-guide/features.textile
```diff
@@ -606,6 +612,13 @@ h4. ClientOptions
 *** @(TO3k3)@ @realtimeHost@ string - for development environments only; allows a non-default Ably Realtime host to be specified
 *** @(TO3k4)@ @port@ integer - for development environments only; allows a non-default Ably non-TLS port to be specified
 *** @(TO3k5)@ @tlsPort@ integer - for development environments only; allows a non-default Ably TLS port to be specified
+** @(TO3l)@ The follow attributes, if set, are used to change the default behaviour of the libary:
+*** @(TO3l1)@ @disconnectedRetryFrequency@ integer - default 15,000 (15s). When the connection is in the @DISCONNECTED@ state, this delay in milliseconds controls how frequently the client library attempts to reconnect automatically
+*** @(TO3l2)@ @suspendedRetryFrequency@ integer - default 30,000 (30s). When the connection is in the @SUSPENDED@ state, this delay in milliseconds controls how frequently the client library attempts to reconnect automatically
+*** @(TO3l3)@ @httpOpenTimeout@ integer - default 4,000 (4s). Timeout for opening the connection, available in the client library if supported by the transport
```
SimonWoolf: @mattheworiordan Not sure I understand why there's an `httpOpenTimeout` separate from `httpRequestTimeout`. Why do we require the connect request to be particularly faster than any other req?

content/client-lib-development-guide/features.textile
```diff
@@ -606,6 +612,13 @@ h4. ClientOptions
...
+*** @(TO3l3)@ @httpRequestTimeout@ integer - default 15,000 (15s). Timeout for any single HTTP request and response
+*** @(TO3l4)@ @httpMaxRetryCount@ integer - default 3. Max number of fallback host retries for HTTP requests that fail due to network issues or server problems
+*** @(TO3l4)@ @httpMaxRetryDuration@ integer - default 10,000 (10s). Max elapsed time in which fallback host retries for HTTP requests will be attempted i.e. if the first default host attempt takes 5s, and then the subsequent fallback retry attempt takes 7s, no further fallback host attempts will be made as the total elapsed time of 12s exceeds the default 10s limit
```
SimonWoolf: TO3l3, TO3l3, TO3l4, TO3l4

mattheworiordan: He he, thanks for spotting

content/client-lib-development-guide/features.textile
mattheworiordan: I deliberated over this one, and decided to leave it in only because most HTTP client libraries that I took a quick look at supported separate timeout values for connect & response. The reason I think it's useful is that if an HTTP request is not opened within 4s, then it's quite possible there is a problem connecting as the HTTP server is probably not "alive" and using a fallback host is therefore sensible.  Having only one timeout means we'd need to configure the two values to be the same, so an unresponsive HTTP server could hold up a single REST HTTP request for 15s and therefore prevent any fallbacks being used at all because the overall `httpMaxRetryDuration` would have been exceeded.  Definitely one for discussion though.

content/client-lib-development-guide/features.textile
```diff
@@ -606,6 +612,13 @@ h4. ClientOptions
...
+*** @(TO3l2)@ @suspendedRetryFrequency@ integer - default 30,000 (30s). When the connection is in the @SUSPENDED@ state, this delay in milliseconds controls how frequently the client library attempts to reconnect automatically
+*** @(TO3l3)@ @httpOpenTimeout@ integer - default 4,000 (4s). Timeout for opening the connection, available in the client library if supported by the transport
+*** @(TO3l4)@ @httpRequestTimeout@ integer - default 15,000 (15s). Timeout for any single HTTP request and response
+*** @(TO3l5)@ @httpMaxRetryCount@ integer - default 3. Max number of fallback host retries for HTTP requests that fail due to network issues or server problems
+*** @(TO3l6)@ @httpMaxRetryDuration@ integer - default 10,000 (10s). Max elapsed time in which fallback host retries for HTTP requests will be attempted i.e. if the first default host attempt takes 5s, and then the subsequent fallback retry attempt takes 7s, no further fallback host attempts will be made as the total elapsed time of 12s exceeds the default 10s limit
```
SimonWoolf: Is this needed given we have `realtimeOpenTimeout`?

content/client-lib-development-guide/features.textile
```diff
@@ -606,6 +612,13 @@ h4. ClientOptions
...
 *** @(TO3k5)@ @tlsPort@ integer - for development environments only; allows a non-default Ably TLS port to be specified
+** @(TO3l)@ The follow attributes, if set, are used to change the default behaviour of the libary:
+*** @(TO3l1)@ @disconnectedRetryFrequency@ integer - default 15,000 (15s). When the connection is in the @DISCONNECTED@ state, this delay in milliseconds controls how frequently the client library attempts to reconnect automatically
+*** @(TO3l2)@ @suspendedRetryFrequency@ integer - default 30,000 (30s). When the connection is in the @SUSPENDED@ state, this delay in milliseconds controls how frequently the client library attempts to reconnect automatically
+*** @(TO3l3)@ @httpOpenTimeout@ integer - default 4,000 (4s). Timeout for opening the connection, available in the client library if supported by the transport
+*** @(TO3l4)@ @httpRequestTimeout@ integer - default 15,000 (15s). Timeout for any single HTTP request and response
+*** @(TO3l5)@ @httpMaxRetryCount@ integer - default 3. Max number of fallback host retries for HTTP requests that fail due to network issues or server problems
```
SimonWoolf: Does this just mean 'truncate the fallbackHosts list to the first httpMaxRetryCount`?

content/client-lib-development-guide/features.textile
mattheworiordan: Are they really related?  This is an HTTP request limit, whereas the other is an acceptable time for a connection to to be `CONNECTED`.

content/client-lib-development-guide/features.textile
mattheworiordan: Sort of, if you assume that all fallback hosts are used, and that the max retries is the total number of fallback hosts.

content/client-lib-development-guide/features.textile
SimonWoolf: Not sure if I'm misunderstading. Under what circumstances could `httpMaxRetryDuration` elapse other than just before the connection attempts would be cut off anyway by `realtimeOpenTimeout` elapsing?

content/client-lib-development-guide/features.textile
```diff
@@ -606,6 +612,13 @@ h4. ClientOptions
 *** @(TO3k3)@ @realtimeHost@ string - for development environments only; allows a non-default Ably Realtime host to be specified
 *** @(TO3k4)@ @port@ integer - for development environments only; allows a non-default Ably non-TLS port to be specified
 *** @(TO3k5)@ @tlsPort@ integer - for development environments only; allows a non-default Ably TLS port to be specified
+** @(TO3l)@ The follow attributes, if set, are used to change the default behaviour of the libary:
+*** @(TO3l1)@ @disconnectedRetryFrequency@ integer - default 15,000 (15s). When the connection is in the @DISCONNECTED@ state, this delay in milliseconds controls how frequently the client library attempts to reconnect automatically
```
paddybyers: This isn't a frequency but a delay.

content/client-lib-development-guide/features.textile
```diff
@@ -606,6 +612,13 @@ h4. ClientOptions
...
+*** @(TO3l2)@ @suspendedRetryFrequency@ integer - default 30,000 (30s). When the connection is in the @SUSPENDED@ state, this delay in milliseconds controls how frequently the client library attempts to reconnect automatically
```
paddybyers: Ditto, it' a delay not a frequency.

content/client-lib-development-guide/features.textile
paddybyers: > most HTTP client libraries that I took a quick look at supported separate timeout values for connect & response.

That's reasonable for a generic library because you might have a request that you expect to be long-running (eg a comet recv) but you'd still want to know very quickly if it had failed to send the request. In the defaults we had in the ably-js library there was a single composite timeout (the connection + the response) for "normal" requests, and a different timeout for "long" requests. For the latter, I'd expect the library could still raise an error if the connection and the sending of the request hadn't occurred within the "normal" time.

content/client-lib-development-guide/features.textile
```diff
@@ -642,3 +655,13 @@ h4. CipherParams
...
 ** @(TZ3b)@ takes an optional key parameter, returning a @CipherParams@ instance created with that key and with a keyLength calculated from that key
+
+h3(#defaults). Client Library defaults
+
+The following default values are configured for the client library:
+
+* @(DF1)@ Realtime defaults:
+** @(DF1a)@ @connectionStateTtl@ integer - default 60s. The duration that Ably will persist the connection state when a Realtime client is abruptly disconnected. When the client is in the @DISCONNECTED@ state, once this TTL has passed, the client should change the state to the @SUSPENDED@ state signifying that the state is now lost i.e. channels need to be reattached manually.
+Note that this default is override by any @connectionStateTtl@ specified in the @ConnectionDetails@ of the @CONNECTED@ @ProtocolMessage@
+** @(DF1b)@ @realtimeCloseTimeout@ - default 10s. Following an explicit request to close a connection from the client library by sending a @CLOSE@ @ProtocolMessage@, this value is the amount of time that the client library will wait before forcibly closing the connection and moving the client's connection state to @CLOSED@
```
paddybyers: I don't really see the justification for making this a distinct timeout. Most operations are "short" operations and we expect them to complete more or less immediately, and I think it would be quite legitimate for these all to be covered by the same default.

content/client-lib-development-guide/features.textile
mattheworiordan: True

content/client-lib-development-guide/features.textile
mattheworiordan: 1. First attempt to `/time` sends a request `rest.ably.io`, and in 4s, because the HTTP request is not even open, the client library aborts the request
2. Time elapsed is checked against `httpMaxRetryDuration`, 4s has passed, therefore another attempt to a fallback host is made.  Second attempt to `a.ably-realtime.com/time` is made, 4 more seconds pass because it ends up, by chance hitting the same server as 1.
3. Time elapsed is checked against `httpMaxRetryDuration`, 8s has passed, therefore another attempt to a fallback host is made.  Second attempt to `b.ably-realtime.com/time` is made, request fails completely after 7s.
4. Time elapsed is checked against `httpMaxRetryDuration`, 15s has passed, so no further attempt is made.

content/client-lib-development-guide/features.textile
mattheworiordan: So you want a generic timeout that applies to any ProtocolMessage sent that expects a response, and thus assumes the request has failed?  I am not sure I would necessarily agree, assuming I sent a message, then the connection dropped, and was resumed 20s later, I would not expect an automatic timeout on the Message publish as it may have been sent before the connection disconnected briefly.

content/client-lib-development-guide/features.textile
SimonWoolf: Possible implementation for ably-js [here](https://github.com/ably/ably-js/commit/d8d9e524e2e43fabb3d6d712f73dd65fcf75c3b6). I restricted it to /connect requests, as they're long (recv) requests but which will return a CONNECTED response immediately (and hold the stream open), so I put in this as an additional timeout for the first data returned.

content/client-lib-development-guide/features.textile
SimonWoolf: But a REST /time request isn't a /connect request, it's a normal REQ_SEND, so would use the (15s) `httpRequestTimeout`...?

mattheworiordan: I think we're confusing terminology.  The open timeout has nothing to do with the `/connect` endpoint, it applies to ANY HTTP request, and from the Ruby docs states `Number of seconds to wait for the connection to open. Any number may be used, including Floats for fractional seconds. If the HTTP object cannot open a connection in this many seconds, it raises a Net::OpenTimeout exception. The default value is nil.`.  So the point is that if the underlying TCP connection is not opened within this time, then the request will fail.  That is totally different to the request timeout which is the total time allowed for the request and response.

content/client-lib-development-guide/features.textile
paddybyers: If a message is published on one transport, the ACK can only ever come on the same transport.

Messages whose ACKs are outstanding when a transport closes must be re-sent.

So I don't that's a counterargument to a single timeout for short operations.

paddybyers: To clarify, however, I'm not necessarily arguing that we add any timers that don't already exist; I'm just saying that operations that expect an "immediate" response can have a common timeout.

mattheworiordan: @paddybyers I don't think that requirements exists in the spec.  I recall discussing something along those lines before, but I don't think we ever agreed on anything or documented it did we?  Should we document it now, or leave that for 0.9 release? 

I agree we could have a generic timeout for any type of ack for a `ProtocolMessage`, however we then probably need a different transport connect timeout which is different because the total time to connect and receive the first `CONNECTED` message is probably greater than the time expected for a `MESSAGE`, `PRESENCE`, `ATTACH`, `DETACH`, `CLOSE` message.  WDYT?

paddybyers: > Should we document it now, or leave that for 0.9 release?

No new timers in this spec version.

> we then probably need a different transport connect timeout which is different because the total time to connect and receive the first CONNECTED message is probably greater than the time expected for a MESSAGE, PRESENCE, ATTACH, DETACH, CLOSE message. WDYT?

I don't think it needs to be different. There's the round trip in each case, which is the same, and the realtime system response in each case should be "instantaneous" except for factors such as:

- establishing account and app master, which requires (potentially) cross-region round trip(s) and may also require a wait for the cluster to become stable if it is momentarily unstable. This applies to `CONNECT` and to all REST operations;
- establishing channel master (which suffers from the same issues and applies to `ATTACH` but not `CONNECT`);

- general delays due to processing loading (which affects both, but arguably channel operations more than `CONNECT` since it is affected by core load as well as frontend load.

mattheworiordan: >> Should we document it now, or leave that for 0.9 release?

> No new timers in this spec version.

I was actually referring to your point about `If a message is published on one transport, the ACK can only ever come on the same transport. Messages whose ACKs are outstanding when a transport closes must be re-sent.`.  That is quite key, yet not a requirement of the client library.  Should I add that as a feature for a 0.9 release?

In terms of timers, I agree with you then, a generic timer is correct.  Any thoughts on the timeout we should use?  10s?

paddybyers: > That is quite key, yet not a requirement of the client library. Should I add that as a feature for a 0.9 release?

If there is a message that has been sent, but for which there has been no ACK at the time the transport closes, then the message either needs to be failed (ie error callback called) or re-sent on resume. I think the library is broken if it does neither.

So does 0.8 say nothing about which one of these options is implemented and leave it up to the implementation?

content/client-lib-development-guide/features.textile
SimonWoolf: Right, sorry, I'd misunderstood.

I'm unsure how to implement that in ably-js though. I might be missing something, but I can't see a way for a browser to access to the underlying TCP connection of an xhr request, only the http readyState. That doesn't have anything in between `1` (`OPENED`, ie you've called `open`, but doesn't imply anything's succeeded, you get `1` even if you call `open` with a nonexistant domain) and `2` (`HEADERS_RECEIVED`, ie response's http status is available -- by which time it'll have started sending a response, which is what cancels the `httpRequestTimeout`, presumably it isn't a common failure mode to send headers and then no response). https://xhr.spec.whatwg.org/#dom-xmlhttprequest-readystate

paddybyers: Any such requirement would only apply to transports that allow you to implement it. In JSONP you can do even less than XHR.

However ... I think the only place where it could potentially be useful to get an indication of a failure to connect in a shorter time than the expected time for the request itself to complete is a non-streaming comet `recv`. This of course uses just those transports that can't give you the feedback you'd want.

content/client-lib-development-guide/features.textile
mattheworiordan: What we have is:

* (RTN7c) If a connection enters the CLOSED or FAILED state, or if the connection state is lost, and a ACK or NACK has not yet been received for a message, the client should consider the delivery of those messages as failed
* (RTN15a) If a Connection transport is disconnected unexpectedly or if a token expires, then the Connection manager will immediately attempt to reconnect and restore the connection state. Connection state recovery is provided by the Ably service and ensures that whilst the client is disconnected, all events are queued and channel state is retained on the Ably servers. When a new connection is made with the correct connection recovery key, the client is able to catch up by receiving the queued ProtocolMessages from Ably. Connection state is only maintained for a brief period, up to a minute, so if a client is disconnected for a longer period connection state cannot be resumed

As such, I have now added:

* https://github.com/ably/docs/pull/52
* 2dc7923
* 5c303e8

Can you comment on these changes please @paddybyers?

content/client-lib-development-guide/features.textile
mattheworiordan: As mentioned in the spec `Timeout for opening the connection, available in the client library if supported by the transport`, it only applies if the transport supports it

mattheworiordan:
@paddybyers you happy for me to merge now?

paddybyers:
> you happy for me to merge now?

Yes, looks good.

PR #52, opened 2015-10-09
Spec to address transport disconnects and ACK messages

mattheworiordan:
@paddybyers following our discussion, do you feel this covers the requirement of the client library to resend pending messages when the transport is changed?

@SimonWoolf FYI


mattheworiordan:
@paddybyers I need to merge before I go, can you review and comment please?

content/client-lib-development-guide/features.textile
```diff
@@ -289,7 +289,7 @@ h3(#realtime-connection). Connection
...
 *** @(RTN15b1)@ @resume@ is the private connection key assigned to the connection when the first @CONNECTED@ @ProtocolMessage@ was received
 *** @(RTN15b2)@ @connection_serial@ is the most recent @ProtocolMessage#connectionSerial@ received from Ably or @Connection#serial@ which should be identical
-** @(RTN15c)@ When a new connection is made using the @resume@ and @connection_serial@ params, the client library must consider the resume to have failed if the @CONNECTED@ @ProtocolMessage@ contains an error or if the @Connection#id@ has changed. When a @Connection@ resume has failed, all channels should be detach with a suitable error reason
+** @(RTN15c)@ When a new connection is made using the @resume@ and @connection_serial@ params, the client library must consider the resume to have failed if the @CONNECTED@ @ProtocolMessage@ contains an error or if the @Connection#id@ has changed. When a @Connection@ resume has failed, all channels should be detached with a suitable error reason, triggering any messages awaiting an ACK to fail
```
paddybyers: I think this is confused.

`ACK` and `NACK` responses for published messages can only ever be received on the transport connection on which those messages were sent. Therefore, once a transport drops, the client library must either fail the publish attempt, or re-attempt by re-sending the messages on a new transport if the resume was successful (ie the `CONNECTED` response includes the expected `connectionId`).

The system's response to a resume request can be a `CONNECTED` message with an `error`, if the error does not prevent the establishement of a connection. The outcome may be a resume of the requested connection (which implies that the connection state is intact, and all attached channels remain attached) but there may have been a non-fatal error (such as that the system is unable to send the backlog in full because the limit of backlog messages has been exceeded), or the outcome may be that a new connection is established (in which case the library must assume that all connection state is lost).

paddybyers:
One comment but otherwise good.

mattheworiordan:
@paddybyers would you mind reviewing these final changes before I merge and update all the specs?

mattheworiordan:
BTW. I will tag this as `v0.8` so that we version the docs as well now

content/client-lib-development-guide/features.textile
```diff
@@ -299,7 +299,9 @@ h3(#realtime-connection). Connection
...
 *** @(RTN15b2)@ @connection_serial@ is the most recent @ProtocolMessage#connectionSerial@ received from Ably or @Connection#serial@ which should be identical
-** @(RTN15c)@ When a new connection is made using the @resume@ and @connection_serial@ params, the client library must consider the resume to have failed if the @CONNECTED@ @ProtocolMessage@ contains an error or if the @Connection#id@ has changed. When a @Connection@ resume has failed, all channels should be detach with a suitable error reason
+** @(RTN15c)@ The system's response to a resume request can be a @CONNECTED@ message with an error. If the @CONNECTED@ message @connectionId@ matches the prior @connectionId@, then Ably has confirmed to the client that the resume has been successful, the connection state is intact, and all attached channels remain attached. However, if the @connectionId@ has changed and optionally an error is present, then there may have been a non-fatal error (such as the system is unable to send the backlog in full because the limit of backlog messages has been exceeded), at which point the client library must assume that all connection state is lost and a new connection has been initialised
```
paddybyers: Sorry, maybe I wasn't clear enough.

- `CONNECTED` + same `connectionId` + no `error`: resume succeeded, all channels attached, all backlog messages available

- `CONNECTED` + same `connectionId` + `error`: resume succeeded but with a non-fatal error, all channels attached, error indicates a resume problem, all backlog messages may not be available

 -`CONNECTED` + new `connectionId` + `error`: new connection established but unable to resume, channels no longer attached, error indicates the resume problem

- `ERROR`: fatal error in connection

mattheworiordan: > `CONNECTED` + same `connectionId` + `error`

You are right, I I really don't like this and assumed you were just not clear in your last comment. We have been talking about guarantees, and this is not a guarantee at all. It means you may have lost some messages. Why are we not sending a new `connectionId` in this instance instead so that the developer can at least decide how they wish to handle the loss of messages?

paddybyers: We're recovering to the maximum extent possible, and the error indicates exactly what happened. For example, there is a max backlog of 10,000 messages, and we recovered that many, but the backlog was larger than that. It's not "maybe we did, maybe we didn't". Why is it better then to start a new connection and recover nothing at all?

mattheworiordan: @paddybyers predictability. We have often discussed that PubNub's claim that messages are in order (mostly) and are resumed (mostly) is a real fault of their product.  This now feels like we are more like them and don't provide guarantees.  I would rather a user knows messages are missed and does a history query knowing the limitations of that, than resumes and has some messages that are lost. In my mind, resume is binary, it either works or it doesn't.

paddybyers: We are deterministic and predictable within the capacity limits of the infrastructure. If the thing that the developer asked to do exceeds the limits of the infrastructure, then we indicate that. We will send the most recent 10,000 (or whatever) messages in the backlog; not a random sample. That IMO is a much more useful default behaviour than killing the connection and expecting the developer to pick up the pieces; what is he supposed to do? Query each attached channel separately(does he know which serial to query from?) and reassemble those messages in connection order; and that will possibly also fail because the limitations of his connection prevent him from retrieving a huge list in a useful timescale.

mattheworiordan: > We are deterministic and predictable within the capacity limits of the infrastructure. If the thing that the developer asked to do exceeds the limits of the infrastructure, then we indicate that.

Well do we, because the spec does not even specify which error code represents which problem, and thus what action the developer will take.

> We will send the most recent 10,000 (or whatever) messages in the backlog; not a random sample. 

And why are the 10k most recent more valuable than the previous ones.  Can we really assume that is what the developer wants?

>  That IMO is a much more useful default behaviour than killing the connection and expecting the developer to pick up the pieces

I would agree that it's more useful in a lot of cases, but not in all cases. Also, if the buffer limit is hit, we can arguably save considerable resources by simply discarding that connection state.  I would prefer to tell users that we retain connection state for 120s and for up to 10k messages, and if either of those thresholds are exceeded they will be responsible for resuming the connection state or using history if the messages are important to them.  We have also recently discussed a system that will guarantee connection state recovery on channels that are persisted.  I feel that is a much better solution.

> Query each attached channel separately(does he know which serial to query from?) and reassemble those messages in connection order; 

As above, this is a planned feature for persisted channels.

> and that will possibly also fail because the limitations of his connection prevent him from retrieving a huge list in a useful timescale.

I am not sure I understand why that is.  If the message history is so important a developer will either need a guarantee that none are lost and he/she can retrieve them, or not care that much if some are lost and therefore losing connection state is not a train smash and they can continue.

I really dislike the idea that our behaviour is unpredictable. Perhaps we could offer an additional option that states if you hit the message limit for a connection we will discard the old messages, but I don't think that should be the default behaviour.

mattheworiordan:
@paddybyers spec has been updated.  Can you review now?

content/client-lib-development-guide/features.textile
```diff
@@ -299,7 +299,13 @@ h3(#realtime-connection). Connection
...
+** @(RTN15c)@ The system's response to a resume request will be one of the following:
+**** @(RTN15c1)@ @CONNECTED@ @ProtocolMessaeg@ with the same @connectionId@ as the current client, and no error. In this case, the server is indicating that the resume succeeded, all channels are still attached, and all backlog messages are available
+**** @(RTN15c2)@ @CONNECTED@ @ProtocolMessaeg@ with the same @connectionId@ as the current client, and an error in @errorInfo@. In this case, the server is indicating that the resume succeeded but with a non-fatal error, all channels are still attached, and some backlog messages may be unavailable. The @ErrorInfo@ received should be emitted on the client @Connection@ as well as all @ATTACHED@ @Channel@ objects. The exact use case for this partial success state is still being reviewed, however this does not effect the client library behaviour
```
paddybyers: - `@ProtocolMessaeg@` typo (and paras above and below)
- `@errorInfo@` the name of the `ProtocolMessage` field is `error`, with type `ErrorInfo`
- `The exact use case for this partial success state is still being reviewed, however this does not effect the client library behaviour`: irrelevant to the spec IMO

paddybyers:
LGTM other than comments above.

content/client-lib-development-guide/features.textile
mattheworiordan: Thanks re: typo.

Re: `errorInfo`, that is worrying.  Looking at the spec TR4h at http://docs.ably.io/client-lib-development-guide/features/, that is certainly not the case.  I will update there as well.

PR #54, opened 2015-11-08
Move "Jump to" nav into right margin

lmars:
Before:

screen shot 2015-11-08 at 22 53 32

After:

screen shot 2015-11-08 at 22 56 47


mattheworiordan:
Great, thanks!

mattheworiordan:
BTW. I will deploy this, but in future you simple `rake deploy`

PR #55, opened 2015-11-09
Jump to nav is now left aligned so will not ever overlap content

mattheworiordan:
@lmars I deployed your changes but then saw that the Jump to nav still overlaps on other pages because it is right aligned to a fixed position.

### Before

voila_capture 2015-11-09_12-13-19_a m

### After

voila_capture 2015-11-09_12-13-38_a m


lmars:
LGTM

Issue #57, opened 2015-11-10
Spec for queued messages following a reconnect


mattheworiordan:
Scenario as follows:

1. Client connects
2. Client attaches to channel `bacon`
3. Client publishes 5 messages to `bacon`
4. Transport disconnects, and client automatically tries to reconnect
5. Whilst reconnecting the developer publishes 5 more messages to the `bacon` channel, and another client publishes 3 messages to the `bacon` channel.
6. Client reconnects, but is unable to resume, so detaches all channels

Question:

* Should the queued messages for the channel that is not attached fail those messages as the channel is now failed?
* Could we publish the messages anyway when we introduce the functionality to publish without being attached?

@paddybyers WDYT?

paddybyers:
> Client reconnects, but is unable to resume, so detaches all channels

Why unable to resume?

This is the current behaviour:

Either:
a) the connection fails - no issue then
b) the resume on the given channel fails with an error the system deems to be fatal: so the channel enters the `failed` state, none of the locally queued messages are published and no new messages are received;
c) the resume on the given channel fails with a non-fatal error, meaning that the channel is reattached but with some discontinuity in the messages; this means that the resume returns as many historical messages on the channel as are available, which might or might not include the other client's 3 messages, and an error is returned indicating the problem. The 5 new queued messages are sent once attached.

When we in future support implicit attach on publish and multiple attach modes (publish, publish+subscribe etc) I don't really see that changing, except that there is a variation of case (b) where the fatal problem might apply to publish only, or to subscribe only, in which case the attach mode could be downgraded without failing the channel.

mattheworiordan:
@paddybyers to be clear, I raised this issue not because I don't agree with your suggested approach above, but because we have not been specific enough in our spec, see http://docs.ably.io/client-lib-development-guide/features/#RTN15c3.  There is NO mention whatsoever of queued messages and how to handle them, so every client library developer will likely implement this differently because we have been unspecific. I am happy with `a` and `b`, still unhappy with `c` but will go ahead and leave that in until we resolve this issue in `0.9`.

> When we in future support implicit attach on publish and multiple attach modes (publish, publish+subscribe etc) I don't really see that changing, except that there is a variation of case (b) where the fatal problem might apply to publish only, or to subscribe only, in which case the attach mode could be downgraded without failing the channel.

I would disagree in fact, I think if a channel that was open becomes `detached`, then it's just easier to simply fail those messages so that the developer can decide what recovery activity they would like to take. 

[...]

PR #58, opened 2015-11-11
Issue 57

mattheworiordan:
This addresses the discussion at https://github.com/ably/docs/issues/57#issuecomment-155377854 around how to handle queued messages when a channel enters a detached or failed state.

@paddybyers and @SimonWoolf you OK to merge this in?  If so, I will update the spec sheet as well, including adding the one new spec item.


SimonWoolf:
:+1:

content/client-lib-development-guide/features.textile
```diff
@@ -301,9 +301,9 @@ h3(#realtime-connection). Connection
...
+**** @(RTN15c1)@ @CONNECTED@ @ProtocolMessage@ with the same @connectionId@ as the current client, and no @error@. In this case, the server is indicating that the resume succeeded, all channels are still attached, and all backlog messages are available. The client should not change the state of attached channels, and immediately process any queued messages for that channel
+**** @(RTN15c2)@ @CONNECTED@ @ProtocolMessage@ with the same @connectionId@ as the current client, and an @error@. In this case, the server is indicating that the resume succeeded but with a non-fatal error, all channels are still attached, and some backlog messages may be unavailable. The @ErrorInfo@ received should be emitted on the client @Connection@ as well as all @ATTACHED@ @Channel@ objects, and the @errorReason@ should be set. The client should not change the state of attached channels, and immediately process any queued messages for that channel
```
paddybyers: I don't think it makes sense to emit the error on every channel.

paddybyers:
One comment, otherwise LGTM.

content/client-lib-development-guide/features.textile
mattheworiordan: Why not?  This is an error on the channel, so if I wanted to subscribe to errors relating to the channel I have one canonical place to do that.  Otherwise, devs need to subscribe to both Connection and Channel errors to get the full set.

paddybyers: Except that we don't right now have a way of determining which channel(s) are affected. I don't think there should be an error emitted on every channel.

mattheworiordan: I really don't like the idea that a) we don't know which channels are effected, b) we don't ever both notifying the user that an error has occurred, when discontinuity is indeed an error if you're expecting continuity on a channel. As a programmer, I think doing something like:

```javascript
channel = client.channels.get('x')
channel.on('attached', => { /* add channel to UI */ })
channel.on('detached', => { /* remove channel */ })
channel.on('failed', => { /* remove channel */ })
channel.on('error', => { /* notify user in channel messages are missing or get history */ })
```

is perfectly reasonable and expected.  However, expecting them to catch a connection error and KNOW that the connection error is related to the channel is just asking too much of a developer.  How will they know that an error message relates to the channel without looking up all our error codes, which is unrealistic for your average dev who just wants to catch an error and handle it.

mattheworiordan: As a use case btw. think about building a simple chat app.  You attach to a channel, and on the assumption that we manage connection state, you assume that as long as the channel remains attached then you can trust that it will capture all messages.  If we don't emit an error (and I would prefer to simply detach the channel personally, but we can revisit in 0.9), then the dev will likely never know that the channel has lost state, and thus missed messages.

mattheworiordan: Any final thoughts on this?  I realise emitting an error no every channel is not ideal, but until we can confirm which channels have suffered from discontinuity, I don't see any way around this.  I would prefer to address this property in 0.9, and at least let developers know that it is quite possible the channel emitting an error has indeed lost continuity of service.

paddybyers: I don't think we should emit an error on channels that don't have an error. I would prefer that we send any per-channel errors on per-channel error messages, and have the client library react to them individually.

mattheworiordan: @paddybyers sure, that would be better, how do you propose we do that as I don't think we ever specified that a protocol `ERROR` message can be directed at a channel, or that it is recoverable

paddybyers: > I don't think we ever specified that a protocol ERROR message can be directed at a channel

`ERROR` messages can have a `channel` member, in which case it signifies that the channel has `failed`. If there is no `channel` member, then the error is connection-wide.

> or that it is recoverable

Right now we don't have a way to indicate that the error is non-fatal.

Or we could send an `ATTACHED` for resumed channels, which may contain an (implicitly non-fatal) error.

mattheworiordan: > ERROR messages can have a channel member, in which case it signifies that the channel has failed. If there is no channel member, then the error is connection-wide.

Sorry, my oversight there.

> Or we could send an ATTACHED for resumed channels, which may contain an (implicitly non-fatal) error.

I think that is better suited to be honest.

mattheworiordan:
@paddybyers please review ce380a7 which hopefully addresses the problem when a connection is resumed but not all channels are resumed correctly.  If you are happy, I will merge in.

Issue #49, opened 2015-10-01
Terminology clean up


mattheworiordan:
Whilst working through the documentation today, I found that referring to clients that have a `clientId` as authenticated clients could be very confusing for users.  This is probably because being `authenticated` has no relation to your authentication with Ably via a token or API key, and the word authentication exists in Basic Authentication as well.

Having had a quick chat about this, @paddybyers and I came up with the following suggestion, which should not have any impact on the client libraries fortunately, but will simply be rolled out in our documentation.

@SimonWoolf and @lmars, do you have any objections or suggestions to improve this terminology?

### Authentication

The process of providing credentials to Ably in the form of a token or API key, which in turn allows you to communicate with Ably if the credentials are valid.

### Authorisation

Ably is responsible for authorising or refusing operations for the authenticated client based on the permissions specified within the API key and/or Token provided when the client authenticated.

### Identified Client

Any client that has connected to Ably and has either explicitly provided a `clientId` when instancing the library, or implicitly been assigned a `clientId` following token authentication, will then assume and be restricted to that `clientId` for all operations. As our customers are responsible for issuing tokens or token requests themselves, and we recommend that they only use token authentication and never share their private keys with clients, an identified client publishing or present on a channel with the `clientId` given in their token can therefore be trusted by other clients.

lmars:
LGTM

paddybyers:
Because people get especially pedantic when it comes to these things, I think we need to make sure the definitions of these terms are accurate and unambiguous in the docs.

(I wrote these and now I'm not sure if they're too pedantic, but whatever.)

Authentication

The process of deciding, based on the presented credentials, whether or not an entity may interact with the service. The credentials may be presented explicitly (eg using HTTP Basic Authentication) or in some cases the entity may prove possession of the credentials (eg with the signed body of a `requestToken` request). In Ably credentials are either an API key or a auth token.

Authenticated client

A client of the Ably service that has been successfully authenticated.

Authorisation

The process of deciding whether or not a given entity (usually authenticated) is allowed to perform a given operation. In Ably, authorisation for most operations is based on the Capabilities associated with the key or token that was used to authenticate a client.

Identified client

An authenticated client with a specific claimed client identity, or `clientId`, whose credentials are verified as confirming that identity. In Ably a client can be identified with a `clientId` in three ways:

- if the client authenticated with a token that is issued for that `clientId`;
- if the client  claims that `clientId` and is authenticated with an API key ("basic auth");
[...]

PR #60, opened 2015-11-18
Wildcard clientId Auth attribute

mattheworiordan:
Following the [ably-ruby PR to support a wildcard clientId](https://github.com/ably/ably-ruby/pull/66), it became apparent that there is no way currently to determine whether an authenticated client can assume any clientId, or is an anonymous client, because in both instances `Auth#clientId` is `null`.

This PR addresses that issue, and also introduces a new `identityConfirmed` attribute that can be used to cover the situation whereby a client is not yet authenticated or connected and as such the `null` value `Auth#clientId` could mean it's anonymous, but could also mean its `clientId` is unknown at the time

@paddybyers @SimonWoolf WDYT?

BTW. Whilst I would like to include this in the spec, this has no impact on the realtime system, and as such, can be progressively introduced into client libraries as and when.

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -89,8 +89,8 @@ h3(#rest-auth). Auth
 * @(RSA4)@ Token Auth is used if @useTokenAuth@ is set to true, or if @useTokenAuth@ is unspecified and any one of the following conditions cause token auth to be selected as the default: a @clientId@ is specified; @authUrl@ or @authCallback@ is provided; an explicit @token@ or @TokenDetails@ is provided
 * @(RSA14)@ If Token Auth is selected, an exception will be raised if a token is not provided or there is no means to generate a token. For example, if only the option @useTokenAuth@ is specified, and a @key@ is not provided, then the client library is unable to authenticate or issue a token
 * @(RSA15)@ If Token Auth is selected and @clientId@ has been set in the @ClientOptions@ when the library was instanced:
-** @(RSA15a)@ Any @clientId@ provided in @ClientOptions@ must match any non @null@ @clientId@ value in @TokenDetails@ or @connectionDetails@ of the @CONNECTED@ @ProtocolMessage@, where applicable
-** @(RSA15b)@ If the @clientId@ from @TokenDetails@ or @connectionDetails@ contains only a wildcard string @'*'@, then the client is permitted to be either unauthenticated (effectively anonymous without a @clientId@) or authenticated providing a @clientId@ when communicating with Ably
+** @(RSA15a)@ Any @clientId@ provided in @ClientOptions@ must match any non wildcard (@'*'@) @clientId@ value in @TokenDetails@ or @connectionDetails@ of the @CONNECTED@ @ProtocolMessage@, where applicable
+** @(RSA15b)@ If the @clientId@ from @TokenDetails@ or @connectionDetails@ contains only a wildcard string @'*'@, then the client is permitted to be either unauthenticated (i.e. anonymous without a @clientId@) or authenticated by providing a @clientId@ when communicating with Ably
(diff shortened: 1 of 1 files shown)


mattheworiordan:
@paddybyers I have removed the `identityConfirmed` attribute change from this PR so that this can land, and moved that to https://github.com/ably/docs/pull/64. Once we have resolved whether we include default attribute values for https://github.com/ably/docs/commit/afb2ad1698f5e42a731be04d0e0d5910a0de3434#diff-11900b395df266bc2bbfc1d11bdfc7a0R648 or not, we can proceed to merge this PR and https://github.com/ably/docs/pull/59

content/client-lib-development-guide/features.textile
```diff
@@ -98,14 +98,14 @@ h3(#rest-auth). Auth
...
+*** @(RSA7b2)@ Token authentication is being used, and the @TokenRequest@ or @TokenDetails@ object, used for authentication, has a @clientId@ value that is not @null@
+*** @(RSA7b3)@ Following a realtime connection being established, if the @CONNECTED@ @ProtocolMessages@ contains a @clientId@ that is not @null@. @clientId@ is an attribute of @ProtocolMessage#connectionDetails@ within a @CONNECTED@ @ProtocolMessage@
+*** @(RSA7b4)@ When a wildcard string @'*'@ is present in the @TokenRequest@, @TokenDetails@, or @ProtocolMessage#connectionDetails@ object, then the client does not have a @clientId@ but is allowed to assume any identity performing operations with Ably. As such, @Auth#clientId@ should contain the string value @'*'@ indicating that the current client is allowed to perform operations on behalf of any @clientId@
+** @(RSA7c)@ A @clientId@ provided in the @ClientOptions@ when instancing the library must be either @null@ or a string, and cannot contain only a wilcard @'*'@ string value as that client ID value is reserved
```
SimonWoolf: `wilcard`

SimonWoolf:
so we now have four different states that the clientId can be in:
- unknown (eg client initialized with a token string, before CONNECTED message)
- anonymous only (eg token auth with clientId null)
- a specific string (eg token auth with clientId *x*)
- anything (eg basic auth; token auth with clientId `*`)

AIUI the proposal is to use `null` for the first two, distinguishing between them with an `identityConfirmed` boolean, then {any string except `*`} for the third, and  `*` for the fourth.

At this point, should we consider just making it a four-valued enum? [`UNKNOWN`, `ANONYMOUS`, `IDENTIFIED`, `ANY`] or something. Might be overkill, but multi-meaning'd-`null` + magic value + boolean to distinguish four mutually exclusive states just seems a bit messy, I dunno

mattheworiordan:
> https://guides.github.com/features/mastering-markdown/so we now have four different states that the clientId can be in:

@SimonWoolf what's the URL to markdown for?  I assume a copy & paste mistake?

SimonWoolf:
no idea how that got there :blush:.

(best guess: palm brushed the trackpad and did an accidental middle-click, which pastes the X selection buffer)

mattheworiordan:
> At this point, should we consider just having an enum? `[UNKNOWN, ANONYMOUS, IDENTIFIED, ANY]`

Sure, could do, makes sense and certainly clears things up. However, I am not sure `IDENTIFIED` is necessarily accurate either.  For example, if a client initialises a library with a specific `clientId`, using the word `IDENTIFIED` implies that it's had some type of confirmation from Ably (see https://github.com/ably/docs/issues/49 for our new terminology).  This may not be the case.

However, thinking about this some more, I do think this does introduce a new set of issues in that a `TokenDetails`, `TokenParams` and `ConnectionDetails` object will not support this enum, so the client library will then have to do the work of mapping the different values to the enum, and if a user did inspect the value of one of those objects, or indeed `clientId` itself, they would still get either `null`, `*` or a string.

On balance, I think I'd prefer to not go down that route for now because I think whilst it provides some clarity in this instance, it introduces behaviour that is different from elsewhere in the system, the protocol and the REST API

SimonWoolf:
Fair enough. In that case, :+1: to this.

PR #64, opened 2015-11-19
Add identityConfirmed attribute to remove ambiguity

mattheworiordan:
null Auth#clientId can mean anonymous only, or potentially also mean the client does not yet know if it has a clientId or not. For example, if a client is initialised with a token string, then the client will have a null Auth#clientId always in the case of REST, or until it is connected in the case of Realtime. The Auth#identityConfirmed attribute disambiguates this

@paddybyers as discussed, if you have a better proposal please do suggest something.


mattheworiordan:
@paddybyers nudge to come up with an alternative or approve tomorrow please.

@SimonWoolf if you have any other suggestions please do shout!

PR #65, opened 2015-11-20
API versioning, client enforced defaults, and REST API updates

mattheworiordan:
@paddybyers and @SimonWoolf here are the API version updates as discussed, plus I updated the REST API to include details on the new custom error headers, and params for controlling REST responses.

@paddybyers one small issue, the curl request to filter fields with a `select` clause does not work.

    # this works
    curl https://rest.ably.io/stats?fields=channels.* \
     -u "xVLyHw.HqKqFA:BEPkQrL2XSdyjg-Z"

    # this doesn't
    curl https://rest.ably.io/stats?select=channels.* \
      -u "xVLyHw.HqKqFA:BEPkQrL2XSdyjg-Z"

Can you advise what I am doing wrong?


paddybyers:
> the curl request to filter fields with a select clause does not work

Unlike `fields`, which returns a subset of the result (ie an object with fewer fields, or an array of objects, each with fewer fields), `select` flattens the matching keys so that the result is a single object, and therefore all returned entries have to have a unique key. Thus if the matching object is an array, the matching member elements are given keys in the result that include the array index. Thus, in turn, the index is part of the path spec.

So, since a picture is worth a thousand words, the example is:

    curl https://rest.ably.io/stats?select=*.channels -u "xVLyHw.HqKqFA:BEPkQrL2XSdyjg-Z"

gives

````
{
	"0.channels": {
		"peak": 3,
		"min": 0,
		"mean": 2.7,
		"opened": 3
	},
	"3.channels": {
		"peak": 5,
		"min": 0,
		"mean": 4
	},
	"4.channels": {
		"peak": 5,
		"min": 2,
		"mean": 2.6,
		"opened": 3
	},
	"5.channels": {
		"peak": 2,
		"min": 2,
		"mean": 2
	},
	"6.channels": {
		"peak": 2,
		"min": 2,
		"mean": 2
	},
	...
}
````

Note that the representation is flattened only to the point that the keys match the path spec. So extending the path spec flattens the result further:

    curl https://rest.ably.io/stats?select=*.channels.* -u "xVLyHw.HqKqFA:BEPkQrL2XSdyjg-Z"

gives

````
{
	"3.channels.peak": 2,
	"3.channels.min": 0,
	"3.channels.mean": 0.2,
	"4.channels.peak": 3,
	"4.channels.min": 0,
	"4.channels.mean": 2.7,
	"4.channels.opened": 3,
	...
}
````

File: content/client-lib-development-guide/features.textile [hunk 1]
```diff
@@ -35,6 +35,7 @@ h2(#test-guidelines). Test guidelines
...
 * @(G3)@ Testing statistics can be tricky due to timing issues and slow test suites as a result of sending requests to generate statistics.  As such, we provide a special stats endpoint in our sandbox environment that allows stats to be injected into our metrics system so that stats tests can make predictable assertions.  To create stats you must send an authenticated @POST@ request to the stats JSON to @https://sandbox-rest.ably.io/stats@ with the stats data you wish to create. See the "Javascript stats fixture":https://github.com/ably/ably-js/blob/4e65d4e13eb8750a375b9511e4dd059092c0e481/spec/rest/stats.test.js#L8-L51 and "setup helper":https://github.com/ably/ably-js/blob/4e65d4e13eb8750a375b9511e4dd059092c0e481/spec/common/modules/testapp_manager.js#L158-L182 as an example.
+* @(G4)@ All REST requests and WebSocket connections to Ably must include the current API version @0.8@. Should any new API version with breaking changes be released, the client library will continue to use the API version explicitly requested
```
paddybyers: I didn't think we were intending that it was mandatory to include the version.

content/client-lib-development-guide/features.textile
```diff
@@ -56,6 +57,7 @@ h3(#restclient). RestClient
 *** @(RSC6b4)@ @unit@ is the period for which the stats will be aggregated by, values supported are @minute@, @hour@, @day@ or @month@; if omitted the unit defaults to the REST API default (@minute@)
 * @(RSC16)@ @RestClient#time@ function sends a get request to @rest.ably.io/time@ and returns the server time in milliseconds since epoch or as a Date/Time object where suitable
 * @(RSC7)@ Sends REST requests over HTTP and HTTPS to the REST end-point @rest.ably.io@
+** @(RSC7a)@ The header @Ably-Version: 0.8@ must be included in all REST requests to the Ably end-point
```
paddybyers: `X-Ably-Version` ?

content/rest-api/index.textile
```diff
@@ -120,6 +125,21 @@ Wherever possible, success response bodies contain links, in "HATEOS":http://en.
...
+h3(#versioning). Versioning
+
+By default, all requests receive the latest version of the API, which is currently @0.8@.
+
+When we make backwards-incompatible API changes to the API, we release new versions. Therefore, we encourage you to explicitly request the version you are interfacing with in all requests using one of the following mechanisms:
+
+1. Include a @Ably-Version@ header. Example:
+
+bc[sh]. curl https://rest.ably.io/time \
+ -H "Ably-Version: 0.8"
```
paddybyers: `X-` ?

content/rest-api/index.textile
```diff
@@ -149,6 +169,67 @@ In principle the link URL might be any valid URL but in practice it will always
...
+  # Response
+  {
+      "0.channels.peak": 2,
+      "0.intervalId": "2015-11-20:15"
+  }
+```
+
+h4. select
+
+Specifying @?select=<path spec>@ returns a representation of the resource containing only the fields that match the given path specification. Like the fields param this permits only specific fields to be obtained, but it the path spec may contain wildcard elements, and all matching fields in the representation are returned.
```
SimonWoolf: `but it the path spec` -> `but the path spec`

content/rest-api/index.textile
```diff
@@ -149,6 +169,67 @@ In principle the link URL might be any valid URL but in practice it will always
...
+
+The result is a partially-flattened representation, as a map whose keys are the long keys that match the path spec, and the values are the corresponding values. When the results are contained in an array, each result key is prefixed with an index in the form @[index].[key]@
```
paddybyers: > When the results are contained in an array

This needs to be worded in a way that applies to arrays at all levels, not just the top-level.

When the results contain array elements, the array index of each element is included as the corresponding component of the result key.

mattheworiordan:
This has been merged

PR #70, opened 2015-11-25
Remove connect callback as not agreed in spec

mattheworiordan:
See https://github.com/ably/ably-js/issues/43#issuecomment-159657024

As we have not agreed the spec, and it is not a core feature, deferring this until 0.9

@SimonWoolf @paddybyers you happy with this?


SimonWoolf:
:+1: yep

paddybyers:
LGTM.

PR #71, opened 2015-11-26
Attribute to confirm if identified

mattheworiordan:
This PR supersedes https://github.com/ably/docs/pull/64

Note I have suggested a new attribute `clientIdValidated`, please see 5a44e84


mattheworiordan:
cc @paddybyers @SimonWoolf

content/client-lib-development-guide/features.textile
```diff
@@ -107,7 +107,8 @@ h3(#rest-auth). Auth
...
-** @(RSA7c)@ A @clientId@ provided in the @ClientOptions@ when instancing the library must be either @null@ or a string, and cannot contain only a wildcard @'*'@ string value as that client ID value is reserved
+** @(RSA13)@ @Auth#clientIdValidated@ boolean is @true@ when the @Auth#clientId@ has been validated through the inspection of a @clientId@ attribute of a signed @TokenRequest@, @TokenDetails@ or a @ProtocolMessage#connectionDetails@ object. When @identityConfirmed@ is @true@ and @Auth#clientId@ is @null@, the client is anonymous and cannot assume a @clientId@ in any operations. However, when @identityConfirmed@ is @false@, a @null@ value @Auth#clientId@ does not necessarily mean that the client is anonymous as the client ID has simply not yet been confirmed by Ably following authentication or establishing a realtime connection
```
SimonWoolf: `@identityConfirmed@` -> `@clientIdValidated@` x2

SimonWoolf:
:+1:

content/client-lib-development-guide/features.textile
mattheworiordan: He he, quality of late night commits not great hey ;)

paddybyers:
It says that the `clientId` is validated as a result of seeing that `clientId` in a `TokenRequest`. Of course that by itself doesn't validate the `clientId`. (Equally you can't necessarily trust a `TokenDetails` unless you know where it came from.) Perhaps for consistency we should only be indicating that the `clientId` is validated when the `ConnectionDetails` confirm that, or if the client itself received the `TokenDetails` from the system.

Also, if we are having this new field I this we should consider @SimonWoolf's suggestion that we have an enum here of validation statuses, instead of simply the true/false of a boolean. So as well as 0=UNKNOWN, 1=VALID we could have ANY and ANONYMOUS/NONE. I agree that the increase in complexity doesn't seem quite commensurate with the extra value you get, but it does potentially allow us to have other statuses in the future (although I don't know what those would be).

mattheworiordan:
> It says that the clientId is validated as a result of seeing that clientId in a TokenRequest. Of course that by itself doesn't validate the clientId. (Equally you can't necessarily trust a TokenDetails unless you know where it came from.) Perhaps for consistency we should only be indicating that the clientId is validated when the ConnectionDetails confirm that, or if the client itself received the TokenDetails from the system.

I disagree here. The point is that the client library can and should assume that any `TokenDetails` and `TokenRequest` object provided to it is valid (it has no means to validate that anyway without communicating with Ably, which in the case of REST is not practical). The reason I am proposing the `clientIdValidated` field should exist is only to eliminate the ambiguity of a `null` value `Auth#clientId` field where that could be `null` because it was not specified, or `null` because it is restricted to an anonymous user.

Why do we need a field that states if the clientId is unknown, valid, anonymous, or any.  That simply adds more confusion:

  * If a user instances a library with a clientId value, but has not yet authenticated.  Is it `valid` or `unknown`?  
  * If a wildcard token request is used with a clientId, is the `clientId` `unknown` because it's not validated by Ably, or is it `any`?
  * If a client library is instanced without a clientId, is it `anonymous`, or `unknown`, because it could be either depending on the authentication.

I really don't think adding that enum helps solve the problem of the `null` value clientId, it just adds additional complexity to the library, and as far as I can tell, does not even solve the problem in hand.

mattheworiordan:
We have not agreed this, so it may be deferred to 0.9

PR #72, opened 2015-11-27
Proofread of how-ably-works draft

SimonWoolf:
Looks good. Couple of trivial style tweaks. (also for the auth diagrams: second one seems to have much smaller arrowheads than the first)

Minor concern: we don't give any guidance on whether to use server-signed token requests or server-requested tokens. I'd guess most devs probably just want to be told 'do it this way except in circumstances X,Y,Z'? (If we really don't want to have a preferred one, should we maybe just pick one for the 'how ably works' page to keep it simple, relegate the other to an "Alternatively, ..." paragraph with no diagram?)

### DIFF
--- a/content/root/how-ably-works.textile
+++ b/content/root/how-ably-works.textile
@@ -15,17 +15,17 @@ Ably's global realtime service enables Internet enabled devices, such as a brows
 
 h2. Key concepts
 
-At its simplest, Ably can be described as a cloud-based "pub/sub":https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern Platform-as-a-Service ensuring any device publishing messages to Ably will be received in real time by any number of subscribing devices. However, in reality, Ably makes it possible for developers to build apps and infrastructure that can communicate in realtime without the worries of managing scale, latency, data durability, integrity and storage, seamless connection recovery, device interopability, network outages, encryption, security and authentication, throttling and denial of service attacks, to name a few.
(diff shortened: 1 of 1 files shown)


mattheworiordan:
> Minor concern: we don't give any guidance on whether to use server-signed token requests or server-requested tokens. I'd guess most devs probably just want to be told 'do it this way except in circumstances X,Y,Z'? (If we really don't want to have a preferred one, should we maybe just pick one for the 'how ably works' page to keep it simple, relegate the other to an "Alternatively, ..." paragraph with no diagram?)

I agree, less is more.  We do have a dedicated authentication page which I will move server generated token auth to then.  Our recommendation is to use signed token requests and pass those to clients.  I did have copy recommending one over the other but it became a bit verbose.  As you say though, if we only show one method and link to others that would be better.

content/root/how-ably-works.textile
```diff
@@ -39,7 +39,7 @@ Ably supports two forms of authentication.
 
 h4. Basic Authentication
 
-Basic authentication is the simplest form of authentication using an app's private API key to connect to Ably. Private API keys and their capabilities (permissions) are managed within an "app's dashboard":http://support.ably.io/support/solutions/articles/3000030053-how-do-i-access-my-app-dashboard. In most cases, we do not recommend that basic authentication is used as it requires that you share your private API key with the client that is connecting to Ably. Our recommendation is to only use basic authentication on your trusted servers when communicating with Ably over TLS.
```
paddybyers: > Our recommendation is

It's more than a recommendation; it's a requirement.

mattheworiordan: Well it's worded wrong, what I meant was it's our recommendation to only use basic auth on your trusted servers, which inadvertently only works over TLS anyway

PR #77, opened 2016-01-15
Clarify naming of reason / errorReason attribute

mattheworiordan:
If a state change is emitted, then the attribute ConnectionStateChange#reason makes sense in the context of the state change.

However, on a Connection or Channel object itself, simply having Connection#reason is not clear, so in this case it is named Connection#errorReason so that the last error on the object is accessible.

@paddybyers as discussed, are you happy with this?

@SimonWoolf shout if you have any views or spot any inconsistencies.


paddybyers:
LGTM

PR #79, opened 2016-01-21
Handling token expired failures more comprehensively

mattheworiordan:
See https://github.com/ably/ably-js/issues/203

@paddybyers are you happy with the proposed change to the spec?


content/client-lib-development-guide/features.textile
```diff
@@ -63,15 +63,15 @@ h3(#restclient). RestClient
...
 ** @(RSC14b)@ Uses token authentication whenever the criteria of "RSA4":#RSA4 are met
-** @(RSC14c)@ When a @token@ or @tokenDetails@ is used to instance the library, and no means to renew the token is provided (either an API key, @authCallback@ or @authUrl@), if the server responds with 40140 (token expired) to a request, the client library should indicate an error and not retry the request
-** @(RSC14d)@ When the client does have a means to renew the token automatically, and the token has expired or the server has responded with 40140 (token expired), then the client should automatically renew the token and resend the request
+** @(RSC14c)@ When a @token@ or @tokenDetails@ is used to instance the library, and no means to renew the token is provided (either an API key, @authCallback@ or @authUrl@), if the server responds with a 401 HTTP status code and an Ably code value @40140 <= @code < 40150@ indicating token is expired or invalid, the client library should indicate an error and not retry the request
```
paddybyers: I don't really like the idea that the library now has to remember the previous error code.

An error in the range [40140 ... 40150) is saying that there was a specific issue with that token, and there is every reason to suppose that a new token will not have the same problem. If on the other hand the key was revoked, say, you'd get a 4013xx error and would therefore not attempt to get a new token.

So although I don't like the possibility of a perpetual cycle of issuing tokens, it should not arise provided we return the correct error code.

mattheworiordan: > So although I don't like the possibility of a perpetual cycle of issuing tokens, it should not arise provided we return the correct error code.

Yup, agreed, but the perpetual cycle worries me...

mattheworiordan:
@paddybyers what about this?

content/client-lib-development-guide/features.textile
```diff
@@ -294,12 +294,13 @@ h3(#realtime-connection). Connection
...
 ** @(RTN14f)@ The connection will remain in the @SUSPENDED@ state indefinitely, whilst periodically attempting to reestablish a connection
 * @(RTN15)@ @Connection@ failures once @CONNECTED@:
+** @(RTN15h)@ If a @DISCONNECTED@ message is received from Ably, then the connection will subsequently be closed by Ably. If the @DISCONNETED@ message contains a token error (@statusCode@ value of 401 and error @code@ value in the range @40140 <= @code < 40150@), if the token is renewable, the error should not emitted, a single attempt to create a new token will be made and a new connection attempt will be initiated using the new token. If the token creation fails, the next connection attempt fails due to a token error, or the token was not renewable, the connection will transition to the @FAILED@ state and the @Connection#errorReason@ will be set
```
paddybyers: I think the expression "the connection will subsequently be closed" is misleading. I think we should refer to a "transport" or "transport connection".

paddybyers:
> what about this?

one comment above but otherwise LGTM

content/client-lib-development-guide/features.textile
mattheworiordan: Yes, I agree, will change now

PR #84, opened 2016-02-09
spec: Further clarification on fallback host behaviour

mattheworiordan:
@paddybyers can you review these changes to the fallback host spec as I specifically wanted to address:

* REST library requests should always be issued to the default host first i.e. the REST lib has no state based on the previous request ensuring the closest data centre is always preferred relying on DNS
* Realtime libraries, when using a fallback host, will use that endpoint first for any REST requests ensuring the same data centre is used

cc @SimonWoolf


content/client-lib-development-guide/features.textile
```diff
@@ -73,9 +73,10 @@ h3(#restclient). RestClient
...
+** @(RSC15b)@ This fallback behaviour only applies to the default @rest.ably.io@ endpoint; if either @host@ or @environment@ options are set resulting in the endpoint not being @rest.ably.io@, no fallback host is used
+** @(RSC15d)@ Automatic retries to fallback hosts will not occur for any HTTP status codes not in the range 500 to 504. For example, if the credentials are invalid (a 401 status code) or the maximum message length has been exceeded, then resending the request to an alternative data centre will not fix the problem and will only increase the load on other data centres. See the "complete list of Ably error codes":https://github.com/ably/ably-common/blob/master/protocol/errors.json#L11
+** @(RSC15e)@ Every new HTTP operation is always first attempted to the primary host, and then, only upon "an appropriate request failure":#RSC15a, attempted against the fallback hosts when "RSC15b":#RSC15b applies. This ensures that should the closest data centre become available at a later time, the client library will automatically resume using the primary data centre at @rest.ably.io@
```
paddybyers: I don't understand why this links to RSC15a.

Perhaps the logic/explanation would be clearer if these were reordered slightly:

RSC15e: by default (provided conditions in RCS15b are met) try primary host;
RSC15a: in the case of an error necessitating use of an alternative host (see RCS 15d), try fallback hosts in random order, continuing to try further hosts if qualifying errors occur, failing when all have been tried;
RSC15d: errors that necessitate use of an alternative host: host unresolvable, or unreachable, or was reachable but responded with an applicable error code
RSC15b: fallback behaviour only applies if the primary host is the default, and no overridden in client options

Also, I would prefer that we specify the qualifying error responses in terms of the `code` first, deferring to `statusCode` if there is no `code`. This will make it more general and consistent with how we specify other error response behaviour eg conditions for token renewal.

Are you unable to reorder the points because you want to preserve consistency with earlier numbering?

content/client-lib-development-guide/features.textile
```diff
@@ -324,8 +325,9 @@ h3(#realtime-connection). Connection
...
+** @(RTN17b)@ The fallback behaviour only applies to the default @realtime.ably.io@ endpoint; if either @host@ or @environment@ options are set resulting in the host endpoint not being @realtime.ably.io@, no fallback host is used
+** @(RTN17c)@ If the realtime client is currently connected to a fallback host endpoint, then for the duration that the transport is open to that host, all HTTP requests, such as history or token requests, should be first attempted to the same data centre the realtime connection is established with i.e. the same fallback host must be used as the default host. If however the HTTP request against that fallback host fails, then the normal fallback host behaviour should be followed attempting the request against another fallback host as described in "RSC15":#RSC15
```
paddybyers: I would prefer that we say the transport is `connected` instead of `open`.

paddybyers:
A couple of comments; otherwise LGTM

content/client-lib-development-guide/features.textile
mattheworiordan: I have no problem re-ordering, yes it's a bit more of a hassle but it's not a big problem...  I will have a go at that and addressing your points

mattheworiordan:
@paddybyers please review now again as I believe I've addressed all your comments

content/client-lib-development-guide/features.textile
```diff
@@ -73,9 +73,10 @@ h3(#restclient). RestClient
...
+** @(RSC15e)@ By default, every new HTTP request is first attempted to the primary host @rest.ably.io@, which, through DNS, is automatically routed to the client's closest data centre. The client library must always prefer the default endpoint (closest data centre), even if a previous request to that endpoint has failed
+** @(RSC15a)@ In the case of an error necessitating use of an alternative host (see "RSC15d":#RSC15d), try fallback hosts in random order, continuing to try further hosts if qualifying errors occur, failing when all have been tried or the configured @httpMaxRetryCount@ has been reached (see "@TO3l5@":#TO3l5). This ensures that a client library is able to work around routing or other problems for the user's closest data centre. For example, if a @POST@ request to @rest.ably.io@ fails because the default endpoint is unreachable or unserviceable, then the @POST@ request should be retried again against the fallback hosts in attempt to find an alternate healthy data centre to service the request. The five fallback hosts are @[a-e].ably-realtime.com@.
```
paddybyers: Perhaps link "qualifying error" to `RSC15d` ?

content/client-lib-development-guide/features.textile
```diff
@@ -73,9 +73,10 @@ h3(#restclient). RestClient
...
-** @(RSC15b)@ This fallback behaviour only applies to the default @rest.ably.io@ endpoint; if either @host@ or @environment@ options are set, no fallback host is used
-** @(RSC15d)@ Automatic retries to fallback hosts will not occur for any HTTP status codes not in the range 500 to 504. For example, if the credentials are invalid (a 401 status code) or the maximum message length has been exceeded, then resending the request to an alternative data center will not fix the problem and will only increase the load on other data centers. See the "complete list of Ably error codes":https://github.com/ably/ably-common/blob/master/protocol/errors.json#L11 and a "discussion in the Java library about this":https://github.com/ably/ably-java/issues/14
+** @(RSC15b)@ The fallback behaviour described below only applies when the default @rest.ably.io@ endpoint is being used and has not been overriden, see "RSC11":#RSC11. If host fallback is not supported, failing HTTP requests result in an error immediately and are not retried
```
paddybyers: This needs to be clear that failing requests do not result in an alternative host being tried; there may still be a retry with the same host for other reasons (eg after token renewal).

paddybyers:
See comments but otherwise LGTM.

PR #85, opened 2016-02-09
spec: Clarify EventEmitter spec

mattheworiordan:
Specifically addressing the ambiguity around whether registrations are sets or arrays, and whether off removes the first match or all matches

@tcard @paddybyers please review


content/client-lib-development-guide/features.textile
```diff
@@ -482,11 +482,11 @@ h3(#realtime-presence). Presence
...
-* @(RTE5)@ @EventEmitter#off@ deregisters a listener for the specified event if given, else it is deregisters that listener for all registrations (global and event-specific)
+* @(RTE3)@ @EventEmitter#on@ allows a listener to be registered for either all events when no event argument is provided, or for only a single type of event when an event argument is provided. If @on@ is called twice with the exact same arguments, two listeners are added to the array of registered listeners. Therefore, it is perfectly valid to register the same listener multiple times and expect the listener to be called multiple times for each event emitted
+* @(RTE4)@ @EventEmitter#once@ registers a listener that is called only once when the first of the applicable events occurs (i.e. the specified event if given, or the first event if it is registered for all events). As with @on@, it is valid to register the same listener more than once which will result in the listener being called once for each registration
```
tcard: This phrasing is still somewhat ambiguous, in that you can interpret "which will result in the listener being called once for each registration" as meaning "all previous (and subsequent?) registrations of the same listener will now be called just once". Maybe something like "all registrations of the same listener will be independent from each other, just as if a different listener had been registered"?

content/client-lib-development-guide/features.textile
```diff
@@ -482,11 +482,11 @@ h3(#realtime-presence). Presence
...
 * @(RTE2)@ Where objects provide @subscribe@ or @unsubscribe@ methods, they should follow the specification for the @EventEmitter#on@ and @EventEmitter#off@ methods respectively
-* @(RTE3)@ @EventEmitter#on@ registers a listener for the specified event if provided, else it is registered for all events
-* @(RTE4)@ @EventEmitter#once@ registers a listener that is called only once when the first of the applicable events occurs (ie the specified event if given, or the first event if it is registered for all events)
-* @(RTE5)@ @EventEmitter#off@ deregisters a listener for the specified event if given, else it is deregisters that listener for all registrations (global and event-specific)
+* @(RTE3)@ @EventEmitter#on@ allows a listener to be registered for either all events when no event argument is provided, or for only a single type of event when an event argument is provided. If @on@ is called twice with the exact same arguments, two listeners are added to the array of registered listeners. Therefore, it is perfectly valid to register the same listener multiple times and expect the listener to be called multiple times for each event emitted
```
tcard: "If @on@ is called twice with the exact same arguments, two listeners are added to the array of registered listeners." I think this is a bit confusing way of putting it; now what constitutes a listener is even more diffuse, as what we treat like the _same_ listener in one place results in _two_ listeners here. "Array" also isn't particularly helpful in describing this.

Maybe it could be mentioned that the EventEmitter keeps a _registry_, each entry in the registry is called a _registration_ as has a listener and possibly a specific event type, and several _independent_ registrations can be made for the _same listener_.

content/client-lib-development-guide/features.textile
```diff
@@ -482,11 +482,11 @@ h3(#realtime-presence). Presence
...
+* @(RTE4)@ @EventEmitter#once@ registers a listener that is called only once when the first of the applicable events occurs (i.e. the specified event if given, or the first event if it is registered for all events). As with @on@, it is valid to register the same listener more than once which will result in the listener being called once for each registration
+* @(RTE5)@ @EventEmitter#off@ deregisters every matching listener for the specified event if given, else it is deregisters every matching listener for any registrations i.e. registrations for any event and registrations for a specific event. If @off@ is called with no arguments, all listeners for all and any events are removed
```
tcard: Following up the above comment, what about "if called with a specific event and a listener, it removes all registrations that match both the given listener and the given event; if called only with a listener, it removes all registrations matching the given listener, no matter their associated specific event or if there is one; if called with no arguments, it removes all registrations, for all events and listeners."

mattheworiordan: I like that, I will amend now

mattheworiordan:
@tcard have a look now and tell me what you think

tcard:
LGTM

content/client-lib-development-guide/features.textile
```diff
@@ -482,12 +482,12 @@ h3(#realtime-presence). Presence
...
-* @(RTE5)@ @EventEmitter#off@ deregisters a listener for the specified event if given, else it is deregisters that listener for all registrations (global and event-specific)
-* @(RTE6)@ @EventEmitter#emit@ emits an event, calling registered listeners with the given event name any other given arguments. If an exception is raised in any of the listeners, the exception is caught by the @EventEmitter@ and the exception is logged to the Ably logger. Tests must exist to ensure exceptions raised in client code do not propagate and inhibit other event processing within the client library
+* @(RTE3)@ @EventEmitter#on@ registers the provided listener for either all events when no @event@ argument is provided, or for only a single type of event when an @event@ argument is provided. If @on@ is called more than once with the same listener, the listener is added multiple times to its listener registry. Therefore, as an example, assuming the same listener is registered twice using @on@, and an event is emitted once, the listener would be invoked twice
```
paddybyers: Suggest "or for only a single type of event" -> "or for only a single named event"

"If @on@ is called more than once with the same listener" -> "If @on@ is called more than once with the same listener and event"

content/client-lib-development-guide/features.textile
```diff
@@ -482,12 +482,12 @@ h3(#realtime-presence). Presence
...
+* @(RTE4)@ @EventEmitter#once@ registers the provided listener for either the first event that is emitted when no @event@ argument is provided, or for only the first occurence of a single type of event when an @event@ argument is provided. If @once@ is called more than once with the same listener, the listener is added multiple times to its listener registry. Therefore, as an example, assuming the same listener is registered twice using @once@, and an event is emitted once, the listener would be invoked twice. However, all subsequent events emitted would not invoke the listener as @once@ ensures that the listener is only invoked once
```
paddybyers: "first occurence of a single type of event" -> "first occurence of a single named event"

"as @once@ ensures that the listener is only invoked once" -> "as @once@ ensures that each registration is only invoked once"

paddybyers:
See comments but otherwise LGTM

mattheworiordan:
Thanks @paddybyers, all merged in now

PR #86, opened 2016-02-09
Client library API IDL

mattheworiordan:
@tcard I have imported your IDL, and then gone through it to double check the definitions.  I found a few small issues and fixed them plus added consistent ordering for classes which is embeds, constructors, attributes, methods, all in alphabetical order.  You can see just these changes at https://github.com/ably/docs/compare/5d31555...api-idl. If you agree with my changes, you will most likely need to roll these changes to your iOS API work.

cc @paddybyers 

BTW. A screenshot of how it looks is below:

voila_capture 2016-02-09_08-01-57_pm

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -754,15 +754,15 @@ class ClientOptions:
   defaultTokenParams: TokenParams? // TO3j11
   echoMessages: Bool default true // RTC1a, TO3h
   environment: String? // RSC15b, TO3k1
-  logHandler: // TO3c
-  logLevel: // TO3b
-  port: Int // TO3k4
+  logHandler: // platform specific - TO3c
+  logLevel: // platform specific - TO3b
+  port: Int default 80 // TO3k4
   queueMessages: Bool default true // RTP16b, TO3g
-  restHost: String? // RSC12, TO3k2
-  realtimeHost: String? // RTC1d, TO3k3
+  restHost: String default "rest.ably.io" // RSC12, TO3k2
+  realtimeHost: String default "realtime.ably.io" // RTC1d, TO3k3
   recover: String? // RTC1c, TO3i
   tls: Bool default true // RSC18, TO3d
-  tlsPort: Int // TO3k5
+  tlsPort: Int default 443 // TO3k5
   useBinaryProtocol: Bool default true // TO3f
   // configurable retry and failure defaults
   disconnectedRetryTimeout: Duration default 15s // TO311
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -707,14 +707,14 @@ The following bespoke IDL (Interface Definition Language) describes the types an
 
 Please note the following conventions:
 
-* Types are capitalized.
-* Variable and method names are lowercase.
-* Values produced by I/O (e. g. a request) are prefixed with @io@. In some platforms (JS) those values are passed as arguments to a callback argument, instead of being returned. I/O always can fail, but how do they fail is undefined in the spec, so it's also undefined here.
-* Functions are denoted as @(Type, argName: Type, argName: Type) -> ReturnType@. (@ReturnType@ may be omitted.)
-* Enums are denoted as @.A | .B | .C@.
-* @Type?@ denotes an optional type, ie. a type that can be null plus all values of type Type.
-* @Type default value@ denotes that the thing being anotated with @Type@ has @value@ as default.
-* Class fields (as opposed to instance fields) are prefixed with a @+@.
+* Types are capitalized
+* Variable and method names are lowercase
+* Values produced by I/O (e. g. a request) are prefixed with @io@. In some platforms (JS) those values are passed as arguments to a callback argument, instead of being returned. I/O always can fail, but how do they fail is undefined in the spec, so it's also undefined here
+* Functions are denoted as @(Type, argName: Type, argName: Type) -> ReturnType@ (@ReturnType@ may be omitted)
+* Enums are denoted as @.A | .B | .C@
+* @Type?@ denotes an optional type, ie. a type that can be null plus all values of type Type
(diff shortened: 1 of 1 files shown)


content/client-lib-development-guide/features.textile
```diff
@@ -696,3 +701,355 @@ The following default values are configured for the client library:
...
+
+class EventEmitter<Event, Data>:
+  on: ((Data) ->) // RTE4
+  on: (Event, (Data) ->) // RTE4
+  once: ((Data) ->) // RTE4
+  once: (Event, (Data) ->) // RTE4
+  off: () // RTE5
+  off: ((Data) ->) // RTE5
+  off: (Event, (Data) ->) // RTE5
+  emit: (Event, Data?)  // RTE6
```
tcard: Sorry to make you fix this now when I got it wrong, but here `Data` shouldn't have the `?`. If whoever uses EventEmitter in a particular case wants possibly null data, it should pass a suitable type as the `Data` type parameter.

tcard: Oh, I noticed this is something you added. Well, if you want to require that `Data` must be nullable, then it should be in every method, not only in `emit`.

content/client-lib-development-guide/features.textile
```diff
@@ -696,3 +701,355 @@ The following default values are configured for the client library:
...
+    limit: int api-default 100, // RSC6b3
+    unit: .Minute | .Hour | .Day | .Month api-default .Minute // RSC6b4
+  ) -> io PaginatedResult<Stats> // RSC6a
+  time: () -> io Time // RSC16
+
+class Realtime:
+  constructor: (keyStr: String) -> // RSC1
+  constructor: (tokenStr: String) -> // RSC1
+  constructor: (ClientOptions) -> // RSC1
+  close: () // proxy for RTN12
```
tcard: Nitpick: missing the arrow, functions without returns are noted like `() ->`.

content/client-lib-development-guide/features.textile
```diff
@@ -696,3 +701,355 @@ The following default values are configured for the client library:
...
+* @Type default value@ denotes that the thing being anotated with @Type@ has @value@ as default. @Type api-default value@ denotes that the Ably server API uses those defaults and therefore it is unnecessary for the client library to send these values by default
+* Class fields (as opposed to instance fields) are prefixed with a @+@
+
+```[python]
+class Rest:
+  constructor: (keyStr: String) -> // RSC1
+  constructor: (tokenStr: String) -> // RSC1
+  constructor: (ClientOptions) -> // RSC1
+  auth: Auth // RSC5
+  channels: Channels<Channel> // RSN1
```
tcard: Should be `<RestChannel>` now.

content/client-lib-development-guide/features.textile
```diff
@@ -696,3 +701,355 @@ The following default values are configured for the client library:
...
+class ConnectionDetails:
+  clientId: String? // RSA12a, CD2a
+  connectionKey: String // RTN15e, CD2b
+  connectionStateTtl: Duration // CD2f, RTN14e, DF1a
+  maxFrameSize: Int // CD2d
+  maxInboundRate: Int // CD2e
+  maxMessageSize: Int // CD2c
+  serverId: String // CD2g
+
+class Message:
```
tcard: By the way, as this class can be instantiated by the user (as `Channel.publish` accepts `[Message]`) the constructors maybe should be also specified.

content/client-lib-development-guide/features.textile
mattheworiordan: Well in fact I always understood the EventEmitter to be able to emit any number of arguments, but also not emit any.  See https://github.com/ably/ably-ruby/blob/master/lib/ably/modules/event_emitter.rb#L88 and https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/util/EventEmitter.java#L75.  Both allow zero+ arguments. How do you want to approach this so that it's flexible for iOS as well then?

content/client-lib-development-guide/features.textile
mattheworiordan: Ah I actually misunderstood the syntax, I thought anything with `functionName: ()` indicated a function, whereas without the `()` it was an attribute.  I assumed (incorrectly) that the arrow indicated a closure / callback of some sort so for example, if `close` took a callback, then it would be shown as `close(() ->)`.  Are you saying every function with this syntax should have a trailing `->`?  If it then wanted to express close with a callback would it be `close(() ->) ->`?

content/client-lib-development-guide/features.textile
mattheworiordan: Thanks, will amend

content/client-lib-development-guide/features.textile
mattheworiordan: Ok, will add

content/client-lib-development-guide/features.textile
```diff
@@ -696,3 +702,362 @@ The following default values are configured for the client library:
...
+Please note the following conventions:
+
+* Types are capitalized
+* Attribute and method names are lowercase
+* Attributes are denoted as @attributeName: Type@
+* Instance methods are denoted as @methodName(argName: Type, argName: Type) -> ReturnType@ (@ReturnType@ may be omitted)
+* Callback functions / closures are denoted as @(argName: Type, argName: Type) -> ReturnType@ (@ReturnType@ is usually omitted)
+* Values produced by I/O (e.g. a request) are prefixed with @=> io@. In some platforms (JS) those values are passed as arguments to a callback argument, instead of being returned. I/O always can fail, but how do they fail is undefined in the spec, so it's also undefined here
+* Enums are denoted as @.A | .B | .C@
+* @Type?@ denotes an optional type, i.e. a type that can be null plus all values of type @Type@
```
paddybyers: > @Type?@ denotes an optional type

Wouldn't we say nullable?

I realise that what you're trying to say is that if an argument to a method has nullable type then  the library can equivalently make that an optional argument. But just saying the type is optional is too confusing.

mattheworiordan: Agreed, will fix and merge

PR #87, opened 2016-02-12
Completion listeners for implicit operations

mattheworiordan:
Following two discussions https://github.com/ably/ably-ios/pull/208#discussion_r52734670 and https://github.com/ably/ably-js/commit/1b151df5558339d8b2fd80bbadd2375d9a138b4a on how to handle implicit failures, this PR aims to address the issue by:

* Ensuring that any failure to implicitly attach can notify the caller
* Treating subscribe operations as atomic i.e. if the channel is in a FAILED state, then report the failure to the caller and do no register the listener

This approach is more future-proof as we look to optimise channel attach states i.e. only for subscribe, only for publish etc. based on which operations are called on a channel.  Whilst we could have simply removed the implicit attach, this will hurt us in the long run.

@SimonWoolf @paddybyers @tcard please review

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -444,7 +444,6 @@ h3(#realtime-presence). Presence
 ** @(RTP8a)@ Enters the current client into this channel, optionally with the data provided
 ** @(RTP8b)@ Optionally a callback can be provided that is called for both success or failure to enter
 ** @(RTP8c)@ A @PRESENCE ProtocolMessage@ with a @PresenceMessage@ with the action @ENTER@ is sent to the Ably service. The @clientId@ attribute of the @PresenceMessage@ must not be present. Entering without an explicit @PresenceMessage#clientId@, implicitly uses the @clientId@ for the current connection
-** @(RTP8d)@ Implicitly attaches the @Channel@. However, if the channel is in the @FAILED@ state, the @enter@ request results in an error
 ** @(RTP8e)@ Optional data can be included when entering a channel that will be encoded / decoded as with normal messages. A test should exist to ensure data used with enter is encoded & decoded correctly. Also, when data is provided when entering, but no data is provided when leaving, the data attribute should be emitted in the @LEAVE@ event for this client
 ** @(RTP8f)@ If the client library is authenticated but unidentified (i.e. @clientId@ is a wildcard @'*'@ or client is anonymous), the @enter@ request results in an error immediately
 ** @(RTP8g)@ If the channel is @DETACHED@ or @FAILED@, the @enter@ request results in an error immediately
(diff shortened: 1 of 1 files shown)


content/client-lib-development-guide/features.textile
```diff
@@ -436,7 +436,7 @@ h3(#realtime-presence). Presence
 * @(RTP6)@ @Presence#subscribe@ function:
 ** @(RTP6a)@ Subscribe with no arguments subscribes a listener to all presence messages
 ** @(RTP6b)@ Subscribe with a single action argument - such as @ENTER@, @LEAVE@, @UPDATE@ or @PRESENT@ - subscribes a listener to receive only presence messages with that action
-** @(RTP6c)@ Implicitly attaches the @Channel@. However, if the channel is in the @FAILED@ state, the @subscribe@ request results in an error
+** @(RTP6c)@ Implicitly attaches the @Channel@. However, if the channel is in or moves to the @FAILED@ state before the operation succeeds, it will result in the listener not being registered and an error being indicated, typically to the optional callback where the language permits
```
SimonWoolf: > typically to the optional callback where the language permits

If this is in the spec, I take it you've changed your mind re documenting the callback arg (cf [prev discussion](https://github.com/ably/ably-js/pull/226#discussion_r51737039))?

content/client-lib-development-guide/features.textile
```diff
@@ -444,7 +444,6 @@ h3(#realtime-presence). Presence
 ** @(RTP8a)@ Enters the current client into this channel, optionally with the data provided
 ** @(RTP8b)@ Optionally a callback can be provided that is called for both success or failure to enter
 ** @(RTP8c)@ A @PRESENCE ProtocolMessage@ with a @PresenceMessage@ with the action @ENTER@ is sent to the Ably service. The @clientId@ attribute of the @PresenceMessage@ must not be present. Entering without an explicit @PresenceMessage#clientId@, implicitly uses the @clientId@ for the current connection
-** @(RTP8d)@ Implicitly attaches the @Channel@. However, if the channel is in the @FAILED@ state, the @enter@ request results in an error
```
SimonWoolf: Newer copy missing here?

SimonWoolf:
:+1: after fixes

content/client-lib-development-guide/features.textile
mattheworiordan: No because every language varies the way this works.  The IDL addresses this with `Values produced by I/O (e.g. a request) are prefixed with => io. In some platforms (JS) those values are passed as arguments to a callback argument, instead of being returned. I/O always can fail, but how do they fail is undefined in the spec, so it’s also undefined here`.  I believe being unspecific here is to our advantage

tcard:
If I understand this correctly, what you say [here](https://github.com/ably/ably-ios/pull/208#discussion_r52734670):

> Subscribing to a channel that is failed is not necessarily wrong, it just means the listener will not be called until the channel is attached again.

is no longer valid, because "if the channel is in a FAILED state, then report the failure to the caller and do no register the listener", right?

Also, following what's been said in the discussions, shouldn't it be explicitly told that exception or exception-like error reporting should be avoided here?

mattheworiordan:
> is no longer valid, because "if the channel is in a FAILED state, then report the failure to the caller and do no register the listener", right?

Yes, after chatting with @paddybyers we felt that was the cleanest i.e. it works or fails completely.

> Also, following what's been said in the discussions, shouldn't it be explicitly told that exception or exception-like error reporting should be avoided here?

Well no, I think we should avoid steering people too much with their implementation as it is language specific

mattheworiordan:
@paddybyers any thoughts before we merge?

paddybyers:
LGTM

PR #89, opened 2016-02-23
Proposed encryption changes

SimonWoolf:
Changes to spec and docs resulting from encryption discussion yesterday

Summary of changes:
- Crypto::getDefaultParams always takes a key, and is always synchronous. Generating a random key is split out into Crypto::generateRandomKey(keyLength), which can be async
- key arg must be either a language-dependent binary type or a string. If a string it must be base64-encoded.
- getDefaultParams must validate key is the correct length
- can initialize a channel with `{encrypted: true, cipherParams: {key: key}}`

The documentation changes assume that ably-js and ably-ruby have implemented all the changes, and ably-java has implemented all except the last (I don't know java, but get the impression it seems to like having all options-hash-type things be explicitly-created dedicated objects, so not sure if the last would be idiomatic java..?)

I also moved the Crypto functions in the spec to their own section in Rest rather than being bolted on to the end of Types > Option types > CipherParams.

Rest#encryption page was very old -- all the improvements had been made to the realtime page only -- so I just copied the realtime page over and tweaked as appropriate. IIRC Matt mentioned changes he has planned to reduce rest/realtime docs duplication, so didn't want to do anything that might conflict - if not could easily eg extract most of it to a partial

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -880,13 +880,13 @@ enum ChannelState:
   FAILED
 
 class ChannelOptions:
-  +withCipherKey(key: Binary)? -> ChannelOptions // TB3
+  +withCipherKey(key: Binary | String)? -> ChannelOptions // TB3
   cipher: (CipherParams | Params)? // RSL5a, TB2b
 
 class CipherParams:
   algorithm: String default "AES" // TZ2a
-  key: String // TZ2d
-  keyLength: Int? // TZ2b
+  key: Binary // TZ2d
+  keyLength: Int // TZ2b
   mode: String default "CBC" // TZ2c
 
 class Crypto:
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -229,11 +229,11 @@ h3(#rest-encryption). Encryption
 * @(RSE1)@ @Crypto::getDefaultParams@ function:
 ** @(RSE1a)@ Returns a complete @CipherParams@ instance, using the default values for any field not supplied
 ** @(RSE1b)@ Takes a hashmap (or language equivalent) consisting of any subset of @CipherParams@ fields that includes a @key@
-** @(RSE1c)@ The @key@ must be either a binary (e.g. a byte array, depending on the language), or a base64-encoded string. If the key is a string, the function should base64-decode it into a binary. Since the conversion to base64 is not under Ably control, this should be done liberally -- in particular, it should work with 'url-safe base64' (which uses @-@ and @_@ instead of @+@ and @/@) in addition to standard base64
+** @(RSE1c)@ The @key@ must be either a binary (e.g. a byte array, depending on the language), or a base64-encoded string. If the key is a string, the function should base64-decode it into a binary. Since the conversion to base64 is not under Ably control, this should be done leniently -- in particular, it should work with base64url (RFC 4648 s.5, which uses @-@ and @_@ instead of @+@ and @/@) as well as base64 (RFC 4648 s.4)
 ** @(RSE1d)@ Calculates a @keyLength@ from the key (its size in bits).
-** @(RSE1e)@ Checks that the provided options are valid and self-consistent as best it can, raises an exception if not. At a minimum, this should include checking the calculated keyLength is a valid key length for the encryption algorithm (for example, 128 or 256 for @AES@)
+** @(RSE1e)@ Checks that the provided options are valid and self-consistent as best it can, raises an exception if not. At a minimum, this should include checking the calculated @keyLength@ is a valid key length for the encryption algorithm (for example, 128 or 256 for @AES@)
 * @(RSE2)@ @Crypto::generateRandomKey@ function
-** @(RSE2a)@ Takes an optional keyLength parameter, which is the length in bits of the key to be generated. If unspecified, this is equal to the default keyLength of the default algorithm: for @AES@, 256 bits.
+** @(RSE2a)@ Takes an optional @keyLength@ parameter, which is the length in bits of the key to be generated. If unspecified, this is equal to the default @keyLength@ of the default algorithm: for @AES@, 256 bits.
 ** @(RSE2b)@ Returns (or calls back with, if the language cryptographic randomness primitives are blocking or async) the key as a binary (e.g. a byte array, depending on the language)
 
 h2(#realtime). Realtime client library features
@@ -699,9 +699,9 @@ h4. ChannelOptions
 
 h4. CipherParams
 * @(TZ1)@ params to configure encryption for a channel, see "Java CipherParams class":https://github.com/ably/ably-java/blob/8a151b4edfa228ddef806fa10d2f9ce2be1ac09c/lib/src/main/java/io/ably/lib/util/Crypto.java as a reference
-* @(TZ2)@ The attributes of @CipherParams@ consist of anything necessary to implement the supported algorithms (eg iv), in addition to the following standardised attributes:
+* @(TZ2)@ The attributes of @CipherParams@ consist of anything necessary to implement the supported algorithms, in addition to the following standardised attributes:
 ** @(TZ2a)@ @algorithm@ string - Default is @AES@. Optionally specify the algorithm to use for encryption, currently only @AES@ is supported
-** @(TZ2b)@ @keyLength@ integer - the size in bits of the @key@
+** @(TZ2b)@ @keyLength@ integer - the length in bits of the @key@
 ** @(TZ2d)@ @key@ binary - private key used to encrypt and decrypt payloads
 ** @(TZ2c)@ @mode@ string - Default is @CBC@. Optionally specify cipher mode, currently only @CBC@ is supported
 
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -202,7 +202,7 @@ h3(#rest-channel). Channel
 *** @(RSL4d3)@ a JSON Message payload is stringified either as a JSON Object or Array and represented as a JSON string and the @encoding@ attribute is set to "json"
 *** @(RSL4d4)@ All messages received will be decoded based on the @encoding@ field and deliver payloads in the format they were sent in i.e. binary, string, or a structured type containing the parsed JSON
 * @(RSL5)@ Message payload encryption
-** @(RSL5a)@ When a @Channel@ is instanced with the option @encrypted@ true (or no such option and a @cipher@ option present), message payloads will be automatically encrypted when sent to Ably and decrypted when received on this channel. The cipher configuration is set with the @cipher@
+** @(RSL5a)@ When a @Channel@ is instanced with a (non-null) @cipher@ channelOption, message payloads will be automatically encrypted when sent to Ably and decrypted when received on this channel, using the @cipher@ configuration
 ** @(RSL5b)@ AES 256 and 128 CBC encryption must be supported
 ** @(RSL5c)@ Tests must exist that encrypt and decrypt the following fixture data for "AES 128":https://github.com/ably/ably-common/blob/master/test-resources/crypto-data-128.json and "AES 256":https://github.com/ably/ably-common/blob/master/test-resources/crypto-data-256.json to ensure the client library encryption is compatible across libraries
 * @(RSL6)@ Message decoding
@@ -231,7 +231,7 @@ h3(#rest-encryption). Encryption
 ** @(RSE1b)@ Takes a hashmap (or language equivalent) consisting of any subset of @CipherParams@ fields that includes a @key@
 ** @(RSE1c)@ The @key@ must be either a binary (e.g. a byte array, depending on the language), or a base64-encoded string. If the key is a string, the function should base64-decode it into a binary. Since the conversion to base64 is not under Ably control, this should be done liberally -- in particular, it should work with 'url-safe base64' (which uses @-@ and @_@ instead of @+@ and @/@) in addition to standard base64
 ** @(RSE1d)@ Calculates a @keyLength@ from the key (its size in bits).
-** @(RSE1e)@ Checks that the provided options are valid and self-consistent as best it can, raises an exception if not. At a mininum, this should include checking the calculated keyLength is a valid key length for the (provided if provided else default) encryption algorithm (for example, 128 or 256 for @AES@)
+** @(RSE1e)@ Checks that the provided options are valid and self-consistent as best it can, raises an exception if not. At a minimum, this should include checking the calculated keyLength is a valid key length for the encryption algorithm (for example, 128 or 256 for @AES@)
 * @(RSE2)@ @Crypto::generateRandomKey@ function
 ** @(RSE2a)@ Takes an optional keyLength parameter, which is the length in bits of the key to be generated. If unspecified, this is equal to the default keyLength of the default algorithm: for @AES@, 256 bits.
 ** @(RSE2b)@ Returns (or calls back with, if the language cryptographic randomness primitives are blocking or async) the key as a binary (e.g. a byte array, depending on the language)
@@ -692,8 +692,7 @@ h4. AuthOptions
 h4. ChannelOptions
 * @(TB1)@ options provided when instancing a channel, see "Java ChannelOptions":https://github.com/ably/ably-java/blob/master/src/io/ably/types/ChannelOptions.java as a reference
 * @(TB2)@ The attributes of @ChannelOptions@ consist of:
-** @(TB2a)@ @encrypted@ boolean - when true, enables automatic encryption & decryption of all messages. If not present it defaults to true if the @cipher@ attribute is present, false if not.
-** @(TB2b)@ @cipher@ - This is required if @encrypted@ is true. It consists of either:
+** @(TB2b)@ @cipher@, which is either:
 *** @(TB2b1)@ A @CipherParams@ instance, or
 *** @(TB2b2)@ an options hash (or language equivalent) consisting of any subset of @CipherParams@ fields that includes a @key@. In this case, the client library should call `getDefaultParams`, passing it the options hash, to obtain a @CipherParams@ instance
 * @(TB3)@ The client lib may optionally provide an alternative constructor @withCipherKey@ for ChannelOptions that takes a @key@ only. (This must be differentiated from the normal constructor such that it is clear that the value being passed in is a key). (This is intended for languages where requiring a hash map is unidiomatic)
@@ -883,7 +882,6 @@ enum ChannelState:
 class ChannelOptions:
   +withCipherKey(key: Binary)? -> ChannelOptions // TB3
   cipher: (CipherParams | Params)? // RSL5a, TB2b
-  encrypted: Bool? // RSL5a, TB2a
 
 class CipherParams:
   algorithm: String default "AES" // TZ2a
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -202,7 +202,7 @@ h3(#rest-channel). Channel
 *** @(RSL4d3)@ a JSON Message payload is stringified either as a JSON Object or Array and represented as a JSON string and the @encoding@ attribute is set to "json"
 *** @(RSL4d4)@ All messages received will be decoded based on the @encoding@ field and deliver payloads in the format they were sent in i.e. binary, string, or a structured type containing the parsed JSON
 * @(RSL5)@ Message payload encryption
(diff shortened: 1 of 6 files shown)


content/client-lib-development-guide/features.textile
```diff
@@ -682,18 +693,17 @@ h4. ChannelOptions
 * @(TB1)@ options provided when instancing a channel, see "Java ChannelOptions":https://github.com/ably/ably-java/blob/master/src/io/ably/types/ChannelOptions.java as a reference
 * @(TB2)@ The attributes of @ChannelOptions@ consist of:
 ** @(TB2a)@ @encrypted@ boolean - when true, enables automatic encryption & decryption of all messages
-** @(TB2b)@ @cipherParams@ @CipherParams@ - when encrypted is true, the cipher params are required
+** @(TB2b)@ @cipherParams@ - This is required if @encrypted@ is true. It consists of either:
+*** @(TB2b1)@ A @CipherParams@ instance, or
+*** @(TB2b2)@ an options hash (or language equivalent) consisting of the key @key@ with value a key. In this case, the client library should call `getDefaultParams`, passing it the key, to obtain a @CipherParams@ instance
```
tcard: If the only option we're getting from the hash is the key, why not take the key directly? This would make it also easier to conform to for nominally typed languages, in which passing stuff around in hash maps is not idiomatic and breaks type safety (as they call it, "stringly typed programming")

tcard: Also, you may be tempted to pass more stuff in the hashmap (e. g. algorithm, IV), which will be ignored, which can be confusing.

SimonWoolf: I did it that way following [@mattheworiordan's example of his preferred api](https://github.com/ably/ably-js/issues/235#issue-134905326) (presumably because that's the way it currently works in ably-ruby); happy to change to `{encrypted: true, key: ...}` if Matt and Paddy agree

content/client-lib-development-guide/features.textile
```diff
@@ -880,7 +890,8 @@ class CipherParams:
   mode: String default "CBC" // TZ2c
 
 class Crypto:
-  +getDefaultParams(key: String?) -> CipherParams // TZ3
+  +getDefaultParams(key: binary) -> CipherParams // RSE1
```
tcard: `binary` should be uppercase.

content/realtime/encryption.textile
```diff
@@ -88,44 +92,91 @@ h6(#get-default-params).
...
+- keyLength := Optional @Int@ with the length of key to generate. For AES, this should be either 128 or 256. If unspecified, defaults to 256.
 
-- <div lang="jsall">callback</div> := is a function of the form @function(err, cipherParams)@ which is called upon completion
+- <div lang="jsall">callback</div> := is a function of the form @function(err, key)@ which is called upon completion
 
 blang[jsall].
   h4. Callback result
 
-  On successfully creating default cipher params, the callback is called with the default "@CipherParams@":#cipher-params object and @err@ is @null@. On failure to create default cipher params, @err@ contains an "@ErrorInfo@":#error-info object describing the failure reason.
+  On successfully generating a key, the callback is called with that key as a <span lang=javascript>@WordArray@</span><span lang=nodejs>@Buffer@</span>, and @err@ is @null@. On failure to create a key, @err@ contains an "@ErrorInfo@":#error-info object describing the failure reason.
```
tcard: `WordArray` is called `ArrayBuffer` everywhere else.

SimonWoolf: `WordArray` and `ArrayBuffer` are different datatypes. `ArrayBuffer` is a native javascript binary type (but isn't supported by some browsers); `WordArray` is a CryptoJS custom type. `generateRandomKey` returns a WordArray always, as that's better than returning a different thing depending on what browser you're using. `getDefaultParams` can take a key as [either a WordArray or an ArrayBuffer](https://github.com/ably/docs/pull/89/files#diff-80469b1c5e0dd2e8ff3da0c178748ae7R92).

content/client-lib-development-guide/features.textile
mattheworiordan: I will have to defer to @paddybyers on this as I didn't design this API originally. I question whether we need `encrypted` at all, and instead state that if cipher details are provided, then the channel is encrypted.  If that is the case, I think a more concise option could be offered such as simply `channel.get("name", cipherKey: "BASE64encodedstring")`. @paddybyers WDYT?

content/realtime/encryption.textile
mattheworiordan: Is it worth linking WordArray to documentation on that then?

content/client-lib-development-guide/features.textile
paddybyers: I agree that if cipherParams are given then implicitly `encrypted = true`.

I question the value of further simplifying it so that you specify the key only as a channel option, because it doesn't really scale well to passing other non-default cipher params.

SimonWoolf: > I question the value of further simplifying it so that you specify the key only as a channel option, because it doesn't really scale well to passing other non-default cipher params.

Is that something we want to support? I'd understood from the skype call that we only wanted to support changing the `key`, not anything else, and wrote this accordingly - I guess I misundestood.

If we want to support an arbitrary subset of cipherParams, then we need to either:

- have `getDefaultParams` be able to take an arbitrary subset of cipherParams and return a complete cipherParams (or throw an exception), so that eg `{encrypted: true, cipherParams: {key: key, mode: 'cbc'}}` becomes just convenient shorthand for `{encrypted: true, cipherParams: Crypto.getDefaultParams({key: key, mode: 'cbc'})}`; or
- specify that `{encrypted: true, cipherParams: {key: key, mode: 'cbc'}}` means calling `getDefaultParams` with the key only, then selectively overwriting what that returns with any non-`key` properties.

I don't really like option (b), it feels a bit weird/inelegant -- and means that we can no longer validate anything in getDefaultParams (eg getDefaultParams can't assert that the keyLength is a valid one for AES if the algorithm might be changed later). So I'd prefer (a). @paddybyers WDYT?

(Though if we do go for (a), then `getDefaultParams` no longer really describes what it's doing. Maybe `getCipherParams`?)

mattheworiordan: Based on trying to keep the API as simple as possible from a developer's perspective, yet allowing the options to scale later, why don't we simply have:

```javascript
channel.get({ cipher: { key: key }});
```

If the dev wants to add more params they can do any of the following:

```javascript
channel.get({ cipher: { key: key, mode: 'cbc' }});
// or
var cipherParams = Cypto.getDefaultParams({key: key, mode: 'cbc'});
channel.get({ cipher: cipherParams });
```

Calling it `cipherParams` as opposed to `cipher` makes it sound like you're configuring it but not necessarily enabling it on the channel unless `encrypted: true` is included.  And we can still specify that the type expected is a `CipherParams` object.

WDYT?

paddybyers: > Calling it cipherParams as opposed to cipher makes it sound like you're configuring it but not necessarily enabling it on the channel unless encrypted: true is included.

I disagree with your comment but I'm happy to change it to `cipher` if you think that's an improvement.

> And we can still specify that the type expected is a CipherParams object.

Yes, a `Cipher` type would be the thing that implements the cipher itself.

mattheworiordan: > I disagree with your comment but I'm happy to change it to cipher if you think that's an improvement.

I do, at least from the perspective that when reading the code without any knowledge of our API it feels clearer to me. However, perhaps @SimonWoolf and @tcard can add their preference here. So far one vote for each :)

SimonWoolf: Don't have strong a opinion either way. If pushed, `cipher` seems a bit cleaner.

Before revising this, I'd also appreciate any opinions anyone has on the question I asked [here](https://github.com/ably/docs/pull/89/files#r53927707).

paddybyers: > I'd understood from the skype call that we only wanted to support changing the key, not anything else

That's true now, but not forever.

tcard: I agree that the `ChannelOptions.encrypted` is redundant.

> I question the value of further simplifying it so that you specify the key only as a channel option, because it doesn't really scale well to passing other non-default cipher params.

If you need more params, you can pass them as a normal `CipherParams`, or a hashmap in dynamic languages. This would just provide a straightforward way to do the most common thing, which is just taking a key. In JavaScript and friends this gives little advantage, because making a hashmap is very convenient and idiomatic, but e. g. in iOS having to construct a CipherParams would lead to something like:

```objective-c
[channels get:@"foo" options:[[ARTChannelOptions alloc] initWithCipher: [ARTCrypto getDefaultParams:@"key"]]]
```
```java
// Java equivalent.
channels.get("foo", new ChannelOptions(Crypto.getDefaultParams("key")));
```

What I'm proposing is providing this as a convenience:

```objective-c
[channels get:@"foo" options:[ARTChannelOptions optionsWithCipherKey:@"key"]]
```
```java
// Java equivalent.
channels.get("foo", ChannelOptions.withCipherKey("key"));
```

Or even just:

```objective-c
[channels get:@"foo" encryptedWithKey:@"key"];
```
```java
// Java equivalent.
channels.get("foo", "key");
```

> Calling it cipherParams as opposed to cipher makes it sound like you're configuring it but not necessarily enabling it on the channel unless encrypted: true is included. And we can still specify that the type expected is a CipherParams object.

I don't have a strong opinion but I tend to think that just `cipher` sounds better.

mattheworiordan: > Before revising this, I'd also appreciate any opinions anyone has on the question I asked here.

I prefer option 1 that allows arbitrary params to be set when calling `getDefaultParams`.  However, I think an overloaded version that allows you to only provide a key would be useful.

tcard: > I prefer option 1 that allows arbitrary params to be set when calling getDefaultParams. However, I think an overloaded version that allows you to only provide a key would be useful.

That means a bunch of constructors in static languages. Or do you want the constructor to take a hash map even in those languages?

mattheworiordan: > `channels.get("foo", new ChannelOptions(Crypto.getDefaultParams("key")));`

I am not mad on this, but can't quite tell you why.  Perhaps it's simply because it's not obvious why you're passing in that object.

> What I'm proposing is providing this as a convenience
> `channels.get("foo", ChannelOptions.withCipherKey("key"));`

I don't have an issue with that although equally not sure it's really needed. If we did go down that route though, I would prefer to have `ChannelOptions.withCipher("key")` because then it would simply map directly to `getDefaultParams` and also support `ChannelOptions.withCipher({ "key": key, "algorithm": "cbc" })` etc (I know that is not quite valid because they would be `Params[]`, but for simplicity it's written that way).  Also, I am not sure what the conventions would be about naming a constructor method like this i.e. should it be prefixed with the word `new` or `get` perhaps?

paddybyers: > However, I think an overloaded version that allows you to only provide a key would be useful.

> ````
> // Java equivalent.
> channels.get("foo", "key");
> ````

This is just a dead-end because we can never have any other string option that takes precedence over a key, so I veto this. Something with a factory constructor for `ChannelOptions` might work instead.

mattheworiordan: > That means a bunch of constructors in static languages. Or do you want the constructor to take a hash map even in those languages?

Yes, a Hash map or equivalent i.e. we use `Params[]` in Java, a Javascript object in JS and Hash in Ruby

SimonWoolf: OK. Attempting to synthesize all of the above comments:

- getDefaultParams must be able to take an arbitrary subset of CipherParams fields as a hashmap or equivalent that includes a key; will validate that the provided options are valid and self-consistent as best it can, and will return a complete CipherParams instance.
- if requiring a hash map is unidiomatic, getDefaultParams may optionally also take a key only.
- The `cipher` field of channelOptions must consist of either a CipherParams object, or an arbitrary subset of CipherParams fields that includes a key. In the latter case, the channel will pass it through getDefaultParams to generate a CipherParams.
- if requiring a hash map is unidiomatic, the client lib may optionally provide an alternative constructor for channelOptions that takes a key only. If it does this, it must be differentiated from the normal constructor such that it is clear that the value being passed in is a key.

Everyone's happy with that? If so I'll add it to the PR

paddybyers: > if requiring a hash map is unidiomatic, getDefaultParams may optionally also take a key only.

I still think this is a dead end, but less of a dead end as having the same thing on `channels.get()`.

mattheworiordan: > I still think this is a dead end, but less of a dead end as having the same thing on channels.get().

I think it's OK given 99.9% of the time devs calling `getDefaultParams()` will only provide a key anyway.  But happy to not include it if you dislike it that much. Thanks @SimonWoolf

tcard: I really can't say I like `Param[]`, for the reasons above. `getDefaultParams` is supposed to be a convenience method for the most usual case; if you are doing something more special than the default, you still can make your own `CipherParams`. I don't think we should complicate `getDefaultParams` for that case. (Plus, as @SimonWoolf notes above, it wouldn't really be "default" anymore.)

I'd rather restrict the hashmap convenience to dynamic languages and, as @SimonWoolf says, take just the key in those, or make your own `CipherParams`.

SimonWoolf: > `getDefaultParams` is supposed to be a convenience method for the most usual case

@tcard In the proposal above, almost no-one will *have* to call `getDefaultParams` directly, since almost everything that can be done with that can also be done by supplying the things straight to ChannelOptions and letting it call `getDefaultParams` -- the proposal allows your suggestion of alternative constructor for that as a convenience for the common case (eg `ChannelOptions.withCipher("key")`, fourth bullet point), so there's less need for `getDefaultParams` to be convenient.

content/realtime/encryption.textile
SimonWoolf: > Is it worth linking WordArray to documentation on that then?

I would, but CryptoJS don't seem to be big fans of documentation...
People shouldn't be doing much with the output except feeding it into `getDefaultParams`, so it's probably not a big issue

SimonWoolf:
Added https://github.com/ably/docs/commit/43b5d4e555c8768dd8431928b1442e9f157069cb with the latest feedback

SimonWoolf:
Remove `encrypted` boolean altogether: e0c2908

content/client-lib-development-guide/features.textile
```diff
@@ -681,19 +692,18 @@ h4. AuthOptions
...
 * @(TZ1)@ params to configure encryption for a channel, see "Java CipherParams class":https://github.com/ably/ably-java/blob/8a151b4edfa228ddef806fa10d2f9ce2be1ac09c/lib/src/main/java/io/ably/lib/util/Crypto.java as a reference
-* @(TZ2)@ The attributes of @CipherParams@ consist of anything necessary to implement the supported algorithms (eg key and iv), in addition to the following standardised attributes:
+* @(TZ2)@ The attributes of @CipherParams@ consist of anything necessary to implement the supported algorithms (eg iv), in addition to the following standardised attributes:
 ** @(TZ2a)@ @algorithm@ string - Default is @AES@. Optionally specify the algorithm to use for encryption, currently only @AES@ is supported
-** @(TZ2b)@ @keyLength@ integer - An integer, from an algorithm-dependent range of possible values. If unspecified, an algorithm-dependent default is used. For AES the allowed values at least include 128 and 256, with a default of 128
-** @(TZ2d)@ @key@ string / byte array - private key used to encrypt and decrypt payloads
+** @(TZ2b)@ @keyLength@ integer - the size in bits of the @key@
+** @(TZ2d)@ @key@ binary - private key used to encrypt and decrypt payloads
```
mattheworiordan: I thought key could be a binary or base664 encoded string in line with https://github.com/ably/docs/pull/89/files#diff-11900b395df266bc2bbfc1d11bdfc7a0R232. Is that meant to not be the case here?

mattheworiordan: Or are you saying that the key attribute is always binary, but when constructing it, it can be binary or base64 as above?

paddybyers: Yes, I think there should be the ability to construct a `CipherParams` with a base64 string for the `key`, but this is decoded and the `key` attribute should be the raw key.

tcard:
LGTM

SimonWoolf:
Reference implementation of this for ably-js: https://github.com/ably/ably-js/pull/240

content/client-lib-development-guide/features.textile
```diff
@@ -225,6 +225,17 @@ h3(#rest-presence). Presence
 *** @(RSP4b3)@ @limit@ supports up to 1,000 items; if unspecified defaults to the REST API default (100)
 * @(RSP5)@ Presence Messages retrieved are decoded in the same way that messages are decoded
 
+h3(#rest-encryption). Encryption
+* @(RSE1)@ @Crypto::getDefaultParams@ function:
+** @(RSE1a)@ Returns a complete @CipherParams@ instance, using the default values for any field not supplied
+** @(RSE1b)@ Takes a hashmap (or language equivalent) consisting of any subset of @CipherParams@ fields that includes a @key@
+** @(RSE1c)@ The @key@ must be either a binary (e.g. a byte array, depending on the language), or a base64-encoded string. If the key is a string, the function should base64-decode it into a binary. Since the conversion to base64 is not under Ably control, this should be done liberally -- in particular, it should work with 'url-safe base64' (which uses @-@ and @_@ instead of @+@ and @/@) in addition to standard base64
```
paddybyers: Please reference the RFCs for url-safe and "standard" base64.

I think "leniently" is more appropriate than "liberally".

content/client-lib-development-guide/features.textile
```diff
@@ -225,6 +225,17 @@ h3(#rest-presence). Presence
...
+** @(RSE1d)@ Calculates a @keyLength@ from the key (its size in bits).
+** @(RSE1e)@ Checks that the provided options are valid and self-consistent as best it can, raises an exception if not. At a minimum, this should include checking the calculated keyLength is a valid key length for the encryption algorithm (for example, 128 or 256 for @AES@)
+* @(RSE2)@ @Crypto::generateRandomKey@ function
+** @(RSE2a)@ Takes an optional keyLength parameter, which is the length in bits of the key to be generated. If unspecified, this is equal to the default keyLength of the default algorithm: for @AES@, 256 bits.
```
paddybyers: `@`-quote `keyLength` ?

content/client-lib-development-guide/features.textile
```diff
@@ -681,19 +692,18 @@ h4. AuthOptions
...
 
 h4. CipherParams
 * @(TZ1)@ params to configure encryption for a channel, see "Java CipherParams class":https://github.com/ably/ably-java/blob/8a151b4edfa228ddef806fa10d2f9ce2be1ac09c/lib/src/main/java/io/ably/lib/util/Crypto.java as a reference
-* @(TZ2)@ The attributes of @CipherParams@ consist of anything necessary to implement the supported algorithms (eg key and iv), in addition to the following standardised attributes:
+* @(TZ2)@ The attributes of @CipherParams@ consist of anything necessary to implement the supported algorithms (eg iv), in addition to the following standardised attributes:
 ** @(TZ2a)@ @algorithm@ string - Default is @AES@. Optionally specify the algorithm to use for encryption, currently only @AES@ is supported
-** @(TZ2b)@ @keyLength@ integer - An integer, from an algorithm-dependent range of possible values. If unspecified, an algorithm-dependent default is used. For AES the allowed values at least include 128 and 256, with a default of 128
-** @(TZ2d)@ @key@ string / byte array - private key used to encrypt and decrypt payloads
+** @(TZ2b)@ @keyLength@ integer - the size in bits of the @key@
```
paddybyers: "length" not size

content/realtime/encryption.textile
```diff
@@ -52,6 +50,8 @@ h2(#getting-started). Getting started
   channel.publish("example", "message data");
 ```
 
+Note that the @key@ should not be a password, but a cryptographic key - randomly generated, 128 or 256 bits long, binary or base64-encoded. If you wish to encrypt messages with a password (for example, one inputted by a user), you should use a "key derivation function":https://en.wikipedia.org/wiki/Key_derivation_function to transform that into a key. The client libraries are also capable of "generating a random key":#generate-random-key for you.
```
paddybyers: I suggest "passphrase" instead of "password".

Instead of just "randomly generated" I think we should say "from a secure random source".

"inputted" - is that a word? Maybe "entered" ?

paddybyers:

SimonWoolf:
Final fixes from Paddy's feedback: 7de057191af1951748cec779a7b3de69e838e818

I now have a :+1: from everyone, so squashed down and merged as 9d30ab9

PR #92, opened 2016-02-25
IDL: Add ChannelEvent enum with ERROR variant.

tcard:
This is a requirement of RTL2c that was overlooked in the IDL.

See https://github.com/ably/ably-ios/pull/253#discussion-diff-54078011.

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -870,12 +870,7 @@ enum ChannelState:
   FAILED
 
 enum ChannelEvent:
-  INITIALIZED
-  ATTACHING
-  ATTACHED
-  DETACHING
-  DETACHED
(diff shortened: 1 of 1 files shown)


content/client-lib-development-guide/features.textile
```diff
@@ -869,6 +869,15 @@ enum ChannelState:
   DETACHED
   FAILED
 
+enum ChannelEvent:
+  INITIALIZED
```
mattheworiordan: Instead of duplication, can we not simply state that it inherits / extends ChannelState?

tcard: Done at ca5ce40.

mattheworiordan:
I am happy with this, however @paddybyers are you?

PR #94, opened 2016-03-02
Add Obj-C and Swift to Quickstart Guide.

tcard:


content/root/quick-start-guide.textile
```diff
@@ -14,6 +14,28 @@ When using Ably in a browser, simply include the Ably Javascript library within
...
+  ```
+
+blang[swift].
+  Add this line to your application's Podfile:
+
+  bc. pod 'Ably', '~> 0.8'
+
+  And install with @pod install@. Then in your files: 
+
+  ```[objc]
```
mattheworiordan: isn't this supposed to `swift`?

mattheworiordan: Also, if you read the text at https://github.com/ably/docs/pull/94/files#diff-f9d1ed36577e92a45ffb881113c2205fL13, you'll see that you need to perhaps change that copy to be included in all languages other than the ones you've suggested

content/root/quick-start-guide.textile
```diff
@@ -41,10 +63,15 @@ ably.connection.on(:connected) do
 end
 
 bc[swift]. let ably = ARTRealtime(key: "{{API_KEY}}")
-ably.connection.on(:connected) do
-  print("That was simple, you're now connected to Ably in realtime")
+ably.connection.on(.Connected) {
+    print("That was simple, you're now connected to Ably in realtime")
```
mattheworiordan: Thanks, I had written that before, so must have lost it in a copy & paste, thanks for picking that up

mattheworiordan:
One or two comments, but LGTM

content/root/quick-start-guide.textile
tcard: > isn't this supposed to swift?

Fixed at 2661702.

> you'll see that you need to perhaps change that copy to be included in all languages other than the ones you've suggested

I don't know how this should be done for all languages. I realize that the browser copy should be only for JavaScript, but shouldn't that be in a different pull request?

mattheworiordan:
@tcard I have updated the PR, tell me if you're happy with the changes

tcard:
LGTM, one small comment.

PR #96, opened 2016-03-04
PHP documentation

mattheworiordan:
Add supports for REST PHP library to the documentation.

To be reviewed by Matus who worked on the PHP lib.

Closing for now as merging and releasing, but comments should be added here and will be fixed in latter commits.


bladeSk:
I think that's it :) Possibly a few more dots instead of arrows and missing dollars.

PR #97, opened 2016-03-04
Change iOS examples to new API.

tcard:


mattheworiordan:
Thanks @tcard all merged in now

PR #109, opened 2016-04-13
Host fallback rest domain

ricardopereira:


tcard:
I don't get this, what's wrong with `[a-e].ably-realtime.com` even for REST? Plus, ably-rest.com doesn't seem to exist...

mattheworiordan:
This is wrong, we don't have specific REST host fallbacks, those fallbacks support REST & Realtime.

PR #112, opened 2016-04-25
Exhaustive state conditions for all realtime operations.

tcard:
When calling `Connection.close()` before reaching the CONNECTED state,
the iOS library was trying to send a WebSocket message before the
handshake was even finished. That caused it to crash sometimes.

That situation wasn't contemplated on the spec. I've gone through it
filling a table and filling the gaps, so that the spec is exhaustive
about all possible combinations of state and operations.

(So, yes, https://github.com/zwopple/PocketSocket/pull/48 wasn't fixing
the right problem.)


tcard:
This is what it looks like, which may not be obvious from the source:

captura de pantalla 2016-04-25 a les 19 41 49

mattheworiordan:
Thanks @tcard, this is really good.  I will double check all the links soon, merge & update the unified spec sheet

content/client-lib-development-guide/features.textile
```diff
@@ -301,12 +302,16 @@ h3(#realtime-connection). Connection
 ** @(RTN11a)@ Explicitly connects to the Ably service if not already connected
 ** @(RTN11b)@ An error will be indicated if the state is @CLOSING@ as the connection must complete the close request before reconnecting
 * @(RTN12)@ @Connection#close@ function:
-** @(RTN12a)@ Sends a @CLOSE@ @ProtocolMessage@ to the server, sets the state to @CLOSING@ and waits for a @CLOSED@ @ProtocolMessage@ to be received
+** @(RTN12f)@ If the connection state is @CONNECTING@, do the operation once the connection state is @CONNECTED@
```
mattheworiordan: Why, why not simply close the connection and move the connection to `CLOSED`?

content/client-lib-development-guide/features.textile
```diff
@@ -301,12 +302,16 @@ h3(#realtime-connection). Connection
...
+** @(RTN12a)@ If the connection state is @CONNECTED@, sends a @CLOSE@ @ProtocolMessage@ to the server, sets the state to @CLOSING@ and waits for a @CLOSED@ @ProtocolMessage@ to be received
 ** @(RTN12b)@ If the @CLOSED@ @ProtocolMessage@ is not received within the "default realtime request timeout":#defaults, the transport will be disconnected and the connection will automatically move to the @CLOSED@ state
 ** @(RTN12c)@ If the transport is abruptly closed following a @CLOSE@ @ProtocolMessage@ being sent, then the connection will automatically move to the @CLOSED@ state
+** @(RTN12d)@ If the connection state is @DISCONNECTED@, aborts the retry process described in "RTN14d":#RTN14d and moves the connection immediately to the @CLOSED@ state
+** @(RTN12e)@ If the connection state is @SUSPENDED@, aborts the retry process described in "RTN14e":#RTN14e and moves the connection immediately to the @CLOSED@ state
```
mattheworiordan: Could we make RTN12d and RTN12e a single spec i.e. `if the connection state is @DISCONNETED@ or @SUSPENDED@..` as they are both pretty much the same?

content/client-lib-development-guide/features.textile
```diff
@@ -301,12 +302,16 @@ h3(#realtime-connection). Connection
...
 * @(RTN13)@ @Connection#ping@ function:
+** @(RTN13c)@ If the connection state is @CONNECTING@ or @DISCONNECTED@, do the operation once the connection state is @CONNECTED@
```
mattheworiordan: Can we change the wording to `If the connection state is @CONNECTING@ or @DISCONNECTED@, perform the ping operation once the connection state is @CONNECTED@`

content/client-lib-development-guide/features.textile
```diff
@@ -378,14 +383,19 @@ h3(#realtime-channel). Channel
 * @(RTL11)@ If a channel enters the @DETACHED@ or @FAILED@ state, then all messages that are still queued for send on that channel should be deleted from the queue triggering a failure for the publish or presence methods invoked for those messages
 * @(RTL4)@ @Channel#attach@ function:
 ** @(RTL4a)@ If already @ATTACHED@ or @ATTACHING@ nothing is done
+** @(RTL4e)@ If the channel is in the @DETACHING@ state, it will result in an error, as the channel must complete the detach request before reattaching
```
mattheworiordan: I am not sure this is a good idea as it makes it a bit tricky for a developer as they need to be sure the detach is complete. Why not simply: `If the channel is in the @DETACHING@ state, the attach operation is executed once the channel enters the @DETACHED@ state, or results in an error if it moves to the @FAILED@ state`

content/client-lib-development-guide/features.textile
```diff
@@ -378,14 +383,19 @@ h3(#realtime-channel). Channel
...
 ** @(RTL4f)@ Once an @ATTACH@ @ProtocolMessage@ is sent, if an @ATTACHED@ @ProtocolMessage@ is not received within the "default realtime request timeout":#defaults, the attach request should be treated as though it has failed
 ** @(RTL4d)@ If the language permits, a callback can be provided that is called when the channel is attached successfully or the attach fails and the @ErrorInfo@ error is passed as an argument to the callback
 ** @(RTL4e)@ If the user does not have sufficient permissions to attach to the channel, the channel state @FAILED@ is emitted with the error @ErrorInfo@. An error event is also emitted, and the @Channel#errorReason@ attribute of the @Channel@ is set to the error @ErrorInfo@ object
 * @(RTL5)@ @Channel#detach@ function:
 ** @(RTL5a)@ If state is @INITIALIZED@, @DETACHED@ or @DETACHING@ nothing is done
 ** @(RTL5b)@ If state is @FAILED@, the @detach@ request results in an error
+** @(RTL5g)@ If the connection state is @CLOSING@ or @FAILED@, the @detach@ request results in an error
```
mattheworiordan: Did yo miss @CLOSED@ from that list?

content/client-lib-development-guide/features.textile
```diff
@@ -397,8 +407,9 @@ h3(#realtime-channel). Channel
...
 ** @(RTL6i3)@ Allows @name@ and or @data@ to be @null@.  If any of the values are null, then key is not sent to Ably i.e. a payload with a @null@ value for @data@ would be sent as follows @{ "name": "click" }@
 ** @(RTL6c)@ Connection state conditions:
-*** @(RTL6c1)@ If the connection is @CONNECTED@ then the messages are published immediately
-*** @(RTL6c2)@ If the connection is @CONNECTING@ or @DISCONNECTED@, and @ClientOptions#queueMessages@ has not been explicitly set to false, then the message will be queued and delivered as soon as the connection state returns to @CONNECTED@
+*** @(RTL6c1)@ If the connection is @CONNECTED@ and the channel is @ATTACHED@ then the messages are published immediately
+*** @(RTL6c2)@ If the connection is @INITIALIZED@, @CONNECTING@ or @DISCONNECTED@ or the channel is @ATTACHING@, and @ClientOptions#queueMessages@ has not been explicitly set to false, then the message will be queued and delivered as soon as the connection state returns to @CONNECTED@
```
mattheworiordan: Slight correction, "....  then the message will be queued and delivered as soon as the connection state returns to @CONNECTED@ and the channel is @ATTACHED@"

content/client-lib-development-guide/features.textile
```diff
@@ -441,8 +452,8 @@ h3(#realtime-presence). Presence
 ** @(RTP5a)@ If the channel enters the @DETACHED@ or @FAILED@ state then all queued presence messages will fail immediately
 ** @(RTP5b)@ If a channel enters the @ATTACHED@ state then all queued presence messages will be sent immediately and a presence @SYNC@ will be initiated implicitly
 * @(RTP16)@ Connection state conditions:
-** @(RTP16a)@ If the connection is @CONNECTED@ then all presence messages are published immediately
-** @(RTP16b)@ If the connection is @CONNECTING@ or @DISCONNECTED@, and @ClientOptions#queueMessages@ has not been explicitly set to false, then all presence messages will be queued and delivered as soon as the connection state returns to @CONNECTED@
+** @(RTP16a)@ If the connection is @CONNECTED@ and the channel is @ATTACHED@ then all presence messages are published immediately
+** @(RTP16b)@ If the connection is @INITIALIZED@, @CONNECTING@ or @DISCONNECTED@ or the channel is @ATTACHING@, and @ClientOptions#queueMessages@ has not been explicitly set to false, then all presence messages will be queued and delivered as soon as the connection state returns to @CONNECTED@
```
mattheworiordan: As above, you are missing the comment about the channel needing to be `@ATTACHED@`

content/client-lib-development-guide/features.textile
```diff
@@ -508,6 +519,268 @@ h3(#eventemitter). EventEmitter mixin / interface
...
+  <td>@connect@</td>
+
+  <!-- When INITIALIZED -->
+  <td>"RTN11a":#RTN11a</td>
+  <!-- When CONNECTING -->
+  <td>No-op</td>
+  <!-- When CONNECTED -->
+  <td>No-op</td>
+  <!-- When DISCONNECTED -->
+  <td>No-op</td>
```
mattheworiordan: This is not correct, if a customer explicitly requests a connect then we attempt to reconnect immediately.  Same applies for the suspended state

content/client-lib-development-guide/features.textile
```diff
@@ -508,6 +519,268 @@ h3(#eventemitter). EventEmitter mixin / interface
...
+  <!-- When SUSPENDED -->
+  <td>No-op</td>
+  <!-- When CLOSING -->
+  <td>"RTN11b":#RTN11b</td>
+  <!-- When CLOSED -->
+  <td>"RTN11a":#RTN11a</td>
+  <!-- When FAILED -->
+  <td>No-op</td>
```
mattheworiordan: I am not sure why the client library should not reconnect if the user explicitly requests this when in the failed state. @paddybyers?

content/client-lib-development-guide/features.textile
```diff
@@ -508,6 +519,268 @@ h3(#eventemitter). EventEmitter mixin / interface
 * @(RTE5)@ @EventEmitter#off@ deregisters a listener. If called with a specific event and a listener, it removes all registrations that match both the given listener and the given event; if called only with a listener, it removes all registrations matching the given listener, regardless of whether they are associated with an event or not; if called with no arguments, it removes all registrations, for all events and listeners
 * @(RTE6)@ @EventEmitter#emit@ emits an event, calling registered listeners with the given event name and any other given arguments. If an exception is raised in any of the listeners, the exception is caught by the @EventEmitter@ and the exception is logged to the Ably logger. Tests must exist to ensure exceptions raised in client code do not propagate and inhibit other event processing within the client library
 
+h2(#state-conditions-and-operations). State conditions and operations
+
+@Connection@ operation effects depending on current @Connnection.state@:
```
mattheworiordan: Should this not be: `@Connection.state@ effect on realtime operations:` as these are not just connection operations but channel and presence operations too?

content/client-lib-development-guide/features.textile
```diff
@@ -508,6 +519,268 @@ h3(#eventemitter). EventEmitter mixin / interface
...
+
+<tr>
+  <td>Channel @attach@</td>
+
+  <!-- When INITIALIZED -->
+  <td>"RTL4h":#RTL4h</td>
+  <!-- When CONNECTING -->
+  <td>"RTL4h":#RTL4h</td>
+  <!-- When CONNECTED -->
+  <td>"See channel table":#channel-states-operations</td>
```
mattheworiordan: I thought the links were broken for a while, could we add titles for both tables so that this makes sense i.e. `Connection states` and `Channel states` perhaps and change this to "See channel states table"?

content/client-lib-development-guide/features.textile
```diff
@@ -508,6 +519,268 @@ h3(#eventemitter). EventEmitter mixin / interface
...
+  <!-- When INITIALIZED -->
+  <td>"RTL5h":#RTL5h</td>
+  <!-- When CONNECTING -->
+  <td>"RTL5h":#RTL5h</td>
+  <!-- When CONNECTED -->
+  <td>"See channel table":#channel-states-operations</td>
+  <!-- When DISCONNECTED -->
+  <td>"RTL5h":#RTL5h</td>
+  <!-- When SUSPENDED -->
+  <td>"See channel table":#channel-states-operations</td>
```
mattheworiordan: I am not sure this is right, if the connection is suspended we cannot detach a channel.  I think we should fail this request, @paddybyers?

content/client-lib-development-guide/features.textile
```diff
@@ -508,6 +519,268 @@ h3(#eventemitter). EventEmitter mixin / interface
...
+  <!-- When CLOSING -->
+  <td>"RTL5g":#RTL5g</td>
+  <!-- When CLOSED -->
+  <td>"See channel table":#channel-states-operations</td>
```
mattheworiordan: I think this is just `RTL5g`

content/client-lib-development-guide/features.textile
```diff
@@ -508,6 +519,268 @@ h3(#eventemitter). EventEmitter mixin / interface
...
+  <th>Detaching</th>
+  <th>Detached</th>
+  <th>Failed</th>
+</tr>
+
+<tr>
+  <td>@attach@</td>
+
+  <!-- When INITIALIZED -->
+  <td>"RTL4a":#RTL4a</td>
```
mattheworiordan: Is this not meant to be RTL4c?

mattheworiordan:
I have added a few comments, great bit of work as these were definitely areas not covered previously.

@paddybyers I would suggest you take a look to be sure.
@SimonWoolf if you can cast your eye too to spot any issues that would be good

content/client-lib-development-guide/features.textile
paddybyers: If the client simply transitions to `CLOSED`, then the realtime system will never get a close event, and then the connection state will remain for 2 minutes. It is preferable that, once a connection has been initiated, it is allowed to connect and is then closed.

It would be acceptable if there was a shorter timeout waiting for the `CONNECTED` state (eg the client waits for 5s) because we would still catch the majority of cases. However, I'm not sure it is worth the complexity of that special case - so if `CONNECTING`, wait for either a `CONNECTED` (then doing an explicit `CLOSE`), or any other state, in which case the client transitions to `CLOSED` without further interaction.

content/client-lib-development-guide/features.textile
paddybyers: I'm not sure either of these is necessary. The `ATTACH` message can be sent before receiving a `DETACHED`.

The client state is:

- set to `ATTACHED`, `DETACHED`, or `FAILED` on receipt of the corresponding message;
- set to `ATTACHING` or `DETACHING` after sending the corresponding request.

An error is definitely bad because the library should be attempting always to relieve the developer of explicitly managing the asynchronous nature of connection and channel state transitions (eg by queueing messages).

content/client-lib-development-guide/features.textile
paddybyers: Correct; the `CLOSED` and `FAILED` states are "terminal" only on the sense that there are never autonomous transitions from those states. If there is an explicit `connect()` by the client, then a new connection will be attempted, even if from the `FAILED` state. The connection might have been forcibly closed because the account is blocked, for example, but the reconnection attempt can succeed one the block is removed.

content/client-lib-development-guide/features.textile
paddybyers: Yes. When `SUSPENDED` the library behaves as it there is an indefinite delay before any new connection; nothing is queued or delayed for the next successful connection.

content/client-lib-development-guide/features.textile
mattheworiordan: Yes, that seems right

tcard: I agree, but as this is akin to calling `connect()` while `CLOSING` the connection, I think we should also change [RTN11b](http://docs.ably.io/client-lib-development-guide/features/#RTN11b). Does that sound right?

content/client-lib-development-guide/features.textile
tcard: Done at 1b08e36.

content/client-lib-development-guide/features.textile
tcard: You could say that but, as the connection being CLOSED necessarily implies that the channel is DETACHED per [RTN18b](http://docs.ably.io/client-lib-development-guide/features/#RTN18b), this case is covered by RTL5a two lines above. The table also reflects that.

content/client-lib-development-guide/features.textile
tcard: Done at 5a5f1dd.

content/client-lib-development-guide/features.textile
tcard: Done at 0675514.

content/client-lib-development-guide/features.textile
tcard: If the connection is SUSPENDED, the channel is DETACHED (RTL3b) or INITIALIZED, and then RTL5a says that nothing is done. I think this is more coherent that failing the request.

content/client-lib-development-guide/features.textile
tcard: Again, if the connection CLOSED the the channel is DETACHED, which is covered by RTL5a (nothing is done).

In general, if a connection state necessarily implies a channel state, we should specify the behavior for the channel state, in order not to fall into conflicting requirements.

content/client-lib-development-guide/features.textile
paddybyers: > I think we should also change RTN11b. Does that sound right?

Yes. From the point of view of the realtime system this is an even simpler case, because the closing connection and the new connection are completely independent. So the only issue is in managing the state of the client - you can `connect()` while the outgoing connection is `CLOSING`, but when you later receive the `CLOSED` you have to make sure that you don't indicate the new connection as `CLOSED` - the library has to check whether the `CLOSED` message applies to the new or the old connection.

content/client-lib-development-guide/features.textile
tcard: True. Fixed at b9a3dfd.

content/client-lib-development-guide/features.textile
tcard: @paddybyers Does the server allow to reuse the same WebSocket connection for two different (non-overlapping) Ably connections?

tcard: Addressed at aa9675e, will fix again if the above isn't true.

content/client-lib-development-guide/features.textile
tcard: I guess we should also immediately move the FAILED channels to INITIALIZED, remove their `errorReason`s, and also remove the connection's `errorReason`, right?

tcard: Addressed at b704de5.

content/client-lib-development-guide/features.textile
tcard: Addressed at b704de5.

content/client-lib-development-guide/features.textile
paddybyers: > we should also immediately

when? On a new `connect()` ? Yes

content/client-lib-development-guide/features.textile
paddybyers: > Does the server allow to reuse the same WebSocket connection for two different (non-overlapping) Ably connections?

No. A transport is uniquely associated with a single connection.

tcard: OK, so we should just instantiate a new transport and ignore the old one until it's closed and garbage-collected.

tcard: 26f7ded

content/client-lib-development-guide/features.textile
tcard: Done at eb8c4c7.

content/client-lib-development-guide/features.textile
tcard: Done at eb8c4c7.

content/client-lib-development-guide/features.textile
```diff
@@ -299,14 +300,19 @@ h3(#realtime-connection). Connection
...
 ** @(RTN11a)@ Explicitly connects to the Ably service if not already connected
-** @(RTN11b)@ An error will be indicated if the state is @CLOSING@ as the connection must complete the close request before reconnecting
+** @(RTN11b)@ If the state is @CLOSING@, the client should make a new connection with a new transport instance and remove all references to the old one. In particular, it should make sure that, when the @CLOSED@ @ProtocolMessage@ arrives for the old connection, it doesn't affect the new one.
+** @(RTN11c)@ If the state is @DISCONNECTED@ or @SUSPENDED@, aborts the retry process described in "RTN14d":#RTN14d and "RTN14e":#RTN14e and immediately tries to reconnect.
+** @(RTN11d)@ If the state is @FAILED@, moves all the channels to @INITIALIZED@, sets their @errorReason@ to null, and sets the connection's @errorReason@ to null.
 * @(RTN12)@ @Connection#close@ function:
-** @(RTN12a)@ Sends a @CLOSE@ @ProtocolMessage@ to the server, sets the state to @CLOSING@ and waits for a @CLOSED@ @ProtocolMessage@ to be received
+** @(RTN12f)@ If the connection state is @CONNECTING@, do the operation once the connection state is @CONNECTED@
```
SimonWoolf: >> Why, why not simply close the connection and move the connection to `CLOSED`?  So I propose `If the connection state is @CONNECTING@, close all transports and immediately transition the state to @CLOSED@`

> If the client simply transitions to `CLOSED`, then the realtime system will never get a close event, and then the connection state will remain for 2 minutes. It is preferable that, once a connection has been initiated, it is allowed to connect and is then closed.

> It would be acceptable if there was a shorter timeout waiting for the `CONNECTED` state (eg the client waits for 5s) because we would still catch the majority of cases. However, I'm not sure it is worth the complexity of that special case - so if `CONNECTING`, wait for either a `CONNECTED` (then doing an explicit `CLOSE`), or any other state, in which case the client transitions to `CLOSED` without further interaction.

I see the point, but the result is still quite weird. Presumably, a lot of the time if someone's trying to close while the client is CONNECTING, it's because the client has network issues and the client never had a chance of getting through to realtime at all. In which case the only effect of this would be to make close() appear to have no effect for the 10s it takes for realtimeRequestTimeout to expire, which is a bit rubbish. 

(Or if the client's network connection fails after doing a /connect but before it receives the CONNECTED - in which case waiting for the realtimeRequestTimeout to expire still doesn't help anything, as realtime keeping the connection state for 2 minutes is inevitable).

So while this does improve behaviour in the case of a slow-but-reliable network (ie where there's a larger chance of the close() happening between the connect request but before the CONNECTED), it's at the expense of making it seem unresponsive when the network is failing/unreliable. :/

content/client-lib-development-guide/features.textile
```diff
@@ -397,8 +408,9 @@ h3(#realtime-channel). Channel
...
 ** @(RTL6i3)@ Allows @name@ and or @data@ to be @null@.  If any of the values are null, then key is not sent to Ably i.e. a payload with a @null@ value for @data@ would be sent as follows @{ "name": "click" }@
 ** @(RTL6c)@ Connection state conditions:
-*** @(RTL6c1)@ If the connection is @CONNECTED@ then the messages are published immediately
-*** @(RTL6c2)@ If the connection is @CONNECTING@ or @DISCONNECTED@, and @ClientOptions#queueMessages@ has not been explicitly set to false, then the message will be queued and delivered as soon as the connection state returns to @CONNECTED@
+*** @(RTL6c1)@ If the connection is @CONNECTED@ and the channel is @ATTACHED@ then the messages are published immediately
+*** @(RTL6c2)@ If the connection is @INITIALIZED@, @CONNECTING@ or @DISCONNECTED@ or the channel is @ATTACHING@, and @ClientOptions#queueMessages@ has not been explicitly set to false, then the message will be queued and delivered as soon as the connection state returns to @CONNECTED@ and the channek is @ATTACHED@
```
SimonWoolf: s/channek/channel/

content/client-lib-development-guide/features.textile
SimonWoolf: (Also, even if we do go with for waiting-for-connected, need wording for if the first state change from CONNECTING is not CONNECTED. eg `on the first state change from @CONNECTING@: if @CONNECTED@ do the operation per RTN12a; else if @DISCONNECTED@ or @SUSPENDED@ set the state to @CLOSED@; else if @FAILED@ do nothing`.
Or more elegantly, maybe `If the connection state is @CONNECTING@, wait for the first state change, then follow the corresponding RTN12 spec for the new state`)

paddybyers: > So while this does improve behaviour in the case of a slow-but-reliable network (ie where there's a larger chance of the close() happening between the connect request but before the CONNECTED), it's at the expense of making it seem unresponsive when the network is failing/unreliable. :/

Well, yes, but if they don't care about closing connections cleanly then they would never need to call `close()`; they just exit. So if they are calling `close()` then they are making a good faith attempt to release the resources associated with the connection, so why not do that?

At the expense of further complexity we might say that the wait only occurs if it is a `resume` and not a new connection?

I did suggest the possibility of a shorter timeout

mattheworiordan: I don't think it's complex to state that calling `#close`, if in the `connecting` state, has a timer for 5s then forcibly moves to the `closed` state. However, would this then apply in the `disconnected` state as well?

paddybyers: > I don't think it's complex to state that calling #close, if in the connecting state, has a timer for 5s then forcibly moves to the closed state. However, would this then apply in the disconnected state as well?

The timeout for `close()` could be 5s unconditionally.

tcard: I think we should keep CONNECTING and CONNECTED as semantically close as possible. It simplifies everything. Assuming the caller means something different calling `close` while CONNECTING than while CONNECTED is reading too much, I think. We provide all this queuing before reaching CONNECTED so that operations (publishes, presence actions, etc.) called while INITIALIZED/CONNECTING transparently happen once CONNECTED and I don't think we should depart from that behavior here.

We should tell the developer that, if she's having connection problems and wants to handle it herself instead of letting Ably's library do its reconnection/recovering thing, just call `close()` to clean up on Ably's side if possible, forget about that instance and make a new one.

SimonWoolf: I'd be against adding another timer (we've had bugs in ably-js before from various timers interacting in unforeseen ways, and annoyingness-to-debug seems to be exponential with number of timers going on at once. Also 5s isn't *that* much less than 10s, which is the current realtimeRequestTimeout)

So, OK, one more suggestion: how about we follow @tcard's proposal, with the small modification that calling close() while CONNECTING will immediately put you in the CLOSING state (but will not touch the transports). Then, if a transport connects within the realtimeRequestTimeout, send a CLOSE on it, and await the CLOSED reply (and do not change state). If not, then go into the CLOSED state upon expiration of that timeout.

That allows realtime to release resources, but avoids the weirdness of staying in a CONNECTING state for a while after someone calls close(), and avoids having to go from CONNECTING to CLOSING via an instant of CONNECTED. And from the user's perspective, it makes CONNECTING behave the same as CONNECTED as calling close on each will immediately put you in CLOSING.

tcard: I don't really see that much of a problem with what you call "weirdness"; to me, it makes for a simpler state machine, which is easier to reason about. But I'm OK with what you say if you think that's more sensible.

content/client-lib-development-guide/features.textile
tcard: Fixed at 248195bffd81c905ba629773877950ef076d768b.

content/client-lib-development-guide/features.textile
mattheworiordan: Ah, yes, correct

content/client-lib-development-guide/features.textile
mattheworiordan: @tcard I agree with you

content/client-lib-development-guide/features.textile
mattheworiordan: ok

tcard:
Should we wrap this up? What do @mattheworiordan @paddybyers think of @SimonWoolf 's proposal?

content/client-lib-development-guide/features.textile
mattheworiordan: 👍 for @SimonWoolf's proposal

mattheworiordan:
@paddybyers I vote we wrap this up, and go with @SimonWoolf's proposal on https://github.com/ably/docs/pull/112#discussion_r61485477.  Can you confirm please?

paddybyers:
> I vote we wrap this up, and go with @SimonWoolf's proposal on #112 (comment). Can you confirm please?

I agree with the proposal overall, and definitely in the connecting with `resume`/`recover` case.

But .... do we want to cancel the connection attempt directly in the case that we're not doing a resume/recover, since there's nothing to tidy up on the realtime side?

Also, if this is the correct action in the `connecting` case, why not also when `disconnected` ?

tcard:
> But .... do we want to cancel the connection attempt directly in the case that we're not doing a resume/recover, since there's nothing to tidy up on the realtime side?

Doesn't cancelling the connection attempt directly leave connection state in the server during `realtimeStateTtl`? Or does breaking the connection before CONNECTED is sent clean up that? I'm thinking now that must be the case, since if you don't get to send the connection key to the client then there's no way of recovering that connection and there's no point on keeping the state.

In that case then it makes perfect sense to just abort and move to CLOSED directly.

> Also, if this is the correct action in the connecting case, why not also when disconnected ?

No reason apart from assumptions about what the user means when closing while in that particular state, which I criticized elsewhere. So, well spotted.

So, calling `close():

* While CONNECTING:
  * If the ClientOptions doesn't have a recovery key, move to CLOSED and close the transport.
  * If the ClientOptions has a recovery key, move to CLOSING, but don't actually close the connection. Instead:
    - If CONNECTING arrives before `realtimeRequestTimeout`, send CLOSE, wait for CLOSED message, move to CLOSED.
    - If not, move to CLOSED and close the transport.
* While DISCONNECTING: move to CLOSING, but continue with the recovery attempt. If it succeeds, send CLOSE, wait for CLOSED message, move to CLOSED. If it doesn't, don't move to SUSPENDED as usual and move to CLOSED instead.

Does that cover everything?

paddybyers:
> Doesn't cancelling the connection attempt directly leave connection state in the server during realtimeStateTtl? Or does breaking the connection before CONNECTED is sent clean up that? I'm thinking now that must be the case, since if you don't get to send the connection key to the client then there's no way of recovering that connection and there's no point on keeping the state.

If this is the first connection attempt then there obviously won't be any connection state at the realtime end. If not the first, as you say, even though there is state, the current attempt won't do anything with it, so there is no point in proceeding.

> While DISCONNECTING

s/`DISCONNECTING`/`DISCONNECTED`

mattheworiordan:
@paddybyers I really think this is unnecessary, I prefer a simple approach to this without a number of exceptions for a use case that is arguably really not important to developers anyway.  I don't see any harm in migrating to the `CLOSING` state in pretty much all cases, which in turn simply publishes a `CLOSING` message when the transport is available, or moves to `CLOSED` based on the standard timeout if it's not possible.  It keeps the spec simple, and IMHO, does the job just fine.

content/client-lib-development-guide/features.textile
tcard: Done at 4c177e8. (Sorry for the delay, I was under the impression I had already done this.)

tcard:
I've incorporated @SimonWoolf's proposal. I've _not_ incorporated @paddybyers' comments about recovery special cases due to @mattheworiordan's objections. If we still want to discuss that, maybe we can do so in a new PR? This one is getting old and kind of bloated.

paddybyers:
> I've not incorporated @paddybyers' comments about recovery special cases due to @mattheworiordan's objections.

I'm ok with this

content/client-lib-development-guide/features.textile
```diff
@@ -378,14 +384,19 @@ h3(#realtime-channel). Channel
 * @(RTL11)@ If a channel enters the @DETACHED@ or @FAILED@ state, then all messages that are still queued for send on that channel should be deleted from the queue triggering a failure for the publish or presence methods invoked for those messages
 * @(RTL4)@ @Channel#attach@ function:
 ** @(RTL4a)@ If already @ATTACHED@ or @ATTACHING@ nothing is done
+** @(RTL4h)@ If the channel is in the @DETACHING@ state, do the operation once it's @DETACHED@.
+** @(RTL4g)@ If state is @FAILED@, the @attach@ request results in an error
```
mattheworiordan: @tcard was just about to merge this, but then noticed this.  This is wrong no, if you call @attach@ on a @FAILED@ channel it will initiate a new @ATTACH@ no?

mattheworiordan: We should also change the wording to avoid ambiguity to `If the channel state is @FAILED@...`

mattheworiordan:
@tcard I've made a few minor additions, can you review before I merge in?

tcard:
@mattheworiordan You say that `attach` on a FAILED channel moves to ATTACHING (and actually tries to attach). But what about connection state? Do we implicitly attempt to connect (and move to CONNECTING) if the connection state is FAILED? What about the other states?

Also, we say that `publish` implicitly attaches, but `publish` in a FAILED channel doesn't. If it did, it would follow above's rules and try to reattach and then publish. Do we want it to?

mattheworiordan:
>  You say that attach on a FAILED channel moves to ATTACHING (and actually tries to attach). But what about connection state? Do we implicitly attempt to connect (and move to CONNECTING) if the connection state is FAILED? What about the other states?

That's a good question!  I don't know the answer as I need views from the others, but I would say calling attach when the connection is failed, suspended, closing, or closed, should result in an error, much like publish does now.

> Also, we say that publish implicitly attaches, but publish in a FAILED channel doesn't. If it did, it would follow above's rules and try to reattach and then publish. Do we want it to?

As above, I think publish in any state will implicitly attach, but only when the channel is in initialized and the connection is in a suitable state (connecting, connected, disconnected).

@paddybyers @SimonWoolf WDYT before I merge and push these changes out?

paddybyers:
> I would say calling attach when the connection is failed, suspended, closing, or closed, should result in an error, much like publish does now.

The connection state never changes as a result of operations performed on a channel. So if a connection isn't either `connected`, or a state that anticipates imminent connection (ie `connecting` when not from `suspended`, or `disconnected`), there must be an immediate error. If `connecting` or `disconnected`, the channel operation waits for the outcome of the connection attempt, and is attempted if connected, and fails with the relevant error if the connection state is `suspended` or `failed`.

> publish in any state will implicitly attach

in any connection state or channel state?

mattheworiordan:
>>  publish in any state will implicitly attach

> in any connection state or channel state?

Oops, let me rephrase.

I thought calling publish would only implicitly attach if the channel was in the initialized state and the connection was in a state that was suitable for publishing i.e. connecting, connected or disconnected.  In any other channel state, publish would not implicitly attach because the channel would have previously been attached and we only implicitly attach the first time?

paddybyers:
ok, so the question I think is:

does `publish()` (a) unconditionally call `attach()`; or (b) does it check the channel state and only conditionally call `attach()` if `initialised` (or `detached`?)

The difference is, at least, in the `failed` channel state; in case (a) `publish()` will trigger an attach attempt, whereas in (b) it will fail.

I think we previously said (b) - ie to move a channel from the `failed` state you need to do an explicit attach, instead of an implicit attach. I think @tcard is pointing out that this is really needless special-casing, and (a) is cleaner.

mattheworiordan:
@paddybyers well I thought the reason we do implicit attach was for the most common use case where a user creates a channel and then performs an operation.  If that channel has been explicitly detached, or has failed, I don't think an implicit attach should happen personally.  If everyone else disagrees with that approach then fine, but I always felt that these implicit behaviours were added to help developers when using the library from an initialised state only.

mattheworiordan:
This PR has been open for way too long.  @paddybyers @tcard should we just jump on a quick call to resolve what we should do implicitly or not?

tcard:
@mattheworiordan OK by me.

Spoiler of my view: I want to reduce special cases and complex conditions. So I'd prefer `channel.attach()` on a FAILED channel to do the same thing regardless of the state of the connection. That means either `CONNECTING` when the connection is FAILED, or not doing that but also not `ATTACHING` when the channel is FAILED.

I think the former would be more dev-friendly, but maybe that has unintended consequences I don't foresee and we should do the conservative thing (the latter).

mattheworiordan:
I understand that view, but am quite concerned by too many implicit side-effects.  

On implicit connection, I don't think we have ever done that in the library so not sure why we would start now.

On implicit channel attach, I feel this is only a convenience for devs who do something like:
```
ably = new Realtime()
channel = ably.channels.get("foo")
channel.subscribe (msg) -> { do something }
```

To me this convenience of attaching when we know the channel has just been initialised (because it is still in the initialised state) is right as it just makes things easier and means less code.  We know what they intended.

However, calling `channel.subscribe` on a channel that failed because of a permissions failure and attaching implicitly to me is just wrong.  This is not a convenience in my mind as we don't know if by attaching they want to subscribe later to events when they fix the attach issue, or if they want to subscribe only once the channel becomes attached successfully and they want to try attach again.

tcard:
I think 6ceb6a3 covers everything: implicit attachments don't happen for DETACHED or FAILED channels, and explicit attach on FAILED channels can fail depending on connection state.

Ready to merge then?

mattheworiordan:
@paddybyers @tcard I can't believe this is still ongoing, but I've added two commits 	747e168 and 		16e9547. Please shout if happy and I will merge it all tomorrow

tcard:
Happy here (after aea51a3).

paddybyers:
LGTM, thanks.

mattheworiordan:
Yay :)

PR #113, opened 2016-04-27
Reauthorise (interim spec)

mattheworiordan:
This PR addresses the discussion at https://github.com/ably/ably-js/pull/261

@paddybyers @SimonWoolf please review to see if you are happy with the changes in the spec.

Once we support protocol auth upgrades, we'll need to revisit this spec and it's possible we'll introduce a breaking a change, but hopefully not

See the related Ruby pull request at https://github.com/ably/ably-ruby/pull/90


content/client-lib-development-guide/features.textile
```diff
@@ -142,13 +142,14 @@ h3(#rest-auth). Auth
...
 ** @(RSA10b)@ Supports all @AuthOptions@ and @TokenParams@ in the function arguments
 ** @(RSA10c)@ Will not create a new token unless no previous token exists or current token has expired. Please note that a buffer of 15s for token expiry is recommended to avoid race conditions where the token is valid at the time of the request, but invalid when it reaches the server. Therefore, we recommend that a token is considered expired 15s before the time field `expires`
-** @(RSA10d)@ If the @AuthOption@ argument's @force@ attribute is true, it will force @Auth#authorise@ to issue a new token even if an existing token exists
+** @(RSA10d)@ If the @AuthOption@ argument's @force@ attribute is true, it will force @Auth#authorise@ to issue a new token even if an existing token exists. A special case convenience exists for @AuthOption@ that states that if all its attributes are @null@ apart from @force@ with the value @true@, then when passed to @authorise@ as an argument, the previously configured authentication options will remain intact. This behaviour takes precedence over "@RSA10j@":#RSA10j
```
paddybyers: This could be a lot simpler if `force` was simply an extra optional argument instead of a member of `AuthOptions`.

content/client-lib-development-guide/features.textile
```diff
@@ -142,13 +142,14 @@ h3(#rest-auth). Auth
...
+** @(RSA10k)@ If the @AuthOption@ argument's @queryTime@ attribute is true, it will obtain the server time once and persist the offset from the local clock. All future token requests generated directly or indirectly via a call to authorise will not obtain the server time, but instead use the local clock offset to calculate the server time
```
paddybyers: This would make a lot more sense if `queryTime` was instead `useServerTime`.

As an aside it might be worth adding

> A client MAY discard the cached local clock offset in situations in which it may have been invalidated, such as if there is a local change to the date, time, or timezone, of the client device.

content/client-lib-development-guide/features.textile
```diff
@@ -142,13 +142,14 @@ h3(#rest-auth). Auth
...
 ** @(RSA10e)@ Adheres to the implementation of @requestToken@ when issuing new tokens
 ** @(RSA10f)@ Returns a @TokenDetails@ object that contains the token string + token metadata
-** @(RSA10g)@ Stores the @AuthOptions@ and @TokenParams@ arguments as defaults for subsequent authorisations with the exception of the attributes @TokenParams#timestamp@ and @AuthOptions#force@
+** @(RSA10g)@ Stores the @AuthOptions@ and @TokenParams@ arguments as defaults for subsequent authorisations with the exception of the attributes @TokenParams#timestamp@, @AuthOptions#queryTime@ and @AuthOptions#force@
```
paddybyers: I think it would be better if we were explicit, in the documentation of `TokenParams` and `defaultTokenParams`, that `timestamp` is a "one-time" value, and is valid in a request, but is not validly a member of `defaultTokenParams` and will be ignored.

Then here it simply says that the `defaultTokenParams` is updated with the given values, with the exception of any value for `timestamp`.

content/client-lib-development-guide/features.textile
```diff
@@ -262,6 +263,7 @@ h3(#realtimeclient). RealtimeClient
 * @(RTC6)@ @RealtimeClient#time@ function:
 ** @(RTC6a)@ Proxy to @RestClient#time@ presented with an async or threaded interface as appropriate
 * @(RTC7)@ The client library must use the configured timeouts specified in the @ClientOptions@, falling back to the "client library defaults":#defaults and defaults described in @ClientOptions@ below
+* @(RTC8)@ Authentication upgrades are only possible when the client is @CONNECTED@ or @CONNECTING@. If @auth.authorise@ is called with @AuthOptions#force@ set to @true@ and the state is @CONNECTED@ or @CONNECTING@, the client will obtain a new token, disconnect the current transport and resume the connection (see "@RTN15c@":#RTN15c for details on resume). A test should exist that performs an upgrade of capabilities without any message loss during the upgrade process. A test should also for an auth failure which should result in the connection entering the @FAILED@ state
```
paddybyers: Let's not use "upgrade" in this context because of the conflict with transport upgrade. We could refer to token changes, updates, or renewals.

If we support this operation in the `CONNECTING` state, why not also when `DISCONNECTED` ?

content/client-lib-development-guide/features.textile
mattheworiordan: I know, and annoyingly it was a long time ago, but we hadn't thought about the implications of this.  We'll have to correct in the 0.9 spec

content/client-lib-development-guide/features.textile
mattheworiordan: I can add, but do you think that is necessary?  Happy to if you think it's better to include as opposed to keeping it simpler.

mattheworiordan:
@paddybyers I've incorporated your feedback now

content/client-lib-development-guide/features.textile
SimonWoolf: Maybe drop "`with the value @true@`" from `if all its attributes are @null@ apart from @force@ with the value @true@`? Having `{force: false}` overwrite stored options might end up confusing developers who write something like
```
function shouldForceReauth() { ... }
client.auth.authorise(null, {force: shouldForceReauth()}, function(){ ... })
```
by breaking things whenever `shouldForceReauth()` returns `false`

mattheworiordan: Thanks @SimonWoolf, done in 3c6d67b

SimonWoolf:
:+1:

SimonWoolf:
Oh - while we're doing this, we might want to clarify the corresponding wording for `requestToken` and `createTokenRequest` (RSA8e and RSA9h) too, to be less ambiguous on whether authOptions & tokenParams are replaced or merged

mattheworiordan:
@SimonWoolf I assume you mean like 7a0c56f.

I am a bit concerned about these changes as I am reasonably confident some of the existing implementations do not work this way :(

SimonWoolf:
> I am a bit concerned about these changes as I am reasonably confident some of the existing implementations do not work this way :(

definitely - made the above comment because I noticed that ably-js and ably-ruby did this differently :/

The other option is leaving it alone for now (letting whatever each client lib has done stand), then standardising it when we change the interface for 0.9?

paddybyers:
Until now `authorise()` has effectively been a private API - there is no reason for anyone to call it explicitly.

So I say we update the spec to say what we think it should say, and raise PRs / document the variance on those libraries that do not conform. They can be addressed in time, and the variance should not cause problems in practice.

SimonWoolf:
> Until now authorise() has effectively been a private API

@paddybyers the last three comments have been about whether we should change the semantics of requestToken and createTokenRequest as well, rather than just authorise. (ably-js and ably-ruby differ in whether tokenParams are merged or replaced for requestToken and createTokenRequest - eg see [this](https://github.com/ably/ably-ruby/blob/reauthorise/lib/ably/auth.rb#L210-L214)).

paddybyers:
> @paddybyers the last three comments have been about whether we should change the semantics of requestToken and createTokenRequest as well

must try harder to keep up ...

mattheworiordan:
Ok, we should finalise this and merge.  I propose we correct the spec, [as suggested by @SimonWoolf](https://github.com/ably/docs/pull/113#issuecomment-215454981), and then raise issues in the respective repos where we believe it does not conform, [as suggested by @paddybyers](https://github.com/ably/docs/pull/113#issuecomment-215456057).  Yay / nay?

paddybyers:
> Ok, we should finalise this and merge. I propose we correct the spec, as suggested by @SimonWoolf, and then raise issues in the respective repos where we believe it does not conform, as suggested by @paddybyers.

yes

PR #116, opened 2016-05-07
Javascript recover and closeOnUnload

mattheworiordan:
I would like to land the work in https://github.com/ably/ably-js/pull/266 as soon as possible, however we need the docs updated.

This PR addresses that.

@SimonWoolf @paddybyers please review


content/realtime/connection.textile
```diff
@@ -281,18 +281,33 @@ There are two modes of connection state recovery:
...
+* @recover@: this addresses the case in which a new client library instance wishes to connect and recover the state of an earlier connection. This occurs typically in a browser environment when the page has been refreshed and therefore the client instance is disposed and no client state is retained. In this case any message listeners associated with channels will no longer exist so it is not possible for the library simply to send the message backlog on reconnection; instead the client must re-subscribe to each channel it is interested in within 15 seconds, and its message listener(s) will be called with any message backlog for that channel. If the previously attached channels are not re-attached within 15 seconds of a connection being recovered, the channels will automatically be detached by Ably. A client requests recovery of connection state by including a recovery string in the "client options":/realtime/usage#client-options when instancing the Realtime library. See "connection state recover options":#connection-state-recover-options for more info.
```
paddybyers: Just to be consistent, use re-subscribe or re-attach consistently. I think re-attach is best.

content/realtime/connection.textile
```diff
@@ -281,18 +281,33 @@ There are two modes of connection state recovery:
...
+  In the browser environment, if a callback is provided in the @recover@ option, when @window beforeunload@ event fires, the connection details, including the "@recoveryKey@":#recovery-key, are stored in the "browser's sessionStorage":https://developer.mozilla.org/en/docs/Web/API/Window/sessionStorage. The provided @recover@ callback is then invoked whenever the connection state can be recovered and just before a connection is established, passing in the "@LastConnectionDetails@":#last-connection-details. The callback is then responsible for confirming whether the connection state should be recovered or not. For example, it is common to recover connection state when the page is reloaded but not for different pages the user has navigated to. The callback allows the developer to decide if the connection should be recovered or not at the time the new connection is established by inspecting the "@LastConnectionDetails@":#last-connection-details and evaluating that against any other application state. Below is a straightforward example:
```
paddybyers: We should use the W3C recommendation as the reference instead of the Mozilla one: https://www.w3.org/TR/webstorage/

Suggest:

> The callback allows the developer to decide whether or not the previous connection should be recovered when establishing a new connection, based on the "@LastConnectionDetails@":#last-connection-details and any other application state.

content/types/_realtime_client_options.textile
```diff
@@ -1,10 +1,14 @@
...
+
+- <div lang="javascript">closeOnUnload</div> := When true, the client library will automatically send a close request to Ably whenever the @window beforeunload@ event fires. By enabling this option, the close request sent to Ably ensures the connection state will not be retained and all channels associated with the channel will be detached. This is commonly used by developers who want presence leave events to fire immediately i.e. if a user navigates to another page or closes their browser, then enabling this option will result in the presence member leaving immediately. Without this option or an explicit call to the "@close@":/realtime/connection#close method of the "@Connection object@":/realtime/connection, Ably expects that the abruptly disconnected connection could later be recovered and therefore does not immediately remove the user from presence. Instead, to avoid "twitchy" presence behaviour an abruptly disconnected client is removed in channels they are present after 15 seconds, and the connection state is retained for two minutes<br>__Type: @Boolean@__
```
SimonWoolf: `is removed in channels they are present after` -> `is removed from channels in which they are present after`

content/types/_realtime_client_options.textile
```diff
@@ -1,10 +1,14 @@
...
+
+- <div lang="java,ruby,nodejs,objc,swift"><span lang="default">recover</span><span lang="ruby">:recover</span></div> := This option allows a connection to inherit the state of a previous connection that may have existed under a different instance of the library. This might typically be used by clients of an app to ensure connection state can be preserved following a reload. See "connection state recovery":/realtime/connection/#connection-state-recovery for further information and example code<br>__Type: @String@__
```
SimonWoolf: This could be more specific about how to use the option -- eg could add `by providing that connection's <span lang="default">"@recoveryKey@":#recovery-key</span><span lang="ruby">"@recovery_key@":#recovery-key</span>` at the end of the first sentence

content/realtime/connection.textile
```diff
@@ -281,18 +281,33 @@ There are two modes of connection state recovery:
...
+        if (lastConnectionDetails.url === document.location.href) {
+          cb(true); /* recover connection as user has reloaded page */
+        } else {
+          cb(false); /* do not recover connection as URL has changed */
+        }
+      }
+    });
+  ```
+
+  Alternatively, if it is necessary to be explicit about the connection @recoveryKey@ , the connection can be recovered by providing the last value of the connection's @recoveryKey@ value in the "client options":/realtime/usage#client-options @recover@ attribute when instancing the library.
```
paddybyers: Just wondering if it should be `lastConnectionDetails.location` or `lastConnectionDetails.location.href` instead of `.url` ?

Suggest:

> Alternatively, if it is necessary to recover a connection based on a client-supplied @recoveryKey@, the  most recent value of the connection's @recoveryKey@ value can be specified in the "client options":/realtime/usage#client-options @recover@ attribute when instancing the library.

SimonWoolf: > Just wondering if it should be lastConnectionDetails.location or lastConnectionDetails.location.href instead of .url ?

:+1: to stringifying & storing the whole location object, might save some devs the effort of parsing the url. If @mattheworiordan agrees I'll change that in the ably-js pr

content/types/_last_connection_details.textile
```diff
@@ -0,0 +1,12 @@
...
+
+h4.
+  default: Properties
+
+- recoveryKey := An opaque string obtained from the "recoveryKey":/realtime/connection/#recovery-key attribute of the "Connection object":/realtime/connection before the page was unloaded. This property is used by the library to recover the connection<br>__Type: @String@__
+
+- disconnectedAt := the time at which the previous library was abruptly disconnected before the page was unloaded. This is represented as milliseconds since epoch<br>__Type: @Integer@__
+
+- url := the URL of the previous page before the page was unloaded. A common use case for this attribute is to ensure that the previous page URL is the same as the current URL before allowing the connection to be recovered. For example, you may want the connection to be recovered only for page reloads, but not when a user navigates to a different page<br>__Type: @String@__
```
paddybyers: This should refer to the fact that we use `sessionStorage`, which has origin scope, and so if the library is instanced in any page in the same origin within the same top-level browsing context then the `lastConnectionDetails` for the previous page will be given in the callback.

paddybyers:

content/realtime/connection.textile
mattheworiordan: Sure, please go ahead with stringifying the whole location object.  When done, please can you update the example above with a commit?  I'll fix the rest...

mattheworiordan: @SimonWoolf when you stringifying the object, I assume devs won't need to JSON parse the object when receiving it.  Just checking because that would be weird, i.e. I would expect a dev to be able to do something like `lastConnectDetails.location.href`

content/realtime/connection.textile
```diff
@@ -281,18 +281,35 @@ There are two modes of connection state recovery:
...
 
-  Alternatively, if it is necessary to be explicit about the connection @recoveryKey@ - for example if there are multiple tabs and the cookies would conflict, or if third-party cookies are disabled - the connection can be recovered by providing the last value of the connection's @recoveryKey@ value in the "client options":/realtime/usage#client-options @recover@ attribute when instancing the library.
+  ```[jsall]
+    var ably = new Ably.Realtime({
+      authUrl: '/obtainToken',
+      recover: function(lastConnectionDetails, cb) {
+        if (lastConnectionDetails.location.href === document.location.href) {
```
mattheworiordan: @SimonWoolf I took a stab at how I think you will be implementing this, please just push a commit yourself if this is wrong

mattheworiordan:
@SimonWoolf @paddybyers have addressed all feedback.  Just waiting on @SimonWoolf now to confirm the new API for the `LastConnectionDetails` object

content/realtime/connection.textile
SimonWoolf: Yup, that's right. The `SessionStorage` abstraction stringifies and parses before it gets passed to the devs

content/realtime/connection.textile
SimonWoolf: :+1:

PR #120, opened 2016-06-09
Use fog creds from key 'ably', rather than 'default'

SimonWoolf:
Tried deploying docs.ably.io just now, noticed this repo expects fog creds to be under a 'default' key. Since the infrastructure repo expects ably fog creds to be under an 'ably' key, no point in duplicating them


mattheworiordan:
Looks good to me, thanks

PR #121, opened 2016-06-13
Clarify that presence map is cleared on detach

SimonWoolf:
See https://github.com/ably/ably-js/issues/284


mattheworiordan:
LGTM

PR #124, opened 2016-06-16
Attach reattach behaviour

mattheworiordan:
First draft of the attach/reattach spec.

Note we need to add new error codes https://github.com/ably/ably-common/blob/master/protocol/errors.json


mattheworiordan:
So we are not quite there as discussed now:

* `SYNC` received, all members are updated, any not mentioned get removed from the set.  It's a definitive enumeration of members in the presence set. Decide what happens on the `LEAVE` events or not from the client lib
* Only re-enter when a `SYNC` is complete, or channel is attached without `PRESENCE`. This happens when a `Channel` is attached with/without a sync, and when a `SYNC` on the channel occurs.

mattheworiordan:
@paddybyers I think this is about right.
@SimonWoolf if you happen to be online and fancy reviewing this please do

content/client-lib-development-guide/features.textile
```diff
@@ -274,10 +276,11 @@ h3(#realtime-connection). Connection
 ** @(RTN2b)@ @echo@ should be @true@ by default; @false@ will prevent messages published by the client being echoed back
 ** @(RTN2d)@ @clientId@ contains the provided @clientId@ option of @ClientOptions@, unless @clientId@ is @null@
 ** @(RTN2e)@ Depending on the authentication scheme, either @accessToken@ contains the token string, or @key@ contains the API key
-** @(RTN2f)@ API version param @v@ should be @0.8@
+** @(RTN2f)@ API version param @v@ should be @0.9@
 * @(RTN3)@ If connection option @autoConnect@ is true, a connection is initiated immediately; otherwise a connection is only initiated following an explicit call to @connect()@
 * @(RTN4)@ EventEmitter and states:
 ** @(RTN4a)@ Implements @EventEmitter@ and emits events for state changes @INITIALIZED@, @CONNECTING@, @CONNECTED@, @DISCONNECTED@, @SUSPENDED@, @CLOSING@, @CLOSED@, @FAILED@
+** @(RTN4g)@ Additionally, an @ERROR@ event is emitted that contains an @ErrorInfo@ object with details on an error that has occurred for the @Connection@
```
paddybyers: s/details on/details of

content/client-lib-development-guide/features.textile
```diff
@@ -332,7 +335,7 @@ h3(#realtime-connection). Connection
...
+**** @(RTN15c3)@ @CONNECTED@ @ProtocolMessage@ with a new @connectionId@, and an error in @error@. In this case, a new connection has been established, the resume has failed, the channels are no longer attached, and the error indicates the resume problem which should be emitted. The client library should reattach all channels that are in the @SUSPENDED@ state. For all @ATTACHING@ or @ATTACHED@ channels, the client library should fail any previously queued messages. An @ERROR@ event should additionally be emitted on each @Channel@ object that was automatically re-attached. Finally, the internal @msgSerial@ counter is reset so that the first message published to Ably will contain a @msgSerial@ value of @0@
```
paddybyers: "the resume has failed" -> maybe in situations like this we say the resume was unsuccessful, instead of failed, so there is no possible confusion with the channels being `failed`.

"The client library should reattach all channels" -> I think "the client library should initiate an attach operation for all channels ..." - to make it clear that this step is about starting the attach, and everything subsequent is governed by the requirements for that.

"For all @ATTACHING@ or @ATTACHED@ channels' -> can a channel be attached at the point of connection?

content/client-lib-development-guide/features.textile
```diff
@@ -375,15 +378,20 @@ h3(#realtime-channels). Channels
 
 h3(#realtime-channel). Channel
 
-* @(RTL1)@ As soon as a @Channel@ becomes attached, all incoming messages and presence messages are processed and emitted where applicable.  @PRESENCE@ and @SYNC@ messages are passed to the @Presence@ object ensuring it maintains a map of current members on a channel in realtime
+* @(RTL1)@ As soon as a @Channel@ becomes attached, all incoming messages and presence messages are processed and emitted where applicable. @PRESENCE@ and @SYNC@ messages are passed to the @Presence@ object ensuring it maintains a map of current members on a channel in realtime
 * @(RTL2)@ EventEmitter and states:
-** @(RTL2a)@ Implements @EventEmitter@ and emits events for state changes @INITIALIZED@, @ATTACHING@, @ATTACHED@, @DETACHING@, @DETACHED@, @FAILED@
+** @(RTL2a)@ Implements @EventEmitter@ and emits events for state changes @INITIALIZED@, @ATTACHING@, @ATTACHED@, @DETACHING@, @DETACHED@, @SUSPENDED@ and @FAILED@
 ** @(RTL2b)@ @Channel#state@ attribute is the current state of the channel
+** @(RTL2d)@ A @ChannelStateChange@ object is emitted as the first argument for every connection state change
```
paddybyers: "for every connection state change" -> channel state change

content/client-lib-development-guide/features.textile
```diff
@@ -375,15 +378,20 @@ h3(#realtime-channels). Channels
...
+** @(RTL2e)@ Additionally, a @ChannelStateChange@ can be emitted that contains a @reason@ which contains an @ErrorInfo@ object with details of the error that has occurred for the @Channel@
+** @(RTL2f)@ When a channel @ATTACHED@ @ProtocolMessage@ is received, the @ProtocolMessage@ may contain a bit flag with value 4 indicating that the channel has been resumed. The @ChannelStateChange@ will then be emitted for the @ATTACHED@ state and will contain a @resumed@ boolean attribute with value @true@ if the bit flag value 4 was included. When @resumed@ is @true@, this indicates that the channel attach resumed the channel state from an existing connection and there has been no loss of message continuity. In all other cases, @resumed@ is false. A test should exist to ensure that @resumed@ is always false when a channel first becomes @ATTACHED@, it is @true@ when the channel is @ATTACHED@ following a successful "connection recovery":#RTN16, and is @false@ when the channel is @ATTACHED@ following a failed "connection recovery":#RTN16
```
paddybyers: "with value 4" -> I will define the flags values and this should be documented in the types section of the docs, and then we should refer to the flag position by name.

"resumed the channel state from an existing connection" -> "resumed the channel attachment from the existing connection, with channel state (including presence) preserved and message continuity"

content/client-lib-development-guide/features.textile
```diff
@@ -375,15 +378,20 @@ h3(#realtime-channels). Channels
...
 ** @(RTL3a)@ If the connection state changes to @FAILED@ then an @ATTACHING@ or @ATTACHED@ channel state will transition to @FAILED@, set the @Channel#errorReason@ and emit the error event
-** @(RTL3b)@ If the connection state changes to @CLOSED@ or @SUSPENDED@ then an @ATTACHING@ or @ATTACHED@ channel state will transition to @DETACHED@
-* @(RTL11)@ If a channel enters the @DETACHED@ or @FAILED@ state, then all messages that are still queued for send on that channel should be deleted from the queue triggering a failure for the publish or presence methods invoked for those messages
+** @(RTL3b)@ If the connection state changes to @CLOSED@ then an @ATTACHING@ or @ATTACHED@ channel state will transition to @DETACHED@
+** @(RTL3c)@ If the connection state changes to @SUSPENDED@ then an @ATTACHING@ or @ATTACHED@ channel state will transition to @SUSPENDED@
+** @(RTL3d)@ If the connection state changes to @CONNECTED@ then a @SUSPENDED@ channel state will transition to @ATTACHING@. If any of the channels automatically attached fail, then the error message received from Ably should be emitted as an @ERROR@ event on the @Channel@ with Ably error code @91200@
```
paddybyers: "then a @SUSPENDED@ channel state will transition to @ATTACHING@." -> again I think this should say that an attach operation is initiated.

"the error message received from Ably should be emitted as an @ERROR@ event on the @Channel@ with Ably error code @91200@" -> I know we discussed this but if possible I'd prefer to avoid special-case behaviour here as much as possible.

content/client-lib-development-guide/features.textile
```diff
@@ -444,17 +453,28 @@ h3(#realtime-channel). Channel
...
+* @(RTP2)@ A "@PresenceMap@":https://github.com/ably/ably-java/blob/master/src/io/ably/realtime/Presence.java#L384-L529 should be used to maintain a list of members present on a channel. As there are no guarantees that during the @SYNC@ phase presence events will arrive in order i.e. a leave event for a member can arrive before that member is later registered as as present as part of the initial @SYNC@ operation. As such, until the @SYNC@ operation is complete, timestamps must be recorded for all members to ensure the most recent present state is used, see the "Java implementation":https://github.com/ably/ably-java/blob/master/src/io/ably/realtime/Presence.java#L418-L431.  Once a @SYNC@ event is complete, the members that are now considered @ABSENT@ can be removed from the map, see the "Java implementation":https://github.com/ably/ably-java/blob/master/src/io/ably/realtime/Presence.java#L504-L509
```
paddybyers: "timestamps must be recorded" -> I know this text was here before, but this sounds to me like the library determines the timestamp, instead of using the timestamp in the `PresenceMessage`. Maybe "the member @timestamp@ value is used to determine the most recent state for each member".

content/client-lib-development-guide/features.textile
```diff
@@ -444,17 +453,28 @@ h3(#realtime-channel). Channel
...
+* @(RTP18)@ The realtime system reserves the right to initiate a sync of the presence members at any point once a channel is attached and presence is in sync. The server initiated sync allows it to send a complete list of members present on the channel at that point in time. A sync is started by the server when it sends a @SYNC@ @ProtocolMessage@ to the client containing a @channel@ attribute <br>**TBC - how does the client lib know definitely it's starting a @SYNC@ and not continuing one, and should the client lib continue the @SYNC@ if disconnected & reconnected usingt the cursor?**.
```
paddybyers: The way this works is that a cursor value sent by realtime (in the `channelSerial` field) is a two-part identifier `<sync sequence id>:<cursor value>`. The `sync sequence id` is a new id when a sync sequence starts.

There are 2 ways we could do this:

- say that the realtime system will only perform a sync after sending an `ATTACHED` message with a `sync` flag. This is certainly true now (except for syncs requested by the client).

- say that the library remembers the `sync sequence id`; that value is cleared on completion of any sync, and set when there is a sync message. This means that the library can detect an incomplete sync if it receives an unexpected sync message with the wrong sequence id.

If the transport drops, the client can continue a sync sequence that was in progress prior to the disconnection provided that the resume connection serial it provides ensures that it has already processed, or gets from the backlog, all presence messages on the channel from the point that the first sync message in the sequence was received. If the realtime system isn't happy continuing that sync sequence, it will start a new one, and the sync sequence identifier will allow the client to know that it's a new sequence.

content/client-lib-development-guide/features.textile
```diff
@@ -444,17 +453,28 @@ h3(#realtime-channel). Channel
...
+* @(RTP17)@ The Presence object is also responsible for keeping a separate copy of an object similar to the @PresenceMap@ containing only members that match the current @connectionId@. Any @ENTER@, @PRESENT@, @UPDATE@ or @LEAVE@ event that matches the current @connectionId@ should be applied to this object in the same way it is done for the @PresenceMap@. This object should be private and is used to maintain a list of members that need to be automatically re-entered by the @Presence@ object when a @Channel@ becomes @ATTACHED@ with some or all continuity lost.
```
paddybyers: "similar to" -> "logically equivalent to"

content/client-lib-development-guide/features.textile
```diff
@@ -444,17 +453,28 @@ h3(#realtime-channel). Channel
...
+** @(RTP5a)@ If the channel enters the @DETACHED@ or @FAILED@ state then all queued presence messages will fail immediately, and the @PresenceMap@ and "internal PresenceMap (see RTP17)":#RTP7 is cleared. The latter ensures members are not automatically re-entered if the @Channel@ later becomes attached
+** @(RTP5f)@ If the channel enters the @SUSPENDED@ state then all queued presence messages will fail immediately, and the @PresenceMap@ is cleared
+** @(RTP5b)@ If a channel enters the @ATTACHED@ state then all queued presence messages will be sent immediately and a presence @SYNC@ may be initiated by the Ably service or the Ably service may indicate there is no @SYNC@ necessary
+** @(RTP5c)@ In addition, when a channel becomes @ATTACHED@, the following must happen in regards to @Presence@:
+*** @(RTP5c1)@ If the "@resumed@ flag":#TH4 is true and no @SYNC@ is initiated as part of the attach, do nothing as there is no loss of continuity on the channel and no change to presence. The @PresenceMap@ is not effected and no members need to be re-entered
```
paddybyers: "not effected" -> "not affected"

content/client-lib-development-guide/features.textile
```diff
@@ -444,17 +453,28 @@ h3(#realtime-channel). Channel
...
+*** @(RTP5c2)@ If the "@resumed@ flag":#TH4 is true and a @SYNC@ is initiated as part of the attach, then once the @SYNC@ is complete, all members not present in the
+@PresenceMap@ but present in the "internal @PresenceMap@":#RTP17 must be re-entered automatically by the client using the @clientId@ and @data@ attributes from each. The members re-entered automatically must be removed from the "internal @PresenceMap@":#RTP17 ensuring that members present on the channel are constructed from presence events sent from Ably since the channel became @ATTACHED@
+*** @(RTP5c3)@ If the "@resumed@ flag":#TH4 is false, then the @PresenceMap@ should be cleared, and once the @SYNC@ is complete, all members in the "internal @PresenceMap@":#RTP17 must be re-entered automatically by the client using the @clientId@ and @data@ attributes from each. The "internal @PresenceMap@":#RTP17 must then be cleared to ensure that members present on the channel are constructed from presence events sent from Ably since the channel became @ATTACHED@
```
paddybyers: ""internal @PresenceMap@":#RTP17 must then be cleared to ensure that members present on the channel are constructed from presence events sent from Ably since the channel became @ATTACHED@" -> I don't see why it's not handled just like any other sync - you don't drain the internal map completely, but remove the members that are outstanding on completion of the sync.

content/client-lib-development-guide/features.textile
SimonWoolf: Is this intended just for (eg resume) errors on `CONNECTED`? Or all errors (including fatal ones in `ERROR`s and nonfatal in `DISCONNECTED`s), duplicating the ErrorInfo in `connectionStateChange.reason`?

If just for `CONNECTED` errors, does that mean that those shouldn't be emitted in the `connectionStateChange.reason`? (I guess it would make sense for them not to be, since a resume failure isn't the reason why you're connected, but if so that should be made explicit in RTN4f)

content/client-lib-development-guide/features.textile
SimonWoolf: > The client library should reattach all channels that are in the @SUSPENDED@ state

If the client attempted a resume, there won't be any channels in the suspended state. Just attached or attaching.

Re the next sentence - in addition to Paddy's suggestion we could be more explicit about the next channel state (ie attaching), eg: "`For all @ATTACHING@ or @ATTACHED@ channels, the client library should fail any previously queued messages, move them to the @ATTACHING@ state, and initiate an attach operation.`"

> can a channel be attached at the point of connection?

The context of this spec item is a resume

content/client-lib-development-guide/features.textile
```diff
@@ -375,15 +378,20 @@ h3(#realtime-channels). Channels
...
+* @(RTL1)@ As soon as a @Channel@ becomes attached, all incoming messages and presence messages are processed and emitted where applicable. @PRESENCE@ and @SYNC@ messages are passed to the @Presence@ object ensuring it maintains a map of current members on a channel in realtime
 * @(RTL2)@ EventEmitter and states:
-** @(RTL2a)@ Implements @EventEmitter@ and emits events for state changes @INITIALIZED@, @ATTACHING@, @ATTACHED@, @DETACHING@, @DETACHED@, @FAILED@
+** @(RTL2a)@ Implements @EventEmitter@ and emits events for state changes @INITIALIZED@, @ATTACHING@, @ATTACHED@, @DETACHING@, @DETACHED@, @SUSPENDED@ and @FAILED@
 ** @(RTL2b)@ @Channel#state@ attribute is the current state of the channel
+** @(RTL2d)@ A @ChannelStateChange@ object is emitted as the first argument for every connection state change
+** @(RTL2e)@ Additionally, a @ChannelStateChange@ can be emitted that contains a @reason@ which contains an @ErrorInfo@ object with details of the error that has occurred for the @Channel@
```
SimonWoolf: > `A @ChannelStateChange@ object is emitted ... Additionally, a @ChannelStateChange@ can be emitted that contains a @reason@`

The wording makes it seem like this is a different object from RTL2d. Maybe just "`That @ChannelStateChange@ may contain a @reason@ ...`"

content/client-lib-development-guide/features.textile
paddybyers: > The context of this is a resume

oops, yes :)

content/client-lib-development-guide/features.textile
SimonWoolf: > `If any of the channels automatically attached fail, then the error message received from Ably should be emitted as an @ERROR@ event on the @Channel@`

Why do we need this? If the channel fails (eg realtime rejects the autoattach), it will emit a `failed` event with the reason realtime gives for the failure. Why emit an `error` as well?

(We discussed emitting an error on the channel it in the other issue, but as I understood it, that was for a failed presence autoreenter (which otherwise wouldn't result in an event emitted, unlike a channel failure))

content/client-lib-development-guide/features.textile
mattheworiordan: I did review this and think it sounded weird, but it's just a copy of http://docs.ably.io/client-lib-development-guide/features/#RTL2c that is currently in the spec.  I will reword as this spec item should simply state that that an `ERROR` event can be emitted, not that it will.  So it's a generic spec stating that a channel / connection has the ability to emit errors, that's it.

content/client-lib-development-guide/features.textile
mattheworiordan: Thanks, will amend

content/client-lib-development-guide/features.textile
mattheworiordan: Thanks, good spot

content/client-lib-development-guide/features.textile
mattheworiordan: Sure

content/client-lib-development-guide/features.textile
mattheworiordan: > "with value 4" -> I will define the flags values and this should be documented in the types section of the docs, and then we should refer to the flag position by name.

Sure, go ahead :)

Otherwise comment is appreciated

SimonWoolf:
RTN18 and RTL3 both deal with how connection state changes propagate to channel state. Is there a reason to have both? There's potential for confusion, especially when they get out of sync.

content/client-lib-development-guide/features.textile
SimonWoolf: Fair enough.

(to be clear - `CONNECTED` errors should be in the `connectionStateChange.reason`, even though they're not really the 'reason' for the state change?)

paddybyers: > (to be clear - CONNECTED errors should be in the connectionStateChange.reason, even though they're not really the 'reason' for the state change?)

Yes. An error associated with the transition is in the `reason` field. Thinking in terms of the channel state changes, for example, the `reason` might include the reason for the flags being the way they are, not just the reason that the transition occurred.

content/client-lib-development-guide/features.textile
mattheworiordan: I assume I will simply add the flag to http://docs.ably.io/client-lib-development-guide/features/#TR4i then.  If so, flags will be:
`HAS_PRESENCE: 1`
`HAS_BACKLOG: 2`
`RESUMED: 4`

OK?

content/client-lib-development-guide/features.textile
mattheworiordan: > "the error message received from Ably should be emitted as an @ERROR@ event on the @Channel@ with Ably error code @91200@" -> I know we discussed this but if possible I'd prefer to avoid special-case behaviour here as much as possible.

I am not sure what special-case we are talking about here.  The library is implicitly calling `Channel#attach`, yet there is no possibility for a dev to provide a callback for that situation.  Hence why we discussed emitting an error if the operation fails.

> Why do we need this? If the channel fails (eg realtime rejects the autoattach), it will emit a failed event with the reason realtime gives for the failure. Why emit an error as well?

Because it can also just not attach, and thus remain in `SUSPENDED` state.  So at that point, the automatic reattach failed, no error would be emitted, and the dev would simply be waiting for the channel to leave the `SUSPENDED` state which will never happen.

If the dev has no way of passing in a callback for an implicit operation, then I think it's important that they at least get the opportunity to catch errors when that implicit operation fails.

content/client-lib-development-guide/features.textile
mattheworiordan: Fine, will amend

content/client-lib-development-guide/features.textile
mattheworiordan: Ok, that makes sense.  I will amend stating that should the `sync sequence id` change, then the client library discards any current sync info and starts a new sync sequence.

content/client-lib-development-guide/features.textile
mattheworiordan: True, no need to treat the case where `resumed` is false differently

mattheworiordan:
@SimonWoolf @paddybyers I believe I've addressed all the feedback.  Please shout if you spot anything else.

SimonWoolf:
:+1:

content/client-lib-development-guide/features.textile
```diff
@@ -332,7 +335,7 @@ h3(#realtime-connection). Connection
...
+**** @(RTN15c3)@ @CONNECTED@ @ProtocolMessage@ with a new @connectionId@, and an error in @error@. In this case, a new connection has been established, the resume was unsuccessful, the channels are no longer attached, and the emitted error indicates the cause of the unsuccessful resume. The client library should initiate an attach for channels that are in the @SUSPENDED@ state. For all channels in the @ATTACHING@ or @ATTACHED@ state, the client library should fail any previously queued messages for that channel and initiate a new attach i.e. a new @ATTACH@ @ProtocolMessage@ must be sent for each channel. An @ERROR@ event should additionally be emitted on each @Channel@ object that was automatically re-attached. Finally, the internal @msgSerial@ counter is reset so that the first message published to Ably will contain a @msgSerial@ value of @0@
```
SimonWoolf: > An @ERROR@ event should additionally be emitted on each @Channel@ object that was automatically re-attached.

Now that we've decided to indicate loss of channel continuity with a flag on the `attached` stateChange, do we need the `error` event as well? (why emit an `error` for some losses of continuity and not others?)

mattheworiordan: I think your are right @SimonWoolf, @paddybyers do you agree?

paddybyers: > do you agree?

yes I think so. We will finally confirm I think when it's implemented and we see any unforeseen circumstances.

mattheworiordan:
@SimonWoolf @paddybyers according to this spec, if a channel becomes `DETACHED` whilst the connection remains `CONNECTED`, then the channel will not be automatically reattached.  @paddybyers has said there are conditions where a channel needs to be come detached such as the frontend not being able to communicate with the channel master for whatever reason, and as such, it will detach the channel and emit an error.  I think we must therefore extend this spec to deal with that situation and automatically re-attach & re-enter.  Do you agree?

SimonWoolf:
>  I think we must therefore extend this spec to deal with that situation and automatically re-attach & re-enter. Do you agree?

Sounds good to me. (More precisely: should go into the `attaching` state and re-attempt an attach. Or maybe into the `detached` state and then immediately the `attaching` if we think that'd be more obvious to devs).

Re retrying attaches that timeout:

> Although it does feel weird that we would keep reattaching a channel when it was never attached in the first place.  The attach/reattach PR is all about continuity, whereas this is not really continuity because the service had never started in the first place. So users could call attach and wonder why it never fails and messages are queued forever

I mean, we've never guaranteed that attach() callbacks will be called in 10s. At the moment, if someone calls `attach` while the client lib is `connecting`/`disconnected` and it doesn't succeed in connecting, it can stay in the `attaching` state for up to 2 minutes until the lib goes into `suspended` (at which point the channel will go back to `detached` and the attach cb will be called).

It's not about continuity, true, but it is about relieving devs from having to micromanage channel states and different types of attach failures (fatal vs nonfatal). Wouldn't it be nice to have an api that guarantees devs that once they've called attach, the lib will keep trying to get into the attached state (and if left, will try to re-gain that state) until either it succeeds or (if there's a good reason for it not to be able to attach) Ably responds with a fatal channel error?

Is there anything a dev might plausibly want to do if it an initial `attach` times out other than just try again? If not, then we may as well do it ourselves, no?

mattheworiordan:
> Sounds good to me. (More precisely: should go into the attaching state and re-attempt an attach. Or maybe into the detached state and then immediately the attaching if we think that'd be more obvious to devs).

I was thinking in terms of the old spec unfortunately when I wrote this and specifically a bit of code I have provided to `dub-track`.  In fact, a channel that is not attached, but should be, should always move into the `suspended` state.  So in my view, if the realtime system needs to detach a channel, it could send that detach with an error and the client will move the channel to the detached state.  Then, if the client is configured to reattach suspended channels automatically, it would do that.

Therefore, in your example @SimonWoolf, we could do the same i.e. if the `ATTACH` request fails, it could move into the suspended state, which would fail all queued messages, and then attempt a new attach.

> Is there anything a dev might plausibly want to do if it an initial attach times out other than just try again? If not, then we may as well do it ourselves, no?

I think there could be some use cases, which brings me to my next point.

Even thought I said before we should not make this an option, I think we should introduce a new `ClientOption` along the lines of `reattachSuspendedChannels: true`, which if disabled, does what it says on the tin.  Not really much of a change to the spec, but does ensure customers who want to handle failures for whatever reason can do so.

mattheworiordan:
Closing this as I will create a new PR for the complete 0.9 spec

PR #127, opened 2016-07-12
Spec update v0.9

mattheworiordan:
Following feedback from our customers, this PR addresses the gotcha in v0.8 whereby we (Ably) expect customers to always handle channel failures (see https://github.com/ably/docs/commit/571dae366ab1aea4c9a0950e511c4c9a7d5da7bc), however it is more common that customers want the client library to handle this automatically.

Note: Where we reference 0.9 version of client libraries that do not support the 0.9 spec at the time of merging, I will add a commit to revert that and address that when those client libraries are ready.

@paddybyers @SimonWoolf please confirm you are happy for me to merge these changes?


content/client-lib-development-guide/features.textile
```diff
@@ -287,9 +292,10 @@ h3(#realtime-connection). Connection
 * @(RTN6)@ A @Connection@ is successful and considered @CONNECTED@ once the websocket connection is open and the intial @CONNECTED@ @ProtocolMessage@ has been received
 * @(RTN19)@ If the @CONNECTED@ @ProtocolMessage@ contains a @connectionDetails@ property, the attributes within @ConnectionDetails@ will be used as the defaults for this client library, overriding any configured options at the time the @CONNECTED@ @ProtocolMessage@ is received
```
SimonWoolf: Not one of the changes you made, but: there are now two `RTN19`s

content/client-lib-development-guide/features.textile
```diff
@@ -287,9 +292,10 @@ h3(#realtime-connection). Connection
...
 ** @(RTN7b)@ Every @ProtocolMessage@ sent must contain a unique serially incrementing @msgSerial@ integer value starting at zero. The @msgSerial@ along with the @count@ for incoming @ACK@ and @NACK@ @ProtocolMessages@ indicates which messages succeeded or failed to be delivered
 ** @(RTN7c)@ If a connection enters the @CLOSED@ or @FAILED@ state, or if the connection state is lost, and a @ACK@ or @NACK@ has not yet been received for a message, the client should consider the delivery of those messages as failed
+** @(RTN7d)@ If a connection enters the @DISCONNECTED@ or @SUSPENDED@ state, and has not become @CONNECTED@ within 15 seconds, then the client should consider the delivery of those messages as failed
```
SimonWoolf: - Don't think we need the `@SUSPENDED@` in this one -- for a client to go into `@SUSPENDED@` it'll have been disconnected for 2 minutes, so per this spec any sent messages will have been failed 1m45s ago

- Is 15s a bit short? The default disconnectedRetryTimeout is 15s, so a common pattern for people going through a short network problems (eg tunnels) will be to be disconnected from ably for exactly 15s, then successfully resuming, at which point realtime will send any pending acks/nacks. Seems a shame for the client lib to pessimistically pre-empt those immediately beforehand. I appreciate that you don't want to wait the full connectionStateTtl (2 minutes) though... maybe 30s? @paddybyers may have an opinion on this

content/client-lib-development-guide/features.textile
```diff
@@ -318,22 +324,25 @@ h3(#realtime-connection). Connection
...
+** @(RTN14b)@ If a connection request fails due to an @ERROR@ @ProtocolMessage@ being sent to the client containing a token error (@statusCode@ value of 401 and error @code@ value in the range @40140 <= @code < 40150@), then if the token is renewable, the error should not be emitted, a single attempt to create a new token will be made and a new connection attempt will be initiated using the newly created token. If the attempt to create a new token fails, or the subsequent connection attempt fails due to another token error, then the connection will transition to the @FAILED@ state and the @Connection#errorReason@ will be set
+** @(RTN14g)@ If an @ERROR@ @ProtocolMessage@ is received for any reason other than "RTN14b":RTN14b, then the connection will transition to the @FAILED@ state and the server will terminate the connetion. Additionally the @Connection#errorReason@ must be set with the error from the @FAILED@ @ProtocolMessage@
```
SimonWoolf: s/`@FAILED@ @ProtocolMessage@`/`@ERROR@ @ProtocolMessage@`

content/client-lib-development-guide/features.textile
```diff
@@ -318,22 +324,25 @@ h3(#realtime-connection). Connection
...
+**** @(RTN15c4)@ @ERROR@ @ProtocolMessage@ indicating a fatal error in the connection. The server will close the transport immediately after. The client should move to the @FAILED@ state triggering all attached channels to move to the @FAILED@ state as well. Additionally the @Connection#errorReason@ will be set should be set with the error received from Ably
+**** @(RTN15c5)@ @DISCONNECTED@ @ProtocolMessage@ indicating a recoverable error in the connection, such as a token token error. The process described in "RTN15h":RTN15h must be followed for a connection being resumed with a token error
```
SimonWoolf: > @(RTN14b)@ If a connection request fails due to an @ERROR@ @ProtocolMessage@ being sent to the client containing a token error

...

> @(RTN15c5)@ @DISCONNECTED@ @ProtocolMessage@ indicating a recoverable error in the connection, such as a token token error.

So a token error on a connection attempt will be sent as an `ERROR` action, but on an existing connection will be sent as a `DISCONNECTED`?

It's not obvious to me which one it should be (as it doesn't have normal `ERROR` fatal semantics or normal `DISCONNECTED` can-immediately-retry-with-the-same-params semantics), but whichever one we go for we should be consistent

SimonWoolf: (also `token token`)

content/client-lib-development-guide/features.textile
SimonWoolf: (Also `RTN15f` also deals with what happens to messages if a transport drops - RTN15f and RTN7d are hard to reconcile at the moment)

content/client-lib-development-guide/features.textile
```diff
@@ -318,22 +324,25 @@ h3(#realtime-connection). Connection
...
 ** @(RTN15g)@ When the connection resume has failed, all channels should be detached with a suitable error reason
```
SimonWoolf: This is no longer correct. It's also subsumed by RTN15c3, so can probably just remove it

content/client-lib-development-guide/features.textile
```diff
@@ -287,9 +292,10 @@ h3(#realtime-connection). Connection
...
 * @(RTN19)@ If the @CONNECTED@ @ProtocolMessage@ contains a @connectionDetails@ property, the attributes within @ConnectionDetails@ will be used as the defaults for this client library, overriding any configured options at the time the @CONNECTED@ @ProtocolMessage@ is received
 * @(RTN7)@ @ACK@ and @NACK@:
-** @(RTN7a)@ All @ProtocolMessage@ @Presence@ and @Message@ objects sent to Ably expect either an @ACK@ or @NACK@ from Ably to confirm successful receipt and acceptance or failure respectively
+** @(RTN7a)@ All @ProtocolMessage@ @Presence@ and @Message@ objects sent to Ably expect either an @ACK@ or @NACK@ from Ably to confirm successful receipt and acceptance or failure respectively. For clarity, it is unnecessary to fail the publish operation of a message using a timer. Instead the client library can rely on the realtime system to send an @ACK@ or @NACK@ when connected, to fail all awaiting messages when @DISCONNECTED@ for too long as described in "RTN7d":#RTN7d, or upon reconnecting, the corresponding @ACK@ or @NACK@ protocol messages will be received as described in "RTN19a":#RTN19a
```
SimonWoolf: > or upon reconnecting, the corresponding @ACK@ or @NACK@ protocol messages will be received as described in "RTN19a":#RTN19a

RTN19a contradicts this ("Any ProtocolMessage that is awaiting an ACK/NACK on the old transport will *not* receive the ACK/NACK on the new transport."). It specifies that you need to resend it

content/client-lib-development-guide/features.textile
mattheworiordan: Thanks, good spot, I think this happens due to multiple branches being created and when merged, there is no way to automatically ensure uniqueness.  I will fix.

content/client-lib-development-guide/features.textile
mattheworiordan: Well I referenced RTN19a so that people understand why the ACK or NACK would be received, but clearly the wording is not clear enough so I'll improve.

content/client-lib-development-guide/features.textile
mattheworiordan: Thanks

content/client-lib-development-guide/features.textile
mattheworiordan: > So a token error on a connection attempt will be sent as an ERROR action, but on an existing connection will be sent as a DISCONNECTED?

I thought about this too.  There is fundamentally a difference:

* An `ERROR` is sent when no connection exists i.e. you are have asked for a connection and your connection request has failed.  This is an error, i.e. your auth details are wrong and nothing has been done on our side.
* A `DISCONNECTED` is sent on existing connection because we're saying there is now a problem with your connection you can correct, so we're disconnecting you so that you can reconnect and fix the problem.

content/client-lib-development-guide/features.textile
mattheworiordan: Good spot

content/client-lib-development-guide/features.textile
```diff
@@ -61,8 +63,9 @@ h3(#restclient). RestClient
 *** @(RSC6b3)@ @limit@ supports up to 1,000 items; if omitted the limit defaults to the REST API default (100)
 *** @(RSC6b4)@ @unit@ is the period for which the stats will be aggregated by, values supported are @minute@, @hour@, @day@ or @month@; if omitted the unit defaults to the REST API default (@minute@)
 * @(RSC16)@ @RestClient#time@ function sends a get request to @rest.ably.io/time@ and returns the server time in milliseconds since epoch or as a Date/Time object where suitable
-* @(RSC7)@ Sends REST requests over HTTP and HTTPS to the REST end-point @rest.ably.io@
-** @(RSC7a)@ The header @X-Ably-Version: 0.8@ must be included in all REST requests to the Ably end-point
+* @(RSC7)@ Sends REST requests over HTTP and HTTPS to the REST endpoint @rest.ably.io@
```
paddybyers: Sending these as a header is required where possible; and they should be sent as a query param when the environment does not permit headers to be specified (eg JSONP).

I think the library version is a SHOULD and not a MUST. Protocol version is a MUST.

content/client-lib-development-guide/features.textile
paddybyers: > I didn't see the harm in including the state

I agree.

> he reason I chose 15 is because if we are unable to deliver in that time, I think it's fair that the developer knows about it. 15s for a realtime system is a long time i.e. if I was building a chat app would I not want to know that my message has not arrived in that time?

I definitely agree that it makes no sense to wait for `ConnectionStateTtl` to fail the messages because we could choose arbitrarily large values for that (eg up to 20 mins).

A single momentary disconnection (where the first zero-wait connection attempt succeeds) would permit the messages to be re-sent within the 15s. To get the next 15s attempt to work then this timeout would need to increase to 25s say, which seems a bit long. I think 15s is probably a good number, and if we want to ensure that we get continuity on the second connection attempt (ie the one after the zero-wait attempt) then the right way to do that would be to have an escalating disconnect timeout (eg 5s, 10s, 15s, 15s, ....).

content/client-lib-development-guide/features.textile
paddybyers: and s/connetion/connection

content/client-lib-development-guide/features.textile
SimonWoolf: > he messages awaiting ACK/NACK may have in fact been sent and if the client reconnects we will find out

Unless I'm misunderstanding, I don't think we will - when it reconnects it'll be a different transport, so no acks/nacks from the previous. We'll never know. What we can do is resend them with the original msgSerial, and realtime will dedup if it already received a msg with that serial (though ably-js does not do that at the moment).

@paddybyers to be clear, would you say we should fail all queued messages after 15s of disconnection then (whether or not a send has previously been attempted for that message)?

content/client-lib-development-guide/features.textile
paddybyers: > There is fundamentally a difference:
> ...
> An ERROR is sent when no connection exists

I would say that an `ERROR` is sent when the system is not able to associate the connection attempt with an existing connection. For example, attempting to resume with expired credentials. It doesn't mean that the connection is definitely in error, but the connection attempt did not get as far as being associated with any existing connection.

> A DISCONNECTED is sent on existing connection

I think you mean sent on an existing transport ... but @SimonWoolf's suggestion last night was that you could indeed get a `DISCONNECTED` response even on a transport that was never connected, if the system wishes to communicate that the connection is not failed, only the transport or connection attempt.

So the `ERROR` case, received in response to a connection attempt with, should be the only situation where you need to look at the specific error code to know whether or not you can still expect the connection to exist and be recoverable.

content/client-lib-development-guide/features.textile
mattheworiordan: Fine, will amend.

content/client-lib-development-guide/features.textile
```diff
@@ -819,7 +853,7 @@ h4. ProtocolMessage
 
 * @(TR1)@ A @ProtocolMessage@ represents the type used to send and receive messages over the Realtime protocol.  A ProtocolMessage always relates either to the connection or to a single channel only, but can contain multiple individual Messages or PresenceMessages.  See the "Ruby ProtocolMessage documentation":http://www.rubydoc.info/gems/ably/Ably/Models/ProtocolMessage, but bear in mind the attributes following underscore naming in Ruby
 * @(TR2)@ @ProtocolMessage@ @Action@ enum has the following values in order from zero: @HEARTBEAT@, @ACK@, @NACK@, @CONNECT@, @CONNECTED@, @DISCONNECT@, @DISCONNECTED@, @CLOSE@, @CLOSED@, @ERROR@, @ATTACH@, @ATTACHED@, @DETACH@, @DETACHED@, @PRESENCE@, @MESSAGE@, @SYNC@
-* @(TR3)@ @ProtocolMessage@ @Flag@ enum has the following values in order from zero: @HAS_PRESENCE@, @HAS_BACKLOG@
+* @(TR3)@ @ProtocolMessage@ @Flag@ enum has the following values in order from zero: @HAS_PRESENCE@, @HAS_BACKLOG@ and @RESUMED@
```
paddybyers: These names and values have to be provisional until the realtime implementation of these features is complete.

content/client-lib-development-guide/features.textile
mattheworiordan: >> he messages awaiting ACK/NACK may have in fact been sent and if the client reconnects we will find out

> Unless I'm misunderstanding, I don't think we will - when it reconnects it'll be a different transport, so no acks/nacks from the previous. We'll never know. As I understand it, what we can do instead is resend them with the original msgSerial, and realtime will dedup if it already received a msg with that serial (though ably-js does not do that at the moment).

Yes, we will never know, but the realtime system will send an ACK or NACK back when the client library publishes the message again on the new transport.  So if we failed all messages already sent to the realtime system, then it's quite plausible devs will be told their message send failed and will send again, only to discover it was published twice.  By resending the original messages on the new transport the realtime system prevents that from happening.  

> @paddybyers to be clear, would you say we should fail all queued messages after 15s of disconnection then (whether or not a send has previously been attempted for that message)?

We have never stated that we resend queued messages, we can do that, but I don't think it's right personally. For example take this scenario:

* User is connected and attached
* User publishes 3 messages 
* Connection becomes disconnected
* User publishes 3 more messages
* Connection becomes disconnected
* First 3 messages auto re-published and next 3 messages that are queued are sent.  ACKs are received for all 6 even though the first 3 messages got through in the old connection.

paddybyers:

content/client-lib-development-guide/features.textile
SimonWoolf: > if we failed all messages already sent to the realtime system, then it's quite plausible devs will be told their message send failed and will send again, only to discover it was published twice. By resending the original messages on the new transport the realtime system prevents that from happening. 

@mattheworiordan I'm sorry, I'm getting really confused here. AFAICT, this is an argument against nacking-after-15s for messages that have already been attempted once, and in favour of nacking-after-15s for messages which haven't been attempted yet. But the spec item (RTN7d) we're commenting on only applies to messages which *have* been attempted (RTN7) and are awaiting ack/nack.

content/client-lib-development-guide/features.textile
mattheworiordan: Sorry, but I don't think that is what we agreed.  I thought we agreed that an `ERROR` ProtocolMessage not associated with a channel indicated a non-recoverable error.   So the permutations are:

### New connection

* Connect with no problems: `CONNECTED` with no `error`
* Connect with an issue but the connection is still available: `CONNECTED` with an `error`
* Could not establish a connection / transport: `ERROR` and transport is terminated

### Resume/recover connection

* Connect with no problems: `CONNECTED` with no `error`
* Connect with an issue but the connection is still available: `CONNECTED` with an `error`
* Could not establish the connection / transport: `ERROR` and transport is terminated

### Open connection errors

* Recoverable problem with the connection: `DISCONNECTED` with `error` (client should now resume connection)
* Token failure: `ERROR` with `error#code` in the token failure range (client should now resume connection)
* Non-recoverable failure of the connection: `ERROR` with `error` (client moves to the `FAILED` state)

content/client-lib-development-guide/features.textile
mattheworiordan: > @mattheworiordan I'm sorry, I'm getting really confused here. AFAICT, this is an argument against nacking-after-15s for messages that have already been attempted once, and in favour of nacking-after-15s for messages which haven't been attempted yet. But the spec item (RTN7d) we're commenting on only applies to messages which have been attempted (RTN7) and are awaiting ack/nack.

Well no, I don't think so.  Lets put queued messages aside for now and focus only on messages published but awaiting an `ACK` / `NACK`.

If you publish a message, bearing in mind we're realtime system, then if the message does not get an `ACK` / `NACK` in a sensible amount of time, we should IMHO notify the developer so that they can decide what to do next. Whilst it is possible that the message was indeed published, holding on indefinitely or for a very long time feels wrong because a) the message may not have been published, b) the UI probably needs updating to state it's probably not sent, c) there is no guarantee we can confirm `ACK` / `NACK` correctly anyway once they've been disconnected for a longer period of time.

So I think our design should aim to:

* Hide any transient network failures from the developer such as changing networks, or a very brief disconnection.  This is going to cover most situations where a connection becomes disconnected
* Fail requests that are not processed in a reasonable amount of time (15s in mind is more than enough)
* When in doubt, prefer a failure notification as opposed to success / silence (we may need to add this to the docs for message publishing)

In regards to messages that have not yet been attempted and are sitting in the queue, in fact I would concur that we should probably apply the same 15s timeout rule and fail immediately if the connection becomes `suspended`, `closed` or `failed` (which we do anyway).   For example, if the connection is in the `DISCONNECTED` state and remains in that state for 2 minutes, all queued messages will just sit there waiting to be sent.  If we try and maintain the premise that if we cannot deliver soon then it's considered a failure, we should really time these messages out after 15s.

@paddybyers WDYT?

content/client-lib-development-guide/features.textile
mattheworiordan: Sure, we can update easily anyway

mattheworiordan:
To be created as a clean PR

PR #130, opened 2016-07-19
heartbeat timeout proposal

SimonWoolf:
Per discussion https://github.com/ably/ably-js/issues/155#issuecomment-233481789

Only realtime change needed for now is sending `heartbeatPeriod: config.realtime.heartbeatTimer` in connectionDetails. Honouring qs-param-requested alternate periods is optional & low-priority so can be postponed

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -327,10 +327,11 @@ h3(#realtime-connection). Connection
 ** @(RTN12c)@ If the transport is abruptly closed following a @CLOSE@ @ProtocolMessage@ being sent, then the connection will automatically move to the @CLOSED@ state
 ** @(RTN12d)@ If the connection state is @DISCONNECTED@ or @SUSPENDED@, aborts the retry process described in "RTN14d":#RTN14d and "RTN14e":#RTN14e and moves the connection immediately to the @CLOSED@ state
 * @(RTN13)@ @Connection#ping@ function:
-** @(RTN13c)@ If the connection state is @CONNECTING@ or @DISCONNECTED@, do the operation once the connection state is @CONNECTED@
+** @(RTN13d)@ If the connection state is @CONNECTING@ or @DISCONNECTED@, do the operation once the connection state is @CONNECTED@
 ** @(RTN13a)@ Will send a @ProtocolMessage@ with action @HEARTBEAT@ the Ably service when connected and expects a @HEARTBEAT@ message in response. If the client library language supports callbacks, then the callback will be called with the response time or error
 ** @(RTN13b)@ Will indicate an error if in, or has moved to, the @INITIALIZED@, @SUSPENDED@, @CLOSING@, @CLOSED@ or @FAILED@ state
 ** @(RTN13c)@ Will fail if a @HEARTBEAT@ @ProtocolMessage@ is not received within the "default realtime request timeout":#defaults
+** @(RTN13e)@ The @ProtocolMessage@ sent should include an @id@ property, with value a random string. If so, only a @HEARTBEAT@ response which includes an @id@ property with the same value should be considered a response to that ping, in order to disambiguate from normal heartbeats and other pings.
 * @(RTN14)@ @Connection@ opening failures:
 ** @(RTN14a)@ If an API key is invalid, then the connection will enter the @FAILED@ state and the @Connection#errorReason@ will be set on the @Connection@ object as well as the emitted @ConnectionStateChange@
 ** @(RTN14b)@ If a connection request fails due to an @ERROR@ @ProtocolMessage@ being received by the client containing a token error (@statusCode@ value of 401 and error @code@ value in the range @40140 <= code < 40150@) and an empty @channel@ attribute, then if the token is renewable, the error should not be emitted, a single attempt to create a new token will be made and a new connection attempt will be initiated using the newly created token. If the attempt to create a new token fails, or the subsequent connection attempt fails due to another token error, then the connection will transition to the @FAILED@ state and the @Connection#errorReason@ will be set
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -374,6 +374,9 @@ h3(#realtime-connection). Connection
 ** @(RTN19a)@ Any @ProtocolMessage@ that is awaiting an @ACK@/@NACK@ on the old transport will not receive the @ACK@/@NACK@ on the new transport. The client library must therefore resend any @ProtocolMessage@ that is awaiting a @ACK@/@NACK@ to Ably in order to receive the expected @ACK@/@NACK@ for that message. The Ably service is responsible for keeping track of messages, ignoring duplicates and responding with suitable @ACK@/@NACK@ messages
 ** @(RTN19b)@ If there are any pending channels i.e. in the @ATTACHING@ or @DETACHING@ state, the respective @ATTACH@ or @DETACH@ message should be resent to Ably
 ** @(RTN19b)@ If a @SYNC@ is underway, ensure the client library adheres to @RTP3@
+* @(RTN21)@ Heartbeats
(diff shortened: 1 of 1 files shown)


content/client-lib-development-guide/features.textile
```diff
@@ -367,6 +367,9 @@ h3(#realtime-connection). Connection
 ** @(RTN19a)@ Any @ProtocolMessage@ that is awaiting an @ACK@/@NACK@ on the old transport will not receive the @ACK@/@NACK@ on the new transport. The client library must therefore resend any @ProtocolMessage@ that is awaiting a @ACK@/@NACK@ to Ably in order to receive the expected @ACK@/@NACK@ for that message. The Ably service is responsible for keeping track of messages, ignoring duplicates and responding with suitable @ACK@/@NACK@ messages
 ** @(RTN19b)@ If there are any pending channels i.e. in the @ATTACHING@ or @DETACHING@ state, the respective @ATTACH@ or @DETACH@ message should be resent to Ably
 ** @(RTN19b)@ If a @SYNC@ is underway, ensure the client library adheres to @RTP3@
+* @(RTN20)@ Heartbeats
+** @(RTN20a)@ If a transport does not receive a heartbeat for a period greater than the sum of the @heartbeatPeriod@ (sent in the @connectionDetails@ in the @CONNECTED@ message received on that transport) and the @realtimeRequestTimeout@, that transport may be disconnected. This requirement is optional; in deciding whether to implement, client library developers should take into account whether the transport in question is susceptible to undetected dropped connections.
```
SimonWoolf: `realtimeRequestTimeout` is 10 seconds, that's a pretty long delay. I mean, if we have evidence that a transport's hasn't received a message that realtime sent 10+ seconds ago, seems pretty reasonable to me to consider it dead, notify the user, and start again with a resume with the last-received connectionSerial. You don't think so?

paddybyers: Buffering in a cellular connection can be 6s. Things don't always travel at the speed of light :)

That said, I dont think this is unreasonable; or we could increase it to the trusty 15s timeout :)

SimonWoolf: > Buffering in a cellular connection can be 6s.

FWIW an approximately constant buffer would be fine no matter how big, since the timer is since the last heartbeat (or the `CONNECTED`); it'd only be a problem if it increased from 0 to >10s between heartbeats.

Also, at the risk of stating the obvious, we already use `realtimeRequestTimeout` for a bunch of stuff in the client libs. If we believe 10s is too tight for heartbeats, doesn't that imply it's also too tight for all that stuff? (Especially since most of the things it's already used for involve a round-trip, unlike heartbeats which are one-way). Which would imply we need to increase `realtimeRequestTimeout`

mattheworiordan: @SimonWoolf I think it's too tight because if for example a frontend was under extreme load and was delayed in sending the Heartbeat + the user was on a bad connection, we'd just end up creating more work for that frontend by terminating the connection and immediately resuming a new one.  Why not state that once `realtimeRequestTimeout` has failed, the client initiates a Heartbeat and if a response is not received within `realtimeRequestTimeout` then it disconnects the connection.  I realise we would prefer that users get feedback more quickly when a connection drops, but TCP/IP and/or the OS should do that in most cases, and where it doesn't, I think we'll have accept that it's a bit harder for us to be sure the connection is indeed gone.

BTW. Why did you not include an option to include an ID in Heartbeats so that clients can correlate the Heartbeat they sent with the one they received?  Seems like a trivial change.

SimonWoolf: > Why not state that once `realtimeRequestTimeout` has failed, the client initiates a Heartbeat and if a response is not received within `realtimeRequestTimeout` then it disconnects the connection.

OK. TBH I'm not convinced the effect of that would be much different from just having a timeout of 20s in the first place, unless a setTimeout callback in node under load can be delayed by 10s *more* than an external request, which I don't think is possible in an event loop. But shrug, it's harmless, so if you'd prefer that scheme, I'm happy to change it. Or go up to a 15/20s timeout. @paddybyers any opinion?

> BTW. Why did you not include an option to include an ID in Heartbeats so that clients can correlate the Heartbeat they sent with the one they received? 

Happy to add it. (Missed it out because these changes are about server-sent heartbeats only, I didn't touch the pings section.)

SimonWoolf: added 46d9e3d7c5456d61ec588801e1d37a8a3d73332d for ping IDs

mattheworiordan: LGTM

SimonWoolf:
@paddybyers @mattheworiordan new version per discussion. Went for `heartbeats` param and `maxIdleInterval` in the connectParams.

I dropped the optional `heartbeatPeriod` connectParam on the grounds that if we find it is needed, it's not a breaking change to add it, and until then no point cluttering up the spec with things we don't plan on implementing any time soon.

We didn't actually discuss what the idle-disconnection period should be in the call -- I've left it at the sum of `maxIdleInterval` and `realtimeRequestTimeout` for now, but happy to make it use its own, larger timeout rather than `realtimeRequestTimeout` if you prefer.

content/client-lib-development-guide/features.textile
```diff
@@ -946,6 +950,7 @@ h4. ConnectionDetails
 ** @(CD2e)@ @maxInboundRate@ is the maximum allowable number of requests per second from a client or Ably. In the case of a realtime connection, this restriction applies to the number of @ProtocolMessage@ objects sent, whereas in the case of REST, it is the total number of REST requests per second
 ** @(CD2f)@ @connectionStateTtl@ is the duration that Ably will persist the connection state when a Realtime client is abruptly disconnected
 ** @(CD2g)@ @serverId@ string is a unique identifier for the front-end server that the client has connected to. This server ID is only used for the purposes of debugging
+** @(CD2h)@ @maxIdleInterval@ is the maximum length of time milliseconds that the server will allow no activity to occur in the server->client direction. After such a period of inactivity, the server will send a @HEARTBEAT@ or transport-level ping to the client. If the value is 0, the server will allow arbitrarily-long levels of inactivity.
```
mattheworiordan: What about if the value is missing?  Is that treated as zero?

content/client-lib-development-guide/features.textile
```diff
@@ -367,6 +368,9 @@ h3(#realtime-connection). Connection
...
 ** @(RTN19b)@ If there are any pending channels i.e. in the @ATTACHING@ or @DETACHING@ state, the respective @ATTACH@ or @DETACH@ message should be resent to Ably
 ** @(RTN19b)@ If a @SYNC@ is underway, ensure the client library adheres to @RTP3@
+* @(RTN21)@ Heartbeats
+** @(RTN21a)@ If a transport does not receive any indication of activity on a transport for a period greater than the sum of the @maxIdleInterval@ (which will be sent in the @connectionDetails@ of the most recent @CONNECTED@ message received on that transport) and the @realtimeRequestTimeout@, that transport should be disconnected. This requirement is not mandatory; in deciding whether to implement, client library developers should take into account whether the transport in question is susceptible to undetected dropped connections. Any message (or non-message indicator, see @RTN21a@) received counts as an indication of activity and should reset the timer, not merely heartbeat messages. However, it must be received (that is, sent from the server to the client); client-sent data does count.
```
mattheworiordan: Think you mean see `@RTN21b@`

content/client-lib-development-guide/features.textile
```diff
@@ -367,6 +368,9 @@ h3(#realtime-connection). Connection
...
+** @(RTN21b)@ When initiating a connection, the client may add send a @heartbeat@ param in the querystring, with value @true@ or @false@. If the value is true, the server will use Ably protocol messages (for example, a message with a @HEARTBEAT@ action) to satisfy the @maxIdleInterval@ requirement. If it is false or unspecified, the server is permitted to use any transport-level mechanism (for example, websocket ping frames) to satisfy this. So for example, for websocket transports, if the client is able to observe websocket pings, then it should send @heartbeats=false@. If not, it should send @heartbeats=true@.
```
mattheworiordan: `s/may add send/may send/g`

mattheworiordan:
Few small comments, but looks good to me

content/client-lib-development-guide/features.textile
SimonWoolf: Why would it ever be missing? Once the realtime changes are made, we'll always send this field. (Before the changes are made it won't, but then there's lots of bits of the 0.9 spec that'll only work after realtime changes are made)

content/client-lib-development-guide/features.textile
SimonWoolf: oops, thx

content/client-lib-development-guide/features.textile
mattheworiordan: Well I suppose it won't, fine to be explicit then

mattheworiordan:
@paddybyers can you please confirm if you are happy with this proposal?

PR #132, opened 2016-07-25
Auth protocol message

mattheworiordan:
This PR addresses the requirement to allow token authentication upgrades/downgrades/renewal without closing the connection.  The realtime system can now send an `AUTH` `ProtocolMessage` a sensible time before the token expires forcing the client to immediately commence re-auth.

@SimonWoolf @paddybyers @tcard please review.

Note:
* I believe we should consider exposing a means to invalidate existing tokens based on a token identifier or even a `clientId` thus forcing all clients to immediately reauth.
* We may also want to allow upgrades from anonymous to identified as that is a very common use case


tcard:
Taken a look, looks good to me although I haven't been that involved in all the reauth-related discussions.

content/client-lib-development-guide/features.textile
```diff
@@ -298,6 +302,8 @@ h3(#realtime-connection). Connection
...
 ** @(RTN7b)@ Every @ProtocolMessage@ sent must contain a unique serially incrementing @msgSerial@ integer value starting at zero. The @msgSerial@ along with the @count@ for incoming @ACK@ and @NACK@ @ProtocolMessages@ indicates which messages succeeded or failed to be delivered
 ** @(RTN7c)@ If a connection enters the @SUSPENDED@, @CLOSED@ or @FAILED@ state, or if the connection state is lost, and an @ACK@ or @NACK@ has not yet been received for a message, the client should consider the delivery of those messages as failed
+* @(RTN22)@ Ably can request that a connected client re-authenticates by sending the client an @AUTH@ @ProtocolMessage@. The client must then immediately start a new authentication process as described in "RTC8":#RTC8
+** @(RTN22a)@ Ably reserves the right to forcibly disconnect a client that does not re-authenticate within acceptable period of time, or at any time the token is deemed no longer valid. A client is forcibly disconnected following a @DISCONNECTED@ message containing an error code in the range @40140 <= @code < 40150@. This will in effect force the client to re-authenticate and resume the connection immediately, see "RTN15h":#RTN15h
```
SimonWoolf: > @40140 <= @code < 40150@

unnecessary middle `@`

content/client-lib-development-guide/features.textile
```diff
@@ -270,7 +272,9 @@ h3(#realtimeclient). RealtimeClient
...
+* @(RTC8)@ Authentication token changes for the current connection are possible when the client is in the @CONNECTED@, @CONNECTING@ or @DISCONNECTED@ states. If @auth#authorise@ is called or Ably requests a re-authentication (see "RTN22":#RTN22) and the state is @CONNECTED@, @CONNECTING@, or @DISCONNECTED@, the client will obtain a new token and send an @AUTH@ @ProtocolMessage@ to Ably. The @data@ attribute of the @ProtocolMessage@ must contain either a token string or an encoded @TokenDetails@ object with the @encoding@ attribute set accordingly
```
SimonWoolf: > Authentication token changes for the current connection are possible when the client is in the @CONNECTED@, @CONNECTING@ or @DISCONNECTED@ states.

For `#authorise`, no reason not to allow it in the `INITIALIZED` state, that's a reasonable thing to want to do. (And server-requested re-authing, can only happen when `CONNECTED` anyway, so no need to gate it)

content/client-lib-development-guide/features.textile
```diff
@@ -270,7 +272,9 @@ h3(#realtimeclient). RealtimeClient
...
+* @(RTC8a)@ If the authentication token change is successful, then Ably will send a new @CONNECTED@ @ProtocolMessage@. The @connectionDetails@ provided in the @CONNECTED@ @ProtocolMessage@ must override any existing defaults, see "RTN21":#RTN21. A test should exist that performs an upgrade of capabilities without any loss of continuity or connectivity during the upgrade process. Another test should exist where the capabilities are downgraded resulting in Ably explicitly detaching the channel. That test must assert that the channel becomes detached soon after the token update and the reason is included in the channel state change event
```
SimonWoolf: > then Ably will send a new @CONNECTED@ @ProtocolMessage@

Do we want to expose to the user that this has happend? If so, how? (Currently the connection object only emits a `connected` event on a state change, which this wouldn't be)

SimonWoolf: (does this spec item mean we may in the future be planning to expose capabilities in the `connected` message, btw?)

content/client-lib-development-guide/features.textile
SimonWoolf: s/within acceptable period of time/within an acceptable period of time/

content/client-lib-development-guide/features.textile
```diff
@@ -148,17 +148,19 @@ h3(#rest-auth). Auth
 ** @(RSA9g)@ A valid HMAC is created using the key secret  to sign the @TokenRequest@ so that it can be used by any client to request a token without having or exchanging any secrets
 ** @(RSA9i)@ Adheres to all requirements in @RSA8@ relating to @TokenParams@
 * @(RSA10)@ @Auth#authorise@ function:
```
SimonWoolf: From https://github.com/ably/wiki/issues/68#issuecomment-188697130 :

> However, given where we are, I think we should get consistency in `initialized` everywhere, and worry about `authorize` perhaps in the 0.9 spec.

This might be a good opportunity? (Though should keep `authorise` as an alias with a deprecated warning till 1.0 at least)

content/client-lib-development-guide/features.textile
mattheworiordan: @SimonWoolf have fixed the unnecessary `@` in https://github.com/ably/docs/pull/133/commits/175e6ccdfe3f68e595a5eadddb8146264e91764c elsewhere

content/client-lib-development-guide/features.textile
mattheworiordan: > Do we want to expose to the user that this has happened?

Can you think of a reason why someone would need to handle this event?  Assuming they cared they'd provide an authCallback, and if not, why should they care?  Happy to include if we can think of a use case.

> (does this spec item mean we may in the future be planning to expose capabilities in the connected message, btw?)

No, and I don't think we should.  We have to expect capabilities for a client could change in future without the client being involved, so I don't see any real benefit to the client knowing what capabilities it has as it could be incorrect anyway.

mattheworiordan:
@SimonWoolf updated to address your comments.  Note however that 9022f2a cannot be merged into `master` as it would break the documentation.  We will only update the main documentation when the client libraries use that new syntax.

Also, we will now need to consider a way for users to view old documentation at www.ably.io/documentation :/

@paddybyers WDYT?

content/client-lib-development-guide/features.textile
SimonWoolf: >> (does this spec item mean we may in the future be planning to expose capabilities in the connected message, btw?)

> No, and I don't think we should.

OK. I was about to ask what the point of even sending a CONNECTED is then if nothing's going to act on it and we're not exposing it (since none of the `connectionDetails` are things that could change with a new token), but I've just realised the answer to that is that realtime might want to change the `connectionKey`. So fair enough.

content/client-lib-development-guide/features.textile
```diff
@@ -270,7 +273,9 @@ h3(#realtimeclient). RealtimeClient
...
-* @(RTC8)@ Authentication token changes for the current connection are possible when the client is in the @CONNECTED@, @CONNECTING@ or @DISCONNECTED@ states. If @auth.authorise@ is called with @AuthOptions#force@ set to @true@ and the state is @CONNECTED@, @CONNECTING@, or @DISCONNECTED@, the client will obtain a new token, disconnect the current transport and resume the connection (see "@RTN15c@":#RTN15c for details on resume). A test should exist that performs an upgrade of capabilities without any message loss during the upgrade process. A test should also exist for a token change that fails due to an incompatible token, which should result in the connection entering the @FAILED@ state
+* @(RTC8)@ Authentication token changes for the current connection are possible when the client is in the @CONNECTED@, @CONNECTING@ or @DISCONNECTED@ states. If @auth#authorize@ is called or Ably requests a re-authentication (see "RTN22":#RTN22), the client must obtain a new token and send an @AUTH@ @ProtocolMessage@ to Ably. The @data@ attribute of the @ProtocolMessage@ must contain either a token string or an encoded @TokenDetails@ object with the @encoding@ attribute set accordingly
```
paddybyers: I understand the motivation, but I don't like having a generic `data` member for this. It means both the type and the content are implicit when they don't need to be; and it's not really that extensible: what would happen for example if we had a message that needed two data items? Would we have `data0` and `data1` ? Or `data` would be an array or a hash of items?

Until we come up with a new protocol format then I think we should stick to named and typed members whenever the types are known, and use the `data`+`encoding` pattern for `any` types, as in `Message` etc now.

mattheworiordan: Ok, can you be a bit more specific about what exactly you are suggesting I change?

mattheworiordan: @paddybyers?

content/client-lib-development-guide/features.textile
```diff
@@ -147,18 +147,21 @@ h3(#rest-auth). Auth
...
+* @(RSA10)@ @Auth#authorize@ function:
+** @(RSA10a)@ Instructs the library to create a token immediately and ensures Token Auth is used for all future requests. See "RTC8":#RTC8 for re-authentication behaviour when called for a realtime client
+** @(RSA10j)@ Method signature is @authorize(TokenParams, AuthOptions)@. @TokenParams@ and @AuthOptions@ are optional. When the arguments are present, even if empty, the @TokenParams@ and @AuthOptions@ supersede any previously client library configured @TokenParams@ and @AuthOptions@. For example, if a client is initialised with @TokenParams#ttl@ configured with a custom value, and a @TokenParams@ object is passed in as an argument to @#authorize@ with a @null@ value for @ttl@, then the @ttl@ used for every subsequent authorisation will be @null@
```
SimonWoolf: > `and a @TokenParams@ object is passed in as an argument to @#authorize@ with a @null@ value for @ttl@`

BTW, could maybe add "or with no `ttl` property" to that? Currently it sortof reads like it's the presence of the `ttl` property that makes it override. (I realise that's not a distinction for strongly typed languages, but for jsthat could be a reasonable interpretation)

content/client-lib-development-guide/features.textile
```diff
@@ -147,18 +147,21 @@ h3(#rest-auth). Auth
...
+** @(RSA13a)@ Instructs the library to update the default authentication details in @TokenParams@ and @AuthOptions@ used for all future authentication requests
+** @(RSA13b)@ Method signature is @update(TokenParams, AuthOptions)@. @TokenParams@ and @AuthOptions@ are optional. When the arguments are present, even if empty, the @TokenParams@ and @AuthOptions@ supersede any previously client library configured @TokenParams@ and @AuthOptions@. For example, if a client is initialised with @TokenParams#ttl@ configured with a custom value, and a @TokenParams@ object is passed in as an argument to @#update@ with a @null@ value for @ttl@, then the @ttl@ used for every subsequent authorisation will be @null@
+** @(RSA13c)@ If a previous token does not exist or the token has expired, then a new token will be created immediately, and in the case of a realtime client, the client will re-authenticate, see "RTC8":#RTC8. Please note that a buffer of 15s for token expiry is recommended to avoid race conditions where the token is valid at the time of the request, but invalid when it reaches the server. Therefore, we recommend that a token is considered expired 15s before the time field @expires@
```
SimonWoolf: BTW, just thinking about this. For js, are you imagining this as a function that takes a callback?

I mean, if it could trigger the creation of a new token per RSA13c, it makes sense to have a callback for when that's done. But if it doesn't, then do we callback with null? or with whatever token the lib has at the moment? If the latter, it's unintuitive that a fn called 'update' could callback with a token that doesn't match the tokenParams that were passed in to it. But if it always forces a new token, it's the same as authorize...

To put it another way, what are you thinking of as the usecase for this?

SimonWoolf:
I'm also thinking we should rethink the current strategy of failing the connection if there's any error getting a token. In particular, even if we assume our servers are always up, customers servers might just be down for a few seconds - I just saw this with dubtrack. Currently if an authCallback/authUrl call fails, that's it, connection's failed. Should we consider a spec item saying maybe something like, if the lib requests a token while trying to connect and the token request fails with an error that's anything other than a realtime error in the 4xxxxx range, we move to `disconnected` rather than `failed`?

content/client-lib-development-guide/features.textile
paddybyers: > can you be a bit more specific about what exactly you are suggesting I change?

For consistency with the existing connect-time params wouldn't it be just an `accessToken` member? Or if we want some structure to limit the proliferation of top-level members in the `ProtocolMessage` then an `auth` member with value `{ accessToken: xxx}`

content/client-lib-development-guide/features.textile
mattheworiordan: Sure, I can amend so it makes sense for typed & non-typed langs

content/client-lib-development-guide/features.textile
mattheworiordan: @SimonWoolf the use case is questionable, but `update` is effectively what `authorise` was without `force: true`, so I didn't want to remove functionality that previously existed.  @paddybyers may need to weigh in here, but this API existed previously I believe so that you could change requirements for the next implicit auth without forcibly disconnecting.  Equally, if you wanted to update the auth params and only perform the auth when the next REST request was made, this would be useful as well.

mattheworiordan:
> Should we consider a spec item saying maybe something like, if the lib requests a token while trying to connect and the token request fails with an error that's anything other than a realtime error in the 4xxxxx range (except 40170), we move to disconnected rather than failed?

I don't like this to be honest, if the client lib is unable to successfully get a token to authenticate, and it needs one, then the connection should become failed.  If the customer provides a callback,  then it is up to them to handle failures, however if they provide an authUrl, then perhaps we could considering retrying the request once before failing.  That way we're not really changing anything fundamental in the spec, and instead stating that we allow for one failure on an authUrl. WDYT?

mattheworiordan:
@paddybyers @SimonWoolf I believe this PR now covers everything we discussed.  Please review, with special attention to the new timeout I have suggested for auth requests.

content/client-lib-development-guide/features.textile
```diff
@@ -92,19 +87,25 @@ h3(#restclient). RestClient
...
+** @(RSA4b)@ When the client does have a means to renew the token automatically, and the token has expired or the server has responded with a token error (@statusCode@ value of 401 and error @code@ value in the range @40140 <= code < 40150@), then the client should automatically make a single attempt to reissue the token and resend the request using the new token. If the token creation failed or the subsequent request with the new token failed due to a token error, then the request should result in an error
+** @(RSA4c)@ If an attempt by the realtime client library to authenticate is made using the @authUrl@ or @authCallback@, and the request to @authUrl@ fails, the callback @authCallback@ results in an error, or the attempt times out after "@realtimeRequestTimeout@":#DF1b, then:
+*** @(RSA4c1)@An @ErrorInfo@ with @code@ @80019@ and description of the underlying failure should be emitted with the state change, in the @errorReason@ and/or in the callback as appropriate
```
SimonWoolf: There is already a code for this - `"40170": "error from client token callback"`. But actually I think you're right that it should be an 8xxxx error and not a 4xxxxx, so happy with 80019.

:+1: for the timeout. (We could even add a fourth condition to RSA4c for if the authCallback calls back with something that isn't a well-formed tokenString/tokenDetails/tokenRequest -- I've seen dubtrack's authCallback call back with `(null, null)` before, leading to an unhandled exception. can't trust customer-provided callbacks :P )

content/client-lib-development-guide/features.textile
```diff
@@ -270,7 +270,9 @@ h3(#realtimeclient). RealtimeClient
...
+* @(RTC8a)@ If the authentication token change is successful, then Ably will send a new @CONNECTED@ @ProtocolMessage@. The @connectionDetails@ provided in the @CONNECTED@ @ProtocolMessage@ must override any existing defaults, see "RTN21":#RTN21. A test should exist that performs an upgrade of capabilities without any loss of continuity or connectivity during the upgrade process. Another test should exist where the capabilities are downgraded resulting in Ably explicitly detaching the channel. That test must assert that the channel becomes detached soon after the token update and the reason is included in the channel state change event
+* @(RTC8b)@ If the authentication token change fails, then Ably will send an @ERROR@ @ProtocolMessage@ triggering the connection to transition to the @FAILED@ state. A test should exist for a token change that fails due to an incompatible (such as sending a new token with an incompatible @clientId@)
```
SimonWoolf: fragment: "due to an incompatible (such as"

SimonWoolf:
:+1:

content/client-lib-development-guide/features.textile
mattheworiordan: Thanks, have added a commit to that effect

content/client-lib-development-guide/features.textile
mattheworiordan: Thanks, fixed

mattheworiordan:
@paddybyers can you confirm you are happy with this proposal to be merged now?

mattheworiordan:
@paddybyers FYI I just rebased this as there were a lot of conflicts so thought it best you review after the rebase in case I have introduced any bugs as a result

content/client-lib-development-guide/features.textile
```diff
@@ -276,8 +276,11 @@ h3(#realtimeclient). RealtimeClient
...
-* @(RTC9)@ @RealtimeClient#request@ is a wrapper around @RestClient#request@ (see "RSC19":#RSC19) delivered in an idiomatic way for the realtime library i.e. in the case of Ruby, with an evented async callback interface
+* @(RTC8)@ Authentication token changes for the current connection are possible when the client is in the @CONNECTED@, @CONNECTING@ or @DISCONNECTED@ states. If @auth#authorize@ is called or Ably requests a re-authentication (see "RTN22":#RTN22), the client must obtain a new token and send an @AUTH@ @ProtocolMessage@ to Ably. The @auth@ attribute of the @ProtocolMessage@ must contain a JSON-like object in the format @{ "accessToken": "<token string>" }@ or @{ "tokenDetails": <TokenDetails> }@
```
paddybyers: Wouldn't we explicitly define a type for this, ie

```
AuthDetails {
  String? accessToken;
  TokenDetails? tokenDetails;
}
```

mattheworiordan: Can do, felt unnecessary given this is only for internal use and every client lib dev can implement a type or not if they wish.  If you want me to add that, I can though.

paddybyers: > Can do, felt unnecessary given this is only for internal use and every client lib dev can implement a type or not if they wish

Writing it in IDL takes fewer words than prose and is also less likely to be ambiguous. It's not a requirement that a type is created in the implementation.

If we're going to stick with prose then I think the expression "a JSON-like object" is wrong anyway, so I think defining a type also avoids the discussion about what to say instead :)

mattheworiordan: @paddybyers what about bd2ed74?

paddybyers: Lgtm

PR #134, opened 2016-07-26
spec: Support custom fallback hosts

mattheworiordan:
Customers have requested custom CNAME endpoints, yet if they do this currently, we cannot support fallback host functionality. This ensures the two are not mutually exclusive.

@SimonWoolf @paddybyers


paddybyers:
Before we go into this is it with considering the simpler option of just having a way to modify the hard-coded defaults in the white-label solution?

mattheworiordan:
> Before we go into this is it with considering the simpler option of just having a way to modify the hard-coded defaults in the white-label solution?

No, as I think this change is quite trivial and applies to all client libraries as opposed to just JS

SimonWoolf:
:+1:

paddybyers:
LGTM

PR #137, opened 2016-07-30
Publish on behalf of a realtime client

mattheworiordan:
A customer has asked how to publish on behalf of a realtime client, a pattern that is often used by Pusher because of their legacy handling of clients typically being receivers only.

This documents how to do that with Ably


mattheworiordan:
I have merged this in, but @paddybyers @SimonWoolf included here in case you spot any issues.  If so, just fix with a new commit please

PR #139, opened 2016-07-31
Protocol adaptors

mattheworiordan:
@SimonWoolf I made some updates to the translator docs in your repo, but then thought that it would be better to be in here and we can more easily link to code examples etc.

The only issue is the Pusher demo does not work.  Can you take a look?  Any feedback appreciated too.

FYI. The MD files are compiled to HTML files in the `output` folder so we can use that HTML in Freshdesk

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ ruby File.read(File.expand_path("../.ruby-version", __FILE__)).strip
 
 gem 'nanoc'
 gem 'RedCloth'
+gem 'kramdown'
 gem 'fog'
 gem 'sass'
 gem 'gist'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -128,6 +128,7 @@ GEM
     ipaddress (0.8.0)
     jsbin-client (0.1.1)
       rest-client (>= 1.6)
+    kramdown (1.11.1)
     listen (2.8.5)
       celluloid (>= 0.15.2)
       rb-fsevent (>= 0.9.3)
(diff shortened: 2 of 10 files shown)


SimonWoolf:
LGTM. Added 5a992a3 to fix the pusher channel example (channel namespacing translation) (though haven't updated the jsbin hash, can't find the api key in the private repo)

SimonWoolf:
Also added 7619fb9 to add some inline styles for output that looks decent when copied into freshdesk (which sets margins to 0 as it expects spacing to be done with wysiwyg-output `<p><br></p>` nonsense...)

mattheworiordan:
@SimonWoolf, nice, it works now.  Should we update the docs to state that `public:` namespace is required if not private?

BTW. I've updated the private repo with the required `jsbin_config.yaml` file now

mattheworiordan:
@SimonWoolf do you want to merge when ready and update the support site?

SimonWoolf:
> Should we update the docs to state that public: namespace is required if not private?

It already does say that, in the 'channels' section. https://github.com/ably/docs/pull/139/files#diff-9bd3e2f604fda992561fb436e94dd147R66, plus an example in the table. Happy to tweak if you have a suggestion on making it clearer

I'll do the jsbin thing and merge when done, cheers

mattheworiordan:
> Happy to tweak if you have a suggestion on making it clearer

Nope, that was me skim reading :)  Looks good on re-read

PR #140, opened 2016-08-01
spec: Allow REST publisher to publish on behalf of another client

mattheworiordan:
See https://www.ably.io/documentation/rest/channels-messages#publish-on-behalf

We need to add `connectionKey` to message definition so that REST clients can publish on behalf of other realtime clients.

@paddybyers I would like to merge this ASAP as it is a blocker for a customer.


paddybyers:
LGTM

PR #143, opened 2016-08-03
Document presence sync etc

SimonWoolf:
As discussed this morning @mattheworiordan @paddybyers


SimonWoolf:
BTW: in writing this I found another implementation difference, which was that in ably-js and ably-java, the presence map stores messages with `present` actions; in ably-ruby, they're stored with whatever action they had when they came in. I figure having them be stored with `present` actions is nicer as it means the actions with realtime presence.get match what you'd get with REST presence.get, so I chose that one to codify; let me know if you disagree.

content/client-lib-development-guide/features.textile
```diff
@@ -468,7 +468,7 @@ h3(#realtime-presence). Presence
...
 ** @(RTP2a)@ All incoming presence messages must be compared for newness with the matching member already in the @PresenceMap@, if one exists, where "matching" means they share the same @memberKey@ (or equivalently, they share both @connectionId@ and @clientId@)
-** @(RTP2b)@ 'Comparing for newness' involves: If the @id@ of an incoming presence message corresponds to its @connectionId@ (defined as the @connectionId@ being an initial substring of the @id@), then the messages are compared lexicographically by @id@. If not, they are compared by timestamp
+** @(RTP2b)@ 'Comparing for newness' involves: If the @id@ of both presence message corresponds to the @connectionId@ (defined as the @connectionId@ being an initial substring of the @id@), then the messages are compared lexicographically by @id@. If not, they are compared by timestamp
```
paddybyers: lexicographic won't work .... `xxx-12 < xxx-2`.

SimonWoolf: Oh. Yeah, you're right. Bugger.

Which means that to do the comparison correctly, we have to split the `id` on `-` and `:` to get the msgSerial and index, pass both through parseInt, and compare first on msgSerial, then index. Though actually we can't even do a simple split because `-` is a valid character in a connectionId as well as the connectionId-msgserial separator, but we could do it with a regex with capture groups - something like `[\w-_]*-(\d+):(\d+)`. Probably put that in the spec, and make sure we never change the format of the `id`.

Is it just me, or is this starting to seem like a crazy thing for us to require the client lib to do for every incoming presence message? Surely we can think of a better way.

@paddybyers how about if realtime adds a logical clock field with every presence message? i.e. a number that defines an ordering between all presence messages that share a connectionId. Effectively an order-preserving injection from {msgSerial, index} to the naturals. Then the client libs could just compare by integer.

paddybyers: I agree that having to parse it to that extent is needlessly complex.

> {msgSerial, index}

Well we could simply use `msgSerial` - ie include that in a `PresenceMessage` - if we also knew that no two `PresenceMessage`s with the same `clientId` would be sent in the same `ProtocolMessage`. In other words, the client ensures that the `PresenceMessage`s within a single `ProtocolMessage` never interfere. I'm not sure we would want to be able to assure that.

However, I agree that the thing that the client `PresenceMap` uses should be opaque to the library, except that it knows it needs to compare it either lexicographically or numerically. We should be able include the timestamp comparison into that check as well.

content/client-lib-development-guide/features.textile
```diff
@@ -465,10 +465,20 @@ h3(#realtime-channel). Channel
...
+* @(RTP2)@ A @PresenceMap@ should be used to maintain a list of members present on a channel. Broadly, this is is a map of "memberKeys":#TP3h to presence messages, all with @PRESENT@ actions (during a sync there may also be ones with an @ABSENT@ action, see "RTP2f":#RTP2f ). Consult the "the Java implementation":https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/realtime/Presence.java#L578-L723 for details.
+** @(RTP2a)@ All incoming presence messages must be compared for newness with the matching member already in the @PresenceMap@, if one exists, where "matching" means they share the same @memberKey@ (or equivalently, they share both @connectionId@ and @clientId@)
+** @(RTP2b)@ 'Comparing for newness' involves: If the @id@ of both presence message corresponds to the @connectionId@ (defined as the @connectionId@ being an initial substring of the @id@), then the messages are compared lexicographically by @id@. If not, they are compared by timestamp
```
mattheworiordan: I think it would be useful to include an actual example of a `ProtocolMessage` and `PresenceMessage` with actual field values to make it easier for devs to understand what to expect without having to inspect actual messages back and forth

content/client-lib-development-guide/features.textile
mattheworiordan: @SimonWoolf @paddybyers are we able to document a solution now based on above?

SimonWoolf: > Well we could simply use `msgSerial` - ie include that in a `PresenceMessage` - if we also knew that no two `PresenceMessage`s with the same `clientId` would be sent in the same `ProtocolMessage`.

Does realtime guarantee to process protocol messages atomically, or could it be made to do so? I.e. if I send a protocol message with two presence messages 'foo enter' and 'foo leave', can it guarantee that no-one will get a SYNC showing a presence set with foo in it?

If so, we could get away with only using `msgSerial` (plus timestamp for synthesized msgs), since SYNCs will only have data from the highest-index presence message for that clientId in the protocol message, and while the PRESENCE messages will have both, we can specify the newness comparison uses >= rather than > and that clients process the presence array in order, so higher-index presence message would just overwrite lower.

SimonWoolf: So, summarising the seven possibilities we have so far:

1. Client library compares msgSerial and index (if comparable, else timestamp). Either:

  1. the client lib parses the msgSerial (and index, for presence messages in `SYNC`s) out of the id (possibly changing `id` format to `aaaaa:0:0` rather than `aaaaa-0:0` so client libs can just split on `:`), or:

  1. realtime starts sending `msgSerial` fields in presence messages in `PRESENCE`s, and both `msgSerial` and `index` fields for presence messages in `SYNC`s. (optionally at this point we can drop the `id` from the wire protocol and have the client lib construct it from the `connectionId`, `msgSerial`, and `index`).

1. Client library compares msgSerial only (if comparable, else timestamp). Either:

  1. Client libs all guarantee that no two `PresenceMessages` with the same `clientId` would be sent in the same `ProtocolMessage`, or:

  1. Realtime guarantees to process `PRESENCE`s atomically, see https://github.com/ably/docs/pull/143#discussion_r73927089 .

1. Client library compares new logical clock field. Either:

  1. Clock is a per-connectionId integer that realtime increments for each new msgSerial and for each index within a msgSerial, or:

  1. Same as previous, but per-channel rather than per-connectionId (so like a channelSerial but per `PresenceMessage` rather than per `ProtocolMessage`), or:

  1. If we guarantee that no more than 999 `presenceMessages` per `protocolMessage`: use `msgSerial` + `index`/1000 (e.g. 34.002); client lib can then compare numerically.

mattheworiordan: @paddybyers I think we agreed we'd digest this over night and try and resolve today.  I'd like to get 0.9 closed down today if possible.  Can you provide your thoughts?

FYI. After thinking about it, I think we should go with `i. 1` personally as parsing an ID in three part is really not an issue AFAIC (we do it already in other places such as the Channel serial), and this system already exists.

paddybyers: > After thinking about it, I think we should go with i.

So, given that:

- we're going for minimum viable change in 0.9 to get the new features we want;

- any logical clock proposal, whilst preferable longer term for various reasons, won't be fully thought-through if we do it immediately, and it'll therefore be wrong.

I agree.

Although I don't like having to do anything non-essential on a per-message basis, I think this is ok; a simple optimisation a library can do is to compare each msg id to verify that they start with the connectionId; if they do, and are the same length, then a lexicographic comparison does the job. If they are different lengths then more work is needed (actually if we guarantee that the string sent by realtime never has leading zeros in the msgSerial part of the id, then a longer msg id is later than a shorter msg id.

mattheworiordan: >  I don't like having to do anything non-essential on a per-message basi

I agree, but it's a very small amount of work, and it only applies to client libraries that are managing a PresenceMap.

mattheworiordan: @SimonWoolf you OK to update?

SimonWoolf: >  if they do, and are the same length, then a lexicographic comparison does the job. If they are different lengths then more work is needed (actually if we guarantee that the string sent by realtime never has leading zeros in the msgSerial part of the id, then a longer msg id is later than a shorter msg id

That doesn't quite work because of the index, e.g. `aaaaaa-10:1` vs `aaaaaa-5:10`. (Solvable by first splitting on `:`, but if you're doing that we may as well split into 3 parts and do it properly).

@paddybyers if we're doing that option, wdyt about changing id format from `aaaaaa-0:0` to `aaaaaa:0:0`, so clients can just split on `:`? Realtime PR with that change: https://github.com/ably/realtime/pull/646

SimonWoolf:
Latest newness-comparison update: 3c31b5c

content/client-lib-development-guide/features.textile
```diff
@@ -465,10 +465,23 @@ h3(#realtime-channel). Channel
...
+* @(RTP2)@ A @PresenceMap@ should be used to maintain a list of members present on a channel. Broadly, this is is a map of "memberKeys":#TP3h to presence messages, all with @PRESENT@ actions (during a sync there may also be ones with an @ABSENT@ action, see "RTP2f":#RTP2f ). Consult the "the Java implementation":https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/realtime/Presence.java#L578-L723 for details.
+** @(RTP2a)@ All incoming presence messages must be compared for newness with the matching member already in the @PresenceMap@, if one exists, where "matching" means they share the same @memberKey@ (or equivalently, they share both @connectionId@ and @clientId@)
+** @(RTP2b)@ To comparing for newness:
+** @(RTP2b1)@ Split the @id@ of both presence messages (which will be of the form @connid:msgSerial:index@, e.g. @aaaaaa:0:0@) on the separator @:@, to give @connectionId@, @msgSerial@, and @index@, parsing the latter two as integers (either eagerly or if/when needed)
+** @(RTP2b2)@ If the @connectionId@s are different, compare numerically by @timestamp@
```
mattheworiordan: When do we compare for newness when the `connectionId` is different as that would mean the `memberKey` is different too?

content/client-lib-development-guide/features.textile
```diff
@@ -465,10 +465,23 @@ h3(#realtime-channel). Channel
...
+** @(RTP2f)@ If a @SYNC@ is in progress, then when a presence message with an action of @LEAVE@ arrives, it should be stored in the presence map with the action set to @ABSENT@. When the @SYNC@ completes, any @ABSENT@ members should be deleted from the presence map. (This is because in a @SYNC@, we might receive a @LEAVE@ before the corresponding @ENTER@).
+** @(RTP2g)@ Any incoming presence message that results in a material change to the presence map (that is: which passes the newness check, and is not a @LEAVE@ message for a member which was not in the presence map or which was in the presence map with an @ABSENT@ action) should be emitted on the @Presence@ object, with an event name set to its original action. Note: this action may not be the same one that it will have when stored in the presence map. For example: an incoming presence message with an @ENTER@ action will be emitted as an @enter@ event, and the emitted presence message will have its action set to @ENTER@. However, it will be stored in the presence map with a @PRESENT@ action.
```
mattheworiordan: I am not sure we agreed on this.  My understanding was that subscribing to raw presence events using `presence.subscribe` would simply receive all events as they arrive.  Why would we change that simply because a member was not already present given there may be use cases when that is needed. Equally I can see this as a useful convenience for users, however preventing a `LEAVE` firing simply because they are not in the MemberMap does not feel right.  Perhaps the later `SYNC` that is ignored because of a later `LEAVE` could be ignored, but that will only ever emit a `present` event anyway.

mattheworiordan:
One or two comments but mostly good. You may want to do a rebase if we make any more changes as it has conflicts now

mattheworiordan:
FYI, if you do rebase, perhaps rebase on https://github.com/ably/docs/pull/132 as there are lots of changes there...

content/client-lib-development-guide/features.textile
SimonWoolf: Synthesized leaves. They have a `connectionId` member that matches the connection that leaves (which is what's used for the `memberKey`), but a 'connectionId-part-of-the-id' (the thing being compared here) that's different. (I'll try rewording it to make it clearer)

mattheworiordan: Still not following how that applies to newness checks.

content/client-lib-development-guide/features.textile
SimonWoolf: Yep, we didn't discuss this - this one was just documenting the current behaviour (for ably-js anyway, I guess not ably-ruby :grinning:). That behaviour was added in [this commit](https://github.com/ably/ably-js/commit/9c4f8d78) to ably-js.

The idea (I'm guessing, @paddybyers can correct me if I'm wrong) is that we can encapsulate the oddness of leaves-arriving-before-presents (due to RTP2f behaviour) inside the library. I.e. someone subscribing to presence events will never see a LEAVE before the corresponding ENTER/PRESENT. (The LEAVE won't be shown because it's a leave for someone not in a present map; then the ENTER won't be shown because it'll fail the newness check against the ABSENT). So we can present a 'cleaned-up' presence interface which actually behaves like a set, in the sense that someone doing things using presence events emitted by the library will not have to worry about implementing tombstones (ABSENT actions); doing the 'naive thing' will just work.

WDYT?

mattheworiordan: Well I believe that approach goes against what was previously agreed which is we're a realtime system so the client libraries should always emit things as they happen.  Any type of queueing or filtering at an event level should never be done, at least that is how I understood it.

mattheworiordan: Also, I am not sure I see the value.  If customers care about who is actually present with the presence map then they will check the presence map.  If they don't, then they're just subscribing to raw events and we shouldn't get in the way of that.

content/client-lib-development-guide/features.textile
SimonWoolf: Example: an item in the presence map:
```
{
  clientId: 'foo',
  connectionId: 'aaaaaa',
  action: 'present',
  id: 'aaaaaa:0:0',
  timestamp: 100
}
```
A synthesized leave arrives:
```
{
  action: 14,
  id: 'xxxxxx:0',
  timestamp: 200,
  presence: [{
    clientId: 'foo',
    connectionId: 'aaaaaa',
    action: 'leave'
}]}
```
On getting this client library adds an id and timestamp to the individual presence message as usual, making it:
```
{
  id: 'xxxxxx:0:0',
  timestamp: 200
  clientId: 'foo',
  connectionId: 'aaaaaa',
  action: 'leave'
}
```
We now compare this for newness with the existing item. Both have a memberKey of `aaaaaa:foo`. So we split the id as usual, but find that the connectionId part of the ids are different, meaning we can't compare the msgSerials. So instead we compare by timestamp.

Would it be clearer if instead I said something like "If either presence message has a connectionId that is not an initial substring of its id, then compare by timestamp"? i.e. comparing the messages to themselves rather than each other. The end result is the same

content/client-lib-development-guide/features.textile
SimonWoolf: Fair enough. I don't feel strongly about this, but we should wait for @paddybyers to weigh in since he was the one who changed the behaviour to this in ably-js.

paddybyers: > subscribing to raw presence events using presence.subscribe would simply receive all events as they arrive

We do, except where the subscriber has already received an event for that member that supersedes the "realtime" event. eg if as a result of a sync we already know someone has left, and we emitted a `leave`, then we won't send an earlier 'update`.

paddybyers: > we shouldn't get in the way of that.

see above; we will get in the way if the out-of-order nature of the events means that the event sequence doesn't make sense.

mattheworiordan: > We do, except where the subscriber has already received an event for that member that supersedes the "realtime" event. 

I have no issue with that, I had an issue specifically with `someone subscribing to presence events will never see a LEAVE before the corresponding ENTER/PRESENT`.  Filtering subsequent "out of date" events after is fine, but filtering a `LEAVE` event because we're waiting for a `SYNC`, `ENTER` or `PRESENT` worries me

SimonWoolf: OK, I'll change it to `Any incoming presence message that passes the newness check`

SimonWoolf:
Updated with latest changes (to RTP2b1, RTP2b2, and RTP2g), rebased on latest source (which includes #132), and squashed down

mattheworiordan:
LGTM.  @paddybyers are you happy for us to merge so that I can lock down on the 0.9 spec now?

paddybyers:
> are you happy for us to merge so that I can lock down on the 0.9 spec now?

yes, lgtm, thanks

PR #148, opened 2016-08-24
Clarify authorize callback behaviour for different connection states

SimonWoolf:


SimonWoolf:
@mattheworiordan @paddybyers Quick question re `Another test should exist where the capabilities are downgraded resulting in Ably explicitly detaching the channel` (which was there before, not one of the things I added):

Currently a channel capabilities failure results in an ERROR action, so  channel fails rather than detaches. I think that's correct (since client is barred from ever attaching to the channel using its current token), and the spec item should be changed to "resulting in Ably failing the channel". Do you agree, or do you think realtime should start using `DETACHED` in that scenario?

SimonWoolf:
@mattheworiordan *poke* I'd like to get this merged asap given various people are now implementing the auth spec.  :+1: / :-1: ?

content/client-lib-development-guide/features.textile
```diff
@@ -277,9 +277,15 @@ h3(#realtimeclient). RealtimeClient
...
+*** @(RTC8a2)@ If the authentication token change fails, then Ably will send an @ERROR@ @ProtocolMessage@ triggering the connection to transition to the @FAILED@ state. A test should exist for a token change that fails (such as sending a new token with an incompatible @clientId@)
+*** @(RTC8a3)@ If the client library language supports callbacks, the callback should called with the new token or error only once realtime has responded to the @AUTH@ with either a @CONNECTED@ or @ERROR@ respectively.
```
paddybyers: Presumably libraries that don't support an async completion mechanism such as callbacks will have a blocking call with a return value indicating success or failure. So this should be worded to accommodate that; ie something like "the `authorise()` call is indicated as complete with the new token or error .... "

content/client-lib-development-guide/features.textile
```diff
@@ -277,9 +277,15 @@ h3(#realtimeclient). RealtimeClient
...
+** @(RTC8b)@ If the connection is in the @CONNECTING@ state when @auth#authorize@ is called, all current connection attempts should be halted, and after obtaining a new token the library should immediately initiate a connection attempt using the new token
+*** @(RTC8b1)@ If the client library language supports callbacks, the callback should called once the connection has moved to either @CONNECTED@, @FAILED@, @SUSPENDED@, or @CLOSED@ (with the new token for the first of those, and an error for the rest)
```
mattheworiordan: `s/callback should called/callback should be called/`, although given @paddybyers comment above, we may want to reword this so that we're not tying the spec to callbacks but instead provide a result.

mattheworiordan: Wouldn't it be easier to to write this like:
`...should be called with the new token once the connection has moved to @CONNECTED@, or with the error once the connection has moved to @FAILED@, @SUSPENDED@, or @CLOSED@`

content/client-lib-development-guide/features.textile
```diff
@@ -277,9 +277,15 @@ h3(#realtimeclient). RealtimeClient
...
+** @(RTC8c)@ If the connection is in the @DISCONNECTED@, @SUSPENDED@, @FAILED@, or @CLOSED@ state when @auth#authorize@ is called, after obtaining a token the library should move to the @CONNECTING@ state and initiate a connection attempt using the new token
+*** @(RTC8c1)@ If the client library language supports callbacks, @RTC8b1@ applies.
```
mattheworiordan: As above, this is unnecessary.

mattheworiordan:
Few comments, otherwise good.  Once this is merged in, please can you:

* [ ] [Update the diff for 0.8 to 0.9](http://messaging-public.realtime.co/documentation/starting-guide/subscription-filters.html) (manual I am afraid, I could not work out how to get Github to do the diff unless you diff everything)
* [ ] Update the unified spreadsheet with these changes
* [ ] Review the iOS and Android prioritisation sheets in case they need to be updated

Thanks @SimonWoolf

SimonWoolf:
> Review the iOS and Android prioritisation sheets in case they need to be updated

I've set RTC8 to priority 2 on the ios priority column. Java doesn't appear to have a priority column, I assume they just look at the ios one

mattheworiordan:
@SimonWoolf not sure you're looking at the right docs, can you see the Client Libraries folder in Google Drive at https://drive.google.com/drive/u/0/folders/0BzHX63l2KaOsWnU5UUVRQTRYRGs?  If so, there are two docs, https://docs.google.com/a/ably.io/spreadsheets/d/1W1WiYqr0_at95kxuyIjE5faxgGGbWnygukEbtbKiPTw/edit?usp=drive_web and https://docs.google.com/spreadsheets/d/1GtLmVBC62u32lXY9rxSjWYmBXKFIZhO3tVv_dwq6ekg/edit#gid=0

mattheworiordan:
@SimonWoolf are you able to do the above?

SimonWoolf:
Ah, sorry, thought you meant the 'priority' column (which is on the ios unified spec sheet only). For that java doc the whole of 0.9 is grouped into a 'Start 0.9 branch' task. The ios doc does have individual spec items for 0.9, but only some of them; added RTC8 to the 'authorize in place' task just in case

mattheworiordan:
Ok, thanks...

PR #152, opened 2016-08-31
Documentation versioning

mattheworiordan:
@SimonWoolf @paddybyers this PR introduces versioning into the docs repo.

You can see this all live at http://docs.ably.io/

How it works loosely is that the supported version info is determined from the latest version of the document spec (see https://github.com/ably/docs/compare/documentation-versioning?expand=1#diff-e5dcf965a880c578eb5d6e1d820d0d02R5 for an example) and also by what versioned files actually exist i.e. you will never be presented with a drop-down for versions where a versioned file does not exist in the `/versions/v0.x/` sub-folder.

The version drop down is language aware i.e. it will only show the versions supported by that lang, and each configured language can have a min or max version supported specified.  A warning is shown if you are looking at an older version of the docs, and also we automatically redirect to a versioned URL if the language is not supported for your current version i.e. if you select Java and you're currently looking at the 0.9 spec, then you are taken to the 0.8 documentation which is at a separate URL.  If you then click back on JS, you will be shown a warning that you are using 0.8 docs.

@renatodeleao / @rafaeljegundo , when you come to implement this in the website, you will need to bear in mind that whilst a lot of this work is done automatically by the docs repo, you will need to implement in the website:
- Render the nav with version links i.e. if you click on History nav when viewing 0.8 docs, you are taken to History 0.8 docs.  See the helper methods for this and implementation in the docs repo, see https://github.com/ably/docs/compare/documentation-versioning?expand=1#diff-e9a0d1b8956e8d3c614f0d579b9d296cR13
- You will need to implement a JS check to see which versions are supported using the global vars and update accordingly.  If the current lang & version is not supported, you will redirect
- Show a warning if looking at older version
- Hide versions in the drop down not supported by the current selected language

All of the above is currently implemented in the docs repo, but it's very hacky...


SimonWoolf:
:+1:

SimonWoolf:
Rebased this on latest `source` in preperation for doing js 0.9 docs

SimonWoolf:
So in doing the js 0.9 docs, more than once now I've been looking at js code examples, switched to ruby or something to check I haven't messed up the formatting and then back to js, then gotten confused until I realised it was now showing me the 0.8 page. There's the `version-not-latest-warning` banner at the top of the page, but that doesn't help much if you're just switching between languages on some code example part way down the page.

I realise we discussed this back when you did this and you felt it was the only sensible behaviour if we wanted the version namespacing to be higher-level (ie in the url) than the language choice. Not sure what a good solution to this is. Perhaps we could at least make the version switcher fixed-position, like the language switcher, to at least make it a bit more obvious when switching languages knocks you back to a different version?

SimonWoolf:
Or.. To be honest, if we're going for a model of version choice at the highest level, and language choice under that, then arguably we shouldn't have switching languages change the version, or at least not silently.

Maybe have version selector appear higher-level than language selector - so eg on the main site, to the left of it on the black bar - and if you select a language that doesn't have a release for the version you're viewing, you get an information banner saying 'The java library does not yet have a 0.9 version; we've switched you to the 0.8 docs'. If you then switch back to the js docs, that's then replaced by the version-not-latest banner.

mattheworiordan:
> Or.. To be honest, if we're going for a model of version choice at the highest level, and language choice under that, then arguably we shouldn't have switching languages change the version, or at least not silently.

Well I am not sure we've even addressed how that will work visually on the website, so I don't think you should assume any visual hierarchy at this stage.

> Maybe have version selector appear higher-level than language selector

Well I am not sure I agree unfortunately as the `Jump to:` and `Show code examples in:` navs will be less effective i.e. three nav in one space.  At least now there is one left aligned & one right aligned.

>  and if you select a language that doesn't have a release for the version you're viewing, you get an information banner saying 'The java library does not yet have a 0.9 version; we've switched you to the 0.8 docs'.

That I have no problem with.  Although perhaps it should be a flash i.e. the user only needs to see it for a short period to be aware of the change.  Having it present on every page would be annoying and unnecessary.

> If you then switch back to the js docs, that's then replaced by the version-not-latest banner. WDYT?

Yes, I agree.

SimonWoolf:
Merged into source as 8575af6

PR #154, opened 2016-09-02
Tutorials example

mattheworiordan:
FYI @paddybyers @SimonWoolf 

I have set up a Pub/Sub Node, JS, and Ruby tutorial.

Now that the framework is in place, it should be a lot easier to add more.  I am also tasking our friends to help translate them into other languages next week so I'll add them as quickly as I can.

@SimonWoolf we'll need tutorials for the adaptors.  Perhaps discuss on Monday, but do give some thought as to what would be nice as a simple starting point, but perhaps also some more fun ones where we convert existing demos for example.

PS. Unfortunately I have not been able to push these to https://docs.ably.io as my connection is too dodgy right now, too many SSL errors.  I'll update later


mattheworiordan:
BTW. I include links to specific SHAs for each example, see https://github.com/ably/tutorials/branches for a list of tutorials that are runnable with one commit for each step.

paddybyers:
LGTM

mattheworiordan:
@SimonWoolf @paddybyers I have added a new client/server token auth demo with Node.js and Ruby.  I think it's good enough to help people understand the paradigm and once ported to other langs it will be pretty useful.  See them all in action now at http://docs.ably.io/tutorials/

@SimonWoolf I could really do with one translator tutorial for PubNub and one for Pusher now.  Can you please figure out a way to perhaps take an existing demo app from both in Javascript (and/or one other lang if you care) and walk through the steps on how to set everything up please?

PR #159, opened 2016-09-11
Fallback host option

mattheworiordan:
@paddybyers @SimonWoolf I propose that we add a new client option `fallbackHostsUseDefault` to simplify the use of our client libraries for larger customers with unique endpoints.

For example, if we have company `foo` and we set up `foo-rest.ably.io` and `foo-realtime.ably.io` for them, they can instance libraries very simply now as follows:

```javascript
var client = new Ably.Realtime({ environment: 'foo' });
```

However, they then lose fallback host functionality.  They could add those back in by passing in the array of fallback hosts manually, but a) that is error prone, b) it seems quite unnecessary.

This PR simply states they can do this:

```javascript
var client = new Ably.Realtime({ environment: 'foo', fallbackHostsUseDefault: true });
```

WDYT?


paddybyers:
LGTM

SimonWoolf:
:+1:

mattheworiordan:
Merged in https://github.com/ably/docs/commit/453baabd16bd5db8b12b732d0d7cfce049054a8e

mattheworiordan:
Thanks, changes moved to unified spec

PR #161, opened 2016-09-14
queryTime and useBinaryProtocol clarifications

SimonWoolf:
queryTime change: [conversation](https://app.intercom.io/a/apps/ua39m1ld/inbox/search/conversations/5931838986?q=querytime) with a customer who was trying to use queryTime on the clientside lib rather than the auth server lib made me think we should explain where you might want to use it

useBinaryProtocol change: disabled by default for browsers, see https://github.com/ably/ably-js/issues/333


paddybyers:
LGTM

mattheworiordan:
Thanks, 👍

PR #174, opened 2016-09-29
tutorials: new SHAs for android tutorials

piotrekinventica:


mattheworiordan:
Does this need to be merged in?  Please advise?

piotrekinventica:
@mattheworiordan not really, since there is a need to adjust Ably Installation instructions in all tutorials, the SHAs will change, so I'll create a new PR with correct ones

PR #181, opened 2016-10-20
Update docs for ably-js 0.9

SimonWoolf:
I'm sure I've missed a few things here and there, but this gets us most of the way there.

(Note: I removed `useTokenAuth` from authOptions -- while nothing changed with 0.9, I think it's misleading to have show this in authOptions, since it has no effect in any method that actually takes an authOptions object directly (i.e. authorize, requestToken, or createTokenRequest), it only actually does anything when used in clientOptions.)

PR is based off of `documentation-versioning` branch, will rebase on master once that's merged.


content/realtime/channels-messages.textile
```diff
@@ -421,9 +433,30 @@ bc[swift]. realtime.channels.get("private:chatroom").attach { error in
...
+
+Note that while fatal errors won't get better on their own, they can be fixable. For example, if a channel goes into the @FAILED@ state due to the client not having the right capabilities to attach to it, that client could use "@Auth#authorize()@":/realtime/authentication#authorize to obtain a new token which does have the right capabilities, then call "@attach()@":/channels-messages#attach on the channel. While the library will not automatically reattach in the @FAILED@ state, explicit calls to "@attach()@":/realtime/channels-messages#attach will make the client try again.
+
+h4(#nonfatal-errors). Nonfatal errors
+
+Other types of error are nonfatal. For example, a client may have network connectivity issues, or a channel may experince a loss of strict message continuity. The library will automatically recover from these events. If channel continuity is lost in the process, the library will notify you though a @resumed@ flag in the @ATTACHED@ event, so you can decide how to handle that failure.
```
paddybyers: s/experince/experience/

and s/The library will automatically recover/The library will automatically attempt to recover/

content/realtime/channels-messages.textile
```diff
@@ -421,9 +433,30 @@ bc[swift]. realtime.channels.get("private:chatroom").attach { error in
...
+
+For every channel @ATTACHED@ event, the "@ChannelStateChange@ object":/realtime/types/#channel-state-change contains a @resumed@ attribute. When true, there has been no loss of continuity from the last time the channel was attached. Where false, there has been a loss of continuity. So for example:
```
paddybyers: s/Where false/When false/

content/realtime/usage.textile
```diff
@@ -194,18 +194,17 @@ The Realtime library needs to have credentials to be able to authenticate with t
...
+The "@ClientOptions#token@":#client-options option takes a token string or "@tokenTetails@":/realtime/types#token-details object, which may have been obtained from some other instance that requested the token. This options is rarely used in production since tokens are short-lived, so generally you would not start with a token without the means to refresh it. The "<span lang="default">@authUrl@ and @authCallback@</span><span lang="csharp">@AuthUrl@ and @AuthCallback@</span><span lang="ruby">@:auth_url@ and @:auth_callback@</span> options":#client-options allow the library to request new tokens or token requests as it needs to; using these options allows the library to be instanced without a <span lang="default">@key@ or @token@</span><span lang="csharp">@Key@ or @Token@</span>, and an initial token will be obtained automatically when required.
```
paddybyers: s/tokenTetails/tokenDetails/

s/This options is rarely/This option is rarely/

content/rest/_request.textile
```diff
@@ -0,0 +1,21 @@
...
+
+- method := either @get@ or @post@.<br>__Type: String__
+- path := the path to query.<br>__Type: String__
+- params := (optional) any querystring parameters needed.<br>__Type: Object__
+- body := (optional; for @post@ method only) the body of the request, as anything that can be serialized into JSON, such as an @Object@ or @Array@.<br>__Type: Serializable__
+- headers := (optional) any headers needed. If provided, these will be mixed in with the default library headers.<br>__Type: Object__
+
+h4. Callback result
+
+On success, @results@ contains an "@HttpPaginatedResponse@":/rest/types#http-paginated-response  containing the @statusCode@ and a @success@ boolean, @headers@, and an @items@ array containing the current page of results. It supports pagination using "@next@":#paginated-result and "@first@":#paginated-result methods, identically to "@PaginatedResult@":/rest/types#paginated-result.
```
paddybyers: Double space?

content/rest/authentication.textile
```diff
@@ -268,11 +268,9 @@ bq(definition).
...
 
-In the event that a new token request is made, the specified <span lang="ruby,python">@token_params@ and @auth_options@</span><span lang="java,csharp">@tokenParams@ and @authOptions@</span> are used.
-
-The <span lang="ruby,python">@token_params@ and @auth_options@</span><span lang="java,csharp">@tokenParams@ and @authOptions@</span> objects will supplement or override the existing client library defaults, and all subsequent explicit or implicit token requests will use these defaults.
+Any <span lang="ruby,python">@token_params@ and @auth_options@</span><span lang="default">@tokenParams@ and @authOptions@</span> objects passed in will entirely replace (as opposed to being merged with) the currently client library saved <span lang="ruby,python">@token_params@ and @auth_options@</span><span lang="default">@tokenParams@ and @authOptions@</span>.
```
paddybyers: s/the currently client library saved/the currently saved/

content/types/_realtime_client_options.textile
```diff
@@ -10,7 +10,7 @@
...
 
-- <span lang="default">queryTime</span><span lang="csharp">QueryTime</span><span lang="ruby">:query_time</span> := _false_ If true, the library will query the Ably servers for the current time instead of relying on a locally-available time of day. Knowing the time accurately is needed to create valid signed token requests, so this option is useful for library instances on auth servers if the server clock is untrustworthy<br>__Type: @Boolean@__
+- <span lang="default">queryTime</span><span lang="csharp">QueryTime</span><span lang="ruby">:query_time</span> := _false_ If true, the library will query the Ably servers for the current time instead of relying on a locally-available time of day. Knowing the time accurately is needed to create valid signed token requests, so this option is useful for library instances on auth servers if the server clock is untrustworthy. The server is queried only once per client library instance, and used to calculate and store the offset from the local clock.<br>__Type: @Boolean@__
```
paddybyers: I think "untrustworthy" gives the wrong implication. There isn't a security risk arising from an inaccurate clock; it is just that a token request will be rejected. So I think it would be more appropriate to say the server clock is insufficiently accurate.

SimonWoolf:
Amended based on Paddy's feedback

content/realtime/authentication.textile
```diff
@@ -112,6 +113,12 @@ Below is a rather contrived yet straightforward example that instances a Realtim
   AblyRealtime realtime = new AblyRealtime(options);
 ```
 
+h4(#token-upgrading). Upgrading the token on a connection
+
+In version 0.9 of the client libraries, the token that a connection uses can be upgraded on the fly. When a token is nearly expired, Ably will prompt the library to request a new token (using the @authUrl@ or @authCallback@). Once the library has a new token, it will upgrade the connection to use the new token. This all happens automatically and seamlessly.
```
mattheworiordan: Technically the token could be issued by the client library itself if it has an API key, but I would agree that it's not relevant in most cases so you can leave like this.

content/realtime/authentication.textile
```diff
@@ -252,12 +257,23 @@ h4. Parameters
...
 
-  On failure to obtain a token, @err@ contains an <span lang="ruby">"@ErrorInfo@":#error-info</span><span lang="objc,swift">@NSError@</span> object with an error response as defined in the "Ably REST API":/rest-api#common documentation.
+  On failure to obtain a token, @err@ contains an @NSError@ object with an error response as defined in the "Ably REST API":/rest-api#common documentation.
+
+blang[jsall].
+  h4. Callback result
+
+  If using the REST client, the callback will be called once the call to obtain a new token returns, with either @tokenDetails@ as a "@TokenDetails@":#token-details object if the call succeeded, or @err@ as an "@ErrorInfo@":#error-info (with an error response as defined in the "Ably REST API":/rest-api#common documentation) if it failed.
+
+  If using the realtime client, if the call to obtain a new token failed, the callback will be called with @err@ as with the REST client. If it succeeded, the callback will be called (with the new @tokenDetails@) only once the realtime connection has been successfully upgraded to use the new token (or if the upgrade failed, once the connection state has moved to @FAILED@).
```
mattheworiordan: `s/realtime client/Realtime client/g` as we refer to the client lib with its namespace generally as `Realtime`.

content/realtime/authentication.textile
```diff
@@ -252,12 +257,23 @@ h4. Parameters
...
+
+  If the client was not previously connected, the callback will be called with @tokenDetails@ once it has successfully connected with the new token (or closed, failed, or suspended, with an @err@, should that happen first).
```
mattheworiordan: I can't recall why we decided on doing the auto-connect.  It does seem a little odd now when reading the docs, although I can equally see almost no situation where someone using the Ably Realtime lib would not want to connect.

content/realtime/authentication.textile
```diff
@@ -252,12 +257,23 @@ h4. Parameters
...
+
+  So for realtime clients, you can use the @authorize@ callback to find out when the new token has taken effect. This can be useful if, for example, you want to do something with some new capability carried by the new token.
```
mattheworiordan: `s/realtime clients/Realtime clients/g`

content/realtime/channels-messages.textile
```diff
@@ -246,28 +246,36 @@ channel.subscribe("action") { message in
 }
 channel.publish("action", data: "boom!")
 
+<div lang="jsall">
```
mattheworiordan: I don't think this needs to be limited to `jsall`?

content/realtime/channels-messages.textile
```diff
@@ -246,28 +246,36 @@ channel.subscribe("action") { message in
...
+The @Channel@ object is an @EventEmitter@ and emits an event whose name is the new state whenever there is a channel state change. <span lang="jsall">Listeners are passed a "ChannelStateChange":#channel-state-change object in the first argument. This contains the @current@ and @previous@ states, a @resumed@ flag which indicated whether message continuity on this channel is preserved (from the last time that the channel was attached), and (sometimes) a @reason@ for the state change. As with all events from an @EventEmitter@ in the Ably library, @this@ within the listener function is a reference to an event object whose @event@ property is the name of the event that fired. This allows a listener to listen for all events with a single registration and know, which event fired when it is entered.</span>
```
mattheworiordan: Most of this is not `jsall` specific so should not be wrapped in a lang span

content/realtime/channels-messages.textile
```diff
@@ -376,7 +388,7 @@ h3(#handling-failures). Handling channel failures
 
 Channel attach and detach operations are synchronous - after initiating an attach request, the client must wait for a response from Ably that confirms that the channel is established on the service.
 
-There are cases where an attach will fail deterministically, such as if the client doesn't have the necessary rights to access the channel.
+The client libraries will attempt to automatically recover from non-fatal error conditions. However, you can handle them yourself if you prefer. <span lang="jsall">For explicit attaches with a callback, both fatal and nonfatal errors will result in the callback being called with an @ErrorInfo@ argument</span>
```
mattheworiordan: It would be better if we can word this so it's not JS specific

content/rest/_request.textile
```diff
@@ -0,0 +1,21 @@
...
+  default: request
+
+bq(definition).
+  default: request(String method, String path, Object params, Object body, Object headers, callback("ErrorInfo":/rest/types#error-info err, "HttpPaginatedResponse":/rest/types#http-paginated-response results)))
+
+Makes a REST request to a provided path. This is provided as a convenience for developers who wish to use bleeding edge REST API functionality that is either not documented or is not yet included in the public API, without having to handle authentication, paging, fallback hosts, MsgPack and JSON support, etc. themselves.
+
+h4. Parameters
+
+- method := either @get@ or @post@.<br>__Type: String__
```
mattheworiordan: why not `put`?

content/types/_http_paginated_response.textile
```diff
@@ -0,0 +1,17 @@
+An @HttpPaginatedResponse@ is a "@PaginatedResult@":#paginated-result that carries additional information about the response to an HTTP request. It is used when "making custom HTTP requests":/rest/usage#request.
+
+h4.
+  default: Properties
+  java:    Members
+  ruby:    Attributes
+  python:  Attributes
+
+Includes all properties that "@PaginatedResult@":#paginated-result has, with the additional constraint that @items@ is guaranteed to be an @Array@. Additionally, it has:
```
mattheworiordan: I am not sure that it's really all that relevant to users that this class derives from PaginatedResult.  It's useful to provide that info perhaps, but I don't think we should expect people to jump back & forth between this documentation and the PaginatedResult documentation.  Instead duplication of the documentation would be better,

content/types/_http_paginated_response.textile
```diff
@@ -0,0 +1,17 @@
+An @HttpPaginatedResponse@ is a "@PaginatedResult@":#paginated-result that carries additional information about the response to an HTTP request. It is used when "making custom HTTP requests":/rest/usage#request.
```
mattheworiordan: I think it's generally unwise to use an anchor link that is not a full path in a partial as you don't know that the other partial exists.

content/rest/usage.textile
```diff
@@ -316,6 +319,10 @@ blang[java,ruby,php,csharp].
 
   Failure to retrieve the Ably server time will raise an "@AblyException@":/rest/types#ably-exception.
 
+<div lang="jsall">
+<%= partial 'rest/_request' %>
+</div>
+
 h2(#related-types). Related types
```
mattheworiordan: You have not included the `HttpPaginatedResponse` type in Related types.  Also, you have not added this method to the Realtime docs.  This method is available for both REST & Realtime

content/client-lib-development-guide/features.textile
```diff
@@ -164,7 +164,7 @@ h3(#rest-auth). Auth
 ** @(RSA10g)@ Stores the @AuthOptions@ and @TokenParams@ arguments as defaults for subsequent authorisations with the exception of the attributes @TokenParams#timestamp@ and @AuthOptions#queryTime@
 ** @(RSA10h)@ Will use the value from @Auth#clientId@ by default, if not @null@
 ** @(RSA10i)@ Adheres to all requirements in @RSA8@ relating to @TokenParams@, @authCallback@ and @authUrl@
-** @(RSA10l)@ Has an alias method @RestClient#authorise@ and @RealtimeClient#authorise@ that will log a deprecation warning stating that this alias method will be removed in @v1.0@ and the user should instead use @authorize@
+** @(RSA10l)@ Has an alias method @RestClient#authorize@ and @RealtimeClient#authorize@ that will log a deprecation warning stating that this alias method will be removed in @v1.0@ and the user should instead use @authorize@
```
SimonWoolf: oops! that was silly of me; danger of automated search and replace :)

content/realtime/authentication.textile
SimonWoolf: I can't remember exactly, but I think the argument was mostly around people doing it in `failed` (specifically when it was in `failed` because of an auth problem) -- we didn't want people calling authorise, seeing the lib was still in the failed state, and thinking it hadn't worked (being that if the lib is connected authorise only calls back once the new token has actually taken effect on the connection, and if it calls back and the lib is in `failed`, that means something went wrong). And if doing it for `failed`, may as well do it for all terminal states for consistency.

(I agree this is a bit counterintuitive for `closed` in particular, given that that's a result of an explicit user action - we could maybe change the behaviour for that state? don't mind either way)

content/realtime/channels-messages.textile
SimonWoolf: Not sure what you're thinking specifically? I mean, once other languages do 0.9 they can replace the js-specific bits with their own language equivalents (e.g. `the callback being called` with `<span lang="jsall">the callback being called</span><span lang="ruby">the onError block being yielded</span>` as necessary, but here I was focusing just on js

content/rest/_request.textile
SimonWoolf: mostly because currently, PaginatedResource, Resource, Http, and Request currently all only support `get` and `post`, so I know those work. Adding `put` would definitely be possible, but not completely trivial (for one thing we'd need to change how jsonp works in realtime as it currently just uses the presence of a body to decide if get/post (probs just add a `method` param), make sure it works with xhr in old versions of IE, etc. etc.)

So we could certainly do it if we think it's important, but if so we should add it to the spec and decide how we want to prioritise it

paddybyers: RIght, but the reason for that is that the paginated queries were added originally for the methods that return large collections - principally `GET`. That never applied to `PUT`. Now for this generic mechanism we want to be able to support all relevant methods, and support pagination in the case of `GET` at least. So we do need to know that `PUT`` works, although we don't necessarily need to handle `PUT` responses with pagination.

content/realtime/authentication.textile
mattheworiordan: Nope I think we should leave as is.  I do worry though we may regret that sometime down the line, but equally I can see reasons why we did this.

content/realtime/channels-messages.textile
mattheworiordan: Ok, fair enough, leave as is then.

paddybyers:
> I thought a generic HTTP request is made using the XHR object as this is simply a REST method

No, because in browsers that don't support CORs we can't use XHR. So each browser comet transport exposes a generic HTTP primitive which is used for REST operations, that that includes JSONP.

So, @SimonWoolf  is right we don't support `PUT` at all right now. We would support it in JSONP I guess by having an optional `method` param (or do we do that already?)

mattheworiordan:
> So, @SimonWoolf is right we don't support PUT at all right now. We would support it in JSONP I guess by having an optional method param (or do we do that already?)

We must have a mechanism like that to support `POST` no?

SimonWoolf:
>  We must have a mechanism like that to support POST no?

 As I said in my last comment, it currently just uses the presence of a body to decide if get or post; we'd need to add support for a method param.

> If the JS lib has some constraints around lack of PUT support then that is a constraint of the library

I don't think it matters much if we have differences in supported methods between libraries, but I do think that if we support a method in ably-js, it needs to be supported for all platforms that ably-js works on. Or people are just going to use it, only test it on chrome, then complain at us when their site is broken for customers using IE8.

> we don't necessarily need to handle PUT responses with pagination.

Currently we return the same type (httpPaginatedResponse) whatever the method used; are we sure that's something we want to change?

Look - sure, we should support PUT. But given we currently have no endpoints that use it -- currently it isn't even a CORS permitted method! -- I really don't think we need to have it as a blocker to 0.9 client libs. Lets add support for it to realtime first (router2 fixes; jsonp method param support, and an endpoint that makes PUT requests testable by client libs - maybe picking some api call that could use it), then we can start thinking about adding it to client libraries.

paddybyers:
> are we sure that's something we want to change?

No, I wasn't suggesting that, just saying that there isn't a requirement for pagination to be supported, in case that makes anything simpler.

> Look - sure, we should support PUT. But given we currently have no endpoints that use it -- currently it isn't even a CORS permitted method! -- I really don't think we need to have it as a blocker to 0.9 client libs. Lets add support for it to realtime first (router2 fixes; jsonp method param support, and an endpoint that makes PUT requests testable by client libs - maybe picking some api call that could use it), then we can start thinking about adding it to client libraries.

I agree.

mattheworiordan:
>> Look - sure, we should support PUT. But given we currently have no endpoints that use it -- currently it isn't even a CORS permitted method! -- I really don't think we need to have it as a blocker to 0.9 client libs. Lets add support for it to realtime first (router2 fixes; jsonp method param support, and an endpoint that makes PUT requests testable by client libs - maybe picking some api call that could use it), then we can start thinking about adding it to client libraries.

> I agree.

Sorry, I don't.  I don't think it's a blocker to ably-js, ably-js can simply raise an exception when using `PUT` or `PATCH` stating is unsupported.  But in all other client libraries, it's trivial to support it.

The point of this API is solely to expose future unkown features to customers.  By stating that we don't currently have any methods that use `PUT` or `PATCH`, so therefore we don't need to worry about it, goes against the very reason this method exists at all.

My vote is to add support to the spec, and if it's tricky in JS to implement, then raise an exception now.  Then when we actually need PUT or PATCH in the future for any new endpoint we release, we can implement it in JS at that time.  For all other libraries, most of them simply support a lit of verbs anyway and will immediately support that functionality out of the box.

SimonWoolf:
> My vote is to add support to the spec, and if it's tricky in JS to implement, then raise an exception now. Then when we actually need PUT or PATCH in the future for any new endpoint we release, we can implement it in JS at that time. For all other libraries, most of them simply support a lit of verbs anyway and will immediately support that functionality out of the box.

Sure, that's fine by me

content/rest/_request.textile
```diff
@@ -0,0 +1,21 @@
...
+- path := the path to query.<br>__Type: String__
+- params := (optional) any querystring parameters needed.<br>__Type: Object__
+- body := (optional; for @post@ method only) the body of the request, as anything that can be serialized into JSON, such as an @Object@ or @Array@.<br>__Type: Serializable__
+- headers := (optional) any headers needed. If provided, these will be mixed in with the default library headers.<br>__Type: Object__
+
+h4. Callback result
+
+On success, @results@ contains an "@HttpPaginatedResponse@":/rest/types#http-paginated-response containing the @statusCode@ and a @success@ boolean, @headers@, and an @items@ array containing the current page of results. It supports pagination using "@next@":#paginated-result and "@first@":#paginated-result methods, identically to "@PaginatedResult@":/rest/types#paginated-result.
```
paddybyers: Are we being clear enough here what constitutes success? Reading this I might think a completed request that returns an "error" statusCode such as 400 would be a success.

Are we going to define what constitutes a success error code for every method? or any 2xx ?

paddybyers:
> My vote is to add support to the spec, and if it's tricky in JS to implement, then raise an exception now.

FIne.

content/rest/_request.textile
SimonWoolf: > Reading this I might think a completed request that returns an "error" statusCode such as 400 would be a success.

Thing is that in js, since you have the `(err, res)` thing, if the request returns a 400, you don't get an `HttpPaginatedResponse` at all (you get an `ErrorInfo`). So the `success` boolean is a bit tautological. (Technically could still be false if realtime returned a 100 or something, I guess). I could probably make that clearer; will edit the docs.

Languages without the same convention (eg where you might get an HPR even on error if that'd be idiomatic for that language) will say what they do when they add themselves to the docs

> Are we going to define what constitutes a success error code for every method? or any 2xx ?

`HP5` defines it as 200 <= statusCode < 300

SimonWoolf:
Made requested changes.

(btw -- in duplicating the @PaginatedResult@, noticed it seems to have @items@ as both an attribute and a method, for all languages. Not sure how that happened, in js it's just an attribute, which is what spec says. Not sure what languages have it as a method, something to check at some point, I'll file an issue)

mattheworiordan:
Thanks, LGTM.

BTW. On JS, I assume we should only throw an exception in a browser env, but allow other environments i.e. Node, React Native, Phonegap etc?

SimonWoolf:
> BTW. On JS, I assume we should only throw an exception in a browser env, but allow other environments i.e. Node, React Native, Phonegap etc?

Once I add support for `put` to ` PaginatedResource`, `Resource`, and `Http`, sure. Js lib has a lot of abstraction layers :)

mattheworiordan:
@SimonWoolf @paddybyers I have added quite a few commits to address some shortfalls I noticed, add Ruby 0.9 support, plus ensure that future languages are arguably easier to do. Whilst the docs for those langs may be wrong with what I've done, at least we'll know that the definitions only need to be checked as opposed to working out what was added in the 0.9 JS docs that was not added for the target lang.

mattheworiordan:
@paddybyers you happy for us to merge this in now?

paddybyers:
LGTM

SimonWoolf:
Merged into source as 7063543

PR #182, opened 2016-10-20
Add Message.fromEncoded and Message.fromEncodedArray specs

SimonWoolf:
as requested


mattheworiordan:
Nice, LGTM, @paddybyers?

paddybyers:
> @paddybyers?

I would say that it decodes to the maximum extent possible, and leaves the residual encoding transforms in the `encoding` property of the returned message. Otherwise LGTM.

tcard:
It's probably worth it to link this with other descriptions in the spec about how messages are encoded/decoded, so that we have it defined in a single place and to specify that what this does and what normal message enc/decoding does is the same.

SimonWoolf:
Changes made as requested.

ably-js implementation: https://github.com/ably/ably-js/pull/344

PR #186, opened 2016-11-01
Clarify that authorize can be given just a token

SimonWoolf:
tpr was pointing out in #client-libraries today that the auth#authorize spec doesn't make clear whether you can give it a token and it should use that -- https://ably-real-time.slack.com/archives/client-libraries/p1478012497000052. Proposed clarification


trenouf:
Thanks, LGTM.

content/client-lib-development-guide/features.textile
```diff
@@ -1057,6 +1058,8 @@ h4. AuthOptions
 * @(AO1)@ A class providing configurable authentication options used when authenticating or issuing tokens explicitly. These options are used when invoking @Auth#authorize@, @Auth#requestToken@, @Auth#createTokenRequest@ and @Auth#authorize@
 * @(AO2)@ The attributes of @AuthOptions@ consist of:
 ** @(AO2a)@ @key@ string - Full Ably key string, as obtained from dashboard, used when signing token requests locally
+** @(AO2h)@ @token@ string - An authentication token string issued for this application
+** @(AO2i)@ @tokenDetails@ @TokenDetails@ - An authentication token issued for this application
```
mattheworiordan: Why `and associated details`?

paddybyers: I mean just saying "an authentication token" might be understood to be the token string, whereas here we're talking about a `TokenDetails`, ie including expiry and other attributes.

SimonWoolf: Added 0203f19d38b9095b3ec9706532dace0184fb416c to clarify tokenDetails

SimonWoolf:
Addressed comment, believe this is ready for merging

mattheworiordan:
I don't specifically have anything wrong with this proposal, but I am wondering why we are introducing this complexity.  Under what conditions will people provide a token value and a means to obtain a token in reality?  Also, should we even encourage people to provide tokens directly as part of AuthOptions when in most cases this will be wrong?

Is there a strong enough use case to warrant this additional complexity versus simply rejecting authorize requests where customers explicitly provide a token?

The only situation I can think is that perhaps a client receives a message (as a normal message or via another means) notifying it it should get a new token, and instead of just calling authorize they instead get the token and pass it to authorize.  In that case though, why don't we rather encourage people to just call `authorize` and be done with it?

content/client-lib-development-guide/features.textile
```diff
@@ -107,6 +107,7 @@ h3(#rest-auth). Auth
 *** @(RSA4c1)@An @ErrorInfo@ with @code@ @80019@ and description of the underlying failure should be emitted with the state change, in the @errorReason@ and/or in the callback as appropriate
 *** @(RSA4c2)@If the connection is @CONNECTING@, then the connection attempt should be treated as unsuccessful, and as such the connection should transition to the @DISCONNECTED@ or @SUSPENDED@ state as defined in "RTN14":#RTN14 and "RTN15":#RTN15
 *** @(RSA4c3)@If the connection is @CONNECTED@, then the connection should remain @CONNECTED@
+** @(RSA4d)@ If multiple @authOptions@ are used to initialize the library, the preference ordering among them is identical to @Auth#authorize@, defined in @RSA10e@
```
tcard: I think "multiple `authOptions`" is confusing. What's an `authOption`? The formatting seems to indicate it's something formal in the code, like a specified field or type.

content/client-lib-development-guide/features.textile
```diff
@@ -159,7 +160,7 @@ h3(#rest-auth). Auth
...
-** @(RSA10e)@ Adheres to the implementation of @requestToken@ when issuing new tokens
+** @(RSA10e)@ If the @authOptions@ contains a way of obtaining a token (an @authCallback@, @authUrl@, or @key@), that should be used to obtain a new token, as per @requestToken@ (@RSA8@). If it contains a token (@token@ or @tokenDetails@), that should be used as-is. If it contains both a token and a way of obtaining a token, the token should be used, with the way of obtaining a token being stored per @RSA10g@ for when the token expires. (Ordering of preference within those groups is not defined and is left up to individual implementations)
```
tcard: Nitpick: either "`authOptions`" or "the `AuthOptions`". (I prefer the latter since we don't name the arguments in RSA10j, just the types.)

tcard:
> I am wondering why we are introducing this complexity

I think this proposal reduces complexity, and instead making a special case out of it like causing `authorize` to fail would increase it.

Plus, if we give them liberty to provide a token on initialization I don't see why we should restrict it afterwards. We ask for `authUrl`, `authCallback`, etc. so that we can reauthorize at any point on our terms but the whole point of `authorize` is to let them reauthorize at any point too, and `authUrl`/`authCallback` is not specially relevant in that case.

mattheworiordan:
> Plus, if we give them liberty to provide a token on initialization I don't see why we should restrict it afterwards. We ask for authUrl, authCallback, etc. so that we can reauthorize at any point on our terms but the whole point of authorize is to let them reauthorize at any point too, and authUrl/authCallback is not specially relevant in that case.

Ok, but let's push this into `0-10` please and base on that branch

mattheworiordan:
Can this be rebased onto `0-10-source` please?

mattheworiordan:
@SimonWoolf this now needs to be rebased on `master-1-1`

content/client-lib-development-guide/features.textile
mattheworiordan: Agreed.

content/client-lib-development-guide/features.textile
mattheworiordan: 👍 for this change.

SimonWoolf:
Merged into master-1-1 as 6ce7fba73c763ac159f967cec43fc7b76dffc353

Issue #176, opened 2016-10-10
Presence tutorial


mattheworiordan:
@SimonWoolf taking a stab at some ideas for a presence tutorial.

We need a client-side tutorial (Javascript + optionally Ruby EM if you have time) & a client-server example for (Javascript + Node.js, optionally Javascript + Ruby).

See https://www.ably.io/tutorials/publish-subscribe#lang-javascript for a client-side related example.
See https://www.ably.io/tutorials/token-authentication#lang-nodejs for a client-server related example.

For the client-side example perhaps we could show:
- Into - simple explanation of what presence is and how it can be used.
- Step 1 - Set up (this is a generic partial)
- Step 2 - Install Ably (generic)
- Step 3 - Enter channel (let's use a room analogy). Explain that attaching to a channel does not mean you've entered. You need to explicitly enter (avoid any mention of multiple client support, too much info at this stage). We also need to demonstrate at this point that a clientId must be specified. Need to mention explicitly that for this example for simplicity we're assigning a clientId with basic auth, but we don't recommend this in production and should issue a token server-side, with a link to the client-server examples.
- Step 4 - Get current members - explain this is done following the initial sync on the channel
- Step 5 - Subscribe to presence changes - explain that you can subscribe to enter, leave, update events. 
- Live demo - I think we should allow people to enter a room, and change their avatar. So they enter their name, choose an avatar, and enter.  They can then leave at any point, or update their avatar.  Nice to have a raw console + nice graphic of the avatars and people in the virtual room (just a div I expect) will do. See http://www.freepik.com/free-vector/black-and-white-avatars_713015.htm for some possible avatars

For the client-server example perhaps we could show:
- Into - simple explanation of what presence is and how it can be used.
- Step 1 - Set up (this is a generic partial)
- Step 2 - Install Ably on your server (generic)
- Step 3 - Set up your web server
- Step 4 - Enter client (don't worry about auth, again just link to the client-server auth example and explain it's not relevant to this example)
- Step 5 - iFrame with button to get current presence members server-side
- Step 6 - iFrame with button to get presence member history

SimonWoolf:
from slack:

> 21:02 Might be worth only doing the client-side one to keep it simple
> 21:02 The client-server one we can perhaps do in the future
> 21:02 Really not needed now, better to have one example

PR #187, opened 2016-11-02
First draft of presence tutorial

SimonWoolf:
Fixes https://github.com/ably/docs/issues/176.

Haven't added the code to the tutorials repo yet; once the code snippets are finalised, then will copy it over (so don't have to keep changing commit hashes). (In particular, wasn't sure whether we want to have explicit attaches or rely on `enter` and `get` implicitly attaching -- went for the former for now)

(The avatars are just emoji and are sent over Ably as unicode chars, which is nice as it means the demo is standalone, anyone copying it to play around with won't need access to images hosted on our website. (Though I did add add an emoji font just to be sure no-one'll get �'s))

presence


content/tutorials/presence.textile
```diff
@@ -183,7 +183,7 @@ h2. Next steps
 <script src="https://apis.google.com/js/api.js"></script>
 <script type="text/javascript">
   window.addEventListener("load", function() {
-    var ably = new Ably.Realtime({ authUrl: 'https://www.ably.io/ably-auth/token/docs' }),
+    var ably = new Ably.Realtime({ authUrl: 'https://www.ably.io/ably-auth/token/docs', closeOnUnload: true }),
```
mattheworiordan: Perhaps a comment here on why closeOnUnload is used?  I appreciate it's not public code, but devs will inspect

content/tutorials/presence.textile
```diff
@@ -0,0 +1,306 @@
...
+
+h2. Step 3 - Enter the presence set
+
+With Ably, you can be "attached":/realtime/channels-messages#channels to a channel (that is: be receiving messages published on that channel, or publishing them yourself) without necessarily being present on the channel. If you want to be present, and be seen by other users, you need to explicitly enter the presence set. For example, for a channel called "chatroom":
+
+blang[javascript,nodejs].
+  ```[javascript]
+    var channel = realtime.channels.get('chatroom');
+    channel.attach(function(err) {
+      console.log('If no err, we're now attached to the channel');
```
mattheworiordan: I kind of feel that you should either do `if(!err) { ... }` or get rid of `If no err, ` because it's just confusing in the console seeing `If not err, we're now attached` i.e. are we attached, are we not, what the hell is going on? 😄 

Perhaps instead just a single line above `if (err) { return console.error("Error attaching", err) }`?

content/tutorials/presence.textile
```diff
@@ -0,0 +1,306 @@
...
+    end
+  ```
+
+  "See this step in Github":https://github.com/ably/tutorials/commit/######################TODO
+
+We're now in the presence set. The name we'll appear as is the @client ID@ that we included in the Realtime constructor in step 2.
+
+Similarly, we can call @channel.presence.leave@ to leave the presence set.
+
+(Side note: while in this tutorial we had free choice of client id, in most situations you don't want that. Instead you want your server to constrain what @client ID@ people can use -- for example, to their logged-in username -- so that no-one can pretend to be someone else. When using "Token authentication":/general/authentication#token-authentication, your server can issue tokens with a specific @client ID@. When such a token is used by someone, their connection is bound to that @client ID@.)
```
mattheworiordan: You refer to `id` as `ID` everywhere else so perhaps remain consistent?

content/tutorials/presence.textile
```diff
@@ -0,0 +1,306 @@
...
+        puts "There are #{members.length} clients present on this channel"
+        first = members[0]
+        puts "The first member is #{first.clientId} and their data is #{first.data}"
+      end
+    end
+  ```
+
+  "See this step in Github":https://github.com/ably/tutorials/commit/######################TODO
+
+As soon as you attached to the channel, it started to sync with Ably. When you make this call to get the presence set, it will (by default) wait until it has finished syncing before completing. For small presence sets this will take only a few milliseconds; for larger ones (with hundreds of members) the full sync can take a little longer.
```
mattheworiordan: s/you attached to the channel, it started to sync/you are attached to the channel, the presence set starts to sync with Ably/g

content/tutorials/presence.textile
```diff
@@ -0,0 +1,306 @@
...
+
+  We recommend that you clone the repo locally:
+
+  bc[sh]. git clone https://github.com/ably/tutorials.git
+
+  Checkout the tutorial branch:
+
+  bc[sh]. git checkout presence-javascript
+
+  And then run the demo locally by adding your "Ably API key":https://support.ably.io/support/solutions/articles/3000030502-setting-up-and-managing-api-keys to @example.html@ and opening the page in your browser.
```
mattheworiordan: You specify this is Node.js yet it's not, it's JS code for a browser?

content/tutorials/presence.textile
```diff
@@ -0,0 +1,306 @@
...
+        });
+      }
+    });
+
+    $('button#enter-presence').on('click', function() {
+      clientId = $.trim($('input#client-id').val());
+      var avatar = $('input[name=avatar]:checked').val();
+
+      if(!clientId) {
+        log('You need to specify a client ID before you can enter presence');
```
mattheworiordan: You refer to client ID yet the field is called Name?

content/tutorials/presence.textile
```diff
@@ -0,0 +1,306 @@
...
+</div>
+
+Enter your name and choose an avatar, then click 'Enter' to enter the presence set, and 'Leave' to leave it again. <a href="#" id="new-browser" target="_blank">Open this demo in other browser windows</a> and enter presence in them as well, and see how every window shows an up-to-date presence set.
+
+Once you're entered, try tapping a different avatar to change it, and seeing it change in all the other windows in real time.
+
+<div>
+  <div>
+    <p>
+      Name: <input type="text" id="client-id" value="">
```
mattheworiordan: <label for="client-id">Name</label>. Same for Avatar below.

content/tutorials/presence.textile
```diff
@@ -0,0 +1,306 @@
...
+      <% ['👦', '👧', '👨', '👩', '😺'].each do |emoji| %>
+        <label><input type="radio" name="avatar" value="<%= emoji %>" /><span class="emoji"><%= emoji %></span></label>
+      <% end %>
+    </p>
+
+    <button id="enter-presence">Enter</button>
+    <button id="leave-presence">Leave</button>
+  </div>
+
+  <ul id="presence-set">
```
mattheworiordan: We should have an empty message "No one is present yet"

content/tutorials/presence.textile
```diff
@@ -0,0 +1,306 @@
...
+      clientId = $.trim($('input#client-id').val());
+      var avatar = $('input[name=avatar]:checked').val();
+
+      if(!clientId) {
+        log('You need to specify a client ID before you can enter presence');
+        return;
+      }
+
+      log("[Entering...] ");
+      channel.presence.enterClient(clientId, avatar, function(err) {
```
mattheworiordan: I find it a bit of an add decision to use `enterClient` when that's not really an API anyone would be using, and not the API we've been talking about in the tutorial either.  I would really prefer if we simplified this significantly by breaking the demo into into two stages.

### 1. Enter your name and connect

Name: [pre-populated]
[Connect to Ably]

(section below is disabled or not visible, and once connected, name becomes disabled and connect to ably changes to a disabled button "Connected to Ably..."

### 2. Enter channel "foo" 

Avatar: 👦' 👧' 👨' 👩' [😺 ]
[Enter]

When they press Enter, the button changes to [Leave] with a comment "Choose a new Avatar above to update presence in real time"

If you do this I think all of this code would be simplified immensely.

content/tutorials/_install_ablyrealtime.textile
```diff
@@ -0,0 +1,336 @@
+<%
+  client_id = 'jamie'
+%>
+h2.
+  default: Step 2 - Install Ably
+  swift: Step 2 - Setup an Xcode project and install Ably
```
mattheworiordan: You're assuming we support Swift / Android in our presence demo, which we won't for now

content/tutorials/_install_ablyrealtime.textile
```diff
@@ -0,0 +1,336 @@
...
+  ```[ruby]
+    require 'ably'
+    EventMachine.run do
+      ably = Ably::Realtime.new(key: api_key<%= ", client_id: #{client_id}" if @use_client_id %>)
+    end
+  ```
+
+  "See this step in Github":https://github.com/ably/tutorials/commit/0ec2c7e
+
+blang[php].
```
mattheworiordan: We're not support PHP / Python etc?

content/tutorials/publish-subscribe.textile
```diff
@@ -14,339 +14,7 @@ Messages published can contain string, JSON object, JSON array or binary data pa
...
-
-  Add the following to the file in which you imported the Ably library:
-
-  ```[swift]
-    let API_KEY = "INSERT-YOUR-API-KEY-HERE"
-    let client = ARTRealtime(key: API_KEY)
-  ```
-
-  "See this step in Github":https://github.com/ably/tutorials/commit/59dcc56
+<%= partial 'tutorials/_install_ablyrealtime', vars(why_realtime: "subscribe to messages") %>
```
mattheworiordan: I think it's a mistake to try and share this.

You're adding complexity to the markup, but also a lot of the shared content means we cannot release this tutorial now until we have presence demos for every language.  We do not want to do that.

SimonWoolf: I'm happy to expand this back out into duplicates, but TBH I'm a bit surprised to see you object to this, seemed like an obvious DRYing up. Having other languages rendered, is fixable with a couple lines of `<% if js %> ... <% end %>` etc. This is a pretty basic partial and we're only going to add more tutorials, do we really want to end up with more or less the same 330 lines of code copied and pasted into a dozen different ones?

SimonWoolf:
Requested fixes are in b81d6f2, including to only show the languages we want in the `install_ablyrealtime` partial. In case you still don't agree with the partial use even with that fix, I've added d43e472 which reverts all the partialisation.

content/tutorials/publish-subscribe.textile
mattheworiordan: > I'm happy to expand this back out into duplicates, but TBH I'm a bit surprised to see you object to this, seemed like an obvious DRYing up

I refer you to https://go-proverbs.github.io/ and `A little copying is better than a little dependency.`

I love DRYing things up, but I objected to this specifically because:

* We have not to date used ERB to do conditional content in templates so we have now made it just a tad more complicated for new devs to work on these emplates
* There is no mechanism using `<%if%>` statements to work out if a block should be shown or not based on the languages supported as the page at that stage is not rendered and thus knowing which languages maybe further down in the page are declared is not possible.
* You had code like this `<% if @use_client_id %>As we'll be using presence, we also include the @clientId@, which is what we'll appear as in the presence set.<% end %>`, which means our DRY partial is now talking about presence, yet the var is `@use_client_id`.  I think the context of the setup will very quickly change per example and we'll probably end up with bugs if we go down this route.

mattheworiordan:
@SimonWoolf btw. I assume you need to provide the links now and merge.  Shout when you do and I'll push to the website when done.

mattheworiordan:
@SimonWoolf any ETA on getting the commit SHAs in so that we can deploy this to production?

SimonWoolf:
I'll go back to docs & tutorials stuff tomorrow. Want to make some headway with realtime queue tests today.

SimonWoolf:
Added a step 5 for subscribing to presence changes; somehow missed that out in the initial PR - ea8230d

Added tutorial repo commit shas: f35e61d

content/tutorials/presence.textile
```diff
@@ -0,0 +1,438 @@
+---
+title: Presence Tutorial
+section: tutorials
+index: 10 # TODO ??
```
mattheworiordan: Nope, it just affects the position in the nav, but the tutorials are not listed in the nav anyway so you can ignore it

PR #194, opened 2016-11-12
Proposal: standardised way to 'refuse to authorize' for authUrl

SimonWoolf:
Thought prompted by [slack discussion](https://ably-real-time.slack.com/archives/support-offer-up/p1478906710000408).

If someone's using authCallback and they want to have a way of downright refusing to authorize a client (and not have the client keep retrying), they can send some custom thing to trigger their authCallback on the client to call connection#close(). But if using authUrl, it's not so simple. So, proposal: have client libraries interpret an HTTP 403 response from an auth server as a refusal to ever authorize that client, which therefore triggers the connection to move to the `failed` state.

(Also, on rereading RSA4c now, realised it's not clear that we meant it to only apply to network errors, auth servers being down, etc. Added an item to clarify that an actual ErrorInfo response from Ably to an authUrl/authCallback is definitely capable of triggering a `failed` state).


content/client-lib-development-guide/features.textile
```diff
@@ -103,10 +103,12 @@ h3(#rest-auth). Auth
...
 *** @(RSA4c2)@If the connection is @CONNECTING@, then the connection attempt should be treated as unsuccessful, and as such the connection should transition to the @DISCONNECTED@ or @SUSPENDED@ state as defined in "RTN14":#RTN14 and "RTN15":#RTN15
 *** @(RSA4c3)@If the connection is @CONNECTED@, then the connection should remain @CONNECTED@
+** @(RSA4d)@ If a request by a realtime client to an @authUrl@ results in an HTTP 403 response, then the client library should transition to the @FAILED@ state, with the connection @errorReason@ being set to an @ErrorInfo@ with @code@ @40300@ and some appropriate message.
+** @(RSA4e)@ If an @authCallback@ attempt or @authUrl@ request by a realtime client results in an Ably @ErrorInfo@, the client library should transition to the @FAILED@ state, with the connection @errorReason@ being set to the error. (This applies only where the response received over the wire is actually an @ErrorInfo@; any errors generated by the client, such as network errors, must be handled per @RSA4c@).
```
mattheworiordan: I am really not sure how the library is going to know it's an Ably @ErrorInfo@ specifically, especially with typed languages where the IDL makes no mention of an ErrorInfo object `authCallback: (String | TokenDetails | TokenRequest) →? // RSC14c, RSA4, TO3j5, AO2b`.

I think this is quite unclear as to why this spec item exists, what the purpose is, and how it would be used.  

I recommend that RSA4d covers only authUrl and we explain that unless an unauthorized response is received from an auth URL, all other failures are not considered terminal and thus the client library moves to the disconnected state so that an auth attempt will be retried.

For RSA4e, we should make it clear that an authCallback failure in most cases would be considered not terminal and therefore the library will move to the disconnected state (in the case of realtime), and reattempt a connection at a later time thus triggering another authCallback.  If we do want to support passing an error to the callback, then we should consider adding a new argument to the callback IMHO which is error, which will be null when successful, and contain an Ably ErrorInfo when unsuccessful.  If the error status code is 403, then it too will be treated as terminal and thus move the connection to the FAILED state

paddybyers:
> I recommend that RSA4d covers only authUrl and we explain that unless an unauthorized response is received
You mean 403 Forbidden?

> For RSA4e, we should make it clear that an authCallback failure in most cases would be considered not terminal and therefore the library will move to the disconnected state (in the case of realtime), and reattempt a connection at a later time thus triggering another authCallback. If we do want to support passing an error to the callback, then we should consider adding a new argument to the callback IMHO which is error, which will be null when successful, and contain an Ably ErrorInfo when unsuccessful. If the error status code is 403, then it too will be treated as terminal and thus move the connection to the FAILED state

In each language the callback already has a way of indicating an error - either an `err` in the callback in js, for example, or an exception in any language in which the call is synchronous. I'm not sure there is a need for a new argument.

I think it's reasonable to say that an error resulting from the `authCallback` is treated opaquely unless it's an `ErrorInfo`. I think these should be non-fatal. If it's an `ErrorInfo` (or at least duck-types as that by having a `statusCode`) then we are able to treat it non-opaquely; in that case it seems appropriate to treat 403 as fatal and everything else as non-fatal.

SimonWoolf:
@mattheworiordan 

> I recommend that RSA4d covers only authUrl and we explain that unless an unauthorized response is received from an auth URL, all other failures are not considered terminal and thus the client library moves to the disconnected state so that an auth attempt will be retried.

Sure, I agree. AFAICS that is what this PR (in conjunction with the existing RSA4c) already says - am I missing something?

> For RSA4e, we should make it clear that an authCallback failure in most cases would be considered not terminal and therefore the library will move to the disconnected state

Agreed; RSA4c already says this

> If we do want to support passing an error to the callback, then we should consider adding a new argument to the callback IMHO which is error, which will be null when successful, and contain an Ably ErrorInfo when unsuccessful.

That is how the authCallback already works, in ably-js anyway. See e.g. https://jsbin.ably.io/ocodaf/1/edit : `callback(err, token)`

> I am really not sure how the library is going to know it's an Ably @ErrorInfo@ specifically, especially with typed languages where the IDL makes no mention of an ErrorInfo

I'm not sure if I'm misunderstanding you here, but - this is a problem that typed libraries already have to solve, surely. They make a rest request, and get back `{"error": {"statusCode": 401,"code": 40110,"message":"Account blocked"}}`. It's their job to recognise that that's an ErrorInfo from Ably, presumably they do already.

I had the last sentence just to make clear, for libs where e.g. a network timeout could results in an exception with a library-created ErrorInfo, that that's covered by RSA4c, not RSA4e (I.e. that it should result in `disconnected`, not `failed`). I'm not sure what alternative could have the behaviour we want: that is, if we get an error that the auth server's passing on from ably (eg "Account blocked"), we want to fail the connection; if we just get an error from the customer's auth server itself, we don't.

> I think this is quite unclear as to why this spec item exists, what the purpose is, and how it would be used.

For RSA4d: so that auth server authors have an easy way to say 'no, permanently' to someone asking for a token. An alternative to RSA4d would be to have an `Auth#refuseTokenRequest()` method or something that generates a `{"error": {"statusCode": 403,"code": 40300,"message":"Token request denied"}}` thing that the auth server can respond with, but that seems unnecessarily complicated compared to replying with a 403 and having the client lib recognise that.

For RSA4e: I didn't intend it to be a change, just a clarification that we intended RSA4c to only apply to network errors, auth servers being down, etc, and not to actual error responses from Ably that the auth server is just passing on to the client. E.g. ably-js satisfies this by filling in network errors with a 40170 error ([in requestToken](https://github.com/ably/ably-js/blob/0-9-spec/common/lib/client/auth.js#L391)), then if it's a realtime client and that error is passed up to connectionManager, [transmuting it into an 80019 error and making it nonfatal](https://github.com/ably/ably-js/blob/0-9-spec/common/lib/transport/connectionmanager.js#L1447-L1456). 

@paddybyers 

> If it's an ErrorInfo ... it seems appropriate to treat 403 as fatal and everything else as non-fatal.

I'm not sure. With a token request, since it's REST we don't get the additional hint of whether it comes through as an `ERROR` or a `DISCONNECTED` to tell us whether realtime thinks it should be fatal; but in practice the times when ably will respond to a rest request with an error (account blocked, key revoked, etc). are times we'd want the connection to go into `failed`, no? (except for 5xxxxx I guess, for which we could have a `err.code >= 40000 && err.code < 50000` test, though that's a bit messy)

mattheworiordan:
Ok, sorry, not quite sure what I was thinking when re-reading my comments, it's almost as if I was looking at something completely different 😕 

Anyway, 👍 by me.  For Ruby I need to fix this issue anyway so can introduce the error callback capability, see https://github.com/ably/ably-ruby/issues/89

content/client-lib-development-guide/features.textile
tcard: This made me realize that the definition of `authCallback` was wrong. See #205.

I think we need to specify a way for `authCallback` to specify non-transient errors. It doesn't make sense that `authUrl` can do that but `authCallback` doesn't. But I don't think that we should expose `ErrorInfo` for this. `authCallback` IO-returning `null` instead of a `Token*` would suffice; it would say "I succeeded in calling the server or whatever, so I'm not failing, but I have nothing to return because the client was unauthorized or something, so you should move to fail". We don't need anything else from it.

tcard:
Sorry, I didn't notice the discussion was continuing here.

As I say [above](https://github.com/ably/docs/pull/194#discussion_r89487980) I don't think it's particularly friendly we ask people to give us an `ErrorInfo` when they can just give us null.

If we do go with `ErrorInfo`, we need to specify a constructor for it in the IDL.

tcard:
> I'm not sure if I'm misunderstanding you here, but - this is a problem that typed libraries already have to solve, surely. They make a rest request, and get back {"error": {"statusCode": 401,"code": 40110,"message":"Account blocked"}}. It's their job to recognise that that's an ErrorInfo from Ably, presumably they do already.

But it's `authCallback` the one receiving such payload, not the library.

content/client-lib-development-guide/features.textile
SimonWoolf: @tcard 

> I think we need to specify a way for authCallback to specify non-transient errors. It doesn't make sense that authUrl can do that but authCallback doesn't.

The reason is that the client can run whatever code they want in an authCallback, including `client.close()`. So if the dev wants to have a way of downright refusing to authorize a client (and not have the client keep retrying), they can send whatever custom thing they like to trigger code in their authCallback to call `client.close()`. (And then probably not bother calling the callback at all, if the connection's closed it's moot). With authUrl that isn't possible. That's why I made RSA4d apply to authUrl only.

mattheworiordan:
> If we do go with ErrorInfo, we need to specify a constructor for it in the IDL.

Good point, there is no definition for this.

Also, please base on 0-10 branch, see https://github.com/ably/wiki/issues/166#issuecomment-262910738

mattheworiordan:
@SimonWoolf will you rebase onto `0-10` so we can merge in?

PR #203, opened 2016-11-23
spec: Add support for extras to Message object

mattheworiordan:
This PR adds support for the `extras` attribute to the `Message` object, in preparation for push.  Please note:

* I have not added this to the `PresenceMessage` as I do not have a use case for this field, or any API that would allow the extras field to be configured.
* I have modified the publish IDL and set the `clientId` and `extras` args as optional arguments.  I am not clear if this is the right way of presenting this, but I expect most languages will unfortunately approach how this is provided differently anyway, so perhaps we leave it up to the language to work out how these additional options are provided.

@SimonWoolf @tcard @paddybyers please review.  

This needs to form part of 0.9 spec


content/client-lib-development-guide/features.textile
```diff
@@ -870,6 +871,7 @@ h4. Message
 ** @(TM2g)@ @name@ string
 ** @(TM2d)@ @data@ string, buffer or JSON-encodable object or array
 ** @(TM2e)@ @encoding@ string
+** @(TM2i)@ @extras@ JSON-like object, used to contain any arbitrary key value pairs which may also contain other primitive JSON types, JSON-like objects or JSON-like arrays. For context, the @extras@ field is planned for us with push notifications as a payload in the @push@ key, and later to allow arbitrary headers in the @header@ key that may be used for routing, filtering or to expose to consumers when retrieving messages from queues
```
paddybyers: AFAIK we haven't decided to include a `headers` member so we should not say this. Just say that other message metadata or ancillary payloads may be included in `extras` in the future.

Also, call me pedantic but I don't like the expression "JSON-like object" because objects are objects and JSON is text. We could say "JSON-encodable object" or simply "map".

content/client-lib-development-guide/features.textile
```diff
@@ -222,6 +222,7 @@ h3(#rest-channel). Channel
...
+*** @(RSL6a2)@ A set of tests must exist to ensure that the client library provides interoperability for the @extras@ field which is a JSON-like object. The test, at a minimum, should publish a message with an @extras@ object such as @{"key":[{"key":"val"}]}@ and ensure it is received with an equivalent JSON-like object
```
tcard: I'm not sure what "interoperability" means here. Which parts operate? It seems like it's the library with itself?

content/client-lib-development-guide/features.textile
```diff
@@ -1229,7 +1231,7 @@ class RestChannel:
   ) => io PaginatedResult<Message> // RSL2a
   publish([Message]) => io // RSL1
   publish(name: String?, data: Data?) => io // RSL1
-  publish(name: String?, data: Data?, clientId: String) => io // RSL1h
+  publish(name: String?, data: Data?, clientId?: String, extras?: JsonObject) => io // RSL1h
```
tcard: The previous line is redundant now, ie. this line includes it.

content/client-lib-development-guide/features.textile
tcard: What happens with fields in `extras` that are not specified to be supported? Does realtime respect them, filter them out, reject the message altogether?

RSL6a2 may need to be changed depending on this too.

mattheworiordan:
@paddybyers @tcard updated

content/client-lib-development-guide/features.textile
tcard: Well, we already have a opaque field in messages, ie. `data`. We do look into `extras` (`push`, so far) so it's not really opaque. So it's maybe cleaner to keep things meaningful to us in a field (`extras`) and things meaningful to the clients in another (`data`).

tcard: To be clear, my point is that such thing should be specified, not really arguing for one way or the other.

mattheworiordan: Doesn't "used to contain any **arbitrary** key value pairs" cover that?

mattheworiordan: > . So it's maybe cleaner to keep things meaningful to us in a field (extras) and things meaningful to the clients in another (data).

I agree with that, but I am not sure that really affects the spec, it only applies to user documentation surely?

tcard: It may be arbitrary for the point view of client libraries (so they don't need to check for existing fields themselves, for example, and thus we can add new extras fields at any point without updating the libraries) but not arbitrary to realtime (and the users).

> I am not sure that really affects the spec, it only applies to user documentation surely?

If realtime rejects unrecognized extras fields then it affects at least the tests for the spec, that would fail if they expect an unrecognized field to be respected.

mattheworiordan: > but not arbitrary to realtime (and the users).

But I think you're assuming that there is some form of validation yet there is no spec / indication this is so, so unless we imply that we do check it, I think this is fine

paddybyers:
> What happens with fields in extras that are not specified to be supported? Does realtime respect them, filter them out, reject the message altogether?

I think that the processing of members of `extras` that are not specified should be undefined: ie the realtime system and/or libraries reserve the right to strip or modify them. There should be no guarantee that they pass through transparently.

mattheworiordan:
> I think that the processing of members of extras that are not specified should be undefined: ie the realtime system and/or libraries reserve the right to strip or modify them. There should be no guarantee that they pass through transparently.

I don't agree the library reserves the right to strip them.  If so, that needs to be specified.

I believe the realtime system reserves the right to do what it likes, but that does not affect this spec in the same way the realtime system can reject a message for any number of reasons, but we don't need to list those reasons out.

tcard:
> I think that the processing of members of extras that are not specified should be undefined: ie the realtime system and/or libraries reserve the right to strip or modify them.

OK, well, let's hope that people don't put their stuff there and then we start rejecting or stripping it and apps start to break.

paddybyers:
> OK, well, let's hope that people don't put their stuff there and then we start rejecting or stripping it and apps start to break.

If we think that people will still try to put things in there, despite us saying that it's unsupported and the behaviour is undefined, then we should explicitly say that any message with unspecified `extras` fields will be rejected.

I don't agree that there should be an expectation that anything in there will pass through transparently; we provide `data` for message payloads, and we don't want to be constrained in what we do with `extras` in the future.

paddybyers:
> I don't agree the library reserves the right to strip them. If so, that needs to be specified.

I'm saying that, at the API, from the point of view of the consuming app, the handling of any unspecified `extras` passed in is undefined. We're not making any distinction at that point about what the library might do and what the realtime system might do.

In the library spec, we should be explicit about what the library is supposed to do with them, obviously, and unless there is a reason to process them they should be passed transparently to realtime.

mattheworiordan:
Ok, so this is approved?

content/client-lib-development-guide/features.textile
```diff
@@ -194,7 +194,7 @@ h3(#rest-channel). Channel
...
 *** @(RSL1g3)@ When publishing a @Message@ with a different @clientId@ attribute value to the identified client's @clientId@, the client library should reject the message, and indicate an error. The connection and channel remain available for further operations
 *** @(RSL1g4)@ When publishing a message with an explicit @clientId@ that is incompatible with the identified client's @clientId@ (either inferred or explicitly configured in the token or @ClientOptions@), the library will reject the message immediately and indicate an error
-** @(RSL1h)@ Where the library language permits, the @Channel#publish(name, data)@ method should provide an optional argument that allows the @clientId@ value to be specified such as @Channel#publish('event', 'data', { clientId: 'John' })@
+** @(RSL1h)@ Where the library language permits, the @Channel#publish(name, data)@ method should provide an optional argument that allows the @clientId@ value to be specified such as @Channel#publish('event', 'data', { clientId: 'John' })@, and/or the @extras@ field to be specified such as @Channel#publish('event', 'data', { extras: { headers: { 'key': 'value' } })@
```
paddybyers: Shouldn't this be a `push` example instead of `headers` ?

content/client-lib-development-guide/features.textile
```diff
@@ -222,6 +222,7 @@ h3(#rest-channel). Channel
...
+*** @(RSL6a2)@ A set of tests must exist to ensure that the client library provides interoperability for the @extras@ field which is a JSON-encodeable object. The test, at a minimum, should publish a message with an @extras@ object such as @{"key":[{"key":"val"}]}@ and ensure it is received with an equivalent JSON-encodeable object
```
paddybyers: encodable

content/client-lib-development-guide/features.textile
```diff
@@ -870,6 +871,7 @@ h4. Message
 ** @(TM2g)@ @name@ string
 ** @(TM2d)@ @data@ string, buffer or JSON-encodable object or array
 ** @(TM2e)@ @encoding@ string
+** @(TM2i)@ @extras@ JSON-encodeable object, used to contain any arbitrary key value pairs which may also contain other primitive JSON types, JSON-encodeable objects or JSON-encodeable arrays. For context, the @extras@ field is planned for use with push notifications as a payload in the @push@ key
```
paddybyers: I think that we should say explicitly that the `extras` field is provided to contain message metadata and/or ancillary payloads in support of specific functionality, eg push. Each of these supported extensions is documented separately; for 0.9 the only supported extension is push, via the `extras.push` member. The processing of any other members is undefined.

As written here it sounds like you can freely put anything in there, which is wrong.

mattheworiordan:
Ok ,comments address

content/client-lib-development-guide/features.textile
mattheworiordan: And this is now addressed too in 959a7d7

content/client-lib-development-guide/features.textile
```diff
@@ -222,6 +222,7 @@ h3(#rest-channel). Channel
...
+*** @(RSL6a2)@ A set of tests must exist to ensure that the client library provides interoperability for the @extras@ field which is a JSON-encodable object. The test, at a minimum, should publish a message with an @extras@ object such as @{"key":[{"key":"val"}]}@ and ensure it is received with an equivalent JSON-encodable object
```
paddybyers: But aren't we now saying that this isn't guaranteed to work?

Do we say to construct a test based on a `push` member? Or we define some limited support for `headers` or `metadata` ? Or we say that a `test` extra is supported on test apps?

PR #205, opened 2016-11-24
Fix type of authCallback in IDL.

tcard:
It was a function taking a TokenDetails, TokenRequest or String, but it should be a function that takes nothing and IO-returns a TokenDetails, TokenRequest or String.

IO-returning functions can always fail in a unspecified but platform-consistent manner, so it's relevant for the ongoing #194.


mattheworiordan:
LGTM

PR #206, opened 2016-11-24
spec: Reject invalid authCallback / authUrl content

mattheworiordan:
We have seen authUrl retrieving HTML files greater than 500kb in size which should have been rejected due to the content type, but also because of the size.

@SimonWoolf @paddybyers @tcard please review


content/client-lib-development-guide/features.textile
```diff
@@ -107,6 +107,7 @@ h3(#rest-auth). Auth
 *** @(RSA4c1)@An @ErrorInfo@ with @code@ @80019@ and description of the underlying failure should be emitted with the state change, in the @errorReason@ and/or in the callback as appropriate
 *** @(RSA4c2)@If the connection is @CONNECTING@, then the connection attempt should be treated as unsuccessful, and as such the connection should transition to the @DISCONNECTED@ or @SUSPENDED@ state as defined in "RTN14":#RTN14 and "RTN15":#RTN15
 *** @(RSA4c3)@If the connection is @CONNECTED@, then the connection should remain @CONNECTED@
+** @(RSA4d)@ If an attempt to authenticate is made using the @authUrl@ and the content type is neither text nor JSON, or if using @authCallback@ and the object passed in is neither a @String@, @JsonObject@, @TokenRequest@ object nor a @TokenDetails@ object, the auth attempt should fail with error code @40143@ and a suitable error message.  Additionally, if the text token string is greater than 384 bytes, or the JSON stringified @JsonObject@, @TokenRequest@ or @TokenDetails@ is greater than 128kb, then the auth attempt should fail with error code @40143@ and a suitable error message.
```
SimonWoolf: > neither text nor JSON

this should probably read "neither `text/plain` or `application/json`" -- technically `text/html` is a kind of text

> or the JSON stringified `JsonObject`, `TokenRequest` or `TokenDetails` is greater than 128kb

what's this protecting against? tokenrequests and tokendetails are easy to detect heruistically, and if it's a valid one but with an 129k capabilities string, that's admittedly a bit ridiculous, but at that point the client has it already, so why reject it? (An object that isn't a valid tokenDetails or tokenRequest should be rejected already)

> 40143

this is inconsistent with the error codes for other reasons for authCallback and authUrl failing (including the token being in an invalid format), which are specified in RSA4c. (40170 for standalone, 80019 on the realtime connection). 40143 is used by realtime.

How about just bundling this in with RSA4c, which already specifies what to do when " the token is in an invalid format", which this is just clarifying:

`(RSA4c4) A returned token shall be taken to be in an invalid format if: the content-type of the response from @authUrl@ is neither @text/plain@ or @application/json@; the object returned by @authCallback@ is neither a @String@, @JsonObject@, @TokenRequest@ object nor a @TokenDetails@ object; or if the text token string is greater than 384 bytes.`

mattheworiordan: > Where do we say it's OK for authCallback to return JsonObject? I don't think it is.

Well I imagine it's going to be quite common for a dev to obtain the token request / details over HTTP, and simply pass the object on.  I don't see why we wouldn't allow this.

> I think you're confusing authCallback with the callback passed to authCallback (in some platforms). You don't pass things to authCallback; authCallback passes things to you, either by returning them in "sync" platforms or by passing them to a callback you give it.

Yes, I will fix the terminology, I meant the callback passed to `authCallback`

SimonWoolf: > Because of a problem we've seen recently whereby a 500kb payload was sent to Ably

I follow that, I just meant that in all of those problems are random html getting treated as a token string, not random json getting treated as a token details, because that shouldn't happen - it wouldn't get detected as one. So not sure that's needed. But shrug, I don't mind it if you think it's needed, so sure.

> The spec item is not describing token failure conditions, it is instead describing side effects of a failure.

True, but we want it to be clear that 4c side-effects *do* apply to the failure in this PR. The way it's phrased in this PR makes it seem like an unrelated requirement. Especially as your new 4d also describes side-effects of the failure, which kinda clash with 4c.

So eg if you wanted to keep this at 4d, you could modify 4c to say "the provided token is in an invalid format **(per RSA4d)**, or ...", then have `(RSA4d) The following conditions imply that a token is in an invalid format: the content-type of the response from @authUrl@ is neither @text/plain@ or @application/json@; the object returned by @authCallback@ is not a @String@, @TokenRequest@ object, or @TokenDetails@ object; or if the text token string is greater than 384 bytes, or the JSON stringified @JsonObject@, @TokenRequest@ or @TokenDetails@ is greater than 128kb.`

mattheworiordan: > it wouldn't get detected as one. So not sure that's needed. But shrug, I don't mind it if you think it's needed, so sure.

@paddybyers was going to go for something much lower!  If we know we are going to reject the request before it's sent, why bother sending it?

Re: your other suggestion, sure, I will amend

tcard:
> Well I imagine it's going to be quite common for a dev to obtain the token request / details over HTTP, and simply pass the object on. I don't see why we wouldn't allow this.

Why did we did `Token*.fromJson` then? Anyway, if we want this we should add it to the IDL too.

content/client-lib-development-guide/features.textile
tcard: > Well I imagine it's going to be quite common for a dev to obtain the token request / details over HTTP, and simply pass the object on. I don't see why we wouldn't allow this.

Why did we did `Token*.fromJson` then? Anyway, if we want this we should add it to the IDL too.

mattheworiordan: > Why did we did Token*.fromJson then?

Because it's a convenience and allows them to be explicit, but also because we can then simply call that method directly for customers?

I can add it to the IDL, will do now.

tcard: Then maybe we should add to the spec a "official", testable way of discerning between TokenRequest and TokenDetails structurally. (The presence of a `token` field, I guess.)

mattheworiordan:
@SimonWoolf @tcard updated

content/client-lib-development-guide/features.textile
```diff
@@ -924,7 +924,7 @@ h4. PaginatedResult
 
 h4. HttpPaginatedResponse // RSC19b
 
-* @(HP1)@ A @HttpPaginatedResponse@ is a type that represents the response from an HTTP request containing an empty or JSON-like response. "Paginated queries":/rest-api/#pagination are supported
+* @(HP1)@ A @HttpPaginatedResponse@ is a type that represents the response from an HTTP request containing an empty or JSON-encodeable response. "Paginated queries":/rest-api/#pagination are supported
```
paddybyers: s/encodeable/encodable/

mattheworiordan: Bollocks, fixed in 	56af867

mattheworiordan:
Rebased onto `source-0-10` but for some reason Github is playing up and showing other commits... simply review all commits since `9055d5c`.  @paddybyers @SimonWoolf @tcard you happy for me to merge into `source-0-10` now?

SimonWoolf:
I'm not sure my objection [here](https://github.com/ably/docs/pull/206#discussion_r89503821) is fixed I'm afraid - the new 4d describes side-effects of the failure, which are incompatible with with the side-effects described in 4c1/2/3. I.e. with this, we have:
```
RSA4c: If the token is in an invalid format (per RSA4d), then A, B, and C should happen
RSA4d: If [some conditions], then D, E, and F should happen
```
I'll add a commit with a proposed fix and you can tell me what you think.

Not sure why github's messing up, as far as I can see you've done everything correctly. I'll try changing the base to something else and back to make it recalculate.

SimonWoolf:
Added 7422159, which does
RSA4c: results of authCallback/url failure for realtime connection
RSA4d: results of authCallback/url failure for rest requests and explicit auth requests
RSA4e: conditions for token to be in an invalid format

@mattheworiordan WDYT?

content/client-lib-development-guide/features.textile
```diff
@@ -103,11 +103,12 @@ h3(#rest-auth). Auth
...
+*** @(RSA4c1)@ An @ErrorInfo@ with @code@ @80019@ and description of the underlying failure should be emitted with the state change, in the @errorReason@ and/or in the callback as appropriate
+*** @(RSA4c2)@ If the connection is @CONNECTING@, then the connection attempt should be treated as unsuccessful, and as such the connection should transition to the @DISCONNECTED@ or @SUSPENDED@ state as defined in "RTN14":#RTN14 and "RTN15":#RTN15
+*** @(RSA4c3)@ If the connection is @CONNECTED@, then the connection should remain @CONNECTED@
+** @(RSA4d)@ If an attempt to authenticate is made using the @authUrl@ or @authCallback@ as a result of a request that uses the REST api, or by an explicit call to @authorize@ or @requestToken@, and the request to @authUrl@ fails, the callback @authCallback@ results in an error, the provided token is in an invalid format (as defined in "RSA4e":#RSA4e), or the attempt times out after "@realtimeRequestTimeout@":#DF1b, then the request should fail with error code @40170@ and a suitable error message.
```
mattheworiordan: `s/REST api/REST API`

content/client-lib-development-guide/features.textile
```diff
@@ -103,11 +103,12 @@ h3(#rest-auth). Auth
...
+** @(RSA4e)@ The following conditions imply that the token is in an invalid format: the content type is neither @text/plain@ or @application/json@; the object passed by @authCallback@ is neither a @String@, @JsonObject@, @TokenRequest@ object or a @TokenDetails@ object; the text token string is greater than 384 bytes; or the JSON stringified @JsonObject@, @TokenRequest@ or @TokenDetails@ is greater than 128kb.
```
mattheworiordan: `s/the content type is neither @text/plain@ or @application/json@/the @authUrl@ response content type is neither @text/plain@ or @application/json@/`

mattheworiordan:
Thanks for the update @SimonWoolf

SimonWoolf:
Added and merged

Issue #75, opened 2015-12-16
Token ClientOption should allow Token, TokenDetails or TokenRequest


mattheworiordan:
Currently `ClientOption#token` only supports a `Token` or `TokenDetails` object.  It should also support a `TokenRequest` given `authUrl` and `authCallback` both support all three

mattheworiordan:
Optionally a client library should be instanceable with just a TokenDetails object, like we do for API Key string or Token string, where the language permits

SimonWoolf:
I'm happy with the first (initing with {token: <tokenRequest>}) :+1:

Not sure about the second (initing with just a tokenDetails) -- means in weakly-typed languages we'd have to guess whether an object passed to the constructor is a tokenDetails or a clientOptions. `{tokenDetails: <tokenDetails>}` isn't much to ask

mattheworiordan:
> Means in weakly-typed languages we'd have to guess whether an object passed to the constructor is a tokenDetails or a clientOptions. 

No, you only need to check if:

* It's a string it's a token string 
* If it's an object, check if it has a `token` attribute, then it's a `TokenDetails`
* Else if it's an object, it's a `TokenRequest` object.

That same logic exists for `authUrl` now.

Not sure why you'd need to determine if its a `ClientOptions` or not?

SimonWoolf:
@mattheworiordan

>> Not sure about the second (initing with just a tokenDetails) --  means in weakly-typed languages we'd have to guess whether an object passed to the constructor is a tokenDetails or a clientOptions

> That same logic exists for authUrl now. Not sure why you'd need to determine if its a ClientOptions or not?

Your're justifying your first proposal (initing with `{token: <tokenRequest>}`). I completely agreed with that, which is why I :+1:'d it. My objection was to your second proposal, that "*Optionally a client library should be instanceable with just a TokenDetails object, like we do for API Key string or Token string*".

If you instance a client library with an object, currently that's assumed to be a clientOptions (in weakly-typed languages). clientOptions and tokenDetails can both have `token` and `clientId` attributes. We could guess based on the presence of `expires` or `issued`, but that's starting to get messy. (Especially if we take it it to its logical conclusion and allow initing with a bare tokenRequest object too - and if we don't, we've just introduced an inconsistency of the same kind that we've just removed with your first proposal!).  `{token: <tokenDetails>/<tokenRequest>}` isn't too much to ask.

(I'd have no objection to the second proposal if it applies only to strongly typed languages where there's no need for guesswork)

[...]

PR #207, opened 2016-11-24
Allow clientoptions.token to be a string, tokendetails, or tokenrequest

SimonWoolf:
Fixes https://github.com/ably/docs/issues/75

Was surprised to find that `clientOptions.token` wasn't specified to accept anything other than a token string at the moment.

Question: now we're formally allowing `clientOptions.token` to be any of the three, should we deprecate `clientOptions.tokenDetails`? Cut down on redundant document clientoptions a bit. (If not, it's a bit inconsistent having `clientOptions.tokenDetails` but not `clientOptions.tokenRequest`)


paddybyers:
> Question: now we're formally allowing clientOptions.token to be any of the three, should we deprecate clientOptions.tokenDetails? Cut down on redundant document clientoptions a bit. (If not, it's a bit inconsistent having clientOptions.tokenDetails but not clientOptions.tokenRequest)

We're not seriously suggesting this for typed languages are we?

tcard:
> We're not seriously suggesting this for typed languages are we?

We do it already for the return value of `authCallback`, and I think it's cleaner than having different fields for things that have the same purpose.

In Objective-C we make TokenRequest, TokenDetails and String all implement a common interface that returns. Sadly in Java you can't implement interfaces for types you don't own, so I guess you have to take Object and downcast.

paddybyers:
> I guess you have to take Object and downcast.

That's so bad I think a better API, in the java case, is to have differently named and typed fields.

mattheworiordan:
> That's so bad I think a better API, in the java case, is to have differently named and typed fields.

I think for ~~crappy~~ limited languages like Java, we should indeed simply deviate where necessary.  But I don't think the entire API should suffer as a result of primarily Java.

mattheworiordan:
I think this should be based on `source-0-10`, see https://github.com/ably/wiki/issues/166#issuecomment-262910738

SimonWoolf:
rebased on source-0-10

content/client-lib-development-guide/features.textile
```diff
@@ -1025,7 +1025,7 @@ h4. ClientOptions
 ** @(TO3i)@ @recover@ string - A connection recovery string, specified with the intention of inheriting the state of an earlier connection
 ** @(TO3j)@ Auth option attributes:
 *** @(TO3j1)@ @key@ string - Full Ably key string as obtained from dashboard
-*** @(TO3j2)@ @token@ string - An authentication token string issued for this application
+*** @(TO3j2)@ @token@ string | @TokenDetails@ | @TokenRequest@ - An authentication token issued for this application, either in the form of a token string, a @TokenDetails@ object, or a @TokenRequest@ object
```
SimonWoolf: Given Paddy's objection, I guess maybe something like this?
`@(TO3j2)@ @token@ string | @TokenDetails@ | @TokenRequest@ - An authentication token issued for this application, either in the form of a token string, a @TokenDetails@ object, or a @TokenRequest@ object. For languages where it's impractical or unidiomatic to take multiple types, this should take a token string only, and TO3j3 and TO3j12 apply`

`(TO3j3) tokenDetails TokenDetails – for languages where @token@ (TO3j2) takes a string only, this is a @TokenDetails@ object`

`(TO3j12) tokenRequest TokenRequest – for languages where @token@ (TO3j2) takes a string only, this is a @TokenRequest@ object`

mattheworiordan: LGTM 👍

PR #208, opened 2016-11-25
Add UPDATE event, remove ERROR

SimonWoolf:


content/client-lib-development-guide/features.textile
```diff
@@ -515,7 +516,7 @@ h3(#realtime-presence). Presence
...
 *** @(RTP5c2)@ If a @SYNC@ is initiated as part of the attach, then once the @SYNC@ is complete, all members not present in the @PresenceMap@ but present in the "internal @PresenceMap@":#RTP17 must be re-entered automatically by the client using the @clientId@ and @data@ attributes from each. The members re-entered automatically must be removed from the "internal @PresenceMap@":#RTP17 ensuring that members present on the channel are constructed from presence events sent from Ably since the channel became @ATTACHED@
-*** @(RTP5c3)@ If any of the automatic @ENTER@ presence messages published in "RTP5c2":#RTP5c2 or "RTP5c3":#RTP5c3 fail, then the error message received from Ably should be emitted as an @ERROR@ event on the @Channel@ with Ably error code @91201@
+*** @(RTP5c3)@ If any of the automatic @ENTER@ presence messages published in "RTP5c2":#RTP5c2 fail, then an @UPDATE@ event should be emitted on the channel, with a @reason@ set to the error message received from Ably, wrapped in an @ErrorInfo@ with error code @91201@
```
SimonWoolf: This bit I wasn't sure of - not sure if an auto-reenter-failure deserves an `UPDATE`, or just log it like we do for decode failures? what do you think?

paddybyers: I think there should be a programmatic way of being notified, and `UPDATE` is the only way

mattheworiordan:
FYI, what about:

```
** @(RTN15h)@ If a @DISCONNECTED@ message is received from Ably, then that transport will subsequently be closed by Ably. If the @DISCONNECTED@ message contains a token error (@statusCode@ value of 401 and error @code@ value in the range @40140 <= code < 40150@), if the token is renewable, the error should not be emitted, a single attempt to create a new token will be made and a new connection attempt will be initiated using the new token. If the token creation fails, the next connection attempt fails due to a token error, or the token was not renewable, the connection will transition to the @FAILED@ state and the @Connection#errorReason@ will be set
```

Should we be mentioned an error being emitted given we have no generic way of doing that?

mattheworiordan:
```
enum ChannelEvent:
  embeds ChannelState
  ERROR // RTL2c
```

should be

```
enum ChannelEvent:
  embeds ChannelState
  UPDATE // [INSERT REF]
```

content/client-lib-development-guide/features.textile
```diff
@@ -301,12 +301,12 @@ h3(#realtime-connection). Connection
 * @(RTN3)@ If connection option @autoConnect@ is true, a connection is initiated immediately; otherwise a connection is only initiated following an explicit call to @connect()@
 * @(RTN4)@ EventEmitter and states:
 ** @(RTN4a)@ Implements @EventEmitter@ and emits events for state changes @INITIALIZED@, @CONNECTING@, @CONNECTED@, @DISCONNECTED@, @SUSPENDED@, @CLOSING@, @CLOSED@, @FAILED@
```
mattheworiordan: Should we not just mention this is a `ConnectionEvent` type that extends `ConnectionState` and adds `UPDATE` event?

content/client-lib-development-guide/features.textile
```diff
@@ -301,12 +301,12 @@ h3(#realtime-connection). Connection
...
-** @(RTN4g)@ Additionally, an @ERROR@ event can be emitted by the @Connection@ that contains details of the error in an @ErrorInfo@ object
+** @(RTL4h)@ Additionally, an @UPDATE@ event can be emitted by the @Connection@. This is used for changes to connection conditions for which the main connection state (e.g. @CONNECTED@) does not change. (The library must never emit an @RTN4a@ state change event for a state equal to the previous state).
```
mattheworiordan: I am not keen on the terminology "main connection state" i.e. what is main and what is not?  I am struggling to find a better description, but perhaps "This is used for changes to connection conditions for which the `ConnectionState` value does not change..."

content/client-lib-development-guide/features.textile
```diff
@@ -406,16 +407,16 @@ h3(#realtime-channel). Channel
 * @(RTL1)@ As soon as a @Channel@ becomes attached, all incoming messages and presence messages are processed and emitted where applicable. @PRESENCE@ and @SYNC@ messages are passed to the @Presence@ object ensuring it maintains a map of current members on a channel in realtime
 * @(RTL2)@ EventEmitter and states:
 ** @(RTL2a)@ Implements @EventEmitter@ and emits events for state changes @INITIALIZED@, @ATTACHING@, @ATTACHED@, @DETACHING@, @DETACHED@, @SUSPENDED@ and @FAILED@
+** @(RTL2g)@ Additionally, an @UPDATE@ event can be emitted by the @Channel@ that contains a @ChannelStateChange@ object. This is used for changes to channel conditions for which the main channel state (e.g. @ATTACHED@, per @RTL2a@) does not change. (The library must never emit an @RTL2a@ state change event for a state equal to the previous state).
```
mattheworiordan: As above, should we not be more explicit about the `ChannelState` and `ChannelEvent` types?

content/client-lib-development-guide/features.textile
mattheworiordan: Yup, but messy, but not sure what else we can do.  Error code @91201@ is not at https://github.com/ably/ably-common/blob/master/protocol/errors.json though?

SimonWoolf: Re the error code, I'm now thinking it's probably be a mistake to wrap the error code from realtime. If someone wants to programatically do something useful with a failed re-enter, they need to know the actual reason it failed (which'll probably be one of the 9100x reasons), wrapping that in a general 'an auto-re-enter failed' code is probably unhelpful.

(of course, can still wrap just the message, `"An automatic re-enter attempt failed: #{error.message}"` etc, to make the log message more explicit)

Unless anyone objects, I'll just remove that bit

SimonWoolf:
Made requested changes:  2561ea4

SimonWoolf:
@mattheworiordan happy for me to merge? Ricardo and Peter are both implementing eventemitter stuff at the moment (see [here](https://ably-real-time.slack.com/archives/client-libraries/p1480597783000137), I've pointed them to this PR for now), so would be good to get this locked down

mattheworiordan:
@paddybyers have you approved?

paddybyers:
> @paddybyers have you approved?

Yes, LGTM, thanks

PR #217, opened 2016-12-05
Update documentation for UPDATE event

SimonWoolf:
Documentation changes for https://github.com/ably/docs/pull/208 (adding `update` event, removing `error`).

Per @mattheworiordan's comment [here](https://github.com/ably/docs/pull/181#issuecomment-260531685) I've made the changes for all languages, making guesses as to what non-js languages will look like, as Matt's done with the rest of the 0.9 spec. (In particular I'm not sure if the typed languages will just have ChannelEvent as a separate enum from ChannelState (which is what I've assumed in this PR), or have some way to inherit enums, or ChannelEvent'd only have one element and the on() methods'd take either a ChannelState or ChannelEvent, or whatever)


content/types/_channel_event.textile
```diff
@@ -1,25 +1,92 @@
...
+  h4. Symbol states
+
+  ```[ruby]
+    :initialized # => 0
+    :attaching # =>   1
+    :attached # =>    2
+    :detaching # =>   3
+    :detached # =>    4
+    :failed # =>      5
+    :update # =>       8
```
mattheworiordan: Weird spacing and `8` should be `6`?

content/types/_channel_event.textile
```diff
@@ -1,25 +1,92 @@
...
-      case Failed
-      case Error
-  }
-```
+blang[jsall].
+  @ChannelEvent@ is a String that can be emitted as an event on the @Channel@ object; either a "@ChannelState@":/realtime/channels-messages#channel-states or an @update@ event.
+
+
+  ```[javascript]
+    var allChannelEvents = allChannelStates.concat(['update']);
```
mattheworiordan: I am not sure I understand what you are demonstrating here given `allChannelStates` is not defined.  Could you rather not just list out all possible string values like we do for other enums making it much clearer to users what to expect?

content/types/_connection_event.textile
```diff
@@ -0,0 +1,117 @@
+blang[jsall].
+  @ConnectionEvent@ is a String that can be emitted as an event on the @Connection@ object; either a "@Realtime Connection@ state":/realtime/connection#connection-states or an @update@ event.
+
+
+  ```[javascript]
+    var allConnectionEvents = allConnectionStates.concat(['update']);
```
mattheworiordan: Same as above, surely listing out all values is better for users?

content/realtime/channels-messages.textile
```diff
@@ -462,14 +464,14 @@ Note that while fatal errors won't get better on their own, they can be fixable.
 
 h4(#nonfatal-errors). Nonfatal errors
 
-Other types of error are nonfatal. For example, a client may have network connectivity issues, or a channel may experience a loss of strict message continuity. The library will automatically attempt to recover from these events. If channel continuity is lost in the process, the library will notify you though a @resumed@ flag in the @ATTACHED@ event, so you can decide how to handle that failure.
+Other types of error are nonfatal. For example, a client may have network connectivity issues, or a channel may experience a loss of strict message continuity. The library will automatically attempt to recover from these events. If channel continuity is lost in the process, the library will notify you though a @resumed@ flag in the @ATTACHED@ or @UPDATE@ event, so you can decide how to handle that failure.
```
mattheworiordan: `Other types of errors`...

content/types/_channel_event.textile
SimonWoolf: It's not defined in this partial, but everywhere `channel_event` is included, `channel_state` is included right above it, which defines it. I figured this way makes it clearer what the difference actually is between the two. But happy to switch out for a full list if you think that'd be better, wdyt?

mattheworiordan: I don't think this makes it any easier for a developer i.e. expecting them to jump around documentation to find out what `allChannelStates` is only makes their job harder.  We have had feedback from developers quite often that they feel they need to jump around too much in docs and they'd prefer we duplicate content to help them avoid that.  So my vote is that a partial should not make assumptions about other partials existing close enough to this content for it to make sense and we should go with simplicity for the reader.

SimonWoolf: Fair enough :+1: 

Changed in  	347880c

paddybyers:
> (In particular I'm not sure if the typed languages will just have ChannelEvent as a separate enum from ChannelState (which is what I've assumed in this PR), or have some way to inherit enums, or ChannelEvent'd only have one element and the on() methods'd take either a ChannelState or ChannelEvent, or whatever)

In Java you can't have derived enum types, so `ChannelEvent` will be an interface.

mattheworiordan:
Would prefer not to merge until https://github.com/ably/docs/issues/225#issuecomment-267461130 is resolved

paddybyers:
Ok, but if in java we rename the interface `ChannelEvent` to, say, `IChannelEvent` (the actual name never really relevant) for, and rename `UpdateEvent` to `ChannelEvent`, then you'd have:

    on(ChannelState.attached, ...);

or

    on(ChannelEvent.update, ...);

mattheworiordan:
@SimonWoolf @tcard @paddybyers please review recent commits to implement what we discussed

paddybyers:
LGTM. Just checking out the java implementation to make sure it's implementable in the way we think

mattheworiordan:
> Your changes to the spec are based on a pre-UPDATE-event version of the spec (which is why it didn't already mention ConnectionEvent)

I know, but I didn't want to rebase this on top because of all the feedback...

> the solution is probably to merge jsall-0.9 in to source already, let me know if you're happy for me to do that

Sure, I'll leave this with you then if that is OK?

mattheworiordan:
Think we just need a thumbs up from @paddybyers and @tcard

paddybyers:
LGTM

SimonWoolf:
Merged into source as 2b0dc8a

PR #222, opened 2016-12-14
Fix duplicated RTL4h

ricardopereira:


SimonWoolf:
Thanks! Sorry, didn't see this before Peter prodded me earlier to fix it straight on the source branch.

PR #229, opened 2016-12-20
WIP: Push notifications spec.

tcard:


tcard:
@SimonWoolf I initially added you as reviewer and then removed you because you haven't been so involved and I didn't want to be imposing, but of course your input is appreciated!

mattheworiordan:
@tcard please note that these notes have been written offline whilst on the train so it's quite possible some of the information I have here is incorrect and needs to be revised.  However, I hope it sparks some conversation on the points at least.

Review of your PR by spec code:

- RSH1: I am not sure I understand why this function exists. My understanding was that later, we may want to allow customers to push raw push notifications.  The functionality at the outset to push notifications directly was to allow:
  - Push directly to a device ID
  - Push directly to a clientId and thus all devices registered to that clientId
  - As above, but with some batching functionality to make this more efficient i.e. so that customers who now use GCM/APNS, can simply use our API for both channel based broadcasts and direct broadcasts to devices.
- RSH1: If a user wanted to push a notification without a channel message, then if this is a proxy to `Rest#channels.get(channel).publish(null, null, extras: {"push": notification})`, then that would not be possible.  I don't think we should be forcing users unnecessarily to publish emtpy messages on channels simply so that they can publish raw notifications.  My view is that this publish method should not be associated with channels at all, and instead the only argument is the complete extras field. The reason I say the complete extras field is that there may very well be use cases in the future where the other fields may be needed such as rules around when to deliver, perhaps metadata that influences delivery etc.
- RSH2b: See my note below on device ID. I believe this could be broader than just being a push target to be honest, and perhaps should just be `AblyRest / AblyRealtime.device` which is the same object? The device attributes could then be associated with a connection and if we exposed that to users (via a new REST API), then we are also solving an additional problem we have had before where customers have asked us how they can set properties on a connection that they can trust (because presence can't be trusted).  So connection details could include device details (issued by the client) and token details (provided in the token for that connection, a bit like `clientId`, but just more info). And every connection request now over WS / Comet could include these details when creating a connetion.
  - RSH2b4 - wouldn't this now just be `deviceId`, and if the user chose to change this, it would invalidate any recipientAddress
  - RSH2b2 and RSH2b3 - shoudl be prepoluated with some default values we can define (we should make recommendations in this spec), however the customer should probably be able to change these?
  - RSH2b1 and RSH2b5 - is transport necessary, isn't this only relevant if there is a recipientAdress and coupled to that?  If so, why not bundle the two together?  BTW. I considered that perhaps we should support multiple receipientAddresses (which includes transport) so that if we for example had more than one way to push notifications in the future, we could easily support that. But the more I thought about this the more I realised this is a bad idea I think because right now it's nice, there is a 1-to-1 relationship between device & registration and it makes things easier if we just think about notifying a device one way.
- RSH2c - I am not sure I like activateeFromDevice.  Should this not simply be `subscribe(transport, clientId)` or `attach(transport, clientId)` with transport having a default on each device, but allowing for new push subscription types in future. I prefer attach over activate because it's conceptually similar to channel attach except that it's persistent, although I think we previously agreed on subscribe. Given we have PushChannelSubescriptions subscribe would be better, but then how does a user register their subscribe handler?  Right now it seems that is done with an EventEmitter but that feels wrong and I am not sure technically that will work anyway for browsers (service workers) and mobile (intents / equivalent for iOS). What is `onRenewal` for?
- RSH2c - what happens if this is called and the device has a registration?  Do we issue a new registration?
- RSH2c1 - instead of returning an error, coudl the user not optionally provide device details as an argument?
- RSH2c3 - shouldn't the renewal secret be provided in the response to the initial registration, or better still generated as part of the device details and used to create a new registration?  If we wanted to be really secure, we could even consider never storing the secret on our servers and instead simply store the hash, so that secrets can only ever be changed, not read
- RSH2c5 - how will passing in an onRenewal callback work in this case given the app will most likely have been shut down before that gets called again. Does the app need to register some type of intent etc. so that it can be woken if that event is generated for example?
- RSH2d - I am not sure I understand why this is a different method. A bit like our Auth, you can let the library handle auth or you can provide a callback so that the token is provided explicitly to the library.  Why not go with a similar strategy i.e. you provide a callback when you call subscribe such as `persistCallback((deviceDetails) -> (err))` or something along those lines so that if you want to explicitly persist the registration to Ably yourself, the library does everything the same?
- RSH2d4 - similar to RSH2c5, doesn't this need to be an intent or something like that to make it work beyond app restarts?
- PPT1 - I expect we'll soon support Microsoft platforms and even OSX platforms, but perhaps leave as is for now as we can add that easily later
- PDT1 - We need to add "embedded" as a device type for IoT type devices
- IDL deactivateFromDevice - as above, why not simply detach?
- IDL deactivateExternally - how is this then different from deactivateFromDevice?  I assume deactivateExternally only calls the OS level methods to deactivate?
- IDL generateNewRenewalSecret - re: above and device registrations that could potentially be distinct from push device registations, should we perhaps have `deviceId` and `deviceKey` whereby `deviceKey` allows updates from the device because it can provide it is that device (like a connectionKey). If that were the case, should we not then in the admin API allow a deviceSecret to be changed, and also allow a local deviceSecret to be changed on the device and simply make it a property of `device`.  Then `generateNewRenewalSecret` could change to `recycleDeviceSecret` or `renewDeviceSecret`. WDYT?
- IDL PushDeviceRegistrations#find - why not only allow find by targetId as they should always be unqiue no?
- IDL Push#embeds EventEmitter - EventEmitter implies that it will be `on(message)` yet `on(state/event)` is used elsewhere for state events. This is not an eventemitter in that sense and should be `subscribe` no? Also, technically, how does this now work between app reloads? How is that handler registered each time the app starts up?
- IDL PushDeviceRegistrations - is this not just a Deviceregistration that can optionally have a push target?
- Just to be clear, are we now saying that we only ever have on target for one device, and as such the old proposed IDL with different targets per DeviceRegistration is no longer valid. If a user registers a new target for a device (renewal / expires etc) we simply replace the odl with new and keep no record of it?

### Device IDs - broader opportunity?

If we have a device ID for a registered device that is decoupled from the specific OS level registration for Push notifications, I wonder if we should instead force that every client library has a device ID (similar to a PubNub UUID perhaps) that is assigned the first time the lib is instanced. In the case of devices which cannot persist a device ID, then this will effectively then be a session ID, but I believe most apps and even websites have some form of local storage we can use. This would also help tackle some more generic problems we have such as https://github.com/ably/wiki/issues/117

### Use case review

I wanted to think more broadly about our use cases to see if this PR is addressing those, at least from a schema and API perspective so that we don't have to change the API as soon as we move forwards with smart notifications.  Here are some scenarios I thought of, not saying this is exhausive and not given much thought as to whether your proposed API will satisfy all of this:

- Try and notify a user that someone is trying to get hold of them, but do this with smart notifications. If connected on the channel, just publish the message without a push notification. If no response in 1 minutes, send a push notification, then if the app is not opened in 5 minutes, send a SMS or email.  So my immediate thoughts on ths are that we need to have arbitrary data that can be associated with a clientId, and pehaps even separate targets that end up just calling a WebHook for example.  So when I publish my message I may simply say 1st try channel if active, fall back any push notification targets if registered (else skip to next step) and wait a minute, then fall back to sms or email target if registered.  SMS target is just a WebHook with sufficient details to send that SMS, and email similar.
- As a developer, I want to know what devices the user has easily.  I query the API by clientId, and see all their devices. I want to augment that metadata about devices which I can do via the REST API, but equally those devices can augment data about themselves too if they have permission to update their device details with the secret.  When I issue tokens for that device to be connected to Ably, the device ID is sent in the connection string, and the token itself can have some additional details on that user.  Therefore, anyone can see in a channel where I am present my public infromation relating to my connection such as device type, platofrm, and any other key/vals assigned.  This is not the same as presence data which varies by channel.
- I am a developer who is moving to Ably but already uses GCM / APN.  I would prefer to use Ably for everyhing so will be pushing messages directly to devices as I did before (with the platform registration IDs I had before). We may or may not have these registration IDs in our database, but we should allow them to use them.  Those push notifiations however will be received with the callbacks they have registered previosuly with their apps, and not with the new Ably callbacks.
- I am dev and simply want to send push notifications to a group of people.  I register their devices from the devices themselves with the token allowing channel registrations, and I publish a message on those channels to broadcast push notifications to both iOS & Android devices.  If the devices are active they won't show the notification and will subscribe to Ably, and if they are not ative a visual notification will be shown.
- I receive feedback from Ably that a device ID / target ID for a push notificatiokn has stopped working. I have mechanism to notify the device next time it logs in to obtain a new push registration ID. Two options, the user does not allow push notifications so the registration ID fails, so the registration should be removed from Ably.  Second option is it succeeds and the old registration must be replaced with the new one.  I hope that simply calling push.subscribe does all of this for me transparently.
- I am a devleoper and don't care about push token expiry, I expect the client library will handle this all transparently for me.

tcard:
About RSH1 (`Push#publish`): I can probably easily implement all that functionality you mentioned (published to device IDs, to client ID, and to channel without sending an Ably message), in that case, should we keep it now?

> My view is that this publish method should not be associated with channels at all, and instead the only argument is the complete extras field. The reason I say the complete extras field is that there may very well be use cases in the future where the other fields may be needed such as rules around when to deliver, perhaps metadata that influences delivery etc.

If such rules affected push, arguably they should be under the "push" section of extras, and if they affect more than push, they shouldn't be under `Push#publish` but in a more general field in `Rest`. `Push#publish({"push": ...})` seems to me painfully redundant.

About device IDs, and in general a more general way of defining a device as an Ably entity: maybe. But for push, we _need_ a push target ID, a push recipient address, a transport type, etc. Those are all push-specific, need to be defined and stored somewhere, and don't conflict at all with any more general thing (`Rest#device`, UUIDs, whatever) we come up later.

So, replying to specifics about this same point:

> RSH2b4 - wouldn't this now just be deviceId, and if the user chose to change this, it would invalidate any recipientAddress

We still need the push-specific target ID. We still can call device ID the generic thing.

> RSH2b2 and RSH2b3 - shoudl be prepoluated with some default values we can define (we should make recommendations in this spec), however the customer should probably be able to change these?

This is about `Push#platform` and `Push#deviceType` and yes, those would be populated by the client library.

> RSH2b1 and RSH2b5 - is transport necessary, isn't this only relevant if there is a recipientAdress and coupled to that?

It's necessary to know how to push. It's strongly associated with the target ID, not with the recipient address, since the recipient address can change. But the three of them are non-optional fields. "Coupled" can mean lots of things, but if we put them together, say, in a string, we then must take them apart anyway to process them for their separate meanings, so I don't see the benefit of defining it that way in this high level and this is just an encoding issue.

> perhaps we should support multiple receipientAddresses (which includes transport) so that if we for example had more than one way to push notifications in the future, we could easily support that

You probably mean target ID, since a recipient address can change for the same push recipient. We already support a way to associate together several target IDs (with their recipient addresses), and that's by client ID; we may introduce down the line new ones (e. g. your proposed device UUID) if we want in this way, rather than make "fat" PushDeviceRegistrations. Think array of registrations rather than registration of arrays. It seems cleaner to me.

About `Push#activateFromDevice(onRenewal, clientId)`:

> Should this not simply be subscribe(transport, clientId) or attach(transport, clientId)

About the naming: `subscribe` is what we call attaching a registration to a channel. There's no channel here. We're activating the device for push.

> with transport having a default on each device

That's done via `Push#deviceDetails`, and having it here too would just duplicate things.

> although I think we previously agreed on subscribe

Well, we previously agreed on pretty much everything that's in there!

> what happens if this is called and the device has a registration? Do we issue a new registration?

Yes, which would have upsert semantics given that the target ID is immutable. Well, we could have an "already activated" flag in the client library to avoid the extra computational work on the server, but it's not crucial.

> Given we have PushChannelSubescriptions subscribe would be better, but then how does a user register their subscribe handler? Right now it seems that is done with an EventEmitter but that feels wrong and I am not sure technically that will work anyway for browsers (service workers) and mobile (intents / equivalent for iOS). 

I don't know what you mean by "subscribe handler". You mean the thing receiving push notifications? Then yes, RSH2a says that "If possible" (and it is possible in iOS, Android and browser, although it needs some platform-specific hooking) "`Push` implements `EventEmitter` and emits all push notifications received" but I'm not sure why does that feels wrong and what would be better.

> What is onRenewal for?

RSH2c5: `Whenever the renewal process described at "RSH2c4":#RSH2c4 happens, it reacts by following "RSH2c2":#RSH2c2 and "RSH2c3":RSH2c3 again, using @onRenewal@ (if provided) as callback instead for the generated error or renewal secret.` So, pretty much to notify the user about the renewal of the registration (typically, a change in its recipient address), or (more interestingly) failure thereof.

> instead of returning an error, coudl the user not optionally provide device details as an argument?

So "if not already populated" is meant for this use case. Users could populate it themselves if they wanted to for some reason. We could easily add it as an additional optional argument, but the "90%" case is that we don't want users to do that, because we do this for them and it would be just boilerplate.

> RSH2c3 - shouldn't the renewal secret be provided in the response to the initial registration [...] we could even consider never storing the secret on our servers

The renewal secret is provided and stored by our users. We don't provide it nor store it, we don't even use it. Its purpose is so that their servers, with `push-admin` permissions, do privileged stuff (typically, register the device) on behalf on their clients, with only `push-subscribe` permissions. But because we're talking about `activateFromDevice`, which imples that the client has `push-admin` permissions, the renewal secret is pretty much useless. It's just provided for completeness, but the 90% case is to just ignore it. (It's still potentially useful if the library updates to a `push-susbscribe` model.)

> RSH2c5 - how will passing in an onRenewal callback work in this case given the app will most likely have been shut down before that gets called again. Does the app need to register some type of intent etc. so that it can be woken if that event is generated for example?

Kind of. So, right now, for our currently supported transports, `onRenewal` will only be called when {GF}CM refreshes the token, and you [natively provide a callback for it](https://firebase.google.com/docs/cloud-messaging/android/client#monitor-token-generation). We would just hook to that callback and call `onRenewal` when it's fired.

About `Push#activateExternally(registerDevice, renewRegistration)`:

> RSH2d - I am not sure I understand why this is a different method. A bit like our Auth, you can let the library handle auth or you can provide a callback so that the token is provided explicitly to the library.

It could be the same method. `activateFromDevice` is actually just `activateExternally` with defaults, really. But we need some configuration to set those defaults (namely, the `clientId` and `onRenewal` arguments) and anyway I think the naming makes the intent clearer. An alternative would be something like this:

```swift
let (registerDevice, renewRegistration) = Push.defaultActivationCallbacks(clientId, onRenewal)
Push.activate(registerDevice, renewRegistration) // <- `activate` would be the single method
```

> PPT1 - I expect we'll soon support Microsoft platforms and even OSX platforms, but perhaps leave as is for now as we can add that easily later

Yes, and it's specified that the library must not check at any point the string against the specified list, so an update to the spec doesn't imply a change in all client libraries in this case.

> PDT1 - We need to add "embedded" as a device type for IoT type devices

OK, done at 71a1cd588fc79cfd8cddd7466f85b0d0dd91be90.

> IDL deactivateExternally - how is this then different from deactivateFromDevice? I assume deactivateExternally only calls the OS level methods to deactivate?

Correct, it deactivates the device from {GF}CM (in iOS I believe you don't need to do this, but I'd have to check) and the calls back so that you unregister the device from Ably.

> IDL generateNewRenewalSecret - re: above and device registrations that could potentially be distinct from push device registations, should we perhaps have deviceId and deviceKey whereby deviceKey allows updates from the device because it can provide it is that device (like a connectionKey). If that were the case, should we not then in the admin API allow a deviceSecret to be changed, and also allow a local deviceSecret to be changed on the device and simply make it a property of device. Then generateNewRenewalSecret could change to recycleDeviceSecret or renewDeviceSecret. WDYT?

I'm not sure I understand. See my comment above about how `onRenewal` is supposed to works, as per our conversations. These tokens are purely a convenience for users to authenticate between their clients and servers.

> IDL PushDeviceRegistrations#find - why not only allow find by targetId as they should always be unqiue no?

Chances are they are, but we can't guarantee that. Target IDs are provided by the transports (GCM, APNs, etc.) and they're only guaranteed to be unique within their transports. So we always, 100 % of the time deal with target ID and transport together. They together uniquely identify a registration.

> IDL Push#embeds EventEmitter - EventEmitter implies that it will be on(message) yet on(state/event) is used elsewhere for state events. This is not an eventemitter in that sense and should be subscribe no?

Yes, it's more like `RealtimeChannel#subscribe`. But `subscribe` would be kind of confusing here, as we already call something else subscriptions.

> Also, technically, how does this now work between app reloads? 

The same way that token refresh does: we hook to the native callbacks.

> IDL PushDeviceRegistrations - is this not just a Deviceregistration that can optionally have a push target?

I really don't know what you mean here. What's a `Deviceregistration`? Also, `PushDeviceRegistrations` doesn't _have_ anything since it's just a collection of static methods.

> Just to be clear, are we now saying that we only ever have on target for one device

Yes. Well, per application (not Ably-application, platform-application) per device. But the meaning of "device" and "target" here is in terms of the native platform. Again, all this is push-specific and forcibly deeply tangled with GCM, APNs, etc. domains.

> and as such the old proposed IDL with different targets per DeviceRegistration is no longer valid.

I can't find that in the old proposed IDL. I haven't changed anything significant that I know of from it in this proposal.

> If a user registers a new target for a device (renewal / expires etc) we simply replace the odl with new and keep no record of it?

As long as the target ID doesn't change (and it never changes, unless you uninstall the application and install it again) the target stays the same, so renewals/expirations update the old registration, so everything about it will be kept (including, most critically, channel subscriptions).

Now, on to the use cases:

> Try and notify a user that someone is trying to get hold of them, but do this with smart notifications. If connected on the channel, just publish the message without a push notification. If no response in 1 minutes, send a push notification, then if the app is not opened in 5 minutes, send a SMS or email.

Can be done with some kind of "smart rule" that associates with a push registration/subscription via client ID. This smart rule would just "use" push registrations as they are.

> As a developer, I want to know what devices the user has easily. I query the API by clientId, and see all their devices. I want to augment that metadata about devices which I can do via the REST API, but equally those devices can augment data about themselves too if they have permission to update their device details with the secret.

Well, as I said above that "with the secret" would more precisely be "with an `push-admin` token", but yes.+

> When I issue tokens for that device to be connected to Ably, the device ID is sent in the connection string, and the token itself can have some additional details on that user. Therefore, anyone can see in a channel where I am present my public infromation relating to my connection such as device type, platofrm, and any other key/vals assigned. This is not the same as presence data which varies by channel.

I'm sorry but I've read this a few times now and I don't get what you're saying or how it relates to the above. Maybe we can chat about it?

> I am a developer who is moving to Ably but already uses GCM / APN. I would prefer to use Ably for everyhing so will be pushing messages directly to devices as I did before (with the platform registration IDs I had before). We may or may not have these registration IDs in our database, but we should allow them to use them.

No problem at all. Registration IDs are unique so our libraries would just naturally get them.

> Those push notifiations however will be received with the callbacks they have registered previosuly with their apps, and not with the new Ably callbacks.

No problem. We provide a native hook ourselves but they're welcome to implement manually whichever parts they want; the API is designed to seamlessly integrate with everything out of the box, but also be fully able to opt-out to any of these integrations. In this case, instead of using our EventEmitter they can subclass our application class, which implements our hooks, and override the platform-specific notification callback method.

> I am dev and simply want to send push notifications to a group of people. I register their devices from the devices themselves with the token allowing channel registrations, and I publish a message on those channels to broadcast push notifications to both iOS & Android devices. If the devices are active they won't show the notification and will subscribe to Ably, and if they are not ative a visual notification will be shown.

Doable. There's nothing special or Ably-specific about this. They would just _not_ send a visual notification, but the data necessary to construct one, and then, on the notification callback, implement whatever they want.

> I receive feedback from Ably that a device ID / target ID for a push notificatiokn has stopped working. I have mechanism to notify the device next time it logs in to obtain a new push registration ID. Two options, the user does not allow push notifications so the registration ID fails, so the registration should be removed from Ably.

We don't remove registrations right now if not one-by-one manually, but it certainly can be addded.

> Second option is it succeeds and the old registration must be replaced with the new one.

No, you would have to re-subscribe to everything. We could provide this functionality though by client ID, and if we add the device ID thing later, also from that.

> I am a devleoper and don't care about push token expiry, I expect the client library will handle this all transparently for me.

You call `activateFromDevice` and don't even provide a callback.

tcard:
@mattheworiordan I think de7d92f covers all the changes in the IDL we discussed on Friday. PTAL

tcard:
Sorry, I obviously was missing the new channel methods. On 4e7e9f2.

content/client-lib-development-guide/features.textile
```diff
@@ -1534,40 +1478,59 @@ enum StatsIntervalGranularity:
...
+class DeviceDetails:
+  + getDeviceInfo() => io DeviceDetails
+  id: String
+  platform: DeviceType
+  type: DeviceType
+  clientId: String?
+  metadata: Dict<String, String>
+  updateTokenId: String
+  resetId(String)
+  resetUpdateToken() => io
```
mattheworiordan: I am assuming therefore there is no way to configure the `updateTokenId`?  I appreciate most of the time this will be unnecessary, however if the `updateTokenId` was invalid for some reason, then I assume `resetUpdateToken` would not work because it cannot authenticate, or would it instead rely on the client being authenticated to obtain a new `updateTokenId`.  If the latter, then this in itself is a bit of a security concern as our servers would then have no way to confirm if the request for a new token is valid or not for that `device.ID` and as such you could in theory take control of someone else's `device.ID`.  Perhaps the thinking is as follows:

* `updateTokenId` is generated by Ably when the device is first registered, and returned in the registration response
* An endpoint will exist that allows any device to generate a new `updateTokenId` and can do this when authenticating with the existing `updateTokenId`
* A client library can reset it's `updateTokenId` by calling `resetUpdateToken()` which handles everything transparently if the existing `updateTokenId` is valid.
* A client library can explicitly reset it's `updateTokenId` by calling `resetUpdateTokenId(string)`, but the token string can only be obtained server-side by a client with `push-admin` rights i.e. that endpoint to reset the `updateTokenId` is callable with that privilege

Note that I think a lot of this functionality is unnecessary for MVP launch, and at launch, we should probably just focus on ensuring that when registered, a device obtains and stores a valid `updateTokenId` only, and we can worry about `resetId, resetUpdateToken, resetUpdateTokenId` post MVP launch.  Not saying we don't define it now in the IDL though...

On a separate note, I think `resetId` should require a new `updateTokenId` too as the old `updateTokenId` will no longer be valid.  So perhaps then given the above, we only have two methods:

`reset(deviceId, updateTokenId)` and `renewUpdateToken() => io`

WDYT?

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
+  errorReason: ErrorInfo?
+  metadata: Dict<String, String>
+
+class Push:
+  publish(clientId: String, JsonObject) => io
+  publish(deviceId: String, JsonObject) => io
+  publish(transportType: String, transportId: String, transportTokenOrKey?: String) => io
+  admin: PushAdmin
+
+  // Only on platforms that support receiving notifications:
```
mattheworiordan: We're missing any comment or indication about how you receive push notifications which is surely needed at this point?

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
+  transportType: DevicePushTransportType
+  transportId: String
+  transportToken: String?
+  state: .Active | .Failing | .Failed
+  errorReason: ErrorInfo?
+  metadata: Dict<String, String>
+
+class Push:
+  publish(clientId: String, JsonObject) => io
+  publish(deviceId: String, JsonObject) => io
```
mattheworiordan: In languages without named params, how would we support this overloaded method with the same signature?

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
+  publish(transportType: String, transportId: String, transportTokenOrKey?: String) => io
+  admin: PushAdmin
+
+  // Only on platforms that support receiving notifications:
+
+  activate(DeviceDetails) => io ErrorInfo? // RSH2c
+  activateExternally(
```
mattheworiordan: Why do we need activeExternally, why not:
`activate(DeviceDetails, registerCallback?: (ErrorInfo?, DeviceDetails)) => io ErrorInfo?`, where registerCallback is called if the user wants to handle registration details of the device themselves with Ably, else the client library handles this for them.  This is the same pattern as `realtime.new({ key: 'foo', useTokenAuth: true })` versus `realtime.new({ key: 'foo', authCallback: (params) => {})` etc.

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
+    DeviceDetails,
+    registerDevice: (ErrorInfo?, DeviceDetails?) -> io String?,
+    renewRegistration: (ErrorInfo?, DeviceDetails?) -> io String?
```
mattheworiordan: I still don't follow how the user can pass in a callback for `renewRegistration` that would be accessible following an app reload / upgrade etc.  The callback needs to be registered as part of an interface surely so that the client library can find the method responsible for renewing or the method for renewing could be overriden by the developer for example.

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
+  )
+  deactivate(deviceId: String) => io ErrorInfo?
+  deactivateExternally
```
mattheworiordan: As above, why not an optional callback instead of two methods?

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
+
+class PushAdmin:
+  deviceRegistrations: PushDeviceRegistrations
+  channelSubscriptions: PushChannelSubscriptions
+
+class JsonObject:
+  // Platform-dependant, typically a Dict-like object
+
+class PushDeviceRegistrations:
+  save(PushDeviceRegistration) => io
```
mattheworiordan: See above, we may need one additional method `renewUpdateToken(PushDeviceRegistration) => io String`

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
+  "gcm" // PTT1
+  "apns" // PTT1
+  "web" // PTT1
+
+enum DevicePlatform:
+  "android" // PPT1
+  "ios" // PPT1
+  "browser" // PPT1
+
+enum DeviceType:
```
mattheworiordan: Type is quite broad.  Should this perhaps be `DeviceFormFactor` instead?  @paddybyers WDYT?

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
+  save([PushChannelSubscription]) => io
+  save(PushChannelSubscription) => io
+  listChannels() => io PaginatedResult<String>
+  get(channelName: String) => io PaginatedResult<PushChannelSubscription>
+  get(type: PushDeviceType, transportId: String, channelName: String?) => io PaginatedResult<PushChannelSubscription>
+  get(clientId: String, channelName: String?) => io PaginatedResult<PushChannelSubscription>
+  remove([PushChannelSubscription]) => io
+  remove(PushChannelSubscription) => io
+
+class PushDeviceRegistration:
```
mattheworiordan: I am not following why this exists.  PushDeviceRegistration, as far as I can tell, is just `DeviceDetails` with the `push` attributes flattened.  We don't have `ApiKey` and `ApiKeyRegistration` simply because it's persisted to disk.  As far as I am concerned, it's just `DeviceDetails` which may be an object in memory that is not yet persisted, or an object in the Ably system that is persisted.  Added to which, should that then not just be `Device`?

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
+class PushAdmin:
+  deviceRegistrations: PushDeviceRegistrations
+  channelSubscriptions: PushChannelSubscriptions
+
+class JsonObject:
+  // Platform-dependant, typically a Dict-like object
+
+class PushDeviceRegistrations:
+  save(PushDeviceRegistration) => io
+  find(type: PushDeviceType, transportId: String) => io PushDeviceRegistration?
```
mattheworiordan: I am finding the differentiation between find & get a bit obscure i.e. `find` returns one, whereas `get` multiple items.  To simplify things, why don't we have a set of params that can be used and have a single method `get` that is used for all queries such as:

```
get(queryParams: Params<String, String>)
```

So it can be used as follows:
```
get() => Page through all devices
get({ "clientId": "bob" }) => give me all devices for bob
get({ "deviceId": "xxx" }) => give me all devices for that device Id, which should only be one
get({ "transportPlatform": "ios", "formFactor": "mobile" }) => assuming we supported this in the REST API, they could get all iOS devices with a mobile form factor
```

I don't see the value in typing these queries to be honest.

Same applies for `remove`, why not only allow remove by `deviceId` which is unique, and if a user needs to look up a `transportId` and delete it, then let them do that, so `remove(deviceId: string) | remove(device: DeviceDetails) | remove(Array<string | DeviceDetails>)`.

I think taking this approach would simplify things immensely and allow us to add support for new query types more easily too.

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
+  get(clientId: String, state?: .Active | .Failed) -> io PaginatedResult<PushDeviceRegistration>
+  remove(type: PushDeviceType, transportId: String) -> io
+  remove(clientId: String) => io
+  remove(PushDeviceRegistration) => io
+
+class PushChannelSubscriptions:
+  save([PushChannelSubscription]) => io
+  save(PushChannelSubscription) => io
+  listChannels() => io PaginatedResult<String>
+  get(channelName: String) => io PaginatedResult<PushChannelSubscription>
```
mattheworiordan: As above, I think we should opt for a more flexible Params approach given this is just a call to the REST API. We can return suitable response codes for invalid queries instead of hard-coding this into each library

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
+  "browser" // PPT1
+
+enum DeviceType:
+  "mobile" // PDT1
+  "tablet" // PDT1
+  "desktop" // PDT1
+  "embedded" // PDT1
+
+class PushChannelSubscription:
+  constructor(transportType: DevicePushTransportType, transportId: String, channelName: String) -> PushChannelSubscription
```
mattheworiordan: Why is this not `constructor(device: DeviceDetails, channelName: String) -> PushChannelSubscription`?  I don't follow why we'd lock in a `transportType` and `transportId` into this interface when the customer thinks of `Devices` and `clientId`s for their customers.

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
+  constructor(clientId: String, channelName: String) -> PushChannelSubscription
+  transportType: DevicePushTransportType? // PCS1, PCS5, PCS6
```
mattheworiordan: I don't think this is relevant personally, I think it should be `deviceId`

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,110 @@ enum StatsIntervalGranularity:
...
   MONTH
 
+class DeviceDetails:
+  + getDeviceInfo() => io DeviceDetails
+  id: String
+  platform: DeviceType
+  type: DeviceType
+  clientId: String?
+  metadata: Dict<String, String>
+  updateTokenId: String
```
paddybyers: I prefer to think of this as a `localDeviceToken` - ie something that authenticates any request as being made by the device, and there may be multiple operations in future that are permitted to be authorised in that way; not just token renewals.

content/client-lib-development-guide/features.textile
paddybyers: I think this requires more thought. For launch I think we should consider the `updateToken` or `localDeviceToken` being generated along with the device ID, and updatable with `push-admin` rights. If we want a robust way for the device to be able to prove that it is what it claims then we could have an asymmetric keypair generated locally on the device and used by the device to sign token requests or whatever. So I don't think we should be putting things in the IDL that we don't intend to implement and release without being more confident that they will be the right thing when we come to implement them.

content/client-lib-development-guide/features.textile
paddybyers: I think things could easily get out of hand if the method names are explicitly qualified, eg `publishForDevice()` etc; it would be more extensible to have `publish(map: options)` if named params aren't supported.

content/client-lib-development-guide/features.textile
paddybyers: Why not `activateDevice()` ?

content/client-lib-development-guide/features.textile
paddybyers: `deactivateDevice()` ?

content/client-lib-development-guide/features.textile
mattheworiordan: Sure, I prefer that

content/client-lib-development-guide/features.textile
mattheworiordan: Ok, sure, so remove any ability for the device to automatically renew the token and only allow it to be set explicitly with `reset`?

content/client-lib-development-guide/features.textile
mattheworiordan: Ok, so to be clear, you are saying where named params are supported we allow it otherwise we simply using a `Map<string, string>`?

content/client-lib-development-guide/features.textile
mattheworiordan: Are you saying `activateDevice` in place of `activate` or `activeExteanally` or both?

content/client-lib-development-guide/features.textile
paddybyers: Yes

content/client-lib-development-guide/features.textile
paddybyers: in place of `activeExternally`.

Ie `activate()` is called on the device.

`activateDevice()` is called by a server, because it wants to activate a device.

mattheworiordan: I think you've misundestood the intention of this method.  The `activate` comes in two forms where in the first form it is responsible for get a push token from the OS, and then registering itself with Ably via the REST API directly.  The second form it does step 2, but invokes a callback provided by the dev so that they can pass the details of that registration to their server which in turn handles that registration via the REST API.

Registration of the device via the REST API is simply done using the current `PushDeviceRegistrations#save` method.

content/client-lib-development-guide/features.textile
tcard: de2352095dd0e56efe333789afcf2dfa9369b579

content/client-lib-development-guide/features.textile
```diff
@@ -1534,40 +1478,59 @@ enum StatsIntervalGranularity:
   DAY
   MONTH
 
+class DeviceDetails:
+  + getDeviceInfo() => io DeviceDetails
+  id: String
+  platform: DeviceType
```
tcard: 4e8db205f42ab7c552c54bd5d4cf4b9a264aca42

content/client-lib-development-guide/features.textile
```diff
@@ -1534,40 +1478,59 @@ enum StatsIntervalGranularity:
   DAY
   MONTH
 
+class DeviceDetails:
+  + getDeviceInfo() => io DeviceDetails
```
tcard: f3262359651dd13c77def4370309a153e2117935

content/client-lib-development-guide/features.textile
tcard: Isn't natively enough? We had the EventEmitter before but it's not necessary really.

content/client-lib-development-guide/features.textile
tcard: Sorry but I don't understand now which parts I should be implementing here.

content/client-lib-development-guide/features.textile
tcard: 9478122808f7cc75110e9a068a28794b96498b36

content/client-lib-development-guide/features.textile
tcard: That's an implementation detail really, and honestly I'm not sure it can be defined in a cross-platform way. But yes, it probably can't be provided as a function object in any platform. So I'll just remove it and mention it in the spec without a IDL.

9478122808f7cc75110e9a068a28794b96498b36

content/client-lib-development-guide/features.textile
tcard: 9478122808f7cc75110e9a068a28794b96498b36

content/client-lib-development-guide/features.textile
tcard: 3a23829

content/client-lib-development-guide/features.textile
tcard: 3a23829

content/client-lib-development-guide/features.textile
tcard: Removed at 99e796f.

content/client-lib-development-guide/features.textile
tcard: Form factor sounds good  to me. b8b59a4

content/client-lib-development-guide/features.textile
tcard: Why not just a device ID? We don't need a full DeviceDetails here.

content/client-lib-development-guide/features.textile
tcard: 7c71069

content/client-lib-development-guide/features.textile
```diff
@@ -1464,6 +1486,82 @@ enum StatsIntervalGranularity:
...
+enum DevicePlatform:
+  "android" // PPT1
+  "ios" // PPT1
+  "browser" // PPT1
+
+enum DeviceFormFactor:
+  "mobile" // PDT1
+  "tablet" // PDT1
+  "desktop" // PDT1
+  "embedded" // PDT1
```
paddybyers: BTW I think this isn't really useful as-is, because `embedded` tells you nothing about what the device is, and it just means "none of the above". I suggest we either:

- define something more meaningful here (not sure what that would be - but could for example include `embedded-headless`);

- be honest have have an `other` value;

- remove this and do nothing until we have actual examples of a device type and use-cases that allow us to define something meaningful.

content/client-lib-development-guide/features.textile
mattheworiordan: I'm not sure either, but given the discussion above, perhaps for this release:

* `activate` - activate device with or without callback responsible for registering the device with Ably.  If no callback is provided, activate will a) request a push notification target from the OS, b) store that target info locally, c) register the device automatically with Ably via the REST API assuming it has permissions, and will store the `localDeviceToken` locally that is received from Ably when registered.  If a callback is provided, then c) is not needed and it is the responsibility of the dev to do this typically through a request to their servers which in turn communicate with Ably.  Note that if activate is called again, and the device has a push registration ID, we need to decide what action is taken i.e. is a new token obtained or not?  Perhaps we'd need `reactivate` instead so that it's explicit.
* `reset` which generates a new `deviceId` and removes all locally stored info relating to push.  I am not sure why we would want to allow a developer to set the `deviceId` as it should be random, and if the `deviceId` has changed, then the token and push target effectively becomes invalid anyway, so to keep things simply why not just discard the lot.
* `deactivate` is the same as `reset` except the `deviceId` does not change and it either contacts Ably and unregisters itself, or optionally takes a callback (like `activate`).
* `updateLocalToken(token)` which allows a a local token string to be set for the client

So the changes as far as I can tell are:

* The client library does not automatically handle anything to do with resetting the local token, it simply takes a token string as an argument
* We need to make sure we have a REST API that can allow devices to be registered and returns a local device token that is stored with the device in Cassandra, can allow a device to be deleted (deactivated) and deletes all associated records, can allow a `localDeviceToken` to be reset by `POST`ing to an endpoint that returns a new `localDeviceToken` ID. The latter API needs `push-admin` rights or to be authenticated with the existing `localDeviceToken`

content/client-lib-development-guide/features.textile
mattheworiordan: > Isn't natively enough? We had the EventEmitter before but it's not necessary really.

I thought we were going to expose a function that the user overrides in the same way it's done to receive native notifications?  I would prefer this personally as it means as much as possible, we have a standard API for receiving notifications even if the implementation details differ by platform (i.e. a ServiceWorker for the browser)

content/client-lib-development-guide/features.textile
mattheworiordan: @tcard sure, device ID or device details, same effect from my perspective.

content/client-lib-development-guide/features.textile
mattheworiordan: :+ for `other` until there is a requirement to support more types which can be added server-side anyway.

mattheworiordan:
One or two small comments, otherwise ready to discuss later

content/client-lib-development-guide/features.textile
mattheworiordan: BTW. I think we should add `watch` to this list

mattheworiordan:
Please rebase this on `master` now

tcard:
Superseded by #257.

PR #234, opened 2016-12-24
spec: Clarify untilAttach behaviour after discontinuity

mattheworiordan:
@paddybyers whilst working on the JSON patch demo and adding functionality to recover the history following discontinuity, it occurred to me that we are really not explicit about what happens for `untilAttach` queries.

I am not sure if client libraries will do the right thing or not at present, so propose adding this to the 0.9 spec

cc @SimonWoolf if you've not left
cc @tcard

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -480,10 +480,11 @@ h3(#realtime-channel). Channel
 ** @(RTL9a)@ Returns the @Presence@ object for this channel
 * @(RTL10)@ @Channel#history@ function:
 ** @(RTL10a)@ Supports all the same params as REST @Channel#history@
-** @(RTL10b)@ Additionally supports the param @untilAttach@, which if true, will only retrive messages prior to the moment that the channel was attached. This bound is specified by passing the querystring param @fromSerial@ with the serial number assigned to the channel in the @ATTACHED@ @ProtocolMessage@. If the @untilAttach@ param is specified when the channel is not attached, it results in an error
+** @(RTL10b)@ Additionally supports the param @untilAttach@, which if true, will only retrieve messages prior to the moment that the channel was attached or emitted an @UPDATE@ indicating loss of continuity. This bound is specified by passing the querystring param @fromSerial@ with the @Channel#attachedSerial@ assigned to the channel in the @ATTACHED@ @ProtocolMessage@ (see "RTL15":#RTL15). If the @untilAttach@ param is specified when the channel is not attached, it results in an error
 ** @(RTL10c)@ Returns a @PaginatedResult@ page containing the first page of messages in the @PaginatedResult#items@ attribute returned from the history request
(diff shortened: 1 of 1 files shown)


paddybyers:
The realtime system will always send the correct `channelSerial` in an `ATTACHED` message, whether `resume` is `true` or `false`. If `resume` is true, it will be the same as the previous `ATTACHED` message (and therefore the same as the most recent `ATTACHED` for which `resume` was false).

So I think it might be simpler to specify here that the `fromSerial` is always updated based on the `channelSerial` in the `ATTACHED` message.

mattheworiordan:
Excellent, good to know, will amend

mattheworiordan:
@paddybyers slightly controversial perhaps, but I propose we expose the `attachedSerial` attribute given it's trivial to do and future proofs us should we wish to expose additional bleeding edge functionality (using our `request` method) without requiring a change to the public API.

WDYT?

paddybyers:
> I propose we expose the attachedSerial attribute given it's trivial to do and future proofs us should we wish to expose additional bleeding edge functionality (using our request method) without requiring a change to the public API.

I would support this, and also think about exposing other channel attributes resolved as part of attach (eg the channel mode flags), when we're sure that this won't change as part of the 0.10 changes.

mattheworiordan:
Ok, so you happy with the proposed **really final** change to the 0.9 spec?

paddybyers:
> you happy with the proposed really final change to the 0.9 spec?

No. Nothing is really final until the realtime part is finished, and we have at least one library fully conforming.

I don't agree (yet) with exposing the `attachSerial`. I'm saying that once the 0.10 changes are implemented in realtime we will know what we can expose, knowing that it will be stable.

mattheworiordan:
> I don't agree (yet) with exposing the attachSerial. I'm saying that once the 0.10 changes are implemented in realtime we will know what we can expose, knowing that it will be stable.

But we're working on 0.9, not 0.10.  We cannot block one PR for future undetermined functionality.

On a separate note, I think if we expose the channel serial at the time of attached, we should probably just expose the current channel serial as well ([like we do for connections](http://docs.ably.io/client-lib-development-guide/features/#RTN10)), so perhaps something like:

```
channel.serials.attach
channel.serials.current
```

WDYT?

mattheworiordan:
@paddybyers please review again

content/client-lib-development-guide/features.textile
```diff
@@ -480,10 +480,12 @@ h3(#realtime-channel). Channel
...
+* @(RTL15)@ @Channel#properties@ attribute is a @ChannelProperties@ object representing properties of the channel state. The @ChannelProperties@ object is a public API. However, it will remain undocumented in our user documentation until its use is finalised in later versions of this specification. It has the following attributes:
```
paddybyers: I think saying it's a public API is confusing. I think we should say that it's a publically accessible member of the channel, but it is an experimental and unstable API.

SimonWoolf:
:+1:

PR #237, opened 2016-12-28
Remove deprecated ProtocolMessage#connectionKey

mattheworiordan:
This property is now present in the connectionDetails and 0.9 is a good time to remove.

Matching issue to remove in the product when 0.9 is detected, see https://github.com/ably/realtime/issues/802

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -907,7 +907,6 @@ h4. ProtocolMessage
 ** @(TR4b)@ @channel@ string
 ** @(TR4c)@ @channelSerial@ string
 ** @(TR4d)@ @connectionId@ string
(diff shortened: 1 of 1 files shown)


mattheworiordan:
> We also need to remove (TM2h).

Why are we removing `TM2h`.  As far as I can tell, that has nothing to do with this PR and allows REST messages to be published on behalf a connected user?  See http://docs.ably.io/client-lib-development-guide/features/#TM2h

paddybyers:
> Why are we removing TM2h.

Sorry, I meant http://docs.ably.io/client-lib-development-guide/features/#TR4e

mattheworiordan:
@paddybyers you mean like the one I removed? https://github.com/ably/docs/pull/237/files#diff-11900b395df266bc2bbfc1d11bdfc7a0L908

PR #238, opened 2016-12-31
spec: PresenceMap behaviour whilst SUSPENDED

mattheworiordan:
Suspended channels are likely to become Attached again at some point. As such, I think the API should reflect this in that whilst a channel is SUSPENDED the state should be maintained, and once the channel becomes ATTACHED any state changes can be emitted. Simply stripping the PresenceMap according to the current spec results in members being present again when the channel becomes ATTACHED, yet they've never left.

@paddybyers what do you think about these changes?

FYI, implemented in https://github.com/ably/ably-ruby/pull/99/commits/554a2e043749c42b172bf7f3cb60d9f65704f8a4 and https://github.com/ably/ably-ruby/pull/99/commits/446082f76a42aa7c5bf9444abf795632d4018a7a


mattheworiordan:
@paddybyers I have now added spec items 3b42fc4 and 08b0cdf to further clarify the behaviour expected from the client libraries.  Please read the comments of 08b0cdf to explain my thinking.

@SimonWoolf perhaps you should review too as I believe you added these specs.

Note most of this new functionality is implemented in recent commits on https://github.com/ably/ably-ruby/pull/99

content/client-lib-development-guide/features.textile
```diff
@@ -558,6 +558,7 @@ h3(#realtime-presence). Presence
 * @(RTP11)@ @Presence#get@ function:
 ** @(RTP11a)@ Returns the list of current members on the channel in a callback. By default, will wait for the @SYNC@ to be completed, see "RTP11c1":#RTP11c1
 ** @(RTP11b)@ Implicitly attaches the @Channel@ if the channel is in the @INITIALIZED@ state. However, if the channel is in or enters the @DETACHED@ or @FAILED@ state before the operation succeeds, it will result in an error
+** @(RTP11d)@ If the @Channel@ is in the @SUSPENDED@ state then the @get@ function immediately returns the members stored in the @PresenceMap@ and does not wait until the channel state changes to @ATTACHED@ irrespective of whether @waitForSync@ is true or false
```
paddybyers: I'm not sure I agree with this. When you're `SUSPENDED` then there is no synchronisation of state between the client and server. I don't see why it's valid to continue to return stale presence state.

paddybyers:
> I question whether we should be emitting an UPDATE when a client enters with the same data and they are already there.

The realtime system determines what message(s) to send based on message metadata, but never message data. We're not going to go comparing message data when there's an update. So if the `msgId` and `timestamp` indicate that a presence message corresponds to the same event, or a superseded event, then there will be no update; otherwise there will be an update.

mattheworiordan:
>  So if the msgId and timestamp indicate that a presence message corresponds to the same event, or a superseded event, then there will be no update; otherwise there will be an update.

That's all well and good, but that does not address the problem as far as I can tell because the two `enter` events sent will have different timestamps and message IDs.  

> We're not going to go comparing message data when there's an update

I'm only suggesting this for the `enter` event explicitly, and why should we not compare message data?  If it's just a blob, why can't we compare it?

mattheworiordan:
@paddybyers please see two new further proposed amendments caffeb9 and 66a569c.  Tests exist for this in https://github.com/ably/ably-ruby/pull/99/commits/6aa91b7c0df7cf05990b37c9bd298f021c155d3f.  Also, 66a569c
 relies on https://github.com/ably/ably-common/pull/20

paddybyers:
> If it's just a blob, why can't we compare it?

Because the cost of that grows with the size of the data. We're not going to compare a 64K message payload when there's an update. It's just wrong. Also, the client might intentionally be updating the state of the member, with a message having a newer timestamp, and might care that there is an update with that even if the data is the same. If we really need to know that the messages correspond to the same enter action, as understood by the client, then there should be metadata present that we can use to determine that.

content/client-lib-development-guide/features.textile
mattheworiordan: Because the state was synchronised previously. If a connection becomes disconnected or is disconnected without the client knowing,  asking for presence with waitForSync returns the current sets waitForSync does not mean it's in sync, it simply means wait for the sync to complete before retiring a list of members so that during an active sync you get the complete list. I don't see how the current connection state has any bearing on that option whatsoever

mattheworiordan:
> Because the cost of that grows with the size of the data.

I don't think that's a reason to be honest. The cost of the number of subscribers grows as the number of subscribers grows, the cost of the number of capabilities grows as the number of capabilities grows, the cost of the number of channels grows as more channels are used. This is inherent in what we do and should not dictate the behaviour of the service AFAIC.

> We're not going to compare a 64K message payload when there's an update.

We're not, I never suggested that.  I said that if a client explicitly tries to `enter`, yet the realtime system has the member present, then it should try it's best to deal with that situation in the way we believe the customer intended.  In reality, this duplicate `enter`will happen very rarely so the cost is probably insignificant, yet the expected behaviour from a customer is not.

>  Also, the client might intentionally be updating the state of the member

We have update for that, why would they be updating the member?

> with a message having a newer timestamp, and might care that there is an update with that even if the data is the same

Why, can you give me one realistic use case where that would be expected?  I can give you plenty the other way around.

>  If we really need to know that the messages correspond to the same enter action, as understood by the client, then there should be metadata present that we can use to determine that.

Why?  Why not just compare the payloads in the very rare occasion that a duplicate `enter` occurs?  In what situation, other than when a client mistakenly enters twice, do you foresee an `enter` legitimately generating an `update` event?

content/client-lib-development-guide/features.textile
```diff
@@ -511,17 +511,18 @@ h3(#realtime-presence). Presence
...
+*** @(RTP5c2)@ If either a @SYNC@ is initiated as part of the attach and the @SYNC@ is complete, or if the @resumed@ flag is false and a @SYNC@ is not expected, all members not present in the @PresenceMap@ but present in the "internal @PresenceMap@":#RTP17 must be re-entered automatically by the client using the @clientId@ and @data@ attributes from each. The members re-entered automatically must be removed from the "internal @PresenceMap@":#RTP17 ensuring that members present on the channel are constructed from presence events sent from Ably since the channel became @ATTACHED@
+*** @(RTP5c3)@ If any of the automatic @ENTER@ presence messages published in "RTP5c2":#RTP5c2 fail, then an @UPDATE@ event should be emitted on the channel with @resumed@ set to true and @reason@ set to an @ErrorInfo@ object with error @code@ value @91004@ and the error @message@ string containing the message received from Ably (if applicable), the @code@ received from Ably (if applicable) and the explicit or implicit @client_id@ of the @PresenceMessage@
```
SimonWoolf: > set to an `@ErrorInfo@` object with error `@code@` value `@91004@`

I'm not sure I agree with this.

IIRC in the original draft of the 0.9 spec the error from realtime was wrapped like this, but we decided to remove the wrapper on the grounds that any of the ways a client might want to programmatically take an action to actually fix the reason that an enter failed -- rather than just observing and shrugging -- require listening for a specific realtime-generated error code.

E.g. someone might be doing [this](https://support.ably.io/support/solutions/articles/3000056545-recommendations-for-incrementally-authorising-new-capabilities) and want to listen for 40160 `operation not permitted with provided capability` in response to an enter, in order to trigger a reauth.

Perhaps we could just prepend the `.message` with `Presence auto-re-enter failed: ` to make it more obvious what's happened in the log message without wrapping the real code?

mattheworiordan: Well expecting a user to look for a specific message text is just horrible for so many reasons and is exactly why we have error codes.

I really dislike the fact that we are proposing to emit an error on the `update` event specifically because we want developers to be able to handle a re-enter failure, yet they have no way of actually knowing that the error relates to that problem reliably i.e. any other types of errors could be emitted by Realtime with an `ATTACHED` with an `error`.  I realise that it's possible that `40160` would require the dev to reauth, but surely they would catch up the re-enter failure, try to enter and then catch the source reason then and handle that in the same way they would normally handle an enter failure?

paddybyers: The way other languages deal with this is to have a `cause` member of an error that can be another error. So the outer error has the specific code indicating enter failure, and the wrapped error is the primary cause.

mattheworiordan: @paddybyers I would be happy to have cause and even some type of metadata so that, for example, the enter presence message that failed was included, however I think we should defer that until a later spec personally as it's just not a big enough problem.  WDYT?

mattheworiordan:
On the realtime doing de-duping:

> Also, the client might intentionally be updating the state of the member, with a message having a newer timestamp, and might care that there is an update with that even if the data is the same.

How about in a future version of the spec, we propose that the client is responsible for keeping track of in-flight enter presence events. If a subsequent enter is then called whilst the previous one is in-flight it simply omits sending a new enter and waits until the in-flight enter succeeds or fails?

content/client-lib-development-guide/features.textile
paddybyers: It was synchronised previously, but is known to be getting stale.

In the `disconnected` or `connecting` state, you get the current (potentially slightly out-of-date) state, but you know that you will also see one or other of the following events within a known and bounded amount of time:

- a `connected` event, after which you know you will see any events for updates that were not in the state you saw when you called `get()`;

- a `suspended` event, which you know means that you no longer have any assurance of synchronisation, and no timescale that you can rely on for any future change in that status.

So the difference I think is that in `suspended` all you know is that it is potentially stale to an indefinite degree. In the other states, you know that within a bounded time you will either see any events necessary to bring it up to date, or an event telling you that it's in the indefinitely stale state.

content/client-lib-development-guide/features.textile
paddybyers: I don't think adding the extra metadata is the way to go, but adding an `ErrorInfo` cause ought to be straightforward enough. But if we defer adding `cause` then that would mean that right now we return only the outer error and code instead of the underlying error. Are we ok with that?

SimonWoolf: > they have no way of actually knowing that the error relates to that problem reliably i.e. any other types of errors could be emitted by Realtime with an ATTACHED with an error

Fair point. In practice I'm not sure that would actually be a problem, not sure there's any intersection - the reasons you can't enter a presence channel are all 910xx. (The `resume` flag also disambiguates: if realtime sends an ATTACHED with `resumed`=`true`, that's not emitted, so an `update` with an error and `resumed`=`true` must not have been an ATTACHED).

But I don't feel strongly about it, so if you strongly feel it should be wrapped, I'm happy to go along with that.

(Not sure about `cause`. Giving ErrorInfo an extra attribute of another ErrorInfo to solve this one relatively obscure situation feels like its overcomplicating the API)

mattheworiordan: Agreed with @paddybyers we'll go with the fixed error code and [later when cause is supported](https://github.com/ably/docs/issues/242), the customer can see the root cause of the error.

mattheworiordan:
@paddybyers @SimonWoolf this PR is now updated following my chat with Paddy.

Please note the revised `RTP11d` spec is now implemented in Ruby at https://github.com/ably/ably-ruby/pull/99/commits/f095a2a0acb7ac097e5c5cdf663802061f992b10, and this PR now depends on a new commit in the ably-common PR at https://github.com/ably/ably-common/pull/20.

Final feedback please.

paddybyers:
> Final feedback please.

It's not final :)

mattheworiordan:
>> Final feedback please.

> It's not final :)

0.9 final feedback please 😄

paddybyers:
LGTM

PR #257, opened 2017-01-20
WIP: Push notifications spec.

tcard:


tcard:
I've added some docs for the push API endpoints.

content/client-lib-development-guide/features.textile
```diff
@@ -1636,6 +1636,7 @@ enum PushDeviceType:
   "mobile" // PDT1
   "tablet" // PDT1
   "desktop" // PDT1
+  "embedded" // PDT1
```
mattheworiordan: I had previously suggested watch, which would probably be better as wearable in fact. However, looking at the docs, it appears you cannot send notifications to wearables, only to the mobile devices which in turn decide whether they display on the wearable or not, so leave as is.

content/client-lib-development-guide/features.textile
```diff
@@ -1473,6 +1495,82 @@ enum StatsIntervalGranularity:
...
 
+class DeviceDetails:
+  + fromLocalDevice() => io DeviceDetails
+  id: String
+  platform: DevicePlatform
+  formFactor: DeviceFormFactor
+  clientId: String?
+  metadata: Dict<String, String>
+  updateTokenId: String
+  resetId(String)
```
mattheworiordan: I thought we had agreed on just `resetId` i.e. not allowing a user to explicitly set the device ID?  Given the registration target is linked to a unique device ID, I am not sure I can think of  a good reason when a dev should be able to set the device ID.

content/client-lib-development-guide/features.textile
```diff
@@ -1473,6 +1495,82 @@ enum StatsIntervalGranularity:
...
   MONTH
 
+class DeviceDetails:
+  + fromLocalDevice() => io DeviceDetails
+  id: String
+  platform: DevicePlatform
+  formFactor: DeviceFormFactor
+  clientId: String?
+  metadata: Dict<String, String>
+  updateTokenId: String
```
mattheworiordan: Is this not just `updateToken` given `resetUpdateToken` below?

content/client-lib-development-guide/features.textile
```diff
@@ -1473,6 +1495,82 @@ enum StatsIntervalGranularity:
...
+class DevicePushDetails:
+  transportType: DevicePushTransportType
+  transportId: String
+  transportToken: String?
+  state: .Active | .Failing | .Failed
+  errorReason: ErrorInfo?
+  metadata: Dict<String, String>
+
+class Push:
+  publish(options: Dict<String, String>, JsonObject) => io
```
mattheworiordan: is options not params?  params used below for get  / remove etc. and these options are required to publish so I would argue options is the wrong word to describe the argument.

content/rest-api/index.textile
```diff
@@ -461,6 +469,251 @@ bc[json]. [{
...
+h3(#post-channel-subscription). Subscribe to a channel
+
+Subscribe either a single device or all devices associated with a client ID to receive push notifications from messages sent to a channel.
+
+h6. POST rest.ably.io/push/channelSubscriptions
+
+The request body contains subscription details and is an object of the form:
+
+bc[json]. {
+  deviceId: <optional, string, cannot be set when clientId is set>,
```
mattheworiordan: Perhaps change this to `<optional, string, must be set if `clientId` is empty, cannot be set when `clientId` is set>`

content/rest-api/index.textile
```diff
@@ -461,6 +469,251 @@ bc[json]. [{
...
+
+h3(#delete-device-registration). Unregister device for push notifications
+
+h6. DELETE rest.ably.io/push/deviceRegistrations/@<device id>@
+
+Unregister device. All its subscriptions will be removed.
+
+Example request:
+
+bc[sh]. curl -X DELETE https://rest.ably.io/push/deviceRegistrations/01ARZ3NDEKTSV4RRFFQ69G5FAV \
```
mattheworiordan: If a device is deleted, are all associated subscriptions deleted?

content/rest-api/index.textile
```diff
@@ -461,6 +469,251 @@ bc[json]. [{
...
+
+h5. Returns
+
+A successful request returns nothing.
+
+An unsuccessful request returns an error.
+
+h3(#delete-channel-subscription). Unsubscribe from push notifications from channel
+
+h6. DELETE rest.ably.io/push/channelSubscriptions
```
mattheworiordan: I suspect common operations required for delete may be:

* Delete all subscriptions for this device 
* Delete all subscriptiosn for this clientId
* Delete all devices and associated subscriptiosn matching this clientId
* Delete all devices and associated subscriptions matching this deviceId

Not sure that is possible from this API unless I am misreading this.

content/rest-api/index.textile
tcard: Yes.

content/rest-api/index.textile
tcard: Done at 1d7571245a90a5d9fe927d0f4c68ab8eb03081b3.

content/client-lib-development-guide/features.textile
tcard: Fixed at 1d7571245a90a5d9fe927d0f4c68ab8eb03081b3.

content/client-lib-development-guide/features.textile
tcard: The original reason was that you might want to inherit all the server-side state if e. g. you reinstalled the app, and not do it by client ID. But yes, I think we agreed on not allowing this in the end. Fixed at 1d7571245a90a5d9fe927d0f4c68ab8eb03081b3.

content/client-lib-development-guide/features.textile
tcard: Yes, fixed at 1d7571245a90a5d9fe927d0f4c68ab8eb03081b3.

content/rest-api/index.textile
tcard: > Delete all subscriptiosn for this clientId

`DELETE /push/channelSubscriptions?deviceId=...`

> Delete all devices and associated subscriptions matching this deviceId

I guess here "all devices" is only one device, given deviceId is unique? Then `DELETE /push/deviceRegistrations/<deviceId>`.

> Delete all subscriptions for this device
> Delete all devices and associated subscriptiosn matching this clientId

Those are not implemented but should be straightforward. Added to the backlog.

tcard:
PTAL

mattheworiordan:
LGTM

tcard:
I've added documentation for extras/push in messages, direct push, etc. Please take a look.

content/rest-api/index.textile
```diff
@@ -17,11 +17,13 @@ jump_to:
...
     - register device#post-device-registration
     - subscribe to a channel#post-channel-subscription
+    - unregister device#delete-device-registration
+    - unsubscribe from channels#delete-channel-subscription
     - get registered device#get-device-registration
     - list registered devices#get-device-registrations
     - list channel subscriptions#get-channel-subscriptions
-    - unregister device#delete-device-registration
-    - unsubscribe from channel#delete-channel-subscription
+    - list channels#get-channels
```
mattheworiordan: Given our recent conversation about range queries in Cassandra being difficult, would list channels not be a range query or do we store the list of channels registered in a bucket?

content/rest-api/index.textile
```diff
@@ -660,57 +815,103 @@ bc[json]. [{
...
+bc[json]. {
+  registrationToken: <string>
+}
+
+For web:
+
+bc[json]. {
+  targetUrl: <string>
+  encryptionKey: <string>
+}
```
mattheworiordan: Surely the user should also be able to specify a recipient type with a `deviceId` or a `clientId`?

content/rest-api/index.textile
tcard: It's in a bucket, yes. Everything about push is in a bucket.

content/rest-api/index.textile
tcard: Ah, I thought direct push was just for specific devices, completely bypassing registrations. OK, I'll add that.

mattheworiordan: > Ah, I thought direct push was just for specific devices, completely bypassing registrations

It is for specific devices, but customers will either have the native registration tokens (if for example they registered the device for push without Ably), or they will have a device ID issued by Ably which they can use to do a direct push to a device.

tcard: Done at 0887e96, implemented at https://github.com/ably/realtime/pull/737/commits/0fc483b658a3ba687ea3801e6622d07553e570b5 (also in the worker in the `wip` branch: https://github.com/ably/realtime/commit/ded0698991753a34a8b55fcfa450e28fd29abc9a but that's a bit unstable).

tcard:
Please see 0ab0811, to update existing registrations with PUT.

tcard:
@mattheworiordan I've adjusted the IDL to the latest changes I've made in Java. Please check if everything looks good to you. I've added some rationale in 8c487a9's commit message.

tcard:
I've also added some specs, specifically for the state machine that describes the activation/deactivation process. That one is a bit tricky, and by trying to be as platform-agnostic as possible it ends up being _too_ abstract, but well, there's the Java and soon iOS examples to follow along.

content/client-lib-development-guide/features.textile
```diff
@@ -1282,8 +1371,10 @@ class RealtimeChannel:
   push: PushChannel
 
 class PushChannel:
-  subscribe(type: .Device | .ClientId) => io
-  unsubscribe(type: .Device | .ClientId) => io
+  subscribeDevice() => io
+  subscribeClientId() => io
+  unsubscribeDevice() => io
+  unsubscribeClientId() => io
```
ricardopereira: So, the type is implicitly added from `options` or `device`?

content/rest-api/index.textile
```diff
@@ -20,6 +20,7 @@ jump_to:
     - unregister device#delete-device-registration
     - unsubscribe from channels#delete-channel-subscription
     - update a device registration#update-device-registration
+    - reset a device's update token#reset-update-token
```
ricardopereira: And for the reset device id?

ricardopereira: @tcard ^

content/client-lib-development-guide/features.textile
ricardopereira: @tcard ^

content/client-lib-development-guide/features.textile
```diff
@@ -1473,6 +1586,85 @@ enum StatsIntervalGranularity:
...
+enum DevicePlatform:
+  "android" // PPT1
+  "ios" // PPT1
+  "browser" // PPT1
+
+enum DeviceFormFactor:
+  "mobile" // PDT1
+  "tablet" // PDT1
+  "desktop" // PDT1
+  "embedded" // PDT1
```
ricardopereira: I can detect many device forms in the Cocoa ecosystem:

 - Phone
 - Tablet
 - Watch
 - TV
 - Car
 - Desktop
 - Simulator

Should we update this or it's fine for now?

content/client-lib-development-guide/features.textile
```diff
@@ -859,6 +864,85 @@ h3(#channel-states-operations). @RealtimeChannel.state@ effects on channel opera
...
+** @(RSH2a)@ @Push#activate@ sends a @CalledActivate@ event to "the state machine":#RSH3.
+** @(RSH2b)@ @Push#deactivate@ sends a @CalledDeactivate@ event to "the state machine":#RSH3.
+** @(RSH2c)@ Whenever, from the platform's APIs, details for sending push notifications to the local device (e. g. GCM's registration token) is received, a @GotPushDeviceDetails@ event is sent to "the state machine":#RSH3.
+
+h3(#activation-state-machine). Activation state machine
+
+* @(RSH3)@ In platforms that support receiving push notifications, in order to connect the device's push features with Ably's, the library must perform the process described in the following abstract state machine. While this process should be implemented in whatever way better fits the concrete platform, it should be taken into account that its lifetime is that of the the _app_ that runs it, which outlives that of the @Rest@ instance or (typically) the process running the app. This typically forces some kind of on-disk storage to which the state machine's state must be persisted, so that it can be recovered later by new instances and processes running the app triggered by external events.
```
ricardopereira: The activation state machine should be a singleton to guarantee consistency of the persistent data. Now, there's a problem because, i.e. in iOS cannot guarantee the singleton if the user can create multiple rest clients with different keys. Which one will the state machine use? A device can only be registered to one Ably key? An app could use different Ably keys?

ricardopereira: @mattheworiordan ^

content/client-lib-development-guide/features.textile
mattheworiordan: Personally I don't see any harm in supporting additional types, so long as they map to the ones we use. I also think `phone` is a better classification than `mobile`, so perhaps we should change this @tcard and add the other ones from the Cocoa ecosystem?

content/client-lib-development-guide/features.textile
mattheworiordan: I think for push there is only one target which is the device, so if state is shared between all regardless of the instance of the library, I don't see that being a problem personally.

content/client-lib-development-guide/features.textile
tcard: Yes.

content/rest-api/index.textile
tcard: That's immutable.

content/client-lib-development-guide/features.textile
```diff
@@ -859,6 +864,86 @@ h3(#channel-states-operations). @RealtimeChannel.state@ effects on channel opera
...
+** @(RSH3a)@ State @NotActivated@ (the initial one).
+*** @(RSH3a1)@ On event @CalledDeactivate@:
+**** @(RSH3a1a)@ Makes @Push#deactivate@ return or call its callback with no error.
+**** @(RSH3a1b)@ Transitions to @NotActivated@.
+*** @(RSH3a2)@ On event @CalledActivate@:
+**** @(RSH3a2a)@ If the local device has @id@ and @updateToken@, transitions to @WaitingForNewPushDeviceDetails@.
+**** @(RSH3a2b)@ If the local device has the necessary push details (registration token, etc.), transitions to @GotPushDeviceDetails@.
```
ricardopereira: `GotPushDeviceDetails` isn't a state so this should be: _send `GotPushDeviceDetails` event_?

content/client-lib-development-guide/features.textile
```diff
@@ -1282,8 +1282,10 @@ class RealtimeChannel:
   push: PushChannel
 
 class PushChannel:
-  subscribe(type: .Device | .ClientId) => io
-  unsubscribe(type: .Device | .ClientId) => io
+  subscribeDevice() => io
+  subscribeClientId() => io
```
mattheworiordan: I assume in the spec this should fail if the user is not authenticated with a `clientId`?

tcard: Yes, that's what happen in Java already too.

content/client-lib-development-guide/features.textile
tcard: @ricardopereira Right. Done at 78db45fde344f381af49e52804a3575431418468, along a few additions; please take a look at those.

content/client-lib-development-guide/features.textile
tcard: Done at 8eea4a1.

tcard:
@mattheworiordan I've finally pushed some docs: https://ably-docs-pr-257.herokuapp.com/rest/push-notifications/ Far for perfect or complete I'm afraid. But it might be good enough to get going.

mattheworiordan:
> I've finally pushed some docs: https://ably-docs-pr-257.herokuapp.com/rest/push-notifications/ Far for perfect or complete I'm afraid. But it might be good enough to get going.

Thanks, will review

mattheworiordan:
@tcard I believe this documentation is almost there in terms of structure and a good way in terms of content.  You can see this at https://ably-docs-pr-257.herokuapp.com/realtime/push/ along with various support articles I have added to the support site which were needed, see https://support.ably.io/solution/folders/3000012235

What's outstanding is:

* [ ] Admin API not documented
* [ ] API for `push.*` and `PushChannel` still needed.  Docs exist, but we need more formal API definitions below the fold.  This can come a little later, but is needed for consistency.
* [ ] Javascript & Ruby examples throughout (where applicable)
* [ ] iOS and Android for admin API (I will do JS & Ruby above)
* [ ] Update Smart Notifications page to link to these docs
* [ ] Duplicate push documentation for REST API 
* [ ] Integrate into the website (I have [updated the push branch](https://github.com/ably/website/pull/990) with a few small fixes)
* [ ] Update client & channel documentation to now link to push documentation for publish methods, push attribute on client, link to PushChannel from Channel.

content/rest-api/index.textile
```diff
@@ -616,11 +614,17 @@ An unsuccessful request returns an error.
 
 h3(#update-device-registration). Update a device registration
 
+Device registrations can be either upserted (the existing registration is replaced entirely) with a @PUT@ operation, or specific attributes of an existing registration can be updated using a @PATCH@ operation:
+
 h6. PUT rest.ably.io/push/deviceRegistrations/@<device id>@
 
-The body must have the same shape as when "registering the device":#post-device-registration, except only fields to be changed should be provided. The @PUT@ operation is in fact a @PATCH@ operation i.e. only fields provided are updated.
+The body must have the same shape as when "registering the device":#post-device-registration. The @PUT@ operation will replace the existing device registration, however please bear in mind that a registered device in Ably is largely immutable. As such, only the following attributes are currently updateable and any attempt to modify other fields will result in the update failing:
```
tcard: s/updateable/updatable

content/rest-api/index.textile
```diff
@@ -616,11 +614,17 @@ An unsuccessful request returns an error.
...
+
+* @clientId@
+* @metadata@
```
tcard: Not updatable.

content/rest-api/index.textile
```diff
@@ -616,11 +614,17 @@ An unsuccessful request returns an error.
 
 h3(#update-device-registration). Update a device registration
 
+Device registrations can be either upserted (the existing registration is replaced entirely) with a @PUT@ operation, or specific attributes of an existing registration can be updated using a @PATCH@ operation:
```
tcard: "using a PATCH operation" could use a link to the h6 below.

content/rest-api/index.textile
```diff
@@ -616,11 +614,17 @@ An unsuccessful request returns an error.
...
 
-The body must have the same shape as when "registering the device":#post-device-registration, except only fields to be changed should be provided. The @PUT@ operation is in fact a @PATCH@ operation i.e. only fields provided are updated.
+The body must have the same shape as when "registering the device":#post-device-registration. The @PUT@ operation will replace the existing device registration, however please bear in mind that a registered device in Ably is largely immutable. As such, only the following attributes are currently updateable and any attempt to modify other fields will result in the update failing:
+
+* @clientId@
+* @metadata@
+* @push.recipient@
 
-Only @metadata@ and @push.recipient@ can be updated. If non-null, the provided new values will be merged with their old values.
+Additionally, if the @push.state@ or @updateToken@ attributes are provided, they will be accepted if they match the existing value. However if they differ (and are not @null@ or @omitted@), then the update will fail.
```
tcard: That isn't just for push.state or updateToken; all non-updatable (but settable) fields must match. Otherwise, this would really be a PATCH. In short, the body must be a valid POST body too, since we're inserting a new registration if it doesn't exist.

> (and are not @null@ or @omitted@)

This isn't true, @null@ is treated like a value just any other. There's (thankfully) no @undefined@. See my point above about omitted fields.

(Also, omitted in monospaced makes it look like a special value or something like that.)

content/rest-api/index.textile
```diff
@@ -630,18 +634,56 @@ bc[sh]. curl -X PUT https://rest.ably.io/push/deviceRegistrations/01ARZ3NDEKTSV4
...
+- Accept := @application/json@ by default, or @application/x-msgpack@, @text/html@
+- Auth required := yes ("basic":#basic-authentication or "token":#token-authentication with @push-admin@ capability)
+
+A successful request returns the updated device details.
+
+An unsuccessful request returns an error.
+
+h6. PATCH rest.ably.io/push/deviceRegistrations/@<device id>@
+
+The body must have the same shape as when "registering the device":#post-device-registration, except only fields to be changed should be provided. Any fields provided replace existing values. Please bear in mind that fields that are object (Hash) types will replace existing values as opposed to be merged together. @metadata@ and @push.recipient@ are examples of these types. Currently only the following attributes are currently updateable and any attempt to modify other fields will result in the update failing:
```
tcard: Hash is pretty Ruby-specific. I'd say something like "fields whose values are structured (JSON-like arrays or objects)".

s/updateable/updatable/, and there's a duplicate "currently" in "Currently only the following attributes are currently" (I'd really remove both; all this documentation is "currently" implicitly).

content/rest-api/index.textile
```diff
@@ -630,18 +634,56 @@ bc[sh]. curl -X PUT https://rest.ably.io/push/deviceRegistrations/01ARZ3NDEKTSV4
...
+
+* @clientId@ (this field is only editable with a @push-admin@ capability)
+* @metadata@ (this field is only editable with a @push-admin@ capability)
```
tcard: Really? I thought you said this was only to be set by Ably, or by the users under controlled circumstances (when we have them). So we're letting users putting whatever they want here?

content/rest-api/index.textile
```diff
@@ -821,7 +863,7 @@ A successful request returns an empty response.
 
 An unsuccessful request returns an error.
 
-h3(#post-channel-subscription). Subscribe to a channel
```
tcard: Forgot to change the HTTP method in the h6 below.

content/client-lib-development-guide/features.textile
```diff
@@ -866,7 +866,20 @@ h3(#channel-states-operations). @RealtimeChannel.state@ effects on channel opera
 
 h2(#push-notifications). Push notifications
 
-* @(RSH1)@ @Push#publish(recipient, data)@ performs a HTTP request to "/push/publish":/rest-api/#push-publish
+* @(RSH1)@ @Push#admin@ object provides the following interface:
+** @(RHS1a)@ @#publish(recipient, data)@ performs an HTTP request to "/push/publish":/rest-api#push-publish. Empty values for @recipient@ or @data@ should be immediately rejected. An end-to-end push notification test can be made using the special test-only @ablyChannel@ recipient. Additionally, tests should exist with valid and invalid recipient details
+** @(RHS1b)@ @#deviceRegistrations@ provides access to the admin @PushDeviceRegistrations@ object with the following methods:
+*** @(RHS1b1)@ @#get(deviceId)@ performs a request to "/push/deviceRegistrations/:deviceId":/rest-api#get-device-registration and returns a @DeviceDetails@ object if the @deviceId@ is found or results in a not found error if the device cannot be found
```
tcard: Not found shouldn't return an error, just return a null value.

content/client-lib-development-guide/features.textile
```diff
@@ -1590,27 +1603,22 @@ enum StatsIntervalGranularity:
...
-  state: .Active | .Failing | .Failed
   errorReason: ErrorInfo?
-  metadata: JsonObject
+  recipient: JsonObject
+  state: .Active | .Failing | .Failed
 
 class LocalDevice extends DeviceDetails:
-  resetId()
-  resetUpdateToken() => io
+  resetUpdateToken(token: String?) => io
```
tcard: `resetUpdateToken` without the parameter is a very different operation that with it (assuming you mean it to set a update token gotten from another request, to handle the 401 case we spoke about). The former will call `resetUpdateToken`, the latter will not. I'd really use two separate methods, don't you think? Otherwise, it may look like passing a token will also try to set it in the server, which isn't possible.

content/client-lib-development-guide/features.textile
```diff
@@ -1628,20 +1636,27 @@ class Push:
...
   deviceRegistrations: PushDeviceRegistrations
   channelSubscriptions: PushChannelSubscriptions
+  publish(recipient: JsonObject, data: JsonObject) => io
 
 class JsonObject:
   // Platform-dependant, typically a Dict-like object
 
 class PushDeviceRegistrations:
+  get(device: DeviceDetails) => io DeviceDetails
+  get(deviceId: String?) => io DeviceDetails
```
tcard: Just `String`, and returns `DeviceDetails?`.

content/client-lib-development-guide/features.textile
```diff
@@ -1628,20 +1636,27 @@ class Push:
...
+  list(params: Dict<String, String>?) => io PaginatedResult<DeviceDetails>
   save(DeviceDetails) => io
-  get(params: Dict<String, String>?) => io PaginatedResult<DeviceDetails>
-  remove(params: Dict<String, String>?) => io
+  remove(device: DeviceDetails) => io
+  remove(deviceId: String?) => io
```
tcard: Just `String`.

content/client-lib-development-guide/features.textile
```diff
@@ -1628,20 +1636,27 @@ class Push:
 class PushAdmin:
   deviceRegistrations: PushDeviceRegistrations
   channelSubscriptions: PushChannelSubscriptions
+  publish(recipient: JsonObject, data: JsonObject) => io
 
 class JsonObject:
   // Platform-dependant, typically a Dict-like object
 
 class PushDeviceRegistrations:
+  get(device: DeviceDetails) => io DeviceDetails
```
tcard: Just `String`, and returns `DeviceDetails?`.

content/client-lib-development-guide/features.textile
```diff
@@ -1628,20 +1636,27 @@ class Push:
...
+  remove(device: DeviceDetails) => io
+  remove(deviceId: String?) => io
+  removeWhere(params: Dict<String, String>?) => io
 
 class PushChannelSubscriptions:
-  save(PushChannelSubscription) => io
-  listChannels() => io PaginatedResult<String>
-  get(params: Dict<String, String>?) => io PaginatedResult<PushChannelSubscription>
-  remove(params: Dict<String, String>?) => io
+  get(pushChannelSubscription: PushChannelSubscription) => io PushChannelSubscription
```
tcard: We don't really need to give a name to this parameter.

Returns `PushChannelSubscription?`.

content/rest-api/index.textile
mattheworiordan: >  So we're letting users putting whatever they want here?

No, it explicitly states they need `push-admin` privileges meaning a user with a `push-subscribe` privilege is never able to update these attributes that may need to be "trusted" attributes. But if you have admin rights, you should be able to update them.  Am I missing something?

content/client-lib-development-guide/features.textile
mattheworiordan: > Not found shouldn't return an error, just return a null value.

Why?  It's a 404 if the url `push/deviceRegistration/:id` does not exist, so why should we return `null`? Calling the equivalent `find` or `get` in most ORMs will raise an exception if the item does not exist as `get` is not an operation to check if the item exists, but a request to get an item that ends up not existing.

content/client-lib-development-guide/features.textile
mattheworiordan: Ok, but if we are to have separate methods, then I would argue `resetUpdateToken` is  misleading regardless of which request you meant to make.

Should it not perhaps be:

`setLocalUpdateToken(updateToken: String)`
and 
`reissueUpdateToken() => io`

The reason I went for `setLocal..` as opposed to just `setUpdateToken` is to attempt to make it clear that this update only affects the LocalDevice.  Saying that, perhaps that's a bit unnecessary given the object is a `LocalDevice` anyway?

content/client-lib-development-guide/features.textile
mattheworiordan: Ok, I'll remove the unnecessary parameter in all methods then where the type make its completely obvious what it is.

mattheworiordan:
@tcard I've pushed an update 5b40ca0. The only part that I think specifically needs your review is the `LocalDevice` method section.

content/client-lib-development-guide/features.textile
mattheworiordan: BTW. So that you know, it already does this, see https://github.com/ably/realtime/blob/master/test/rest/push.js#L155-L168

mattheworiordan:
@tcard note that this is now merged. However, the `push` docs are far from complete at this stage.  I will simply add a new issue to keep track of outstanding items.

Issue #251, opened 2017-01-18
REST Channel enumeration API docs


mattheworiordan:
Are missing, we should add beta docs for this.

We will also need to document the request method

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/EDU-1793) by [Unito](https://www.unito.io)

mattheworiordan:
Also need to add channel release documentation

mattheworiordan:
See https://github.com/ably/docs/issues/119#issuecomment-226884777 in regards to JS enumeration

PR #260, opened 2017-01-20
1.0 release

mattheworiordan:
@paddybyers @SimonWoolf @tcard this PR introduces version 1.0 ready for release when the libraries are ready.  Note that:

* Docs milestones are updated, see https://github.com/ably/docs/milestones.  As are [wiki milestones](https://github.com/ably/wiki/milestones)
* Before we can release this Ricardo will need to check the docs for iOS 1.0 and Paddy / Peter will need to check the docs for Android/Java 1.0. Simon to review JS and Matt to review Ruby to make sure it's all up to date for REST & Realtime dos.
* We will need to bump all the client libs to match the 1.0 tag


mattheworiordan:
* [ ] When we release 1.0 we need to update tutorials that reference `0.9.0-beta` and `0.8.0` to `1.0.0`.  Actually updating the source tutorials themselves is probably unnecessary as they will still work.

SimonWoolf:
Should we merge the `source-0-10` branch into this?

mattheworiordan:
No, because `0-10` is actually `1.1` no?

SimonWoolf:
> No, because 0-10 is actually 1.1 no?

I just figured because the only thing in that branch that isn't in master is [the authUrl response validation](https://github.com/ably/docs/pull/206), and you've indicated in https://github.com/ably/infrastructure/issues/1017 that you'd like implementing in ably-js now (i.e. for 0.9/1.0 release).

If you'd rather it be done in ably-js as a special case and kept out of the spec, that's completely fine, just wanted to make sure that was a deliberate decision and not an oversight

mattheworiordan:
@paddybyers @tcard @SimonWoolf if you do get a chance, please review my changes to the docs I have made tonight, see https://github.com/ably/docs/compare/5d87c9b...155fa73.  Lots of fixes for inconsistencies and other issues we had open in this docs repo.

I am merging this into `master` now so that any issues we see can simply be added as commits on `master`

content/realtime/presence.textile
```diff
@@ -899,6 +901,12 @@ h4. Parameters
...
+* The registered <span lang="default">listener</span><span lang="ruby">block</span><span lang="csharp">handler</span> remains active on the presence channel regardless of the underlying channel state. For example, if you call @subscribe@ when the underlying channel is @attached@ and it later becomes @detached@ or even @failed@, when the channel is reattached and a presence message is received, the <span lang="default">listeners</span><span lang="ruby">blocks</span><span lang="csharp">handlers</span> originally registered will still be invoked. Listeners are only removed when calling "@unsubscribe@":#unsubscribe or when the underlying channel is @released@ using the <span lang="default">@Realtime.channels.release(name)@</span><span lang="csharp">@Realtime.Channels.Release(name)@</span> method
```
SimonWoolf: > `default">@Realtime.channels.release(name`

The js docs generally use `realtime` rather than `Realtime` for an instance of the client lib, `Realtime` is the constructor (ie `var realtime = new Ably.Realtime`)

content/tutorials/queue-stomp-neutrino-profanity.textile
```diff
@@ -233,7 +233,7 @@ We're now ready to start consuming messages from the queue. There are "plenty of
 
 ```[json]
   "dependencies": {
-    "ably": ">=0.9.0-beta",
+    "ably": "~1.0",
```
SimonWoolf: > `"~1.0"`

I don't think this is right, assuming we want something that matches `>= 1.0.0 && < 2.0.0` -- you want either `~1` or `^1.0`; `~1.0` locks into the minor version

(similarly for the others)

content/tutorials/token-authentication.textile
```diff
@@ -607,7 +607,7 @@ blang[android].
   ...
   dependencies {
       ...
-      compile 'io.ably:ably-android:0.8.7'
+      compile 'io.ably:ably-android:1.0.0'
```
SimonWoolf: Do we want to lock into the .0.0 patch version? At a quick google looks like gradle uses https://ant.apache.org/ivy/history/2.1.0/settings/version-matchers.html, so I think we should be able to specify `1.0+` or `1+` to lock into the minor or major versions resp

SimonWoolf:
:+1:

content/tutorials/token-authentication.textile
paddybyers: Best practice is a matter of debate: http://blog.danlew.net/2015/09/09/dont-use-dynamic-versions-for-your-dependencies/

Last time we discussed it we decided to use a fixed version reference.

content/realtime/presence.textile
mattheworiordan: > The js docs generally use realtime rather than Realtime for an instance of the client lib,

Well as far as I can tell, we haven't done either. We only use `realtime.*` in code example, never in documentation itself. I searched the docs now, cannot find one instance of that or channel.* etc either.

Unfortunately the issue here is in fact that we don't have docs for the Channels collection, see https://github.com/ably/docs/issues/251. I think I'm going to leave this as is until we address that problem and then we can use the pattern we have used in the docs such as "... if either @publish@ or @subscribe@ for the @Channel@ is called ..."

content/tutorials/queue-stomp-neutrino-profanity.textile
mattheworiordan: Well we previously locked into minor versions, I don't really see the harm, but sure I can change to `~1`

content/tutorials/token-authentication.textile
mattheworiordan: I must admit @paddybyers I really don't see why we're special casing Java over every other library?  The developer referencing the Java lib should decide if they want to lock to a specific version or not, however the documentation itself, unless we update it for every Java release, will simply encourage people to use old unpatched versions of the library. My vote is to add a note to say it's up to the developer to decide if they want to lock into a major, minor or patch release, but the docs should encourage a user to lock into the version that matches this documentation which is `1.0+`

mattheworiordan: Thoughts?

PR #261, opened 2017-01-23
api: WebHook HMAC SHA-256 instructions + demo

mattheworiordan:
This PR fixes a typo in our WebHooks docs (we previously incorrectly mentioned AES256 instead of SHA-256 for the HMAC algorithm), and also adds some instructions to test the signature verification to help developers check this themselves more easily in their chosen language.

@paddybyers @SimonWoolf you happy for me to merge in?

### DIFF
--- a/content/code/reactor/webhook-hmac-sha-256.code
+++ b/content/code/reactor/webhook-hmac-sha-256.code
@@ -1,0 +1,67 @@
+[--- Javascript ---]
+$('#check-hmac-btn').on('click', function() {
+  var stripRegex = /^\s+|\s+$/;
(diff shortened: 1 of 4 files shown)


paddybyers:
LGTM

SimonWoolf:
:+1: 

(Would be nice if we could use webcrypto rather than cryptojs, but browser support is apparently still pretty bad, no support from IE at all. ah well)

mattheworiordan:
> Would be nice if we could use webcrypto rather than cryptojs

Agreed, but as this is just a demo I think it's fine

PR #262, opened 2017-01-24
Documentation intros + some restructuring

mattheworiordan:
@paddybyers @SimonWoolf @tcard I've been working on the docs in preparation for queue documentation.

These changes loosely cover:

* REST & Realtime index (home) pages are now no longer just an index to the content but instead serve as the starting place for people to understand a bit more about how the library is typically used.  
* I mention alternatives now for REST & Realtime client libraries such as queues, Webhooks & adapters.
* Quickstart guide aims to cover key concepts rather than just some code
* Authentication has been restructured quite significantly so that:
  * It's more approachable at the outset outlining key concepts
  * REST specific details are moved into REST API docs as not needed for most people
  * Identified clients blurb added in 
  * Docs fix for [scope intersection issue](https://github.com/ably/realtime/issues/780) until we resolve that
* Attempt to generally dumb things down at the outset by avoiding getting into too much detail too early on

Please remember you can review this in HTML format using the Heroku review app link that will appear below.


content/general/authentication.textile
```diff
@@ -181,7 +190,7 @@ Given an API key exists with the following capabilities:
 
 ```[json]
 {
-  "chat": ["publish", "subscribe", "presence"],
+  "chat*": ["publish", "subscribe", "presence"],
```
SimonWoolf: should be `chat:*` not `chat*`

content/general/authentication.textile
```diff
@@ -226,176 +235,59 @@ auth.requestToken({ capability: {
...
 
-h3. Parameter canonicalisation
+```[javascript]
+auth.requestToken({ capability: {
+  "chat:*": ["*"]
+}}, tokenCallback)
+```
 
-The parameters of the token request are normalized/canonicalized as follows:
+Then Ably will be unable to issue a token because the requested channel scope is greater than the API key's capabilities i.e. @chat:*@ is requested yet the API key only has acccess to the narrower @chat:team:*@ channel scope. In the example above, @requestToken@ will call the callback with an error.
```
SimonWoolf: > `requestToken` will call the callback with an error.

might be worth clarifying that this is only because the resulting token would have literally no capabilities. If the token requests two things and one is satisfiable, it'll happily issue a token with just that one

content/general/authentication.textile
```diff
@@ -226,176 +235,59 @@ auth.requestToken({ capability: {
...
-First the canonicalized request text, constructed as follows:
+In Ably a client can be identified with a @client ID@ in three ways:
 
-* start with the empty string
-* for each of the following fields in order: key ID, ttl, capabilities, clientId, timestamp, nonce, even when empty
-** append the canonicalized string value for that field
-** append a newline (0xa) character.
-Note that a newline character is added for each field, including any empty client ID string, and the last (nonce) field.
+* if the client is authenticated with a token that is issued for that @client ID@;
+* if the client claims that @client ID@ and is authenticated using "Basic Authentication":#basic-authentication (using an API key);
```
SimonWoolf: I thought basic auth clients were never identified? At least for the purposes of the 'only allow identified clients' namespace rule (of course, they can still publish / enter presence with a clientId)

content/general/authentication.textile
```diff
@@ -226,176 +235,59 @@ auth.requestToken({ capability: {
...
-  "token": "{{APP_ID}}.HHZNjgqmC-ACW....truncated",
-  "keyName": "{{API_KEY_NAME}}",
-  "issued": 1449745478956,
-  "expires": 1449749078956,
-  "capability":
-    "{\"*\":[\"subscribe\"],\"private\":[\"presence\",\"publish\",\"subscribe\"]}",
-  "clientId": "unique_identifier"
-}
+```[javascript](code-editor:realtime/auth-client-id)
+  var realtime = new Ably.Realtime({ key: '{{API_KEY}}' });
```
SimonWoolf: Might consider using the Rest client for consistency with the other serverside examples

SimonWoolf:
:+1: Looks good :)

content/general/authentication.textile
mattheworiordan: Thanks

content/general/authentication.textile
mattheworiordan: OK

content/general/authentication.textile
mattheworiordan: > I thought basic auth clients were never identified? At least for the purposes of the 'only allow identified clients' namespace rule (of course, they can still publish / enter presence with a clientId

Well technically no, you can append `clientId` to a websocket URL and I believe you become identified, but it's unsupported in the client libs so I will change.

content/general/authentication.textile
mattheworiordan: Ok, thanks

PR #266, opened 2017-01-31
Fix broken ably-js simple encryption example

SimonWoolf:
generateRandomKey() doesn't return a value, it takes a callback.
Plus convert the examples from 128-bit to 256-bit keys


mattheworiordan:
LGTM, merge if you want.

Issue #270, opened 2017-02-16
Document new protocol for WebHooks


mattheworiordan:
@SimonWoolf see https://app.intercom.io/a/apps/ua39m1ld/inbox/all/conversations/8272535202

Please can we:

* Document change in Webhooks format for PresenceMessage
* Include an example of processing a WebHook message in code using PresenceMessage.fromEncoded etc
* Publish to the website.

I will then send an email to everyone using the old v0.8 protocol for WebHooks

PR #271, opened 2017-02-16
Update webhook documentation

SimonWoolf:
Fixes https://github.com/ably/docs/issues/270

### DIFF
--- a/content/general/versions/v0.8/webhooks.textile
+++ b/content/general/versions/v0.8/webhooks.textile
@@ -44,7 +44,7 @@ Each channel lifecycle event in the @items@ array contains a JSON object with th
 - name := the event type "@channel.opened@" or "@channel.closed@"
 - data := a JSON object containing a single attribute @name@ which is the name of the channel that the lifecycle event applies to
 - webhookId := an internal unique ID for the configured WebHook
-- queueId := This is the WebHook type "@channel.lifecycle@"
+- source := This is the WebHook type "@channel.lifecycle@"
 - timestamp := a timestamp represented as milliseconds since epoch for the channel event
 
 h4. Example @channel.opened@ JSON payload
@@ -54,7 +54,7 @@ h4. Example @channel.opened@ JSON payload
    "items":[
       {
          "webhookId":"M8T4-Q",
-         "queueId":"channel.lifecycle",
+         "source":"channel.lifecycle",
          "serial":"4ab22b4793288073:1",
          "timestamp":1467280480250,
          "name":"channel.opened",
@@ -73,7 +73,7 @@ h4. Example @channel.closed@ JSON payload
    "items":[
       {
          "webhookId":"M8T4-Q",
-         "queueId":"channel.lifecycle",
+         "source":"channel.lifecycle",
          "serial":"4ab22b4793288073:5",
          "timestamp":1467280554968,
          "name":"channel.closed",
@@ -102,11 +102,13 @@ Each presence event in the @items@ array contains a JSON object with the followi
 - data.presence := an @Array@ of presence events with fields for each event described below
 - data.presence.clientId := the client identifier for the member. Find out more about "identified clients":https://support.ably.io/solution/articles/3000038608-authenticated-and-identified-clients
 - data.presence.connectionId := the public unique identifier for the connection the member is using. Find out more about @connectionId@":/realtime/connection#id
-- data.presence.action := a string representing the presence event being one of "@enter@", "@update@" or "@leave@"
+- data.presence.action := a string representing the presence event being one of "@enter@", "@update@" or "@leave@". (Note that in newer versions of the webhook api (identifiable by an @X-ABLY-VERSION@ header of greater than @"0.8"@) this field is an integer; if you are using a newer webhook please view a newer version of this document)
 - data.presence.data := an optional payload if provided when present. Find out more about "member data":/realtime/presence/#member-data
 - data.presence.encoding := if the data payload is present, this encoding specifies the format of the data. If missing or empty, then the data payload is a string type. For example, if JSON data is added to a presence member's "@data@":/realtime/presence/#member-data, then the @encoding@ would be "@json/utf-8@"
+- data.presence.id := unique ID assigned by Ably to this presence event
+- data.presence.timestamp := the time in milliseconds since the epoch when this presence event was received by Ably
 - webhookId := an internal unique ID for the configured WebHook
-- queueId := This is the WebHook type "@channel.presence@"
+- source := This is the WebHook type "@channel.presence@"
 - timestamp := a timestamp represented as milliseconds since epoch for the presence event
 
 h4. Example @channel.presence@ @enter@ JSON payload
@@ -116,7 +118,7 @@ h4. Example @channel.presence@ @enter@ JSON payload
    "items":[
       {
          "webhookId":"N6A8rg",
-         "queueId":"channel.presence",
+         "source":"channel.presence",
          "serial":"4ab22b4737770514:0",
          "timestamp":1467280498532,
          "name":"presence.message",
@@ -127,7 +129,9 @@ h4. Example @channel.presence@ @enter@ JSON payload
                {
                   "clientId":"Sam",
                   "connectionId":"JYNDvRnLPc",
-                  "action":"enter"
+                  "action":"enter",
+                  "id":"JYNDvRnLPc:0:0",
+                  "timestamp":1487253048676
                }
             ]
          }
@@ -143,7 +147,7 @@ h4. Example @channel.presence@ @update@ and @leave@ JSON payload
    "items":[
       {
          "webhookId":"N6A8rg",
-         "queueId":"channel.presence",
+         "source":"channel.presence",
          "serial":"4ab22b4737770514:2",
          "timestamp":1467280513129,
          "name":"presence.message",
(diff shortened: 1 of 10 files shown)


content/general/versions/v0.8/webhooks.textile
```diff
@@ -190,15 +225,86 @@ h4. Example @channel.presence@ @update@ and @leave@ JSON payload
 
 h2(#messages). Messages
 
-Receiving messages via WebHooks is currently in beta. As such, this functionality is currently only available to Enterprise customers. If you are interested in receiving messages via WebHooks, please "get in touch with us":https://www.ably.io/contact.
+When a message is published on a channel, a message WebHook event is triggered.
+
+Every WebHook request will contain one or more enveloped messages in the root level attribute @items@, each of which in turn contains one or messages. See "WebHook transport details":#webhook-transport below to find out more about batching and expected throughput.
+
+Each message envelope in the @items@ array consists of a JSON-like object with the following attributes:
```
paddybyers: I thought we banished "JSON-like"

SimonWoolf: Looks like that change happened in the spec, but not the webhook docs: https://github.com/ably/docs/commit/f960efe15a641efcb76de481971dc142abde546f . I'll get rid of it. (got added here just cause it was already "json-like" in the 0.9 version, and I copied the whole section to the 0.8 version)

SimonWoolf:
> This looks good, but I don't follow why you changed the v0.8 documentation for WebHooks. The whole point of having different versions of docs is so that we can reference old documentation depending on the client lib / protocol version?

Sure, but the 0.8 version was just out of date. Eg doesn't matter what client lib or webhook version you have, you're not going to get a `queueId` attribute in webhooks any more, it's `source` for everyone. And everyone can use message webhooks now, etc. Surely we want the 0.8 page to be documentation for people using 0.8 client libs now, not a historical record of what the service did in 2014.

The only difference in webhooks for a customer using a 0.8 client lib is that they don't have access to `fromEncoded`, and there's an extra para in the 0.8 version explaining that. Everything else is the same whatever client lib you're using -- in particular people using 0.8 client libs who make a new webhook are going to get a 1.0 webhook, so the presence action changes needs to be in both v0.8 and v1.0 docs versions, surely. (AIUI the docs version people select is for their client lib version, *not* the webhook version - for a start it's shared across different docs pages)

mattheworiordan:
> Eg doesn't matter what client lib or webhook version you have, you're not going to get a queueId attribute in webhooks any more, it's source for everyone

Sure, if that's the case, we should correct it.

>  And everyone can use message webhooks now, etc. Surely we want the 0.8 page to be documentation for people using 0.8 client libs now, not a historical record of what the service did in 2014.

Nope, it's the documentation for that version of the protocol.  We can't just selectively decide which version of the protocol people most likely want when they're reading the docs, the user can do that by selecting it.

> The only difference in webhooks for a customer using a 0.8 client lib is that they don't have access to fromEncoded, and there's an extra para in the 0.8 version explaining that

Except that `fromEncoded` won't work in 0.8 because of the string value, which is what prompted us to change this in the first place?

> (AIUI the docs version people select is for their client lib version, not the webhook version - for a start it's shared across different docs pages)

Yes, but right now, our version numbers are linked, and until they're not, that's the behaviour people get.  We could easily add a comment to the webhook on the website now showing the version number of the webhook so at least when they jump to the docs, they'll know what to be looking for.

SimonWoolf:
> Nope, it's the documentation for that version of the protocol. 

Well, that's at odds with how the website works at the moment. It stores one version per language across all pages. If you change it in authentication docs, it'll change in the webhook docs. If someone opens the language docs for js and sets the version to 0.8 because they're using ably-js 0.8, then goes to the webhook docs, they'll see 0.8. If we then show them docs for webhook version 0.8 with a string action, we'll be showing them docs that do not match what they'll get if they create a webhook.

I think that's the correct behaviour. On every other page of our documentation, the rule for the 'API version' dropdown is: "select the number that matches your client library version". That's a simple, consistent rule, and is helped by the website remembering it for you. If I understand correctly, what you're suggesting is having the webhook page be the one exception where the same dropdown, with the same values, means something completely different. (And orthogonal to what it means on every other page). ISTM that that's at best confusing.

For the few people who have presence webhooks currently, their ones won't match the docs, true, but that's why we're emailing them to explain it, no? Better that than confuse new users by showing them docs for a thing they have no possible way to access.

> Yes, but right now, our version numbers are linked, and until they're not, that's the behaviour people get.

But they're not linked. Client lib version has no correlation with webhook version. You can have any combination.

mattheworiordan:
Sorry @SimonWoolf but I just don't agree.

They are not orthogonal, they are intrinsically linked (until we decide not to link them, which we've not done), and the reason we decided to make the WebHook protocol 1.0 is because it matches our client libraries 1.0.  So how they are orthogonal I don't understand.

Added to which, here is what our docs say:

voila_capture 2017-02-17_06-43-51_pm

You have assumed that the drop down is a client library version, but why?  It's not in the client library section at all, it's in the Reactor section and there is an API version.  We will have two versions of the WebHook API, 0.8 and 1.0 so that seems like a sensible drop down.

Added to which, if you actually navigate around, because WebHooks is not language specific, you are always taken to the most recent version.  So your concern of going from a 0.8 lib to 0.8 WebHook will not happen, you will go from a 0.8 lib to 1.0 lib and unless you then explicitly choose 0.8.  See below:

gif

I vehemently dislike your approach because what you've done is broken our versioning system completely.  Now in our 0.8 docs you've mixed in some 1.0 functionality, and if we follow that policy we'll have to do a linear amount of more work relative to the number of versions we release every time we make a change.  It does not scale, and I don't see the point. People are actively encouraged to use the latest version unless they so choose to use an old version. Added to which, we'll be on 1.0 for all libs other than C# imminently anyway

SimonWoolf:
> They are not orthogonal, they are intrinsically linked

I'm honestly about what you mean by this. You can have a webhook version 0.8 (if you happen to have set up a webhook before last week), but be using a 0.9 client lib. You can have a webhook version 1.0 (if you set one up now), while still using a 0.8 client lib. The client lib version you're using has neither knowledge of nor any way of influencing the webhook version.

> Added to which, if you actually navigate around, because WebHooks is not language specific, you are always taken to the most recent version. So your concern of going from a 0.8 lib to 0.8 WebHook will not happen, you will go from a 0.8 lib to 1.0 lib and unless you then explicitly choose 0.8. See below:

Unfortunately the gif doesn't show what you think it does. The language docs are showing 0.8 not because you've selected 0.8, but because 0.8 is currently the only version we've enabled. Your stored version will still be `latest` or empty -- look at your `preferred_lang_version` cookie.

You can set a version in e.g. the general authentication docs, where we have 0.8 and 0.9 enabled. This sets the version for `javascript`, since currently, the non-language-specific pages just use your language preference for javascript -- see [here](https://github.com/ably/website/blob/master/app/assets/javascripts/modules/documentation/helpers.js.coffee#L39). Try changing the version in the auth docs and examining the `preferred_lang_version` cookie to see what changed.

Having done this, you'll see that this changes the version for all the non-language-specific pages, including webhooks.

> You have assumed that the drop down is a client library version, but why? It's not in the client library section at all, it's in the Reactor section and there is an API version. We will have two versions of the WebHook API, 0.8 and 1.0 so that seems like a sensible drop down.

Well, mostly because the current website functionality strongly implies that it is, as above.

If we want to make it a webhook version, we should at the very least have the website store it for that page separately, in a `webhook` attr in the cookie val.

Even if that's done, I still think this is needlessly confusing. We have the same dropdown on every page called 'API version'. In your vision, the API that that refers to depends on what page you're on -- sometimes the client library version (which we know implicitly uses the corresponding protocol version behind the scenes, but customers have no direct insight into that), sometimes the webhook structure version, etc. If having it be the webhook version, can we at least change it to 'Webhook version' on that page, then?

mattheworiordan:
> Try changing the version in the auth docs and examining the preferred_lang_version cookie to see what changed.

Well I didn't check that, but if that is the case, then the website documentation selector is wrong.  When the page is not language specific, then it should not be tied to a language preference.  Assuming that JS is the same as none is wrong. I don't think that should have any impact on how we structure our docs because we can fix that.

> If we want to make it a webhook version, we should at the very least have the website store it for that page separately, in a webhook attr in the cookie val.

Nope, I think we should simply have a preference for `none` language, and use that for all docs outside of this.  For example, if we wanted in a future release to change the payloads for Reactor, with your strategy this would never be possible.

Instead, we'd release 2.0 along with everything else we release as 2.0, or perhaps even 1.1 if we don't jump to a major new version, and we could expose the protocol version in the rules that are set up. This is going to happen and it will be needed in the future, so I don't see why we'd want to go with a design that assumes this won't happen.

Documentation is versioned. Like SEMVER, if we find a bug in the docs that is serious, we'll fix it, but we will not continue to go back to old docs to make them easier to use when we should instead be aiming to keep the latest docs current and get people using those.

> can we at least change it to 'Webhook version' on that page, then?

No, because then we'll have Queue version, and Firehose version, and Lambda version, Adapter version etc.  If you are concerned about this, we should simply add an `API version` into all rules, even if it's read-only for now so that people understand our API version of the docs applies to rules as well.

SimonWoolf:
I'm still not convinced (I think that having our docs assume everyone with 1.0 webhooks is using 1.0 client libraries and everyone using 0.8 webhooks is using 0.8 client libraries, while it does make our job easier, is optimistic at best), but shrug. I've updated the PR per your preference. 0.8 changes reduced to minimal correctness fixes, plus a small note in `data.presence.action` line that it's changed in newer versions.

mattheworiordan:
> I think that having our docs assume everyone with 1.0 webhooks is using 1.0 client libraries and everyone using 0.8 webhooks is using 0.8 client libraries

I've not made that assumption. If someone is looking at v0.8 webhooks documentation, they can use v1.0 Python libs and potentially v0.8 C# libs.  What you have done is introduce v1.0 client library features into v0.8 documentation and that is what I object to.  So conversely, you're assuming they are using v1.0 in v0.8 docs, whereas I am saying legacy docs are just that, and if someone uses them then they won't be made aware of new features introduced in later versions.  So if they are interested in that, they should use the latest docs.

> I've updated the PR per your preference. 0.8 changes reduced to minimal correctness fixes, plus a small note in data.presence.action line that it's changed in newer versions.

Thanks.

LGTM

PR #285, opened 2017-03-11
[objc/swift] Remove ARTPresenceLast case type

ricardopereira:


mattheworiordan:
LGTM. @tcard you agree?

Issue #242, opened 2017-01-03
ErrorInfo `cause` attribute to be added as part of 1.1 spec


mattheworiordan:
See https://github.com/ably/docs/pull/238/files#r94307627 for why we should introduce this.

paddybyers:
Included in https://github.com/ably/docs/tree/integration-spec-1-1

PR #293, opened 2017-05-02
Exception reporting and error help

mattheworiordan:
@tcard @paddybyers please see my recommendation for how we capture and report unhandled exceptions as a variation to our specification.

Please note that I bundled in the optional `source` attribute for ErrorInfo we've talked about before, and also mentioned that the `href` attribute may be present, but I opted to construct a help URL when logging so that errors generated by the client library too can help help links.


content/client-lib-development-guide/features.textile
```diff
@@ -86,6 +86,15 @@ h3(#restclient). RestClient
...
+** @(RSC20b)@ Exception reporting is enabled by default, but can be disabled using the @logExceptionReportingUrl@ @ClientOption@ documented in "@TO3m@":#TO3m. When enabled, the client library must:
+*** @(RSC20b1)@ At startup log a message with the equivalent of @info@ log level with the message "Ably client library exception@reporting service started. Unhandled failures will be automatically submitted to errors.ably.io to help improve our service. To find out more about this feature, see https://help.ably.io/exceptions". The @info@ log level is preferred as customers can hide this entry by configuring the client library log level to @warning@, yet will, by default, see this notice when using a client library with exception reporting enabled
```
SimonWoolf: typo: `exception@reporting`

> `with the equivalent of @info@ log level`

AFAIK we don't standardise log levels, it's left up to the library implementers, so technically 'eqivalent of `info`/`warning`' isn't well-defined, but tbh I don't think that matters

content/client-lib-development-guide/features.textile
```diff
@@ -86,6 +86,15 @@ h3(#restclient). RestClient
...
+** @(RSC20c)@ Exceptions reported must additionally include the following tags: @ably-lib@ with the value defined in "@RSC7b@":RSC7b, @ably-version@ with the value defined in "@RSC7a@":RSC7a, @appId@ if known from either the token or API key currently being used.
+** @(RSC20d)@ All personally identifiable information, as much as is practicable, must be redacted or stripped completely before being submitted to Ably. Our intent is only to capture necessary information to debug issues in our own code
+** @(RSC20e)@ Failures to log exceptions to the @errors.ably.io@ endpoint must be handled gracefully. This includes for example DNS failures, TCP/HTTP requests rejected, slow requests and internal failure errors. Additionally, as specified in @RSC20b2@, a failure to log an exception is logged with log level @info@ i.e. an exception reporting failure is not consider a client library @error@ or @warning@
+** @(RSC20f)@ Any errors emitted by the library as a result of an internal failure must contain status code @500@, an error code in the range @51000@ to @51999@ and a suitable error message
```
SimonWoolf: > `an error code in the range @51000@ to @51999@`

It's a bit unclear if you're inviting client lib devs to choose their own error code from that range, which is a bit against the error code approach we've taken so far. Should we maybe pick a single code to mean 'unhandled exception' and specify it here?

content/client-lib-development-guide/features.textile
```diff
@@ -86,6 +86,15 @@ h3(#restclient). RestClient
 ** @(RSC19c)@ The library will configure the @Accept@ and @Content-Type@ type headers to reflect whether the client is configured to use a binary or JSON based protocol (see "RSC8":#RSC8). All requests are encoded and decoded into Json or MsgPack as appropriate automatically by the library. Binary @body@ payloads are not supported at this time
 ** @(RSC19d)@ @request@ method returns an @HttpPaginatedResponse@ object that inherits from the @PaginatedResult@ object to provide details on the response plus paging support where applicable. See "HP1":#HP1 for more details
 ** @(RSC19e)@ If the HTTP network request fails for reasons such as a timeout (after the underlying fallback host attempts have failed where applicable, see "RSC15":#RSC15), then the @request@ method should indicate an error in an idiomatic way for the platform
+* @(RSC20)@ Unexpected internal library exception handling:
```
SimonWoolf: Should we specify the consequences of unexpected exceptions? Since they leave the lib in an unknown and possibly inconsistent state, should we maybe specify that it should terminate any active websockets and move the connection and all channels to the `failed` state?

paddybyers:
> callbacks registered by customers

s/customers/applications/

paddybyers:
> reporting service started

I don't think the library has a "reporting service". I'd say ".. reporting enabled".

content/client-lib-development-guide/features.textile
paddybyers: I'm not sure that it's possible to specify the correct action to take in any given situation because it depends on the error and the platform. I think generically the aim should be to ensure that no invalid or inconsistent state can potentially be left after handling the exception; depending on circumstances the remedial action could include failing the transport, failing the connection, rejecting a message, reinitialising the library completely, etc.

content/client-lib-development-guide/features.textile
mattheworiordan: I know. Do you think we should define log levels in a future revision?  I would say so, seems odd we never did that. @paddybyers?

content/client-lib-development-guide/features.textile
mattheworiordan: Ok, so this is not relevant as far as I can tell to REST as it's effectively stateless, hence I added [`RTC11 `](https://github.com/ably/docs/pull/293/commits/51cf0c357b274ce9058188c9d85adc41d73a59c7#diff-11900b395df266bc2bbfc1d11bdfc7a0R296). I am happy to revise `RTC11` though to accommodate this feedback.

content/client-lib-development-guide/features.textile
mattheworiordan: I know, I was reluctant to specify any codes here because a) we may want to have separate codes in future, b) I am waiting on @paddybyers  to respond to https://github.com/ably/ably-common/pull/21. @paddybyers thoughts?

mattheworiordan:
I've addressed some feedback at a44c1a5.

@tcard as you are implementing this already, do you have any feedback?

content/client-lib-development-guide/features.textile
```diff
@@ -974,9 +984,11 @@ h4. Stats
 
 h4. ErrorInfo
 
-* @(TI1)@ Provides a generic Ably @ErrorInfo@ object that contains Ably @code@, @statusCode@ (analogous to HTTP status code) and @message@ attributes
+* @(TI1)@ Provides a generic Ably @ErrorInfo@ object that contains Ably @code@, @statusCode@ (analogous to HTTP status code), @message@ and @source@ attributes
 * @(TI2)@ Errors returned from the Ably server are compatible with the @ErrorInfo@ structure and should result in errors that inherit from @ErrorInfo@
-* @(TE3)@ "Ably-common":https://github.com/ably/ably-common should be included as a submodule so that "consistent error codes":https://github.com/ably/ably-common/blob/master/protocol/errors.json can be used
+* @(TI3)@ "Ably-common":https://github.com/ably/ably-common should be included as a submodule so that "consistent error codes":https://github.com/ably/ably-common/blob/master/protocol/errors.json can be used
```
tcard: Does this mean that the JSON should be parsed and error messages should be taken from there? If so, maybe say that explicitly? Should that take precedence over any message decoded from a ErrorInfo from the server, or not?

tcard: (I'm realizing now that this was already there...)

content/client-lib-development-guide/features.textile
```diff
@@ -1475,7 +1489,9 @@ enum StatsIntervalGranularity:
 
 class ErrorInfo:
   code: Int // TI1
+  href: String // TI4
   message: String // TI1
+  source: ErrorInfo // TI1
```
tcard: `ErrorInfo?`

content/client-lib-development-guide/features.textile
```diff
@@ -1475,7 +1489,9 @@ enum StatsIntervalGranularity:
 
 class ErrorInfo:
   code: Int // TI1
+  href: String // TI4
```
tcard: `String?`

content/client-lib-development-guide/features.textile
```diff
@@ -974,9 +984,11 @@ h4. Stats
 
 h4. ErrorInfo
 
-* @(TI1)@ Provides a generic Ably @ErrorInfo@ object that contains Ably @code@, @statusCode@ (analogous to HTTP status code) and @message@ attributes
+* @(TI1)@ Provides a generic Ably @ErrorInfo@ object that contains Ably @code@, @statusCode@ (analogous to HTTP status code), @message@ and @source@ attributes
```
tcard: How is `source` set?

mattheworiordan: I don't think you were part of the conversation we had a while, and I am unable to find the issue unfortunately, but we have plenty of situations where the error sent to the user does not accurately reflect the underlying reason. So we suggested having a source attribute in errors sent from Ably that can be used to indicate the underlying source reason for the error.  So only Ably, at present, will use this.

content/client-lib-development-guide/features.textile
mattheworiordan: Yup, was there, and I don't think we should add that as a requirement to be honest...

content/client-lib-development-guide/features.textile
mattheworiordan: Yes, will change

content/client-lib-development-guide/features.textile
mattheworiordan: Yes, will change

content/client-lib-development-guide/features.textile
paddybyers: I think we agreed `cause`

content/client-lib-development-guide/features.textile
paddybyers: If the error from the server contains a message, that should be used. It will contain detail that the generic message text doesn't have.

content/client-lib-development-guide/features.textile
mattheworiordan: Ok, sure, happy with `cause`.  That may explain why I could not find the original issue. See https://github.com/ably/docs/issues/242!

SimonWoolf:
merged into master-1-1

PR #294, opened 2017-05-16
Add note about matching Host being required for fallback host requests

mattheworiordan:
Whilst this should be assumed, in the Ruby library there was in fact a bug, fixed in https://github.com/ably/ably-ruby/commit/093ef58b056303e92661f700236cba227099836c. Hopefully by adding this to the spec we can minimise the likelihood of this happening in other libraries mistakenly.


SimonWoolf:
Never seen a connect method before that takes both a full URI and a host before, weird.

this seems unnecessary as that bug looks very specifically tied to the way you've done fallback hosts in ably-ruby, but shrug, doesn't harm anything, so if you think it'd be helpful, sure :+1:

mattheworiordan:
> this seems unnecessary as that bug looks very specifically tied to the way you've done fallback hosts in ably-ruby

I would agree, but given WebSocket libraries are often quite different to standard REST libraries (across all our libs I see very little in common between them in terms of APIs), I do think it's worth adding a note about this personally.  I appreciate it's perhaps adding verbosity unnecessarily in most cases, but may help fix issues like this for some.

Issue #296, opened 2017-06-01
authUrls containing queryString params


SimonWoolf:
cf https://support.ably.io/helpdesk/tickets/1901

We should be much clearer in the docs than we currently are that the authUrl must not contain queryString params, and those should be passed in separately as `authParams`.

We may also want to put some more validation in the client libraries for this. Currently, ably-js [is fairly lenient about this](https://github.com/ably/ably-js/blob/2ee678b3cb22e4e672d08cfbe5b247cd6f956b0e/common/lib/client/auth.js#L300-L306), if you have an authUrl containing some params those'll be used if you don't also specify some authParams. ably-java is a lot stricter; any params in the authUrl are discarded (currently silently -- would be nice if that printed a warning at least)

In future versions of the spec we might also want to standardise that behaviour. Not sure whether we should standardise on leniency (i.e. parse any querystring in the authUrl and either merge with authParams) or strictness. If strictness, that's a breaking change for ably-js, so will need to wait till 2.0.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/EDU-1788) by [Unito](https://www.unito.io)

PR #297, opened 2017-06-01
Clarify that authUrl should not include querystring params

SimonWoolf:
cf https://github.com/ably/docs/issues/296


mattheworiordan:
@SimonWoolf this is fine for now, but seems odd we have that constraint at all.  Why can't query params be used?  I think in Ruby I parse the URI and then reconstruct it.

SimonWoolf:
> Why can't query params be used? I think in Ruby I parse the URI and then reconstruct it.

They could. We do something similar in ably-js. But not all client libs currently do that - in particular ably-java doesn't, which was the source of the original support issue. Hence https://github.com/ably/docs/issues/296 : "In future versions of the spec we might also want to standardise that behaviour. Not sure whether we should standardise on leniency or strictness".

I'd be fine with standardising on leniency, but until we do that, given that doesn't work in all client libs, we should tell people to use the way that's guaranteed to work, which is authParams. See the original support issue linked from #296, which was a customer using an authUrl that worked in ably-js but not in ably-java.

mattheworiordan:
> I'd be fine with standardising on leniency

I wouldn't even say that's leniency, it's correctly. `authUrl` is a URL, yet we don't accept URLs in all libraries.  That's a bug IMHO, and not a lenient interpretation.

> we should tell people to use the way that's guaranteed to work

Agreed for now, but the issue should be updated to say a URL should in fact be allowed, of which a set of query string params may be included.

tcard:
I vote for "leniency" (which I agree is "correctness", really) and mark `authParams` for deprecation in 2.0. Instead, `authBody` can be added for when `authMethod` is POST; or not support POST bodies at all (if you really want them you can use `authCallback`).

mattheworiordan:
> and mark authParams for deprecation in 2.0. Instead, authBody can be added for when authMethod is POST; or not support POST bodies at all (if you really want them you can use authCallback).

Yes, perhaps, we also have `authHeaders`.  Seems odd we're going to all that effort to provide a mechanism to make HTTP requests when in reality, those requests are pretty useless in all envs other than browsers i.e. an immutable set of `authParams` is rarely what is needed, and instead those `authParams` (and other auth*) should be set at the time the request is being made.  So instead of building in all this complexity, `authCallback` is probably the preferred way we want customers to do this.

SimonWoolf:
replaced by https://github.com/ably/docs/pull/297

PR #299, opened 2017-06-12
Add transportParams clientOption to features spec

SimonWoolf:


SimonWoolf:
> Should we not simply make it a String and be done with it?

I figured that the fact that we need to stringify booleans/numbers (and parse them at the other end) is an implementation detail of the fact that we're sending them over querystring parameters. Would be nice not expose that to the user (and make them do `{ heartbeats: "true", heartbeatInterval: "15000" }` etc. rather than `{ heartbeats: true, heartbeatInterval: 15000 }` ), no?

(happens implicitly in ably-js as everything's just coerced to a string when the querystring's constructed)

tcard:
One way or the other, authParams should do the same.

I tend to agree that specifying a "stringifiable" type is worth the hassle, as people will pass us things that aren't strings in dynamic languages at least no matter what we document, so we rather handle that nicely than crashing or worse. Maybe add a note that that only applies to dynamic languages/if idiomatic, otherwise Stringifiable is just String?

mattheworiordan:
> Maybe add a note that that only applies to dynamic languages/if idiomatic, otherwise Stringifiable is just String?

👍  I like that suggestion

paddybyers:
I think it depends on the IDL. For me it would be more natural to declare the method as accepting a `String`, and the rules that determine which types can be coerced to a `String` are the rules for the language binding to the IDL. You would only introduce a new type/interface in the API itself if the IDL rules didn't do what you want.

SimonWoolf:
> For me it would be more natural to declare the method as accepting a String, and the rules that determine which types can be coerced to a String are the rules for the language binding to the IDL.

I don't think I agree. With most of the apis we have that take strings, we do want to be strict about it: we don't want people specifying number or boolean `clientId`s and coercing those, because while it'd be fine technically, that wouldn't make conceptual sense -- a `clientId` is an inherently stringy thing. Most `transportParams`, by contrast, are conceptually numbers or booleans, it's only an implementation detail on our part (use of a querystring) that means we have to stringify and parse them.

> Maybe add a note that that only applies to dynamic languages/if idiomatic, otherwise Stringifiable is just String?

Sure

SimonWoolf:
Addressing Toni's suggestion (also [this](https://ably-real-time.slack.com/archives/C030APSH3/p1497348141547631)): 658be16

mattheworiordan:
Nice, LGTM

PR #301, opened 2017-06-13
clarify that authUrl qs params should be preserved

SimonWoolf:
Replaces https://github.com/ably/docs/pull/297


mattheworiordan:
LGTM.

PR #304, opened 2017-08-07
iOS is now thread-safe.

tcard:


content/realtime/usage.textile
```diff
@@ -150,80 +150,7 @@ blang[objc,swift].
...
-          } else {
-              self.ablyQueue.sync(execute: operation)
-          }
-      }
-  }
-  ```
-
-  We're working on improving this situation to be more developer-friendly and less error-prone.
-
+  **Note: The iOS client libraries are thread-safe.** Check "the README":https://github.com/ably/ably-ios#thread-safety for details.
```
mattheworiordan: This is good. Given we have older 1.* versions that are not thread-safe, do you think it would make sense to say it was thread-safe since version `1.0.8`?

content/realtime/versions/v0.8/usage.textile
```diff
@@ -129,79 +129,7 @@ blang[objc,swift].
...
-              operation()
-          } else {
-              self.ablyQueue.sync(execute: operation)
-          }
-      }
-  }
-  ```
-
-  We're working on improving this situation to be more developer-friendly and less error-prone.
+  **Note: The iOS client libraries are thread-safe.** Check "the README":https://github.com/ably/ably-ios#thread-safety for details.
```
mattheworiordan: Why did you change this?  `v0.8` isn't thread-safe right?

tcard:
@mattheworiordan PTAL

mattheworiordan:
Thanks @tcard

PR #305, opened 2017-08-09
Fix Heroku builds

mattheworiordan:
Heroku builds are failing, this PR aims to fix that issue, however the problem appears to be that the [Nanoc build pack fork](https://github.com/ably-forks/heroku-buildpack-nanoc) we have does not install the correct version of Ruby


mattheworiordan:
@tcard @SimonWoolf @paddybyers the docs PRs are now working again, although I fear they're a bit brittle because of the Nanoc build pack we forked.  To get this working I had to modify the [existing build pack](https://github.com/ably-forks/heroku-buildpack-nanoc/compare/f1e1397...master) and add a new Ruby build pack to Heroku.  Additionally, [Travis will now complain if people introduce duplicate spec IDs in the client lib spec](65989e2).

I will merge this now but wanted to keep you guys in the loop.

PR #320, opened 2017-10-13
Proposal for minimal transient publish spec

SimonWoolf:
This is a proposal for doing transient publishes in a way that doesn't break API backward-compatibility, so it can be done in a minor version spec change (i.e. 1.1).

Other related changes (a properly-designed API for channel attach flags, batch publishing, etc.) can wait for 2.0.

Summary:

- When a user publishes to a channel in the `initialized` or `detached` state, the client lib does not initiate an implicit attach, it just does the publish
- If a client lib receives an ATTACHED with the TRANSIENT flag set, it ignores it and stays in whatever state it was already in
- If a client lib receives an DETACHED on a channel in the `initialized` state, it ignores it and stays in the `initialized` state; if in any other state, behaviour remains as it is now

The idea is that being that the client lib shouldn't actually care whether or not it's transiently publish-only-attached, since there's no case where it will do anything differently in that state vs being initialized/detached. If it wants to do some more publishes it will just do them whether or not it's currently transiently attached, and if it wants to subscribe it needs to send an ATTACHED either way. (There's an argument that in this case realtime shouldn't bother sending them, but, shrug)

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -453,10 +453,10 @@ h3(#realtime-channel). Channel
 *** @(RTL6i2)@ When an array of @Message@ objects is provided, a single @ProtocolMessage@ is used to publish all @Message@ objects in the array. However, a yet to be implemented feature should limit the total number of messages bundled in a single ProtocolMessage based on the default max message size
 *** @(RTL6i3)@ Allows @name@ and or @data@ to be @null@.  If any of the values are @null@, then key is not sent to Ably i.e. a payload with a @null@ value for @data@ would be sent as follows @{ "name": "click" }@
 ** @(RTL6c)@ Connection and channel state conditions:
-*** @(RTL6c1)@ If the connection is @CONNECTED@ and the channel is @ATTACHED@ then the messages are published immediately
-*** @(RTL6c2)@ If the connection is @INITIALIZED@, @CONNECTING@ or @DISCONNECTED@ or the channel is @INITIALIZED@, @ATTACHING@ or @ATTACHED@, and @ClientOptions#queueMessages@ has not been explicitly set to false, then the message will be queued and delivered as soon as the connection state enters the @CONNECTED@ state and the channel is @ATTACHED@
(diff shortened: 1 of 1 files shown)


SimonWoolf:
> (There's an argument that in this case realtime shouldn't bother sending them, but, shrug)

I raised this point at the standup. @paddybyers 's response was that that would be ok as long as we're sure we can do it in a way that can be extended to support a full channel mode API.

I don't see any reason it couldn't be. Under that proposal, a transient attach (and subsequent detach) isn't something a client library would ever request explicitly, or even know about -- it's just a serverside optimisation of what to the client library would be indistinguishable from just always being able to publish without attaching.

Paddy also raised the possibility that a client lib that's aware of channel modes and is attached to a channel without publish mode trying to publish. There's three ways we could deal with that: first, we could to reject it; second, we would just permanently automatically upgrade the mode to publish (as long as we have the permissions to); third, we could transiently add the publish permission, then remove it after a timeout. The second and third ones look the same from the client's point of view, so if we want to not reject it, we could maybe do the second for now in the knowledge that we can implement the third later, should we have a customer that has a lot of channels that could benefit from that optimisation (i.e. who want to subscribe all the time and only very occasionally publish from the same connection). None of the three seem unreasonable to me.

If we all agree, then I'll drop this PR in favour of a simpler one that drops the bits about ignoring ATTACHED/DETACHEDs, and do a PR for realtime that stops sending ATTACHEDs and DETACHEDs for transient channel state changes. WDYT @mattheworiordan @paddybyers ?

paddybyers:
> If we all agree, then I'll drop this PR in favour of a simpler one that drops the bits about ignoring ATTACHED/DETACHEDs, and do a PR for realtime that stops sending ATTACHEDs and DETACHEDs for transient channel state changes. WDYT?

On balance, I think this is right. It's poor that for a transient publish we have 4 messages going in the server->client direction for each message in the client->server direction.

SimonWoolf:
Replaced by https://github.com/ably/docs/pull/323

PR #323, opened 2017-10-23
Proposal 2 for minimal transient publish spec

SimonWoolf:
Replaces https://github.com/ably/docs/pull/320.

This is a proposal for doing transient publishes in a way that doesn't break API backward-compatibility, so it can be done in a minor, or even patch spec change. Other related changes (a properly-designed API for channel attach flags, batch publishing, etc.) can wait for 2.0.

Basically, the only thing that's changed is that when a user publishes to a channel in the initialized or detached state, the client lib does not initiate an implicit attach, it just does the publish. External API is identical. People who before were relying on clientside implicit attaches will now rely on serverside transient attaches, but the user experience is unchanged.

Relies on realtime PR https://github.com/ably/realtime/pull/1338 .

The idea is that being that the client lib shouldn't actually care whether or not it's transiently publish-only-attached, since there's no case where it will do anything differently in that state vs being initialized/detached. If it wants to do some more publishes it will just do them whether or not it's currently transiently attached, and if it wants to subscribe it needs to send an ATTACHED either way. So we just don't tell it. See https://github.com/ably/docs/pull/320 for more discussion.

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -453,10 +453,10 @@ h3(#realtime-channel). Channel
 *** @(RTL6i2)@ When an array of @Message@ objects is provided, a single @ProtocolMessage@ is used to publish all @Message@ objects in the array. However, a yet to be implemented feature should limit the total number of messages bundled in a single ProtocolMessage based on the default max message size
 *** @(RTL6i3)@ Allows @name@ and or @data@ to be @null@.  If any of the values are @null@, then key is not sent to Ably i.e. a payload with a @null@ value for @data@ would be sent as follows @{ "name": "click" }@
 ** @(RTL6c)@ Connection and channel state conditions:
-*** @(RTL6c1)@ If the connection is @CONNECTED@ and the channel is @ATTACHED@ then the messages are published immediately
-*** @(RTL6c2)@ If the connection is @INITIALIZED@, @CONNECTING@ or @DISCONNECTED@ or the channel is @INITIALIZED@, @ATTACHING@ or @ATTACHED@, and @ClientOptions#queueMessages@ has not been explicitly set to false, then the message will be queued and delivered as soon as the connection state enters the @CONNECTED@ state and the channel is @ATTACHED@
-*** @(RTL6c4)@ If the connection is @SUSPENDED@, @CLOSING@, @CLOSED@, or @FAILED@, or the channel is @DETACHING@, @DETACHED@, @SUSPENDED@ or @FAILED@, the operation will result in an error
-*** @(RTL6c3)@ Implicitly attaches the @Channel@ if the channel is in the @INITIALIZED@ state. However, if the channel is in enters to the @DETACHED@ or @FAILED@ state before the operation succeeds, it will result in an error
+*** @(RTL6c1)@ If the connection is @CONNECTED@ and the channel is @INITIALIZED@, @ATTACHED@, @DETACHED@, @ATTACHING@, or @DETACHING@ then the messages are published immediately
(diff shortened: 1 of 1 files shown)


paddybyers:
So if the channel is not yet attached, it doesn't trigger an attach, doesn't wait for attach, just publishes.

If the channel is attaching, it waits for that to complete before publishing. Really if it's in attaching state, triggered by an explicit `attach()`, then that's understandable. If it's attaching implicitly as a result of a `subscribe()`, then there's really no reason why that ought to delay the publish, because the publish and subscribe operations ought to be independent.

This is an area that will be clearer once the channel mode is explicit.

SimonWoolf:
> Really if it's in attaching state, triggered by an explicit attach(), then that's understandable. If it's attaching implicitly as a result of a subscribe(), then there's really no reason why that ought to delay the publish

I'm not a fan of requiring libraries to have to keep state on whether an `attaching` state is from an explicit or implicit attach and behaving differently in each case, that feels very messy.

Having added https://github.com/ably/realtime/commit/f5ccdbb22498f18196bfe2e9265b8a5814c87938, I'd be ok with allowing publishes during attaching, as realtime will now just wait until the attach is finished. But I do think we should either always allow it during `attaching` or never, it shouldn't depend on what triggered the attach.

Not sure about publishes during detaching. Looking at the code, I *think* it should be fine because the attachment will still be around (it's only disposed on clearChannel, which is synchronous with removing it from this.channels), and `Channel.message` doesn't actually care whether or not `isAttached` is true, but I'm not certain. @paddybyers can you confirm?

mattheworiordan:
@SimonWoolf this seems good to me.  However, I am not sure why publishing now has any bearing on the channel state at all, other than for perhaps the FAILED state. Under what circumstances would someone want to wait?

SimonWoolf:
> However, I am not sure why publishing now has any bearing on the channel state at all

@mattheworiordan last few comments on this PR (as well as https://github.com/ably/realtime/pull/1338#issuecomment-338689359 ) have been about exactly that (whether we should allow publishes during ATTACHING/DETACHING). Welcome any thoughts you have

paddybyers:
> Not sure about publishes during detaching. Looking at the code, I think it should be fine because the attachment will still be around (it's only disposed on clearChannel, which is synchronous with removing it from this.channels), and Channel.message doesn't actually care whether or not isAttached is true, but I'm not certain. @paddybyers can you confirm?

I'm not sure what you're saying here. `DETACHING` is a state of the client, not realtime, and extends until the moment that the client receives the `DETACHED`. It's therefore possible for the realtime resources associated with the attachment, especially the `ref` on the client, to be gone before the client sees the `DETACHED`.

All of that is beside the point though, I think. The `Channel` attachment, in realtime,  is maintaining a reference to the client, and its `mode` varies as transient publishes happen. If the client has requested a `DETACH`, but then requests a transient attach implicitly via `MESSAGE`, realtime will reestablish the attachment if it needs to. This is different from handling concurrent attach requests because of the possibility of conflict in mode, but with concurrent detach+attach I don't think that conflict arises.

So I don't think the client needs to wait before requesting a implicit-attach publish if it is `DETACHING`. Realtime might need to wait, and can do so if necessary.

mattheworiordan:
> last few comments on this PR ... whether we should allow publishes during ATTACHING/DETACHING ... Welcome any thoughts...

Sorry I was not clear, I knew that was where you were headed.  I was effectively confirming that I cannot think of any practical reason why we would disallow publishes unless the channel is `FAILED`.  I did wonder if there may be some race condition for clients that rely on `echoMessages: true` when a channel becomes suspended, but either it will succeed and provide continuity or it won't and the developer has to deal with that anyway.

SimonWoolf:
Tweaked to allow publishing in ATTACHING or DETACHING.

PR #324, opened 2017-10-28
complete tutorial

ammezie:
The docs for React Realtime Commenting Tutorial. https://github.com/ably/tutorials/pull/40 is PR for the tutorial repo.


mattheworiordan:
Thanks @ammezie. Sorry for the delay, we will make time to review soon.

mattheworiordan:
@ammezie sorry for the continued delay. We've been a bit busy. We'll provide feedback very soon.

content/tutorials/reactjs-realtime-commenting.textile
```diff
@@ -0,0 +1,344 @@
...
+
+h2. Step 12 - Create App Component
+
+The App component will server as the parent component. Open @components/App.js@ and replace it with the code below:
+
+```[javascript]
+import React, { Component } from 'react';
+import CommentBox from './CommentBox';
+import Comments from './Comments';
+import Ably from '../ably';
```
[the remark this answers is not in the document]
mattheworiordan: Oh I see it's used in a later step. So firstly this is a problem because it does result in two connections to Ably, but also because it's just not right i.e. there should be a global Ably object shared between the components, or perhaps a separate object that handles publishing & receiving events.

content/tutorials/reactjs-realtime-commenting.textile
```diff
@@ -0,0 +1,344 @@
...
+h2. Step 13 - Subscribing to Comments
+
+We have seen how we can publish newly added comments. Now, we need to subscribe to those comments being published. We’ll do this within React lifecycle @componentDidMount()@ hook. Still within @components/App.js@, add the code below just before the @render()@ function:
+
+```[javascript]
+componentDidMount() {
+  // Subscribe to comments
+  const channel = Ably.channels.get('comments');
+
+  channel.subscribe('add_comment', (comment) => {
```
mattheworiordan: Why is this logic in the App, when the publish logic is in the CommentBox?  Surely the CommentBox should handle publishing & subscribing?

content/tutorials/reactjs-realtime-commenting.textile
```diff
@@ -0,0 +1,344 @@
...
+    this.setState((prevState) => {
+      return {
+        comments: prevState.comments.concat({
```
mattheworiordan: Should the comments not be in reverse order i.e. newest at the top?

content/tutorials/reactjs-realtime-commenting.textile
```diff
@@ -0,0 +1,344 @@
...
+}
+```
+
+The @componentDidMount()@ hook runs after the component output has been rendered to the DOM. It is a good place to subscribe to comments. We subscribe to the @comments@ channel and listen for the @add_comment@ event. We then update the @comments@ state with the comment published from the channel we are subscribed to.
+
+<a href="/images/tutorials/reactjs-realtime-commenting-final-output.png" target="_blank">
+  <img src="/images/tutorials/reactjs-realtime-commenting-final-output.png" style="width: 100%" alt="Final output screenshot">
+</a>
+
+That’s it. We now have a realtime commenting system. To test it out, start the app:
```
mattheworiordan: Should we not use history to prepopulate the comments?

For example, if you add the following to `componentDidMount`:

```js
channel.attach();
    channel.once('attached', () => {
      channel.history((err, page) => {
        // page.items contains list of up to 100 previous items
      });
    });
```

See https://www.ably.io/tutorials/history#lang-javascript, you could create a similar step to Step 2, and then even link to the history tutorial (with `./history` as the path) if the user wants to understand how to use history.

content/tutorials/reactjs-realtime-commenting.textile
```diff
@@ -0,0 +1,344 @@
...
+Checkout the tutorial branch:
+
+```[sh]
+git checkout reactjs-realtime-commenting
+```
+
+Install NPM dependencies. Be sure to switch into project's directory and then run this command in your terminal:
+
+```[sh]
+npm install
```
mattheworiordan: This won't work if Ably is not included in the `package.json` file

content/tutorials/reactjs-realtime-commenting.textile
```diff
@@ -0,0 +1,344 @@
...
+```
+
+And then run the app locally by adding your "Ably API key":https://support.ably.io/support/solutions/articles/3000030502-setting-up-and-managing-api-keys to @src/ably.js@ and run
+
+```[sh]
+npm start
+```
+
+to start the app.
```
mattheworiordan: start the web server and open the browser.

content/tutorials/reactjs-realtime-commenting.textile
```diff
@@ -0,0 +1,344 @@
...
+
+h2. Next steps
+
+1. If you would like to find out more about how channels, publishing and subscribing works, see the "Realtime channels & messages documentation":https://www.ably.io/documentation/realtime/channels-messages
```
mattheworiordan: Perhaps add in a link to the history tutorial and history documentation as well, see https://www.ably.io/tutorials/history#lang-javascript

content/tutorials/reactjs-realtime-commenting.textile
```diff
@@ -0,0 +1,344 @@
...
+```[html]
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bulma/0.6.0/css/bulma.min.css">
+```
+
+h2. Step 7 - Install Ably
+
+To start using Ably you first need to install the NPM module. The NPM module can be installed as follows:
+
+```[sh]
+npm install ably
```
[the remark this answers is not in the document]
ammezie: The command will do that.

content/tutorials/reactjs-realtime-commenting.textile
```diff
@@ -0,0 +1,344 @@
...
+```
+
+The client library must be instanced with the API key you copied in Step 1. We’ll create a dedicated file for this so we can easily use it in multiple places. Within @src@ folder, create a new file named @ably.js@ and add the code below into it:
```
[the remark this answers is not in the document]
ammezie: Since the app is set up to use WebPack, it makes to use it this way to everything can be bundled together. This will reduce the number of HTTP requests made to fetch the library individually.

content/tutorials/reactjs-realtime-commenting.textile
```diff
@@ -0,0 +1,344 @@
...
+
+```[javascript]
+import { Realtime } from 'ably/browser/static/ably-commonjs.js';
+
+export default new Realtime('INSERT-YOUR-API-KEY-HERE');
```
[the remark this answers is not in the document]
ammezie: `import` can be used to import anything (functions, variables, objects, arrays) not just classes.

content/tutorials/reactjs-realtime-commenting.textile
```diff
@@ -0,0 +1,344 @@
...
+        </form>
+      </div>
+    );
+  }
+}
+
+export default CommentBox;
+```
+
+This simply renders a comment form as the image below. Once the form is submitted, we trigger an @onSubmit@ event which in turn call the @addComment()@ function (which we’ll create shortly).
```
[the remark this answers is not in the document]
ammezie: I already mentioned it renders a comment form.

content/tutorials/reactjs-realtime-commenting.textile
ammezie: That is already added to `package.json` when we ran `npm install ably`

content/tutorials/reactjs-realtime-commenting.textile
ammezie: Wasn't aware of history before.

content/tutorials/reactjs-realtime-commenting.textile
ammezie: Does it matter?

content/tutorials/reactjs-realtime-commenting.textile
ammezie: Because I want to update the `state` with the newly added comment. And `state` is only available at the parent/root component, hence the need to push it up to App component.

content/tutorials/reactjs-realtime-commenting.textile
ammezie: Yeah, using history can be one way. I think it might not really suit the implementation of this tutorial. I'm adding the newly created comment to the `state`, that way, we can do whatsoever we like with the data in `state` like persisting to a database.

content/tutorials/reactjs-realtime-commenting.textile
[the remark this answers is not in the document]
mattheworiordan: Nope it doesn't.  Just tried that again.  Either way, you should be stating that `ably` should be added to the package.json file because it means you're explicitly stating it's a dependency.

content/tutorials/reactjs-realtime-commenting.textile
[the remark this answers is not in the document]
mattheworiordan: OK

content/tutorials/reactjs-realtime-commenting.textile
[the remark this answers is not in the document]
mattheworiordan: Sure, but it's wrong in this context because every import creates a new Realtime object, which in turn creates a new connection.  Ably connections are multiplexed and you should not have, in almost all cases, multiple concurrent connections from a single browser to the same app.

content/tutorials/reactjs-realtime-commenting.textile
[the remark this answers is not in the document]
mattheworiordan: Well sort of.  The problem is at this point it does not because its not invoked or reference anywhere.  I don't think this explains to a noob why you're doing this first and how it renders.

content/tutorials/reactjs-realtime-commenting.textile
mattheworiordan: But that's just wrong I am afraid as your logic is unnecessarily split simply because you want to work around state encapsulation in the component. You could simply expose a new function that allows a message to be added, and call that from the child, see https://stackoverflow.com/questions/35537229/how-to-update-parents-state-in-react.

content/tutorials/reactjs-realtime-commenting.textile
mattheworiordan: Sure, because then you can't see new comment updates as they come in once there are more than a few.

content/tutorials/reactjs-realtime-commenting.textile
mattheworiordan: > I think it might not really suit the implementation of this tutorial.

But the point of the tutorial is to expose Ably functionality, and without a server in this tutorial, this is an obvious way to solve two problems.

content/tutorials/reactjs-realtime-commenting.textile
mattheworiordan: Nope, as above.

content/tutorials/reactjs-realtime-commenting.textile
ammezie: Thanks for pointing this out. This is exactly what I need initially, only to remove it because comments were been added twice due to the code I have in `componentDidMount`. I think I can revert to it since I can now use history.

content/tutorials/reactjs-realtime-commenting.textile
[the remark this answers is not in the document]
ammezie: Guess I assumed people are running on npm 5 and above which automatically adds the package to `package.json`. Anyways, I will add the `--save` flag to it so the package is added to `package.json`.

Issue #287, opened 2017-03-17
Spec: Be more explicit about when to discard connection state?


SimonWoolf:
Currently, RTN15a says "Connection state is only maintained for a brief period, up to two minutes, so if a client is disconnected for a longer period connection state cannot be resumed". We should probably be more explicit about this and formalise under what conditions clients should and should not attempt to resume, possibly with reference to the last known server-sent activity (see RTN23a), based off the `connectionDetails.connectionStateTtl`. cf https://github.com/ably/ably-js/issues/388

mattheworiordan:
How about we make it a pre-requisite that on resume, the timestamp from the last heartbeat / activity from Ably is included?  That way realtime can decide if it wants to log this as a warning or not i.e. whilst the client library may think there is connection state, the realtime system will in most cases know based on the timestamp.  I appreciate there are clock sync issues, but we could address that quite easily too even by insisting the client lib uses the timestamp received in the connection message perhaps?

paddybyers:
Didn't we already say that the client wouldn't bother attempting a resume if it knew that the interval is too long?

When we move to using `timeSerial`, realtime will know without the client having to specify any additional parameter so I think we can defer this until then.

mattheworiordan:
> Didn't we already say that the client wouldn't bother attempting a resume if it knew that the interval is too long?

Yes, but there are conditions I am sure where it fails to connect, reattempts the connection as part of our fallback strategy, then connects say 15s later with a timestamp that has definitely expired.  In that  case, the realtime system no longer needs to log warnings about connection state not being found because it's completely valid.  

> When we move to using timeSerial, realtime will know without the client having to specify any additional parameter so I think we can defer this until then.

But that won't apply to heartbeats, and especially native WS heartbeats

SimonWoolf:
> Yes, but there are conditions I am sure where it fails to connect, reattempts the connection as part of our fallback strategy, then connects say 15s later with a timestamp that has definitely expired. 

True, but I can't imagine that happens very often -- the client lib would have to go through multiple unsuccessful connection cycles, then have the cycle that ends up succeeding start before the two minute mark but only succeed after. Seems a shame to spend another connect param just to avoid a warning in that event.

(We could perhaps decide to have realtime send a connectionStateTtl that's just a couple of seconds less than the time it'll actually keep the connection state for, to allow for connection time etc?)

paddybyers:
> But that won't apply to heartbeats, and especially native WS heartbeats

[...]

PR #331, opened 2017-11-02
Add spec for connection state freshness check

SimonWoolf:
Cf https://github.com/ably/realtime/issues/940#issuecomment-286985784
Prioritised due to analysis in https://github.com/ably/realtime/issues/1289#issuecomment-341152556
Fixes https://github.com/ably/docs/issues/287
Ably-js implementation https://github.com/ably/ably-js/blob/c55b4518d6f096e4bde90afcd04c5ba35d466f3c/common/lib/transport/connectionmanager.js#L702-L712


paddybyers:
> Ably-js implementation https://github.com/ably/ably-js/blob/c55b4518d6f096e4bde90afcd04c5ba35d466f3c/common/lib/transport/connectionmanager.js#L702-L712

But here: https://github.com/ably/ably-js/blob/c55b4518d6f096e4bde90afcd04c5ba35d466f3c/common/lib/transport/connectionmanager.js#L709

You unconditionally go into `suspended` without knowing that the `suspended` time has passed?

SimonWoolf:
> You unconditionally go into suspended without knowing that the suspended time has passed?
Merge state

If I haven't had any evidence of websocket liveness from realtime for longer than connectionStateTtl + maxIdleInterval, then I do know that the suspended time has passed (assuming that realtime hasn't heard from me for the same duration I haven't heard from realtime -- which isn't certain as there can be different latencies in both directions, but it's a decent approximation). True, I haven't had 2 minutes of timers elapse, but the whole point of RTN15g1 is that I can't rely on 2 minutes of timers having elapsed if the device has been sleeping

SimonWoolf:
(And it doesn't make sense to clear the connection state and not have the connecting failstate be `suspended` -- the `suspended` state is for when we've lost connection state)

paddybyers:
> longer than connectionStateTtl + maxIdleInterval

Yes, sorry, I missed the `connectionStateTtl` in there ...

content/client-lib-development-guide/features.textile
```diff
@@ -360,7 +360,11 @@ h3(#realtime-connection). Connection
...
+*** @(RTN15g1)@ This check should be made before each connection attempt. It is generally not sufficient to merely clear the connection state when moving to @SUSPENDED@ state (though that may be done too), since the device may have been sleeping / suspended, in which case it may have been many hours since it was last actually connected, even though, having been in the @CONNECTED@ state when it was put to sleep, it has only moved out of that state very recently (after waking up and noticing it's no longer connected)
+*** @(RTN15g2)@ Anothing consequence of that is that the measure of whether the client been disconnected for too long (for the purpose of this check) cannot just be whether the client left the @CONNECTED@ state more than @connectionStateTtl@ ago. Instead, it should be whether the difference between the current time and the last activity time is greater than the sum of the @connectionStateTtl@ and the @maxIdleInterval@, where the last activity time is the time of the last known actual sign of activity from Ably per "RTN23a":#RTN23a
```
mattheworiordan: @SimonWoolf typo, `Anothing`?

SimonWoolf: Oops, fixed

PR #333, opened 2017-11-02
Apply maxMessageSize limit to protocolmessages, not messages, and firm up the definition

SimonWoolf:
Context: https://github.com/ably/wiki/issues/222#issuecomment-334567652

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -185,10 +185,11 @@ h3(#rest-channel). Channel
 * @(RSL1)@ @Channel#publish@ function:
 ** @(RSL1a)@ Expects either an array of @Message@ objects or a @name@ string and @data@ payload
 ** @(RSL1b)@ When @name@ and @data@ is provided, a single message is published to Ably
-** @(RSL1c)@ When an array of @Message@ objects is provided, a single request is made to Ably. When publishing multiple messages, this approach is more efficient. However, a yet to be implemented feature should limit the total number of messages bundled in a single POST based on the default max request size, and would reject the publish and indicate an error if any single message exceeds that limit
+** @(RSL1c)@ When an array of @Message@ objects is provided, a single request is made to Ably
 ** @(RSL1d)@ Indicates an error if the message was not successfully published to Ably
 ** @(RSL1e)@ Allows @name@ and or @data@ to be @null@.  If any of the values are @null@, then key is not sent to Ably i.e. a payload with a @null@ value for @data@ would be sent as follows @{ "name": "click" }@
 ** @(RSL1f)@ Unidentified clients using "Basic Auth":https://en.wikipedia.org/wiki/Basic_access_authentication (i.e. any @clientId@ is permitted as no @clientId@ specified):
+** @(RSL1g)@ If the total size of the message or (if publishing an array) messages, calculated per "TO3l7":#TO3l7, exceeds the @maxMessageSize@, then the client library should reject the publish and indicate an error
 *** @(RSL1f1)@ When a @Message@ with a @clientId@ value is published, Ably will accept and publish that message with the provided @clientId@. A test should assert via the history API that the @clientId@ of the published @Message@ is populated
 ** @(RSL1g)@ Identified clients with a @clientId@ (as a result of either an explicitly configured @clientId@ in @ClientOptions@, or implicitly through Token Auth):
 *** @(RSL1g1)@ When publishing a @Message@ with the @clientId@ attribute set to @null@:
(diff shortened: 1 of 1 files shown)


content/client-lib-development-guide/features.textile
```diff
@@ -1168,6 +1171,8 @@ h4. ClientOptions
...
 *** @(TO3l6)@ @httpMaxRetryDuration@ integer - default 15,000 (15s). Max elapsed time in which fallback host retries for HTTP requests will be attempted i.e. if the first default host attempt takes 7s, and then the subsequent fallback retry attempt takes 10s, no further fallback host attempts will be made as the total elapsed time of 17s exceeds the default 15s limit
+*** @(TO3l7)@ @maxMessageSize@ integer - default 65536 (64kB). The maximum size of messages that can be published in one go. That is, the size of the @ProtocolMessage.messages@ or @ProtocolMessage.presence@ array for a realtime publish or presence action, or the message or array of messages being published for a REST publish. The size is defined as the sum over all messages being published of the message @name@, @data@, @clientId@, and @extras@. The size of the @data@ should be measured after being stringified if an @Object@ or @Array@, but before the entire frame is serialized to JSON or msgpack (so e.g. if a binary payload using the JSON encoding, before it is encoded as base64 per @RSL4d1@). For realtime publishes, the default will be overridden by the @maxMessageSize@ in the @connectionDetails@, see "CD2c":#CD2c
```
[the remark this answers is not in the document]
SimonWoolf: > How do you measure `Extras`?

Good point, thanks, I'll clarify that. (Answer: as the length of its json representation, as has to match what realtime does, see https://github.com/ably/realtime/blob/master/common/lib/types/protocolmessage.js#L111. Does not depend on `useBinaryProtocol`.)

> maybe worth mentioning that null has a size of zero, and that binary refers to its real size, not that of its base64 representation for JSON (if that's indeed the case).

Good point re `null`. Re binaries, I mentioned that in the second-last sentence, but I'll try to rephrase to make that clearer.

content/client-lib-development-guide/features.textile
```diff
@@ -184,10 +184,11 @@ h3(#rest-channel). Channel
...
-** @(RSL1c)@ When an array of @Message@ objects is provided, a single request is made to Ably. When publishing multiple messages, this approach is more efficient. However, a yet to be implemented feature should limit the total number of messages bundled in a single POST based on the default max request size, and would reject the publish and indicate an error if any single message exceeds that limit
+** @(RSL1c)@ When an array of @Message@ objects is provided, a single request is made to Ably
 ** @(RSL1d)@ Indicates an error if the message was not successfully published to Ably
 ** @(RSL1e)@ Allows @name@ and or @data@ to be @null@.  If any of the values are @null@, then key is not sent to Ably i.e. a payload with a @null@ value for @data@ would be sent as follows @{ "name": "click" }@
 ** @(RSL1f)@ Unidentified clients using "Basic Auth":https://en.wikipedia.org/wiki/Basic_access_authentication (i.e. any @clientId@ is permitted as no @clientId@ specified):
+** @(RSL1g)@ If the total size of the message or (if publishing an array) messages, calculated per "TO3l7":#TO3l7, exceeds the @maxMessageSize@, the the client library should reject the publish and indicate an error
```
mattheworiordan: s/the the/then the/g

content/client-lib-development-guide/features.textile
```diff
@@ -446,18 +447,20 @@ h3(#realtime-channel). Channel
...
 *** @(RTL6c4)@ If the connection is @SUSPENDED@, @CLOSING@, @CLOSED@, or @FAILED@, or the channel is @SUSPENDED@ or @FAILED@, the operation will result in an error
 *** @(RTL6c5)@ A publish should not trigger an implicit attach (in contrast to earlier version of this spec)
-** @(RTL6d)@ Messages are delivered using a single @ProtocolMessage@ where possible by bundling in all messages for that channel into the @ProtocolMessage#messages@ array. However, a yet to be implemented feature should limit the total number of messages bundled per @ProtocolMessage@ based on the default max message size, and would reject the publish and indicate an error if any single message exceeds that limit
+** @(RTL6d)@ Messages for a single channel that have been queued may be sent in a single @ProtocolMessage@ by bundling them into the @ProtocolMessage#messages@ array, subject to the following constraints:
```
mattheworiordan: Why bundle at all then?  What do we really gain?

content/client-lib-development-guide/features.textile
```diff
@@ -1168,6 +1171,8 @@ h4. ClientOptions
...
+*** @(TO3l7)@ @maxMessageSize@ integer - default 65536 (64kB). The maximum size of messages that can be published in one go. That is, the size of the @ProtocolMessage.messages@ or @ProtocolMessage.presence@ array for a realtime publish or presence action, or the message or array of messages being published for a REST publish. The size is defined as the sum over all messages being published of the message @name@, @data@, @clientId@, and @extras@. The size of the @data@ should be measured after being stringified if an @Object@ or @Array@, but before the entire frame is serialized to JSON or msgpack (so e.g. if a binary payload using the JSON encoding, before it is encoded as base64 per @RSL4d1@). For realtime publishes, the default will be overridden by the @maxMessageSize@ in the @connectionDetails@, see "CD2c":#CD2c
+*** @(TO3l8)@ @maxFrameSize@ integer - default 524288 (512kB). The maximum size of a single POST body or WebSocket frame. This is mostly only relevant for `Rest#request` (e.g. for batch publishes), since publishes will hit the @maxMessageSize@ limit before this
```
mattheworiordan: I don't recall agreeing we would have a max frame size at 500kb?  @paddybyers are we going to support batch publishes of up to 500kb then by default?  Doesn't this just provide a way for people then to work around a sensible max size limit we want per request?

SimonWoolf: @mattheworiordan Realtime can limit endpoints to whatever we like. Maybe batch publish 256kB, push publish 16kB, whatever. Having a client lib limit for unknown request types with `Rest#request` doesn't mean the realtime has to allow up to the same value for any and all experimental endpoints we introduce, it's just a backstop to stop the client from sending something several megabytes big to an experimental endpoint, using up our bandwidth for something we know we're going to reject.

So 512kB was picked as a limit that we're very unlikely to want to introduce an experimental endpoint that allows requests larger than that.

I was originally just going to remove the `maxFrameSize` entirely, but then figured it was probably a good idea to have \*something\* to stop the client lib sending multi-megabyte bodies to experimental endpoints (while still giving us flexibility to introduce an endpoint that takes requests larger than the single-message-publish-limit should we want to). Makes sense?

content/client-lib-development-guide/features.textile
SimonWoolf: We could disallow in-protocolmessage bundling entirely, sure.

We do gain something from bundling, being efficiency in the case of someone sending a very high rate of tiny messages (say, 50 per second, our published max rate), which can build up quickly during queueing. If each is only a few bytes we might easily be able to send them in one protocolmessage (or at least a few); sent separately (but all at once when connected), they'll be rejected for rate limit breaches.

But if we don't think that's a sufficient reason, then sure, we should get rid of bundling altogether.

content/client-lib-development-guide/features.textile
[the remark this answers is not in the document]
SimonWoolf: f8f32b8e8754a08b20830906ebabd3337b794470

content/client-lib-development-guide/features.textile
mattheworiordan: @SimonWoolf well that is a good point, especially when we have considered how Ably could be used an event streaming tool for offline clients that later come online and stream all events.  Given we have the functionality already there and it could be leveraged in the future if we want to offer offline event publishing, let's leave

content/client-lib-development-guide/features.textile
mattheworiordan: @SimonWoolf Ok, fair enough.

PR #342, opened 2017-11-21
Clarify spec on client auth failures

SimonWoolf:
On implementing https://github.com/ably/ably-js/pull/434, the current spec implies that the causal errorinfo should have its code changed from whatever it was originally to 80019. Given the new `errorinfo.cause` mechanism in 1.1, clarified this to say that the original errorinfo should be preserved intact, and set as the `cause` of a new ErrorInfo 80019/401 (or 80019/403 for when going to `failed`)


paddybyers:
LGTM

Issue #312, opened 2017-09-13
Should `useTokenAuth: false` force basic auth?


SimonWoolf:
Noticed an inconsistency between ably-java and ably-js just now.

RSA4 requires that: `Token Auth is used if useTokenAuth is set to true, or if useTokenAuth is unspecified and any one of the following conditions are met: a clientId is specified; authUrl or authCallback is provided; an explicit token or TokenDetails is provided `

Ably-java takes advantage of the fact that the condition is an `if`, not an `if and only if`, and continues to use token auth if `useTokenAuth` is `false` as long as one of the other conditions is true. Wheras ably-js [disables token auth if `useTokenAuth` is `false`](https://github.com/ably/ably-js/blob/master/common/lib/client/auth.js#L57). (The distinction is mostly only relevant if the user specifies a `clientId`, which would otherwise force token auth). We should standardise on one behaviour and clarify the spec accordingly.

FWIW I'd prefer we standardise on the ably-js behaviour -- if someone specifies `useTokenAuth: false`, they presumably have a good reason for wanting basic auth, so why should we stop them? (Not to mention it would have been useful for https://support.ably.io/helpdesk/tickets/2248 , we could just have told the customer to specify `useTokenAuth: false` to solve the problem)

cc @mattheworiordan @paddybyers

paddybyers:
One issue here is that in java you can't have `useTokenAuth == undefined`. So there is a default value in `ClientOptions` which (I assume) happens to be false. There is no way to indicate that the user chose a non-default value. If we want the behaviour to be that specifying a `clientId` by default triggers token auth, then we have to keep the java behaviour, at least for java.

SimonWoolf:
So e.g. see https://support.ably.io/helpdesk/tickets/2526 for a usecase. Customer is doing REST publishes, with a different clientId each time (as it's on behalf of different clients). Before every publish it gets a token, for itself, which it then uses exactly once.

This is needlessly high-latency for the customer, and needlessly resource-intensive for us.

> If we want the behaviour to be that specifying a clientId by default triggers token auth

If we're happy to drop that behaviour, then sure, that'd be a good and easy solution. (I don't see the justification for it anyway - what behaviour are we protecting against by forcing someone who has an api key and voluntarily wants to publish with a client id to do it with a token?)

SimonWoolf:
@mattheworiordan any opinion?

paddybyers:
I agree if someone has a key there's no point in making them create a token

[...]

PR #345, opened 2017-11-28
Stop specifying a clientId forcing token authentication

SimonWoolf:
Per discussion in https://github.com/ably/docs/issues/312.

(This section is numbered a bit weirdly -- 'Auth#clientId attribute is not null when:' is RSA7b, but its sibling 'Auth#clientId attribute is null when:', also a child or RSA7, is RSA12. But I'm a bit hesitant to renumber more sensibly as that'd make updating the spec spreadsheet and any references in client lib tests even more of a PITA...)


mattheworiordan:
I must admit I am nervous about this change as I know we consciously decided to do this.  I know that in the case of REST publishes now, we would (in the current proposal) now have to mutate the messages to now include an explicit clientId when a clientId is specified in ClientOptions if API key auth is used.  This feels very wrong.

@paddybyers has suggested we could solve this with a new param or header perhaps.  We'd need to implement this though in the spec.  I'll wait for him to review :)

paddybyers:
We don't currently, but we can easily bind a `clientId` to the `AuthParams` if basic auth is used and a `clientId` is specified as a query param

mattheworiordan:
> but we can easily bind a clientId to the AuthParams

Which `AuthParams`?  You mean for realtime connections?  Or headers using basic auth?

SimonWoolf:
>> but we can easily bind a clientId to the AuthParams

> Which AuthParams? You mean for realtime connections?

For realtime connections we already do this, I checked before writing rsa7e1.

```
$ wscat -c 'wss://sandbox-realtime.ably.io:443/?key=mz3G9w.G3yQww:Rjo9T6rLAKaK64wu&heartbeats=false&v=1.0&clientId=foo'

< {"action":4,"connectionId":"-hpRl0mPFM","connectionKey":"e02PL5F4gAEDaU!-hpRl0mPFMIsZrKQ-6242e02PL5F4gAEDaU","connectionSerial":-1,"connectionDetails":{"clientId":"foo","connectionKey":"e02PL5F4gAEDaU!-hpRl0mPFMIsZrKQ-6242e02PL5F4gAEDaU","maxMessageSize":65536,"maxInboundRate":999,"maxFrameSize":262144,"serverId":"frontend.7310.2.eu-west-2-A.i-0f710a35b4d4f7fc4","connectionStateTtl":120000,"maxIdleInterval":15000}}
> {"action":10,"channel":"channel"}
< {"action":11,"flags":983040,"channel":"channel","channelSerial":"e02uzpWSQAEDaH91882123:-1"}
> {"action":14,"channel":"channel","presence":[{"action":2}],"msgSerial":0}
< {"action":14,"id":"-hpRl0mPFM:0","connectionId":"-hpRl0mPFM","connectionSerial":0,"channel":"channel","channelSerial":"e02uzpWSQAEDaH91882123:0","timestamp":1511862389988,"presence":[{"id":"-hpRl0mPFM:0:0","clientId":"foo","connectionId":"-hpRl0mPFM","timestamp":1511862389988,"action":2}]}
```

If you meant for REST messages, I'm not sure I see the point. The way REST messages can have a clientId is with `Message.clientId`, what do we gain by adding a second way of specifying that in the querystring, in a stateless protocol?

mattheworiordan:
> If you meant for REST messages, I'm not sure I see the point.

I am only referring to REST as, as you say, realtime WS endpoints support clientId.

> The way REST messages can have a clientId is with Message.clientId, what do we gain by adding a second way of specifying that in the querystring, in a stateless protocol?

Currently, if you specify a clientId in the ClientOptions and publish a message like `channel.publish(name, payload)` over REST, it is received by all others with a `clientId` in the message because it is implicit in the publish operation.

Now, with your proposal, it would be received with no `clientId`, so it's a breaking change.

If the client lib then sets the `clientId` in the messages, it would be mutating the message objects which just feels wrong.

Hence why I was saying over REST we need to set the `clientId` as part of the REST operation so that an implicit `clientId` is included in the REST operations.

SimonWoolf:
> If the client lib then sets the clientId in the messages, it would be mutating the message objects which just feels wrong.

The message object delivered to the subscribers will include a `clientId` field either way. The only difference is whether it gets added to the message by the publishing client lib or by realtime (a difference completely hidden from the user, it makes no difference to the client api). Why does it feel wrong to you in the former case but not the latter?

Seems odd to me that we're considering adding more complication to the rest api in order to avoid to make it so that technically it's realtime that adds the field on the instruction of the client lib, rather than the client lib just adding it itself.

mattheworiordan:
> Seems odd to me that we're considering adding more complication to the rest api in order to make it so that technically it's realtime that actually adds the field, on the instruction of the client lib, rather than the client lib just adding it itself.

Because this:

* Introduces complexity in a client library i.e. this is bad when we have to roll this out multiple ties
* Is inconsistent with how realtime connections work where `clientId` is implicit, and in REST `clientId` is not. 
* Is trivial to add `clientId` to the request as a param or header if `clientId` is set in `ClientOptions`, and then as a result, the message object provided to the REST API endpoint exactly matches the message object provided to the API. In no other cases, as far as I am aware, other than for encoding purposes, do we ever mutate a customer provided message object before it arrives at Ably. Why would we want do that and then move the business logic and what is / is not allowed into the client libraries thus adding yet more client library logic in all client libs.

SimonWoolf:
OK, fair enough, I see where you're coming from. Added 3fed1e067bd9f9b8593dd23edc491b2e8d6e2bf8 accordingly.

(By analogy with `Authorization`, I've documented it as a header in the spec, but realtime should also supports it as a qs param for when custom headers aren't supported, e.g. jsonp)

mattheworiordan:
Thanks, LGTM 👍

content/client-lib-development-guide/features.textile
```diff
@@ -129,9 +129,12 @@ h3(#rest-auth). Auth
...
-*** @(RSA7a2)@ If @clientId@ is provided in @ClientOptions@, and an API key is provided along with no other means to generate a token, the client library will authenticate with Ably and obtain a token using the provided @clientId@ ensuring the token is restricted to operations for that @clientId@
+** @(RSA7d)@ If @clientId@ is provided in @ClientOptions@ and @RSA4@ indicates that token auth is to be used, the @clientId@ field in the @TokenParams@ (@TK2c@) should be set to that @clientId@ when requesting a token
+** @(RSA7e)@ If @clientId@ is provided in @ClientOptions@ and @RSA4@ indicates that basic auth is to be used, then:
+*** @(RSA7e1)@ For realtime clients, the connect request should include the @clientId@ as a querystring parameter, @clientId@
+*** @(RSA7e2)@ For REST clients, all requests should include an @X-Ably-ClientId@ header with value set to the @clientId@, Base64 encoded
```
mattheworiordan: @SimonWoolf I assume you don't think it's necessary to specify how JS handles this?  Good suggestion re: Base64 btw.

paddybyers: What's the reasoning behind having a header instead of a query param?

SimonWoolf: as in, for when custom headers aren't supported (i.e. IE<10)? Nah, I don't think it's necessary to put it in the spec, it just complicates things for everyone else. Similar to how we do the `Authorization` header -- realtime also supports auth by qs param for old IEs, but the spec only mentions the main (header) way.

(I think this is justified because it's only temporary -- at some point, maybe not too far away, non-websocket webbrowsers will drop below ε% usage share, and we can drop all the jsonp, comet, upgrade etc. crap from ably-js entirely)

SimonWoolf: > What's the reasoning behind having a header instead of a query param?

Consistency with the `Authorization: Basic ...` header (RSA11). Since that's how auth is already done for REST requests, adding a second header can be done just add it to the existing `Auth#getAuthHeaders` (or equivalent in another client lib) method.

Since there's currently no query param that should be applies to all rest requests, many client libs probably don't have an existing mechanism to add one, so doing it that way would be a bit more work.

(Though realtime will still have to support both for jsonp, so I'm trading off realtime ease of implementation for client lib ease of implementation)

mattheworiordan: I too prefer headers, although I don't have a good reason why 😉 
It just feels a whole load cleaner that the REST URL endpoint is the same regardless of your identity.

PR #348, opened 2017-12-01
Fixes some iOS Swift docs (handles renames)

mezis:
Looks like the Swift docs were slightly out of date?
(certainly not an iOS expert here)


tcard:
Good catch, this changed from Swift 2 to 3.

mezis:
CI fails — but the failures are present on the `push` feature branch as well, so I'm taking the liberty to merge.

PR #363, opened 2018-02-01
Added extra buildpack

tomczoink:
Added https://github.com/heroku/heroku-buildpack-static to buildpacks


tomczoink:
Build wasn't finishing due to missing Buildpack for heroku-buildpack-static. Added it in.

PR #409, opened 2018-03-28
Allow presence messages to include extras

paddybyers:


content/client-lib-development-guide/features.textile
```diff
@@ -1014,6 +1014,7 @@ h4. PresenceMessage
 ** @(TP3d)@ @connectionId@ string. If a presence message received from Ably does not contain a @connectionId@, it should be set to the @connectionId@ of the encapsulating @ProtocolMessage@
 ** @(TP3e)@ @data@ string, buffer or JSON-encodable object or array
 ** @(TP3f)@ @encoding@ string
+** @(TP3i)@ @extras@ JSON-encodable object, used to contain any arbitrary key value pairs which may also contain other primitive JSON types, JSON-encodable objects or JSON-encodable arrays. The @extras@ field is provided to contain message metadata and/or ancillary payloads in support of specific functionality. For 1.1 no specific functionality is specified for @extras@ in presence messages; the processing all members is undefined
```
SimonWoolf: "the processing ^of all members is undefined"

content/client-lib-development-guide/features.textile
```diff
@@ -561,23 +561,23 @@ h3(#realtime-presence). Presence
...
 ** @(RTP8d)@ Implicitly attaches the @Channel@ if the channel is in the @INITIALIZED@ state. However, if the channel is in the @DETACHED@ or @FAILED@ state, the @enter@ request results in an error
-** @(RTP8e)@ Optional data can be included when entering a channel that will be encoded / decoded as with normal messages. A test should exist to ensure data used with enter is encoded & decoded correctly. Also, when data is provided when entering, but no data is provided when leaving, the data attribute should be emitted in the @LEAVE@ event for this client
+** @(RTP8e)@ Optional data and/or extras can be included when entering a channel that will be encoded / decoded as with normal messages. A test should exist to ensure data and extras used with enter are encoded & decoded correctly. Also, when data and/or extras is provided when entering, but neither data nor extras are provided when leaving, the data attribute should be emitted in the @LEAVE@ event for this client
```
paddybyers: From Slack:

> I have noticed that we oscillate between `word/word` and `word / word`

Where it makes sense, instead of a slash, lets have English: "... that will be encoded and decoded as with normal messages..."

Otherwise I think we don't need spaces around a slash.

PR #412, opened 2018-03-28
Log message warning about lack of token renewability

paddybyers:
…a token without any means to renew


content/client-lib-development-guide/features.textile
```diff
@@ -111,7 +111,9 @@ h3(#rest-auth). Auth
...
-** @(RSA4a)@ When a @token@ or @tokenDetails@ is used to instance the library, and no means to renew the token is provided (either an API key, @authCallback@ or @authUrl@), if the server responds with a token error (401 HTTP status code and an Ably error value @40140 <= code < 40150@), then the client library should indicate an error, not retry the request and in the case of the realtime library, transition the connection to the @FAILED@ state
+** @(RSA4a)@ When a @token@ or @tokenDetails@ is used to instance the library, and no means to renew the token is provided (either an API key, @authCallback@ or @authUrl@):
+*** @(RSA4a1)@ if the server responds with a token error (401 HTTP status code and an Ably error value @40140 <= code < 40150@), then the client library should indicate an error, not retry the request and in the case of the realtime library, transition the connection to the @FAILED@ state
+*** @(RSA4a2)@ a message at @info@ log level should be logged indicating that no means has been provided to renew the supplied token, with a link to a "solution article":https://help.ably.io/a_suitable_article with configuration advice relating to token authentication.
```
mattheworiordan: This is wrong IMHO for two reasons:

* `help.ably.io` follows a format for automatic redirects based on an Ably error code. All solutions are are `support.ably.io`, however I fear that coding that URL into the client libs is a big mistake.  At least with an error code and `help.ably.io/errors/[CODE]` we can guarantee that URL won't change, even if we changed support systems down the line.
* Why can't we instead simply state that all errors emitted with an error code automatically include a URL to `https://help.ably.io/errors/[CODE]` as a single new requirement and thus we won't need a requirement like this for every possible error in this spec.

paddybyers: > help.ably.io follows a format for automatic redirects based on an Ably error code. All solutions are are support.ably.io, however I fear that coding that URL into the client libs is a big mistake. At least with an error code and help.ably.io/errors/[CODE] we can guarantee that URL won't change, even if we changed support systems down the line.

Right, that URL was clearly provisional. We can assign an error code for this.

> Why can't we instead simply state that all errors emitted with an error code automatically include a URL to https://help.ably.io/errors/[CODE] as a single new requirement and thus we won't need a requirement like this for every possible error in this spec.

For situations where we emit or return an error then we can, but this isn't an error.

paddybyers:
Now rebased

content/client-lib-development-guide/features.textile
```diff
@@ -118,7 +118,9 @@ h3(#rest-auth). Auth
...
```
SimonWoolf: > with a link to a solution article

help.ably.io currently works on error codes, so we should probably define a code for this (we can use the same one for rsa4a1 and rsa4a2, both'd be directed to the same article about renewability)

cf https://github.com/ably/wiki/issues/142

mattheworiordan: Yes, agreed, we just need to add a new error code.

paddybyers: I was sure we had added this but now I don't see it.

mattheworiordan: @paddybyers any progress on an error code?  Once done, @tomczoink please add a suitable solution article, or link to one if it exists (it should)

SimonWoolf:
This has stuck around for too long. Rebased on master, added an error code ([40171](https://github.com/ably/ably-common/blob/master/protocol/errors.json#L53)) and tweaked wording. Will merge in a day or two unless anyone objects.

(https://github.com/ably/docs/pull/412/commits/2027862ffe9f68d37ef5e496d911e5d2672df94b also appeared in the PR after the master rebase, which means it had been mistakenly merged into integration-spec-1-1 after that had been merged into master, so that can go in too)

PR #413, opened 2018-04-04
Spec: add explanation of supportsPong transportParam 

SimonWoolf:
cf https://github.com/ably/realtime/pull/1646


SimonWoolf:
> I am probably not fully understanding the problem here

Context was https://github.com/ably/infrastructure/issues/1655 . Realtime connection that (possibly due to a router bug) was broken (apparently fine but not receiving any data either way down the tcp socket, at least after shortly after connection). Realtime currently autodetects whether a transport supports ws pings by seeing if it's ever responded to one, if not then it assumes it doesn't, so doesn't have any way of knowing whether a connection's broken, so stays in a broken state indefinitely.

> why are we allowing this mixing and matching of ping/pong mechanisms when as far as I can tell, realtime could just rely on one mechanism only i.e. Ably Protocol heartbeats which require pong support, or native heartbeats which require client pong support as well as client ping detection? 

Because protocol heartbeats are only replied to when sent by the client. When sent by the server, the client doesn't respond, they're one way. So that doesn't work as a mechanism for the server to detect when the connection's gone.

(There's an issue somewhere from a couple years or so back when we were considering whether to change this)

If we think that every websocket implementation we might plausibly use supports pongs then we could just always set it to true, but presumably we don't or we wouldn't have implemented the autodetection mechanism.

(For why not just having supportsPong be forced to be true iff heartbeats=false, the thing is that whether the client can detect pings in application code isn't necessarily correlated with whether the websocket library it's using will send a pong. E.g. [ably-ruby currently can't detect when a ping has happened](https://github.com/ably/ably-ruby/issues/116), but its websocket lib still replies to them with pongs, it just doesn't tell the app code it's done so, so ably-ruby may want to set `heartbeats=true&supportsPong=true`).

SimonWoolf:
> Why add this additional complexity to the spec?

(If you think it's just not worth documenting this due to the complexity, feel free to close, we don't have to spec everything we implement, just figured it'd be nice to have it be documented)

mattheworiordan:
>  Realtime currently autodetects whether a transport supports ws pings by seeing if it's ever responded to one, if not then it assumes it doesn't, so doesn't have any way of knowing whether a connection's broken, so stays in a broken state indefinitely.

Ah, OK, I understand much better now.  Apologies, I didn't understand the issue sufficiently in the first place.

> Because protocol heartbeats are only replied to when sent by the client. When sent by the server, the client doesn't respond, they're one way. So that doesn't work as a mechanism for the server to detect when the connection's gone.

If we're changing the spec, why don't we change that then instead which IMHO is far simpler overall i.e. if a WS connection receives a PING, then send a PONG.  All WebSockets should do that automatically by default anyway, however if realtime chooses to a send a ProtocolMessage HeartBeat with a `PONG` flag or perhaps an ID that it knows did not originate from the client, then clients should respond to that.  This approach feels more flexible to me as we're empowering the realtime service to use whatever ping/pong mechanism it chooses, and adding a pong Heartbeat mechanism should be trivial in the clients.

> If we think that every websocket implementation we might plausibly use supports pongs then we could just always set it to true, but presumably we don't or we wouldn't have implemented the autodetection mechanism.

I am surprised if this is the case, and if it is, should we not be fixing that upstream instead. See https://tools.ietf.org/html/rfc6455#section-5.5.2, this is a requirement for WebSockets.

> (For why not just having supportsPong be forced to be true iff heartbeats=false, the thing is that whether the client can detect pings in application code isn't necessarily correlated with whether the websocket library it's using will send a pong. E.g. ably-ruby currently can't detect when a ping has happened, but its websocket lib still replies to them with pongs, it just doesn't tell the app code it's done so, so ably-ruby may want to set heartbeats=true&supportsPong=true).

Sure, but realtime could still use WebSocket heartbeats if it chooses, although this seems pointless if native heartbeat pong detection is not supported in that we will then be sending both native heartbeats as well as Ably Protocol heartbeats.  Why double up the work?

SimonWoolf:
> I am surprised if this is the case, and if it is, should we not be fixing that upstream instead. See https://tools.ietf.org/html/rfc6455#section-5.5.2, this is a requirement for WebSockets.

In client libs where 'fixing upstream' exists, sure, but that doesn't work for e.g. ably-js in web browsers. Pings are a requirement in the final version, but weren't in early websocket drafts, e.g. https://tools.ietf.org/html/draft-hixie-thewebsocketprotocol-76 , and some web browsers supported drafts that didn't support pings (in particular: Firefox 4-5, chrome 4-6, safari 5, opera 11).

FWIW, while [the version of the node ws lib we're using](https://github.com/websockets/ws/tree/1.1.0#protocol-support) supports websocket drafts old enough to not do ping/pong, [the current version drops support for them](https://github.com/websockets/ws#protocol-support). So perhaps upgrading the ws lib (so forcing those few old browsers listed above back to xhr, but who cares) and just always requiring pong support would be a simple and reasonable solution. @paddybyers wdyt?

> If we're changing the spec, why don't we change that then instead which IMHO is far simpler overall i.e. if a WS connection receives a PING, then send a PONG. All WebSockets should do that automatically by default anyway, however if realtime chooses to a send a ProtocolMessage HeartBeat with a PONG flag or perhaps an ID that it knows did not originate from the client, then clients should respond to that. This approach feels more flexible to me as we're empowering the realtime service to use whatever ping/pong mechanism it chooses, and adding a pong Heartbeat mechanism should be trivial in the clients.

Surprised you'd call that simpler -- since that'd be a change that client libs would have to implement to conform to the spec (they'd need to parse the flags field to know whether to echo the heartbeat), surely that alone makes it much a bigger thing than a completely optional qs param, since it'd then be a new 1.1 blocker for every realtime client lib, which seems like something we want to avoid at this stage of the 1.1 spec

Also if we think that (per first point above) the set of realtime clients that don't support ws pings is either non-existent or will become so in the next few years, doesn't seem much point to put effort into changing the spec to accommodate them. So lets decide the first point first

paddybyers:
> FWIW, while the version of the node ws lib we're using supports websocket drafts old enough to not do ping/pong, the current version drops support for them. So perhaps upgrading the ws lib (so forcing those few old browsers listed above back to xhr, but who cares) and just always requiring pong support would be a simple and reasonable solution. @paddybyers wdyt?

Sooner or later we'll want to upgrade the ws lib, and this issue won't stop us from doing that. So we are going to end up in the situation that the only ws clients we support are those "conforming" to the later spec versions that include pong support. That's not to say that they actually conform, however.

Also, supposing we do make pong support mandatory, so we kill any transport in which the client fails to respond, how does that sit with the upgrade process? Will we have accepted an upgrade transport as having suceeded, and closed the comet transport, only for us to discover that the ws transport isn't supported, and then what? Our current approach assumes that if you succeed in establishing a connection then the transport is viable. Do we want to revist that?

SimonWoolf:
> Also, supposing we do make pong support mandatory, so we kill any transport in which the client fails to respond, how does that sit with the upgrade process? Will we have accepted an upgrade transport as having suceeded, and closed the comet transport, only for us to discover that the ws transport isn't supported, and then what? Our current approach assumes that if you succeed in establishing a connection then the transport is viable. Do we want to revist that?

If it turns out there exist browsers that claim to support the post-ping ws spec but don't actually support pongs, then we'll have to address that - maybe by blacklisting by useragent in the wsserver or something. But doesn't seem much point putting effort into solving problems that might not exist, surely

OK, how about we gather some more data -- I could do a realtime PR that does a sentry warning if either someone connects with Hixie websockets or a modern websocket lib that doesn't seem to support pong per our autodetection, along with the user-agent etc, and we can let that run for a few weeks and see if it triggers and if so on what platforms?

paddybyers:
> I could do a realtime PR that does a sentry warning if either someone connects with Hixie websockets or a modern websocket lib that doesn't seem to support pong per our autodetection, along with the user-agent etc, and we can let that run for a few weeks and see if it triggers and if so on what platforms?

sgtm, thanks

mattheworiordan:
> Surprised you'd call that simpler -- since that'd be a change that client libs would have to implement to conform to the spec (they'd need to parse the flags field to know whether to echo the heartbeat), surely that alone makes it much a bigger thing than a completely optional qs param, since it'd then be a new 1.1 blocker for every realtime client lib, which seems like something we want to avoid at this stage of the 1.1 spec

I mean simpler in that it's much easier to rationalise i.e. ping/pong is either at the WebSocket level or at the Ably Protocol level, but not both as it's unnecessary.  Right now we have to do both whenever client-side WebSocket ping detection is not supported.

funkyboy:
Need I audit the libs we use for iOS/Java to check if they implement/support ping detection?

PR #415, opened 2018-04-12
Fix RSL1g duplicate

funkyboy:


funkyboy:
This is to *prevent* duplicates in the spreadsheet, so I’ll update it after this is merged.

BUT, I didn’t think of repos. 
Do you have in mind a less disruptive relabeling that doesn’t require to update/check all the repos?

paddybyers:
> Do you have in mind a less disruptive relabeling that doesn’t require to update/check all the repos?

no :(

paddybyers:
> that doesn’t require to update/check all the repos?

If this is a new requirement being added for 1.1 then you only need to check ably-js because none of the others will have been updated.

funkyboy:
@paddybyers Lastly, AFAICT ably-js doesn't perform any check on the message size, right @SimonWoolf? 
There's just a check on the size of token requests: https://github.com/ably/ably-js/blob/master/common/lib/client/auth.js#L428

SimonWoolf:
> Lastly, AFAICT ably-js doesn't perform any check on the message size, right @SimonWoolf?

not yet, it's a todo for 1.1

Issue #396, opened 2018-03-12
Best practice and handling errors


mattheworiordan:
@funkyboy suggested we add a best practice guide for developers.  This is a good idea and could sit in the main navigation.

Additionally, having a section specifically about handling errors for the SDKs makes a lot of sense. That should probably sit within the SDK reference itself.  Best practice guide would link to this and advise that developers handle errors.

funkyboy:
To elaborate a bit more, this "guide" should contain information like:

- there's no point in calling `attach()` on a channel that is in a `FAILED` state
- make sure that the log level is `.error` when your app is in production
- are you really sure you want to `subscribe` twice on the same channel?

mattheworiordan:
> there's no point in calling attach() on a channel that is in a FAILED state

That is not always correct as far as I am aware.  A `FAILED` state is terminal in that it won't reattach automatically.  But if for example it became `FAILED` because you don't have permission, you could potentially listen for `FAILED` channel attach events, then go and get a new token with up to date permissions and try again.

> make sure that the log level is .error when your app is in production

Not warn?  I can't recall if we actually have a warn log level though :)

mattheworiordan:
I think best practice guide needs to cover off:

* When subscribe filters are appropriate (mostly never)
* Handling connection state change events
* Handling channel state change events
* Almost never right to use an API key on an untrusted device
* Ensure tokens are renewable
* Understanding REST vs Realtime and the other protocols we support
* It's mostly never correct to subscribe to events server-side and that is what the Reactor exists
* If you want to scale your server-side consumption of messages, then you should probably consider not using WebHooks or Reactor Functions.
* Multiplexing when using the Realtime lib
* Follow principle of least privilege for untrusted clients using tokens and capabilities
* Using channels correctly - they are your unit of distribution and we encourage people to use lots of them, but channels consume resources, so you have to be mindful of that too.

@funkyboy @Srushtika can you two please chat with the team to see what other best practice suggestions we can come up with?

Srushtika:
Sure @mattheworiordan.

PR #418, opened 2018-04-19
best practices guide added

Srushtika:
Initial idea referred from https://github.com/ably/docs/issues/396

### DIFF
--- a/content/root/best-practice-guide.textile
+++ b/content/root/best-practice-guide.textile
@@ -22,29 +22,29 @@ Ably Realtime is a cloud-based platform that allows you to easily integrate real
 
 h4(#pubsub). Publish and Subscribe (Pub/Sub)
 
-You can publish data to a named "channel":/realtime/channels-messages on the Ably Realtime platform, and all the clients subscribed to that very channel will receive this data as it is being published.
+You can publish data to a named "channel":/realtime/channels-messages on the Ably Realtime platform, and all the clients subscribed to that very channel will receive this data in realtime, as it is being published.
 
 h4(#notifications). Notifications 
 
-You can use "Push Notifications":/realtime/push to send notifications to your clients even when they are currently not using your application, or in other words, not connected to Ably. You can do this both on mobile devices as well as browsers. You can also use "Smart Notifications":/general/smart-notifications for your notifications to be most effective depending on your user's connection statte, their active device or even their current context within your application.
+You can use "Push Notifications":/realtime/push to send notifications to your clients even when they are currently not using your application, or in other words, not connected to Ably. You can do this on mobile devices as well as browsers. You can also use "Smart Notifications":/general/smart-notifications for your notifications to be most effective depending on your user's connection statte, their active device or even their current context within your application.
 
 h4(#presence). Presence
 
-You can have your clients attach to the presence channel using "Presence":/realtime/presence in order to indicate if these clients (or devices in case of IoT clients) are online or connected. The changes in this connection status also can be tracked by optionally subscribing to the same presence channel.
+You can have your clients attach to the presence channel using "Presence":/realtime/presence in order to indicate if these clients (or devices in case of IoT clients) are online or connected. The changes in this connection status, if any, can also be tracked by optionally having your clients subscribe to the same presence channel.
 
 h4(#rest-or-realtime). Realtime vs REST
 
-With the availability of both the "Realtime":/realtime and "REST":/rest libraries, choose whichever suits your application's needs the best and get started. While using the realtime library lets your clients subscribe to certain data that Ably can push to your clients as it is being created, REST library offers simple stateless data publish or retrieval to and from Ably respectively.
+With the availability of both the "Realtime":/realtime and "REST":/rest libraries, choose whichever suits your application's needs the best and get started. Using the realtime library, your clients can subscribe to certain data that Ably can then push to your clients as it is being created. On the other hand, the REST library offers simple stateless data publish or retrieval to and from Ably respectively.
 
 h4(#auth). Choose the right auth scheme
 
-You can use "Basic Authentication":/general/authentication#basic-authentication to easily authenticate your clients using an API key or use "Token Authentication":/general/authentication#token-authentication to authenticate your clients using a token request sent to Ably via your auth server. You can also "encrypt your messages":/realtime/encryption for added security of your application.
+You can use "Basic Authentication":/general/authentication#basic-authentication to easily authenticate your clients using an API key or use "Token Authentication":/general/authentication#token-authentication to authenticate your clients using a token request sent to Ably via your auth server. Furthermore, you can use JWT authentication as part of Tokent Authentication ensuring that you use a single authentication strategy for all of your third party services. You can also "encrypt your messages":/realtime/encryption for added security of your application. 
 
 h3(#using-ably-like-a-pro). Using Ably like a pro
 
 h4. 1. Attached channels automatically receive messages from Ably even without a subscribe listener
 
-*Once a channel is attached, assuming the client has permission to subscribe to messages, the Ably platform will immediately start sending published messages to that client irrespective of whether there are any subscribe listeners registered on that client or not. A common misunderstanding is that only once a client "subscribes":/realtime/channels-messages#message-subscription to a channel will Ably start sending messages to the client. This is not the case as registering a listener with a Channel subscribe operation is a client-side operation only and is not communicated to Ably. 
+* Once a channel is attached, assuming the client has permission to subscribe to messages, the Ably platform will immediately start sending published messages to that client irrespective of whether there are any subscribe listeners registered on that client or not. A common misunderstanding is that only after a client has "subscribed":/realtime/channels-messages#message-subscription to a channel will Ably start sending messages to that client. This is not the case as registering a listener with a channel subscribe operation is a client-side operation only and is not communicated to Ably. 
 
 h4. 2. Use separate channels when you want messages to be sent to different sets of subscribers
 
@@ -59,15 +59,15 @@ sportsChannel.publish("add", { "team": "Chelsea" });
 //subscribe filters
 sportsChannel.subscribe("update", function(){
   //do one thing
-})
+});
 sportsChannel.subscribe("add", function(){
   //do another thing
-})
+});
 ```
 
 h4. 3. Being present without subscribing to presence events
 
-* If you want lots of people to be present, but not necessarily listen for "presence change events":/realtime/presence#presence-states, you should use token capabilities to prevent all users from receiving presence notifications while allowing them to enter only. This is a common scenario when you just want to know how many people are present on a channel at any point in time. An example of such a token capability is shown below:
+* If you want lots of people to be present, but not necessarily listen for "presence change events":/realtime/presence#presence-states, you should use token capabilities to prevent all users from receiving presence notifications while allowing them to enter presence set only. This is a common scenario when you just want to know how many people are present on a channel at any point in time. An example of such a differential token capability is shown below:
 
 ```json
 {
@@ -78,9 +78,9 @@ h4. 3. Being present without subscribing to presence events
 
 h4. 4. Handling connection and channel state-change events
 
-Ably's Realtime SDKs automatically handle disruptions such as network disconnections, so you don't need to do manually handle reconnection. The library emits:
+Ably's Realtime SDKs automatically handle disruptions such as network disconnections, so you don't need to manually handle reconnection. The library emits:
 
-* "Connection":/realtime/connection#connection-states state change events such as @initialized@, @connecting@, etc. You can set up listeners for these state changes in order to handle connection failures such as those occuring due to failed authentication. An example of handling connection state-change event is shown below:
+* "Connection":/realtime/connection#connection-states state change events such as @initialized@, @connecting@, etc. You can set up listeners for these state change events in order to handle connection failures such as those occuring due to failed authentication. An example of handling connection state-change event is shown below:
 
 ```javascript
 realtime.connection.on(function(stateChange) {
@@ -88,7 +88,7 @@ realtime.connection.on(function(stateChange) {
 });
 ```
 
-* "Channel":/realtime/channels-messages#channel-states state change events for various channel states such as @initialized@, @attaching@, etc. Similar to connection state changes, you can set up listeners for channel state changes as well, in case you'd like be informed about what's happening. For example, if you want to know whenever the library has lost message continuity on a channel due to being disconnected for more than two minutes. An example of handling channel state-change event is shown below:
+* "Channel":/realtime/channels-messages#channel-states state change events for various channel states such as @initialized@, @attaching@, etc. Similar to connection state changes, you can set up listeners for channel state changes in case you'd like be informed about what's happening. For example, if you want to know whenever the library has lost message continuity on a channel due to being disconnected for more than two minutes. An example of handling channel state-change event is shown below:
 
 ```javascript
 var myListener = function(stateChange) {
@@ -97,7 +97,7 @@ var myListener = function(stateChange) {
 });
 channel.on(myListener);
 ```
-However, while registering listeners for both connection and channel state change events, bear in mind that certain repeating states might add new listeners everytime. For instance, registering a listener @on(connected)@ adds a new listener every time the client becomes connected, even if this is a reconection after being offline for a while.
+However, while registering listeners for both connection and channel state change events, bear in mind that certain repeating states might add new listeners everytime. For instance, registering a listener @on(connected)@ adds a new listener every time the client becomes connected, even if this is a reconection after being offline for a while, which in most case might not be something you intended to do.
 
 h4. 5. Using API key on the client side
 
@@ -111,18 +111,18 @@ h4. 7. Using the correct client library SDK - Choosing between Realtime, REST, M
 
 * Want to maintain a persistent connection with Ably's Realtime platform, so you can subscribe to messages sent on a channel, or be notified whenever the presence set changes? Use the "realtime library":/realtime.
 * Want to just publish data to the platform and don't need to subscribe or manage channel attachment? Use the "REST library":/rest.
-* Want to publish at very high volumes or with the lowest possible latency? "The realtime libary can publish at higher volumes (due to message pipelining) and lower latencies than is possible with the REST library":https://support.ably.io/solution/articles/3000060319-should-i-use-the-rest-or-realtime-library.
+* Want to publish at very high volumes or with the lowest possible latency in realtime? "Ably's realtime libary can publish at higher volumes and lower latencies than is possible with the REST library due to message pipelining":https://support.ably.io/solution/articles/3000060319-should-i-use-the-rest-or-realtime-library.
 * Want to get one-time info about the presence set of a channel? Use the "REST library":/rest.
 * Want to set up realtime communication with IoT devices on platforms for which there's no native Ably client library available? Use the "MQTT protocol adapter":/mqtt.
-* Want to migrate from PubNub/Pusher? Use the relevant "protocol adapter":/adapters to "ease the pain of migration":https://support.ably.io/solution/articles/3000054127-can-you-help-ease-the-pain-of-migration-from-another-realtime-service: you can switch to Ably right away while sticking with your current client libraries, and gradually migrate to native Ably clients.
+* Want to migrate from PubNub/Pusher? Using the relevant "protocol adapter":/adapters you "ease the pain of migration":https://support.ably.io/solution/articles/3000054127-can-you-help-ease-the-pain-of-migration-from-another-realtime-service: and switch to Ably right away and gradually migrate to using native Ably client SDKs.
 
 h4. 8. Subscribing to events on the server-side
 
-* Subscribing to events on the server side using the standard Pub/Sub pattern can be a bad idea for a number of reasons such as additional layer of latency, duplicate events among multiple servers etc, as explained in this "blog article":https://blog.ably.io/message-queues-the-right-way-to-process-and-work-with-realtime-data-on-your-servers-2d15985301f8. With Message Queues, you can set up multiple worker servers to distribute the load of the messages received from Ably while ensuring each of the messages gets processed exactly once by any one of your worker servers.
+* Subscribing to events on the server side using the standard Pub/Sub pattern can be a bad idea for a number of reasons such as an added layer of latency, duplicacy of events among multiple servers etc, as explained in this "blog article":https://blog.ably.io/message-queues-the-right-way-to-process-and-work-with-realtime-data-on-your-servers-2d15985301f8. With Message Queues, you can set up multiple worker servers to distribute the load of the messages received from Ably while ensuring each of the messages gets processed exactly once by any one of your worker servers.
 
 h4. 9. WebHooks/Reactor Functions vs. Message Queues when to use which?
 
-* "WebHooks":/general/webhooks and "Reactor Functions":/general/functions work well for low or medium message volumes, but "Reactor Queues":/general/queues will scale better to high message volumes. For very high volumes, we recommend using "Reactor Firehose":/general/firehose.
+* "WebHooks":/general/webhooks and "Reactor Functions":/general/functions work well for low or medium message volumes i.e ~20/s, but "Reactor Queues":/general/queues will scale better to high message volumes i.e ~200/s. For volumes higher than that, we recommend using "Reactor Firehose":/general/firehose.
 
 h4. 10. Multiplexing channels
 
@@ -130,7 +130,7 @@ h4. 10. Multiplexing channels
 
 h4. 11. Limiting the effect of a security compromise using the principle of least privilege
 
-* Using the available "authentication methods":/general/authentication in Ably, you can choose the permissions or capabilities to be given to a client. You can do this by restricting the permissions on the API key as well as restricting the permissions your auth server grants to that API key whilst granting a token request. You can restrict the channels that a client can access (including with wildcards), as well as the operations the client can perform within those channels, using "capabilities":/general/authentication#capabilities-explained.
+* Using the available "authentication methods":/general/authentication in Ably, you can choose the permissions or capabilities to be given to a client. You can do this by restricting the permissions on the API key as well as restricting the permissions that your auth server grants to this API key whilst sending a token request. You can restrict the channels that a client can access (using wildcards), as well as the operations the client can perform within those channels, using "capabilities":/general/authentication#capabilities-explained.
 * Always make sure to limit the capabilities given to a client to only what's required by the client to do their job. This prevents any unexpected behaviour by a client and increases the security of your application. This is called the principle of least privilege.
 
 h4. 12. Efficient use of channels
--- a/content/root/best-practice-guide.textile
+++ b/content/root/best-practice-guide.textile
@@ -67,7 +67,14 @@ sportsChannel.subscribe("add", function(){
 
 h4. 3. Being present without subscribing to presence events
 
-* If you want lots of people to be present, but not necessarily listen for "presence change events":/realtime/presence#presence-states, you should use token capabilities to prevent all users from receiving presence notifications while allowing them to enter only. This is a common scenario when you just want to know how many people are present on a channel at any point in time.
+* If you want lots of people to be present, but not necessarily listen for "presence change events":/realtime/presence#presence-states, you should use token capabilities to prevent all users from receiving presence notifications while allowing them to enter only. This is a common scenario when you just want to know how many people are present on a channel at any point in time. An example of such a token capability is shown below:
+
+```json
+{
+  "presenceChannel": ["publish", "presence"],
+  "chat": ["presence", "history", "subscribe],
+}
+```
 
 h4. 4. Handling connection and channel state-change events
 
@@ -99,7 +106,7 @@ h4. 6. Ensuring tokens are auto-renewable
 
 * It is almost never correct to provide a token to a client library without ensuring the client library has a means to get a new token automatically when it needs one (tokens expire as they are intentionally short lived). The Ably client library SDKs provide two mechanisms in the "ClientOptions":/realtime/authentication#token-authentication so that they can obtain tokens automatically, via "@authUrl@":/realtime/authentication#token-upgrading and "@authCallback@":/realtime/authentication#token-upgrading.
 
-h4. 7. Using teh correct client library SDK - Choosing between Realtime, REST, MQTT and others 
+h4. 7. Using the correct client library SDK - Choosing between Realtime, REST, MQTT and others 
 
 * Want to maintain a persistent connection with Ably's Realtime platform, so you can subscribe to messages sent on a channel, or be notified whenever the presence set changes? Use the "realtime library":/realtime.
 * Want to just publish data to the platform and don't need to subscribe or manage channel attachment? Use the "REST library":/rest.
--- a/content/root/best-practice-guide.textile
+++ b/content/root/best-practice-guide.textile
@@ -8,7 +8,7 @@ jump_to:
     - Using Ably like a pro
 ---
 
-h2(#title). Ably Best Practice Guide
+h1(#title). Ably Best Practice Guide
 
 This best practice guide is designed to help you make the most out of Ably's Realtime platform for your applications. These recommendations include some important considerations that may otherwise be easily missed when reading through our documentation. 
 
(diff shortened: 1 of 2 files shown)


mattheworiordan:
I'll come back to this, but two key things missing:

* Attaching to channel means that Ably will start sending message data to you the client.  So bear that in mind as if you attach to publish, then even if you never call subscribe, the client will receive all message data.  This is because subscribe simply registers a listener with the client, but Ably has no knowledge of this, and only listens for a client to attach to a channel, and will then send messages once attached.
* If you want lots of people to be present, but not necessarily listen for presence (common when you want to know how many people are present), then you should use token capabilities to prevent all users receiving presence (subscribe) and only allow then to enter.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,89 @@
...
+
+h4. 2. Handling connection and channel state&#45;change events
+
+* Ably's Realtime SDKs automatically handle disruptions such as network disconnections, you don't need to do anything. The library emits "connection":/realtime/connection#connection-states and "channel":/realtime/channels-messages#channel-stateschannel state change events that you can listen to in case you'd like be informed about what it's doing (for example, if you want to know whenever the library has lost message continuity on a channel due to being disconnected for more than two minutes). But there's no need to take any particular action in response to any of these, the library will automatically reconnect and reattach to channels after the network disruption is over.
```
tomczoink: '...you don't need to do anything' => 'so you don't need to do anything'

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,89 @@
...
+
+h4. 3. Using API key on the client side
+
+* Ably's "Basic Authentication":/general/authentication#basic-authentication method is the easiest way to authenticate your clients on Ably Realtime platform. However, it directly exposes the API key in the source code. When you use it on the client side, any entity that interacts with the client could view your API key and potentially misuse it. If you can run an auth server, we recommend using "Token Authentication":/general/authentication#token-authentication on the client side. Read more about selecting an authentication mechanism.
```
tomczoink: 'clients on Ably Realtime platform' => 'client on the Ably Realtime platform'.

A link for 'Read more' potentially?

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,89 @@
...
+* Want to just publish data to the platform, don't care about subscribing, and don't want to manage channel attachment? Use the "REST library":/rest.
+* Want to publish at very high volumes or with the lowest possible latency? "The realtime libary can publish at higher volumes (due to message pipelining) and lower latencies than is possible with the REST library":https://support.ably.io/solution/articles/3000060319-should-i-use-the-rest-or-realtime-library-.
+* Want to get one&#45;time info about the presence set of a channel? Use the "REST library":/rest.
+* Want to set up realtime communication with IoT devices on platforms for which there's no native Ably client library available? Use the "MQTT protocol adapter":/mqtt.
+* Want to migrate from PubNub/Pusher? Use the relevant "protocol adapter":/adapters to "ease the pain of migration":https://support.ably.io/solution/articles/3000054127-can-you-help-ease-the-pain-of-migration-from-another-realtime-service-: you can switch to Ably right away while sticking with your current client libraries, and gradually migrate to native Ably clients.
```
tomczoink: Remove - on end of link.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,89 @@
...
+
+h4. 5. Choosing between Realtime/ REST libraries and Protocol Adapters 
+
+* Want to maintain a persistent connection with Ably's Realtime platform, so you can subscribe to messages sent on a channel, or be notified whenever the presence set changes? Use the "realtime library":/realtime.
+* Want to just publish data to the platform, don't care about subscribing, and don't want to manage channel attachment? Use the "REST library":/rest.
+* Want to publish at very high volumes or with the lowest possible latency? "The realtime libary can publish at higher volumes (due to message pipelining) and lower latencies than is possible with the REST library":https://support.ably.io/solution/articles/3000060319-should-i-use-the-rest-or-realtime-library-.
```
tomczoink: Remove - on end of link

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,89 @@
...
+
+h4. 7. WebHooks/Reactor Functions vs. Message Queues &#45; when to use what?
+
+* "WebHooks":/general/webhooks and "Reactor Functions":/general/functions work well for low or medium message volumes, but "Reactor Queues":/general/queues will scale better to high message volumes. For very high volumes, we recommend using "Reactor Firehose":/general/firehose.
+
+h4. 8. Multiplexing channels
+
+* In almost all cases you should always instantiate a single Ably client library instance per web page or server. Don't forget that the Client Library SDK can "multiplex":https://support.ably.io/support/solutions/articles/3000058406-do-you-support-channel-groups-and-multiplexing- many channels over a single connection.
```
tomczoink: Remove - on end of link

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,89 @@
...
+* Want to get one&#45;time info about the presence set of a channel? Use the "REST library":/rest.
+* Want to set up realtime communication with IoT devices on platforms for which there's no native Ably client library available? Use the "MQTT protocol adapter":/mqtt.
+* Want to migrate from PubNub/Pusher? Use the relevant "protocol adapter":/adapters to "ease the pain of migration":https://support.ably.io/solution/articles/3000054127-can-you-help-ease-the-pain-of-migration-from-another-realtime-service-: you can switch to Ably right away while sticking with your current client libraries, and gradually migrate to native Ably clients.
+
+h4. 6. Subscribing to events on the server&#45;side
+
+* Subscribing to events on the server side using the standard Pub/Sub pattern can be a bad idea for a number of reasons as explained in this "blog article":https://blog.ably.io/message-queues-the-right-way-to-process-and-work-with-realtime-data-on-your-servers-2d15985301f8. With Message Queues, you can set up multiple worker servers to distribute the load of the messages received from Ably while ensuring each of the message gets processed exactly once by any one of your worker servers.
```
tomczoink: 'each of the message' => 'each of the messages'

content/root/best-practices-guide.textile
tomczoink: 'In almost all cases you should always...', 'almost' and 'always' contradict, could simply go for 'In most cases you should instantiate...', or if you want emphasis on how often maybe 'You should almost always instantiate...'.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,89 @@
...
+
+h4. 8. Multiplexing channels
+
+* In almost all cases you should always instantiate a single Ably client library instance per web page or server. Don't forget that the Client Library SDK can "multiplex":https://support.ably.io/support/solutions/articles/3000058406-do-you-support-channel-groups-and-multiplexing- many channels over a single connection.
+
+h4. 9. Limiting the effect of a security compromise using the principle of least privilege
+
+* Using the available "authentication methods":/general/authentication in Ably, you can choose the permissions or capabilities to be given to a client. You can do by restricting the permissions on the API key as well as restricting the permissions your auth server grants to that API key whilst granting a token request. You can restrict the channels that a client can access (including with wildcards), as well as the operations the client can perform within those channels, using "capabilities":/general/authentication#capabilities-explained.
```
tomczoink: 'You can do by...' => 'You can do this by...'

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,89 @@
...
+* Always make sure to limit the capabilities given to a client to only what's required by the client to do their job. This prevents any unexpected behaviour by a client and increases the security of your application. This is called the principle of least privilege.
+
+h4. 10. Efficient use of channels
+
+Channels consume resources and hence cost you money. Make sure that you are using them efficiently and in accordance to the limit you have as per your package. However, bear in mind the following:
+
+* Don't use use different channels just to indicate different types of data or different events, if they're all going to the same set of clients. Use one channel, and distinguish between the different events using the message name.
```
tomczoink: 'Don't use use...' => 'Don't use...'

Srushtika:
@mattheworiordan, I've added those two points. Thanks!

@tomczoink, thanks for the review, most of the links were messed up as I automated the HTML -> textile process :) But all fixed now. Typos and grammar fixed too. wrt the structure, most of the multiple points are explanations themselves actually unlike the last one. I agree that it would be a better structure indeed but with the current one clearly distinguishes a 'best practice' title from a description about the same. However, if this doesn't seem sensible, I'll be happy to edit it. wdyt?
Another thing is that the current structure then makes the last section inconsistent as you mentioned. Still thinking how this can be made better

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
+---
+title: Best Practices Guide
```
mattheworiordan: Best Practice Guide, not Best Practices Guide.  Same applies below.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+index: 30
+jump_to:
+  Contents:
+    - Ably 101
+    - Using Ably like a pro
+---
+
+h2(#title). Ably Best Practices Guide
+
+In order to avail the best advantage of our platform, be sure to follow this best practices guide while building your applications using Ably.
```
mattheworiordan: I don't like this intro I am afraid.  I think it would be better to explain that we've created this handy best practice guide in two sections:
- "Ably 101 - key concepts" - before you get started, make sure you understand 
- "Ably implementation checklist" - once you've started implementing Ably, we've compiled a quick checklist to help you avoid common mistakes or misunderstandings we see when developers implement Ably.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h3(#ably-101). Ably 101
+
+Ably Realtime is a cloud-based platform that allows you to easily integrate realtime functionality into your applications. Just plug and play.
+
+h4(#pubsub). Publish and Subscribe
```
mattheworiordan: I think this should be "Pub/Sub (publish and subscribe)".  Pub/Sub is the common terminology.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+Publish data to a named "channel":/realtime/channels-messages on the Ably Realtime platform, and everyone subscribed to that same channel will receive it.
```
mattheworiordan: I don't think we should refer to people, but instead should refer to clients.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h4(#notifications). Notify
+
+You can use "Push Notifications":/realtime/push to inform your users about the cool new features you are building in your application.
```
mattheworiordan: This is not a good description, it doesn't explain what a notification is.  The title is also misleading, you could use realtime sockets to Notify. I don't think "cool new features" is right according to our ToV.  Please consider the tone, and consider that we don't know what people use push notifications for.  They are there to enable native notifications to be sent to devices when an app is potentially not open / connected to Ably, same applies to browsers.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h4(#presence). Be present
```
mattheworiordan: Why not just "Presence"?  I appreciate this is meant to be 101,  but why is "Be present" (an action) simpler than "Presence" the feature and the same terminology we use everywhere else in our documentation.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+Go online and let others in your application know this by using "Presence":/realtime/presence. You can also see who else is present on a channel even when you yourself are NOT present.
```
mattheworiordan: They don't go online with presence, a client becomes present on a channel.  Again, why are talking about people as  opposed to devices or clients.  Presence could be used by an IoT device for example to indicate it is online and connected.  Why is `NOT` in capitals?  I am not sure of the relevance of explaining in this description that you can do that, this high level summary should explain what presence is and how it can be used.  Whether a client is present or not IMHO is not all that relevant or helpful at this point.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h4(#rest-or-realtime). Remain connected, or not
```
mattheworiordan: I don't know what this means.  Why would someone want to remain connected or not?  Whether you use Realtime or REST has nothing to do with remaining connected or not, it has to do with the functionality that is required and having a stateful connection or not.  A REST library technically may keep an idle HTTP connection open until the next request is issued.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+With the availability of both the "Realtime":/realtime and "REST":/rest libraries, choose what suits your application's needs best and get started.
```
mattheworiordan: You've not made any attempt to explain why one would chose Realtime or REST.  In addition, you've not linked to any content that helps a user understand the difference.  This section does not help the user but confuses them further.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h4(#auth). Choose the right auth scheme
+
+Use "Basic Authentication":/general/authentication#basic-authentication or "Token Authentication":/general/authentication#token-authentication as well as ensure security by "encrypting your messages.":/realtime/encryption
```
mattheworiordan: Authentication and encryption are not related concepts.  Why have you grouped them together?
Also, security is not "encryption", security is a far broader concept that may encompass many features including auth.
You have not explained here why one should use Basic auth or Token auth, but instead have linked to the documentaiton on eahc.  The point of this document is surely to help people understand best practice, yet this document does not do that. It's just mirroring our documentation links.  Best practice would for example state that typically in app, device or browser clients, we recommend you use Token Authentication because the device is untrusted and you can dynamically assign capabilities to each client from your servers, and a realtime client library should be used to maintain a connenction to Ably and subscribe to realtime updates.  Typically on your servers, we most often see people using API keys as this is a simpler authentication scheme when used on private servers you can trust, and REST libraries are often simpler to use because servers often have no need to maintain a connection to Ably as they are publishing messages, or issuing tokens for example.

The above is "best practice".  What you have provided is a list of features with links to our docs, which they have already.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h3(#using-ably-like-a-pro). Using Ably like a pro
+
+h4. 1. Attaching vs subscribing to a channel
+
+* When your client merely attaches to a channel on Ably, the platform will start sending message data even if your client has not subscribed to the channel. Subscribing to a channel simply registers a listener using which your client can perform additonal operations, however, Ably has not knowledge about this as it only listens for a client to attach to a channel. Bear this in mind while attaching to a channel for publish-only purpose.
```
mattheworiordan: I realise this is largely a copy & paste of my points, but it was not my intention that this would be used in this doc.  This is not a good description of the problem I am afraid.  Think about what we're trying to do here which is help people avoid common gotchas / understand things they may not otherwise obtain from our detailed documentation.  So for example, if this is a checklist of things a user should go through, would something like the following be better:

Title: "Attached channels automatically messages from Ably, even without subscribe listeners"
Body: Once a client has attached to a channel, Ably will automatically start sending messages to that client regardless of whether a subscribe listener has been registered.  A common misunderstanding is that only once a client [subscribes](/realtime/channels-messages#message-subscription) to a channel will Ably start sending messages to the client.  This is not the case as registering a listener with a Channel [subscribe](/realtime/channels-messages#message-subscription) operation, is a client-side operation only and is not communicated to Ably.  The Ably platform, once a channel is attached and if the client has permission to subscribe to messages, will immediately start sending published messages to that client irrespective of whether there are any subscribe listeners registered on that client or not.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h4. 2. Using separate channels when you want messages to be sent to different sets of subscribers.
+
+* Ably pushes all the data published on a channel to every subscriber attached to that channel. Subscribers can choose to only subscribe to events with a certain message name, but that filtering is done on the client side (however, it's useful for things like letting the subscriber trigger different listeners for different events). If you have messages that don't need to be sent to some clients at all, you're often better off using a separate channel for those instead of implementing a filter. This way, you won't be sending unnecessary messages that count towards your package limit.
```
mattheworiordan: Please remove the "(however, it's useful)..." and focus on the problem you are trying to explain to the user.  You have just written a sentence explaining how subsctibe works, focus on that concept, explain it in full, and then perhaps in a separate sentence or paragraph explain where the filtering is useful.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h4. 3. Being present without subscribing to presence events
+
+* If you want lots of people to be present, but not necessarily listen for "presence change events":/realtime/presence#presence-states, you should use token capabilities to prevent all users from receiving presence notifications while allowing them to enter only. This is a common scenario when you just want to know how many people are present on a channel at any point in time.
```
mattheworiordan: Could you provide a very simple code example?

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h4. 4. Handling connection and channel state&#45;change events
+
+* Ably's Realtime SDKs automatically handle disruptions such as network disconnections, so you don't need to do anything. The library emits "connection":/realtime/connection#connection-states and "channel":/realtime/channels-messages#channel-stateschannel state change events that you can listen to in case you'd like be informed about what it's doing (for example, if you want to know whenever the library has lost message continuity on a channel due to being disconnected for more than two minutes). But there's no need to take any particular action in response to any of these, the library will automatically reconnect and reattach to channels after the network disruption is over.
```
mattheworiordan: Well they do need to do things, just not in regards to manually reconnecting.  Be explicit, not vague please

mattheworiordan: I don't think you should bundle connection and channel events into one section.  They have different states, different events, and we should aim to explain this.  Perhaps provide a simple example in code for each (Javasceript is fine).  
`/realtime/channels-messages#channel-stateschannel` is not a validf link.

"But there's no need to take any particular action in response to any of these, the library will automatically reconnect and reattach to channels after the network disruption is over."  Well there is if it fails.  This is really not achieving the goal of this document I am afraid. We should be advising them that handling state change events may be useful to them because they need to reflect this state to their users, and also they need to handle failures such as connection failures (because of auth problems etc) or channel failures (permission issues etc).

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h4. 5. Using API key on the client side
+
+* Ably's "Basic Authentication":/general/authentication#basic-authentication method is the easiest way to authenticate your client on the Ably Realtime platform. However, it directly exposes the API key in the source code. When you use it on the client side, any entity that interacts with the client could view your API key and potentially misuse it. If you can run an auth server, we recommend using "Token Authentication":/general/authentication#token-authentication on the client side. "Read more":/general/authentication#selecting-auth about selecting an authentication mechanism.
+
+h4. 6. Ensuring tokens are auto&#45;renewable
```
mattheworiordan: We have lots of `&#45;`  thoughout, please remove

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+* Tokens are short lived and need to be constantly renewed. Rather than getting a token from your auth servers yourself and initializing the Client Library SDK with it, we recommend you instead initialize the SDK with an authUrl or authCallback. It will use this information to automatically request a token from your auth server's endpoint whenever it needs a new one, thus seamlessly switching to the new token without disconnecting, hence ensuring an uninterrupted service.
```
mattheworiordan: `authUrl` and `authCallback` should be in code format `@authUrl` and should be links to the docs.

mattheworiordan: I don't think this description is right.
We should really be saying "It is almost never correct to provide a token to a client library without ensuring the client library has a means to automatically get a new token automatically when it needs one (tokens expire as they are intentionally short lived). The Ably client library SDKs provide two mechanisms in the `ClientOptions` (linked) so that it can obtain tokens automatically, via `authUrl` and `authCallback`.  If you are using token authentication, ensure your client has a means to renew tokens by using `authUrl` or `authCallback`."  Perhaps then provide a simple example and links to appropriate documentation.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h4. 7. Choosing between Realtime/ REST libraries and Protocol Adapters 
+
+* Want to maintain a persistent connection with Ably's Realtime platform, so you can subscribe to messages sent on a channel, or be notified whenever the presence set changes? Use the "realtime library":/realtime.
+* Want to just publish data to the platform, don't care about subscribing, and don't want to manage channel attachment? Use the "REST library":/rest.
```
mattheworiordan: It's not that they don't care, it's that they don't need that feature.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+* Want to publish at very high volumes or with the lowest possible latency? "The realtime libary can publish at higher volumes (due to message pipelining) and lower latencies than is possible with the REST library":https://support.ably.io/solution/articles/3000060319-should-i-use-the-rest-or-realtime-library.
```
mattheworiordan: Typos.  I am not sure we've really quantified high volumes or low latencies.  I think in most cases REST is still the preferred route, but in some cases a realtime publish is preferred.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+* Want to get one&#45;time info about the presence set of a channel? Use the "REST library":/rest.
+* Want to set up realtime communication with IoT devices on platforms for which there's no native Ably client library available? Use the "MQTT protocol adapter":/mqtt.
+* Want to migrate from PubNub/Pusher? Use the relevant "protocol adapter":/adapters to "ease the pain of migration":https://support.ably.io/solution/articles/3000054127-can-you-help-ease-the-pain-of-migration-from-another-realtime-service: you can switch to Ably right away while sticking with your current client libraries, and gradually migrate to native Ably clients.
```
mattheworiordan: You've not mentioned subscribe to events on a server.  I realise it's mentioned below, but is this not about chosing the right library?

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+* Ably's "Basic Authentication":/general/authentication#basic-authentication method is the easiest way to authenticate your client on the Ably Realtime platform. However, it directly exposes the API key in the source code. When you use it on the client side, any entity that interacts with the client could view your API key and potentially misuse it. If you can run an auth server, we recommend using "Token Authentication":/general/authentication#token-authentication on the client side. "Read more":/general/authentication#selecting-auth about selecting an authentication mechanism.
+
+h4. 6. Ensuring tokens are auto&#45;renewable
+
+* Tokens are short lived and need to be constantly renewed. Rather than getting a token from your auth servers yourself and initializing the Client Library SDK with it, we recommend you instead initialize the SDK with an authUrl or authCallback. It will use this information to automatically request a token from your auth server's endpoint whenever it needs a new one, thus seamlessly switching to the new token without disconnecting, hence ensuring an uninterrupted service.
+
+h4. 7. Choosing between Realtime/ REST libraries and Protocol Adapters
```
mattheworiordan: Should this perhaps be: Are you using the correct client library SDK? Choosing between REST, Realtime, MQTT or others.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+* Want to get one&#45;time info about the presence set of a channel? Use the "REST library":/rest.
+* Want to set up realtime communication with IoT devices on platforms for which there's no native Ably client library available? Use the "MQTT protocol adapter":/mqtt.
+* Want to migrate from PubNub/Pusher? Use the relevant "protocol adapter":/adapters to "ease the pain of migration":https://support.ably.io/solution/articles/3000054127-can-you-help-ease-the-pain-of-migration-from-another-realtime-service: you can switch to Ably right away while sticking with your current client libraries, and gradually migrate to native Ably clients.
+
+h4. 8. Subscribing to events on the server&#45;side
+
+* Subscribing to events on the server side using the standard Pub/Sub pattern can be a bad idea for a number of reasons as explained in this "blog article":https://blog.ably.io/message-queues-the-right-way-to-process-and-work-with-realtime-data-on-your-servers-2d15985301f8. With Message Queues, you can set up multiple worker servers to distribute the load of the messages received from Ably while ensuring each of the messages gets processed exactly once by any one of your worker servers.
```
mattheworiordan: I don't like for a number of reasons. Be explicit about one or two of those problems.

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h4. 9. WebHooks/Reactor Functions vs. Message Queues &#45; when to use what?
```
mattheworiordan: When to use which?

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+* "WebHooks":/general/webhooks and "Reactor Functions":/general/functions work well for low or medium message volumes, but "Reactor Queues":/general/queues will scale better to high message volumes. For very high volumes, we recommend using "Reactor Firehose":/general/firehose.
```
mattheworiordan: How will a user know what's low, medium or high?

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h4. 10. Multiplexing channels
+
+* You should almost always instantiate a single Ably client library instance per web page or server. Don't forget that the Client Library SDK can "multiplex":https://support.ably.io/support/solutions/articles/3000058406-do-you-support-channel-groups-and-multiplexing many channels over a single connection.
```
mattheworiordan: surely device not server?

mattheworiordan: We haven't explained why they should do this and the benefit they get

content/root/best-practices-guide.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h4. 11. Limiting the effect of a security compromise using the principle of least privilege
+
+* Using the available "authentication methods":/general/authentication in Ably, you can choose the permissions or capabilities to be given to a client. You can do this by restricting the permissions on the API key as well as restricting the permissions your auth server grants to that API key whilst granting a token request. You can restrict the channels that a client can access (including with wildcards), as well as the operations the client can perform within those channels, using "capabilities":/general/authentication#capabilities-explained.
+* Always make sure to limit the capabilities given to a client to only what's required by the client to do their job. This prevents any unexpected behaviour by a client and increases the security of your application. This is called the principle of least privilege.
+
+h4. 12. Efficient use of channels
+
+Channels consume resources and hence cost you money. Make sure that you are using them efficiently and in accordance to the limit you have as per your package. However, bear in mind the following:
```
mattheworiordan: I think we should try and explain that whilst we recommend they use channels to distribute work more evenly across the cluster, there is an associated cost for a high number of channels, so common sense must prevail in regards to channel use. Here are some considerations that may help you with your channel usage design:

Srushtika:
Thanks for the review @mattheworiordan, I now more clearly understand the purpose of this guide, shall update it accordingly.

mattheworiordan:
Somter other best practice thoughts we should mention:

* Don't register listeners on a connection or channel state change event.  We often see customers do something like `on(connected) { channel.subscribe(...) }`.  The problem with this is every time the connection becomes connected it will add a new listener.  Calling subscribe more than once is not necessary as regardless of connection or channel state the listeners remain registered.
* Sending messages to an individual client should be done using a single channel unique for that client, and preferably with a capability configured in their token to ensure only that client can subscribe to messages on that channel.  
* Embedding a client ID in the token issued to a client ensures all published messages from that client contain their trusted client ID, and allows them to be present on channels with that client ID..

content/root/best-practices-guide.textile
Srushtika: I understand your concern, @mattheworiordan. However, this is explained in point 5 within 'Using Ably like a pro' section. This section was really meant for the readers to know broadly what all features Ably offers (so they can go understand it from the docs if they hadn't known about it before) whereas the actual best practices/ recommendations are provided in the next section. This is the same reason why I provided links within the previous points too. However, I've updated these points so they make sense when read as a standalone document too.

content/root/best-practices-guide.textile
Srushtika: I'm not sure I understand this comment @mattheworiordan

Srushtika:
@mattheworiordan FYI - this is not ready for review yet. I've pushed a subset of the changes so I can continue in a different system later. Will let you know when done.

Srushtika:
@mattheworiordan I've made changes to the doc as per your feedback. Awaiting your final review.

Thanks!

content/root/best-practice-guide.textile
```diff
@@ -0,0 +1,146 @@
...
+index: 30
+jump_to:
+  Contents:
+    - Ably 101
+    - Using Ably like a pro
+---
+
+h2(#title). Ably Best Practice Guide
+
+This best practice guide aims to quickly get you to understand the best way to make use of Ably's Realtime platform for your applications. These recommendations include some important considerations that are otherwise prone to easy ignoration while skimming through the docs.
```
mattheworiordan: Replace "quickly get you to" with "quickly help you to".

ignoration is an odd word IMHO, I believe it generally means the state of being ignorant, which is not what we mean and could be considered offensive.

Perhaps: "These recommendations include some important considerations that may otherwise be missed when reading through our documentation"

content/root/best-practice-guide.textile
```diff
@@ -0,0 +1,146 @@
...
+    - Ably 101
+    - Using Ably like a pro
+---
+
+h2(#title). Ably Best Practice Guide
+
+This best practice guide aims to quickly get you to understand the best way to make use of Ably's Realtime platform for your applications. These recommendations include some important considerations that are otherwise prone to easy ignoration while skimming through the docs. 
+
+This guide consists of the following two sections:
+* Ably 101 - Key concepts that you must understand before moving on to the best practice recommendations.
```
mattheworiordan: I believe in Textile you need an empty line here before starting a new block type (bullets), have you checked how this is rendered?

content/root/best-practice-guide.textile
```diff
@@ -0,0 +1,146 @@
...
+* Using Ably like a pro - Actual recommendations that we've compiled as per our experience with our users, in order for you to avoid common mistakes or misunderstandings whilst using Ably.
```
mattheworiordan: Why "Actual" for recommendations.  Actual makes it sound like other recommendations are not.

Perhaps: "Recommendations that we've compiled from our experience with developers to help you avoid common mistakes or misunderstandings"

mattheworiordan: Perhaps anchor link these to the sections?

content/root/best-practice-guide.textile
```diff
@@ -0,0 +1,146 @@
...
+
+Ably Realtime is a cloud-based platform that allows you to easily integrate realtime functionality into your applications. Just plug and play.
+
+h4(#pubsub). Publish and Subscribe (Pub/Sub)
+
+You can publish data to a named "channel":/realtime/channels-messages on the Ably Realtime platform, and all the clients subscribed to that very channel will receive this data in realtime, as it is being published.
+
+h4(#notifications). Notifications 
+
+You can use "Push Notifications":/realtime/push to send notifications to your clients even when they are currently not using your application, or in other words, not connected to Ably. You can do this on mobile devices as well as browsers. You can also use "Smart Notifications":/general/smart-notifications for your notifications to be most effective depending on your user's connection statte, their active device or even their current context within your application.
```
mattheworiordan: Type "statte".  I think replace "You can do this on mobile devices as well as browsers." with this instead which is clearer "You can publish native notifications to mobile devices on Android and iOS, and also deliver push notifications to browsers".  
Smart notifications are not ready, so you should say perhaps state "In the future you will also be able to use our "Smart Notifications"..."

content/root/best-practice-guide.textile
```diff
@@ -0,0 +1,146 @@
...
+
+h4(#presence). Presence
+
+You can have your clients attach to the presence channel using "Presence":/realtime/presence in order to indicate if these clients (or devices in case of IoT clients) are online or connected. The changes in this connection status, if any, can also be tracked by optionally having your clients subscribe to the same presence channel.
```
mattheworiordan: "You can have your clients attach to the presence channel using "Presence":/realtime/presence in order to indicate if these clients (or devices in case of IoT clients) are online or connected." is confusing.

Try:

"Your clients can attach to a presence channel using "Presence":/realtime/presence and announce they are present on a channel.  This is a useful for example in chat apps to indicate a member is online, or with IoT devices to indicate a device is online and connected. Clients can subscribe to realtime presence changes on channels, such as members entering or leaving, and respond to these events."

content/root/best-practice-guide.textile
```diff
@@ -0,0 +1,146 @@
...
+
+h4(#rest-or-realtime). Realtime vs REST
+
+With the availability of both the "Realtime":/realtime and "REST":/rest libraries, choose whichever suits your application's needs the best and get started. Using the realtime library, your clients can subscribe to certain data that Ably can then push to your clients as it is being created. On the other hand, the REST library offers simple stateless data publish or retrieval to and from Ably respectively.
```
mattheworiordan: I am not sure what "With the availability of both the "Realtime":/realtime and "REST":/rest libraries, choose whichever suits your application's needs the best and get started" means.

Is this not "Ably provides two types of client libraries. Realtime libraries maintain a persistent connection to Ably and can be considered stateful, where REST libraries issue HTTP requests to the Ably platform and are stateless".

I don't like this wording I am afraid: "Using the realtime library, your clients can subscribe to certain data that Ably can then push to your clients as it is being created." for the following reasons:

* Realtime not realtime, this is a reference to the library
* Why certain data?  Clients subscribe on channels.
* Data is not created, it is published?

content/root/best-practice-guide.textile
```diff
@@ -0,0 +1,146 @@
...
+
+h4(#auth). Choose the right auth scheme
+
+You can use "Basic Authentication":/general/authentication#basic-authentication to easily authenticate your clients using an API key or use "Token Authentication":/general/authentication#token-authentication to authenticate your clients using a token request sent to Ably via your auth server. Furthermore, you can use JWT authentication as part of Tokent Authentication ensuring that you use a single authentication strategy for all of your third party services. You can also "encrypt your messages":/realtime/encryption for added security of your application.
```
mattheworiordan: Tokent typo.  I don't think this explains why someone should use one auth scheme over another,  It does not explain how tokens work, and then it introduces encryption which I previously mentioned has nothing to do with authentication,.

Srushtika:
@mattheworiordan, apologies, I only just saw this. Sure, let's review this on the Friday when I'm back from the conference.

Srushtika:
@mattheworiordan @tomczoink all feedback implemented after @Affricacook finished her review. Good to merge?

PR #422, opened 2018-04-25
JWT Additions and Token Terminology Changes

tomczoink:
**This is not finished yet**

Currently I've added JWT throughout the rest/auth and general/auth pages, and will go through the realtime/auth and all other possible mentions of auth following feedback on this.

There's a few use-cases I'm uncertain on for JWT with Ably, for example can you use a JWT in tokenDetails, or can TokenParams be used with Ably JWT? If I've not stated the inclusion of JWT in a structure or function, or verbally excluded it by specifying it for Ably Tokens, please say.

As a separate point from JWT, when going through the docs, I also ended up struggling to interpret sentences due to the various ways words are used, for example "request token requests or request tokens". From a pure readability standpoint I think we should try to distinguish each type of word more. In this, I've tried doing the following:
- “Ably token” is all tokens acceptable with Ably (Ably Token, Ably JWT)
- “Ably Token” is the base token Ably produces
- “Ably Token Request” is the structure given to clients to request an Ably Token
- “Ably JWT” is a JWT formatted appropriately for Ably usage
- “token” is for referring to tokens not-specific to Ably

I'm happy for any and all of those to change, but I do think we need to consider using clearer wording for specific structures and types. The above example with this would now be "request Ably Token Requests or request Ably Tokens", which I feel is more readable.


paddybyers:
Initial comments without having read all the text yet.

Definitely the mantra here needs to be "every word has one meaning, and every meaning has one word".

> “token” is for referring to tokens not-specific to Ably
yes

> “Ably Token” is the base token Ably produces
yes - ie Ably "native" opaque token issued by Ably 

> “Ably JWT” is a JWT formatted appropriately for Ably usage
I think we've agreed that this is a JWT whose signing key is an Ably key. I don't think this term should include External JWTs that embed Ably-related claims.

Then I think we need to add:

"External JWT" is a JWT issued by some external scheme and whose signing key belongs to that external scheme.

An External JWT can embed an Ably Token or an Ably JWT.

> “Ably token” is all tokens acceptable with Ably (Ably Token, Ably JWT)

I think I'd prefer "Ably-compatible token" to mean any one of:
- Ably Token;
- Ably JWT;
- External JWT containing an embedded Ably Token or Ably JWT as a claim or header.

> “Ably Token Request” is the structure given to clients to request an Ably Token

I think “Ably TokenRequest” would be even better for that.

We still don't have terminology for the act of requesting a token; ie passing a TokenRequest to Ably and getting an Ably Token in response. We could say just "Ably Token request" for that, but I still don't like the potential ambiguity there, unless it's always explicitly a verb, eg "make an Ably Token request" or "perform an Ably Token request".

> this would now be "request Ably Token Requests or request Ably Tokens", which I feel is more readable.

Here it's confusing because the word "request" is overloaded and since it's being used here not as a defined term but it is normal meaning then we should try to use a synonym, or make it obviously from the context that the usage of the word is not in a specially defined way, eg by referring to the "authURL request" or even "authURL query".

> There's a few use-cases I'm uncertain on for JWT with Ably, for example can you use a JWT in tokenDetails

Not according to the spec, no. A `TokenDetails` contains details of an Ably Token.

> can TokenParams be used with Ably JWT?

No  `TokenParams` are things you use to create a `TokenRequest`, and that in turn is used to make an Ably Token request; you don't do that if you're creating an Ably JWT. Instead, you just use a JWT library, and you don't need Ably to issue the token.

content/general/authentication.textile
```diff
@@ -18,7 +18,7 @@ jump_to:
...
+Both the "REST client library":/rest/authentication and the "Realtime client library":/realtime/authentication use common authentication mechanisms. The two schemes supported by Ably are "Basic Authentication":#basic-authentication, which uses your private "API key":https://support.ably.io/solution/articles/3000030054-what-is-an-app-api-key, and "Token Authentication":#token-authentication, which relies on an "Ably Token":#tokens you request from Ably, an "Ably Token Request":/realtime/types#token-request that you sign and issue from your servers to your clients, or an "Ably JWT":#ably-jwt signed with your private API key. "Token Authentication":#token-authentication, in most cases, is the recommended strategy on the client-side as it provides robust access control and stringent security measures.
```
paddybyers: Lets split this up. (Re-add links etc.)

> .... 
> The two schemes supported by Ably are "Basic Authentication":#basic-authentication, which uses your private "API key":https://support.ably.io/solution/articles/3000030054-what-is-an-app-api-key, and "Token Authentication":#token-authentication, which uses short-lived tokens for access which are periodically renewed. Different token-issuing mechanisms can be used with Ably; the default is to use Ably Tokens which you request from Ably based on an Ably TokenRequest that you sign an issue from your servers to clients; or a Json Web Token (JWT) which you generate on your servers using your private API key. Token Authentication, in most cases, is the recommended strategy on the client-side as it provides more fine-grained access control and limits the risk of exposed or compromised credentials.

content/general/authentication.textile
```diff
@@ -84,26 +84,26 @@ Whilst Basic Authentication is simple, we recommend it to be only used on the se
...
+Token authentication is the recommended authentication scheme for client devices as tokens are short-lived and may more readily be distributed to clients where there is a risk of compromise. Ably-compatible tokens may also be issued with a particular scope - such as a limited set of "access rights or capabilities":#capabilities-explained or be limited to a "specific identity (<span lang="default">@clientId@</span><span lang="csharp">@ClientId@</span>)":#identified-clients. Ably-compatible tokens are most commonly issued server-side, although the type of Ably-compatible token issued varies. At present, Ably supports "Ably Tokens":#token-process, "Ably TokenRequests":#token-request-process, "JWT with an embedded Ably Token":#jwt-embed-process, and "Ably JWT":#ably-jwt-process.
```
paddybyers: I don't understand why "Ably TokenRequests" is mentioned in the final sentence here. The other listed things are kinds of token.

content/general/authentication.textile
```diff
@@ -84,26 +84,26 @@ Whilst Basic Authentication is simple, we recommend it to be only used on the se
...
-* an "<span lang="default">@authUrl@</span><span lang="csharp">@AuthUrl@</span>":/realtime/types#client-options or "<span lang="default">@authCallback@</span><span lang="csharp">@AuthCallback@</span>":/realtime/types#client-options is provided that returns a token or token request;
+* an "<span lang="default">@authUrl@</span><span lang="csharp">@AuthUrl@</span>":/realtime/types#client-options or "<span lang="default">@authCallback@</span><span lang="csharp">@AuthCallback@</span>":/realtime/types#client-options is provided that returns an Ably-compatible token or Ably TokenRequest;
 * a "<span lang="default">@clientId@</span><span lang="csharp">@ClientId@</span>":/realtime/types#client-options is provided;
 * "<span lang="default">@useTokenAuth@</span><span lang="csharp">@UseTokenAuth@</span>":/realtime/types#client-options is true
 
-Token authentication is typically done in one of two ways:
+Token authentication is typically done in one of four ways:
 
-h4. Signed token request is created by your servers and passed to clients
+h4(#token-request-process). Signed Ably TokenRequest is created by your servers and passed to clients
```
paddybyers: I don't think we should use the term "signed TokenRequest", because it seems to suggest that there might be a kind of TokenRequest that's not signed. I would prefer that we say that a TokenRequest is signed where we define that term, and then elsewhere we just refer to a TokenRequest.

content/general/authentication.textile
```diff
@@ -139,22 +139,91 @@ ably.requestToken({ clientId: 'client@example.com' }, function(err, token) {
...
 
-API keys and their capabilities are "configured using the dashboard":https://support.ably.io/support/solutions/articles/3000030502-setting-up-and-managing-api-keys, they cannot be added or removed programmatically. Tokens on the other hand are designed to be shared with un-trusted clients, are short-lived, and significantly, they are configured and issued programmatically using the Ably client libraries or directly from the "Ably REST API":/rest-api. See "selecting an authentication scheme":#selecting-auth to understand why token authentication, in most cases, is the preferred authentication scheme.
+This is similar to signing an "Ably TokenRequest":/realtime/authentication#request-token, but the client does not need to then request an Ably Token, instead being able to use the Ably JWT as a token in itself. Any third-party library may be used to create the JWT without a connection ever needing to be established with Ably. This can be useful for situations where an Ably client is not available, such as an embedded MQTT client.
```
paddybyers: ".. to then request ..." -> " ... then to request ... ".

Any compliant third-party JWT library may be used to create the JWT without requiring the token to be issued by Ably. This can be useful for situations where an Ably client library is not available, such as an embedded device connecting to Ably via MQTT.

content/general/authentication.textile
```diff
@@ -139,22 +139,91 @@ ably.requestToken({ clientId: 'client@example.com' }, function(err, token) {
...
+  var base64Claims = btoa(claims);
+  /* Apply the hash specified in the header */
+  var signature = hash((base64Header + "." + base64Claims), SECRET);
+  var ablyJwt = base64Header + "." + base64Claims + "." + signature;
+
+*Note:* At present Ably does not support asymmetric signatures based on a keypair belonging to a third party. If this is something you'd be interested in, please "get in touch":https://www.ably.io/contact.
+
+h4(#jwt-embed-process). Ably-compatible token is embedded in a External JWT from your server and passed to clients
+
+If a system has an existing "JWT":https://jwt.io/ scheme, it can be useful to embed an Ably-compatible token into it. To do this, the External JWT and need to ensure it meets the following requirements:
```
paddybyers: "embed an Ably-compatible token into it" -> "embed an Ably-compatible token as a claim within it"

content/general/authentication.textile
```diff
@@ -139,22 +139,91 @@ ably.requestToken({ clientId: 'client@example.com' }, function(err, token) {
...
+
+* The Ably-compatible token is an "Ably JWT":#ably-jwt-process, an "Ably Token":#token-process, or an "Ably TokenRequest":#token-request-process
```
paddybyers: No, a TokenRequest cannot be embedded in a JWT.

content/general/authentication.textile
```diff
@@ -139,22 +139,91 @@ ably.requestToken({ clientId: 'client@example.com' }, function(err, token) {
...
+
+"Ably Tokens":#tokens are issued from an existing "API key":https://support.ably.io/solution/articles/3000030054-what-is-an-app-api-key, and their capabilities can, at most, match the capabilities of the issuing API key. "Ably JWTs":#ably-jwt have a similar restriction on capabilities, correlating to the API key they've been signed with. If an API key must be shared with a third party, then it is recommended that "the principle of least privilege":http://en.wikipedia.org/wiki/Principle_of_least_privilege is considered, assigning only the capabilities needed by that third party. Thus, any Ably requests authenticated using that API key or Ably-compatible tokens associated with that API key, will be restricted to the capabilities assigned to the API key.
+
+Capabilities when requesting Ably Tokens are determined as follows:
 
+* If no capability is specified in the Ably Token request, then the Ably Token will be given the full set of capabilities assigned to the issuing key, "see example":#capabilities-explained-all;
```
paddybyers: I think this should say "specified in the Ably TokenRequest".

Grammatically: I think this is a sentence. It starts with "If ..." and so should end with a full stop. I realise there's a global style issue here, but it either has to be "if .... ;" or "If ... .".

Also I don't like the comma after "issuing key". You can't use a comma to join two sentences. I think it should be a semicolon.

content/general/authentication.textile
```diff
@@ -139,22 +139,91 @@ ably.requestToken({ clientId: 'client@example.com' }, function(err, token) {
...
+* If a set of capabilities are requested, then the Ably Token will be assigned the intersection of the requested capability and the capability of the issuing key, "see example":#capabilities-explained-intersection;
+* If a set of capabilities are requested, and the intersection between those and the API key's capabilities is empty (i.e. they are entirely incompatible), then the Ably Token request will result in an error, "see example":#capabilities-explained-error.
```
paddybyers: While we're on the subject of grammar I think we should be using open punctuation, so for example we would have "ie" instead of "i.e.".

See for example https://www.cam.ac.uk/brand-resources/guidelines/editorial-style-guide for a standard.

content/rest/authentication.textile
```diff
@@ -29,7 +29,7 @@ jump_to:
...
+Ably clients can authenticate with Ably in one of two ways. They can use "Basic Authentication":/general/authentication#basic-authentication or "Token Authentication":/general/authentication#token-authentication. Basic Authentication makes use of the customer's "API key":https://support.ably.io/solution/articles/3000030054-what-is-an-app-api-key) to connect with Ably. Token Authentication requires a server to provide an "Ably Token":/general/authentication#tokens, an "Ably TokenRequest":#request-token, an "Ably JWT (JSON Web Token)":/general/authentication#ably-jwt, or an "External JWT":https://jwt.io with an embedded Ably--compatible token to the client. "Token Authentication":/general/authentication#token-authentication, in most cases, is the recommended strategy due to it providing robust access control and stringent security measures.
```
paddybyers: I've not gone through this in detail but obviously some of the changes in the general doc need to be reflected here.

content/rest/authentication.textile
```diff
@@ -130,26 +134,38 @@ Below is a rather contrived yet straightforward example that instances a REST li
...
+
+There are a couple of Ably-compatible tokens that you can have your server return to a client:
+
+h4(#ably-token). Ably Token
+
+An Ably Token is obtained through a "request":#request-token to Ably. This can either be done by the server, which then passes on the resulting Ably Token to the respective client, or by the client making use of an "Ably TokenRequest":#request-token. This token can also be embedded into an "External JWT":https://jwt.io as the @x-ably-token@ claim in the JWT's payload.
+
+h4(#ably-jwt). Ably JWT (JSON Web Token)
+
+"JWT (JSON Web Token)":https://jwt.io/ is an open standard (RFC 7519) that allows for compact secure information as JSON objects. Ably supports a JWT when the JWT is signed with an Ably API key. This type of JWT is called an "Ably JWT":/general/authentication#ably-jwt. A server can generate a JWT, and pass it to a client to be used. Unlike an Ably TokenRequest, it can be used as an actual token by the client, and the server can generate this token without the Ably library. Any third party software can be used to generate this JWT using the Ably private API key to sign it.
```
paddybyers: "it can be used as an actual token by the client" -> "it can be used directly as a token by the client"

paddybyers:
Having read the edits I'm not sure I'm happy with "Ably-compatible token". I think there are plenty of places where the context makes it clear that you're talking about a token that relates to Ably, so it would be possible to simply say "token".

tomczoink:
@paddybyers I've added JWT references to all the auth pages, along with pages such as 'usage', in addition to attempting to apply the new token terminology across the site. If you could see if my general usage of token terminology is correct (notably in the 'Rest-API' pages), it'd be much appreciated.

content/realtime/authentication.textile
```diff
@@ -117,15 +121,27 @@ Below is a rather contrived yet straightforward example that instances a Realtim
 
 h4(#token-upgrading). Upgrading the token on a connection
 
-The token that a connection uses can be upgraded whilst remaining connected. When a token is nearly expired, Ably will prompt the library to request a new token (using the <span lang="ruby,python">@auth_url@ or @auth_callback@</span><span lang="default">@authUrl@ or @authCallback@</span>). Once the library has obtained a new token, it will upgrade the connection to use the new token. This all happens automatically and seamlessly.
+The token that a connection uses can be upgraded whilst remaining connected. When a token is nearly expired, Ably will prompt the library to request a new token (using the <span lang="ruby,python">@auth_url@ or @auth_callback@</span><span lang="default">@authUrl@ or @authCallback@</span>). Once the library has obtained a new Ably-compatible token, it will upgrade the connection to use the new token. This all happens automatically and seamlessly.
```
paddybyers: I think this (final sentence) is example where simply saying "token" is quite clear, and "Ably-compatible" seems to be suggesting that there's something special about the token it's referring to whereas (in this context, where everything is Ably-compatible) there isn't.

content/realtime/authentication.textile
```diff
@@ -117,15 +121,27 @@ Below is a rather contrived yet straightforward example that instances a Realtim
...
 
-A subsidiary use-case for the @Auth@ object is to preemptively trigger renewal of a token or to acquire a new token with a revised set of capabilities by explicitly calling "<span lang="default">@authorize@</span><span lang="csharp">@Authorize@</span>":#authorize.
+A subsidiary use-case for the @Auth@ object is to preemptively trigger renewal of an Ably Token or to acquire a new Ably Token with a revised set of capabilities by explicitly calling "<span lang="default">@authorize@</span><span lang="csharp">@Authorize@</span>":#authorize.
```
paddybyers: Token renewal can apply to any token type. Again, I would just remove the "Ably" qualiifer.

content/realtime/authentication.textile
```diff
@@ -244,13 +259,13 @@ bq(definition).
...
   swift,objc: authorize(tokenParams: "ARTTokenParams":#token-params?, authOptions: "ARTAuthOptions":#auth-options?, callback: ("ARTTokenDetails":#token-details?, NSError?) -> Void)
 
-Instructs the library to get a new token immediately. Once fetched, it will upgrade the current realtime connection to use the new token, or if not connected, will initiate a connection to Ably once the new token has been obtained. Also stores any <span lang="ruby">@token_params@ and @auth_options@</span><span lang="default">@tokenParams@ and @authOptions@</span> passed in as the new defaults, to be used for all subsequent implicit or explicit token requests.
+Instructs the library to get a new Ably Token immediately. Once fetched, it will upgrade the current realtime connection to use the new token, or if not connected, will initiate a connection to Ably once the new token has been obtained. Also stores any <span lang="ruby">@token_params@ and @auth_options@</span><span lang="default">@tokenParams@ and @authOptions@</span> passed in as the new defaults, to be used for all subsequent implicit or explicit Ably Token requests.
```
paddybyers: This and much of the discussion that follows (relating to authCallback and renewal) applies to all token types.

content/rest-api/token-request-spec.textile
```diff
@@ -146,13 +146,13 @@ bc[json]. {
...
 
-h3. Signed token request example
+h3. Signed Ably TokenRequest example
 
 bc[sh]. curl -X POST "https://rest.ably.io/keys/{{API_KEY_NAME}}/requestToken" \
  -H "Content-Type: application/json" \
  --data '{{SIGNED_TOKEN_REQUEST_EXAMPLE}}'
 
-Responds with JSON token:
+Responds with JSON Ably Token:
```
paddybyers: Should this be `TokenDetails` ?

content/general/authentication.textile
```diff
@@ -84,26 +86,26 @@ Whilst Basic Authentication is simple, we recommend it to be only used on the se
...
 
 Token authentication is the default authentication scheme when a client library is instanced with any of the following options:
 
 * a "<span lang="default">@token@</span><span lang="csharp">@Token@</span>":/realtime/types#client-options or "<span lang="default">@tokenDetails@</span><span lang="csharp">@TokenDetails@</span>":/realtime/types#client-options property is provided;
-* an "<span lang="default">@authUrl@</span><span lang="csharp">@AuthUrl@</span>":/realtime/types#client-options or "<span lang="default">@authCallback@</span><span lang="csharp">@AuthCallback@</span>":/realtime/types#client-options is provided that returns a token or token request;
+* an "<span lang="default">@authUrl@</span><span lang="csharp">@AuthUrl@</span>":/realtime/types#client-options or "<span lang="default">@authCallback@</span><span lang="csharp">@AuthCallback@</span>":/realtime/types#client-options is provided that returns an Ably-compatible token or Ably TokenRequest;
```
mattheworiordan: This should be `.. token or Ably "@TokenRquest@":/realtime/types#token-request` (check the URL, just guessed that)

content/general/authentication.textile
```diff
@@ -84,26 +86,26 @@ Whilst Basic Authentication is simple, we recommend it to be only used on the se
...
 * a "<span lang="default">@clientId@</span><span lang="csharp">@ClientId@</span>":/realtime/types#client-options is provided;
 * "<span lang="default">@useTokenAuth@</span><span lang="csharp">@UseTokenAuth@</span>":/realtime/types#client-options is true
 
-Token authentication is typically done in one of two ways:
+Token authentication is typically done in one of four ways:
 
-h4. Signed token request is created by your servers and passed to clients
+h4(#token-request-process). Ably TokenRequest is created by your servers and passed to clients
```
mattheworiordan: When referring to Ably objects, always hyperlink them please and put them in `code style`

content/general/authentication.textile
```diff
@@ -84,26 +86,26 @@ Whilst Basic Authentication is simple, we recommend it to be only used on the se
...
+Using our REST or Realtime client libraries, a "Ably TokenRequest is generated from your servers":/realtime/authentication#create-token-request and handed to the client-side client library. The client-side client library then uses that "Ably TokenRequest":/realtime/types#token-request to "request an Ably Token":/realtime/authentication#request-token from Ably and subsequently authenticate using that Ably Token. This is the recommended approach for authentication as: a "Ably TokenRequest":/realtime/types#token-request can be generated securely by your servers without communicating with Ably; your secret API key is never shared with Ably or your clients. Ably TokenRequests cannot be tampered with due to being signed, must be used soon after creation and can only be used once. This process is depicted in the following diagram:
```
mattheworiordan: Should be "an Ably TokenRequest", not a Ably TokenRequest. As above, should be code styled, I won't mention that again and will assume you will change everywhere.  Also, might be worth checking with @paddybyers, but these are signed TokenRequests, which I realise one could assume all token requests are, but they are not.  So I think we should be consistent with that in that a signed token request can be trusted by the Ably platform from an untrusted and unauthenticated client.

content/general/authentication.textile
```diff
@@ -139,22 +141,91 @@ ably.requestToken({ clientId: 'client@example.com' }, function(err, token) {
      } */
 });
 
-h3(#capabilities-explained). Capabilities and Token Security explained
+h4(#ably-jwt-process). Ably JWT is created by your servers and passed to clients
 
-"API keys":https://support.ably.io/solution/articles/3000030054-what-is-an-app-api-key, like "tokens":#tokens, have a set of capabilities assigned to them that specify which "operations":#capability-operations (such as subscribe or publish) can be performed on which channels. However, unlike tokens, API keys are long-lived, secret and typically not shared with un-trusted clients.
+It is possible to use a "JWT":https://jwt.io as a form of token for authentication with Ably, so long as it is structured appropriately, in what will be referred to as an *"Ably JWT":#ably-jwt*. t is possible for an Ably JWT to contain claims indicating its clientId, capabilities and expiry - in an analogous way to an "Ably Token":#tokens - and it is signed with an applicable Ably API key.
```
mattheworiordan: Missing `I` in ". t is possible for an.."
clientId, capabilities, expiry are attribnutes, so style them as code please, i.e. `@clientId@, @capabilities@...`
Why an applicable Ably API key, why not just an Ably API key?

content/general/authentication.textile
```diff
@@ -139,22 +141,91 @@ ably.requestToken({ clientId: 'client@example.com' }, function(err, token) {
...
+  var base64Claims = btoa(claims);
+  /* Apply the hash specified in the header */
+  var signature = hash((base64Header + "." + base64Claims), SECRET);
+  var ablyJwt = base64Header + "." + base64Claims + "." + signature;
+
+*Note:* At present Ably does not support asymmetric signatures based on a keypair belonging to a third party. If this is something you'd be interested in, please "get in touch":https://www.ably.io/contact.
+
+h4(#jwt-embed-process). Ably-compatible token is embedded in a External JWT from your server and passed to clients
+
+If a system has an existing "JWT":https://jwt.io/ scheme, it can be useful to embed an Ably-compatible token as a claim within it. To do this, the External JWT and need to ensure it meets the following requirements:
```
mattheworiordan: Sentence "To do this, the External JWT and need to ensure it meets the following requirements" is missing something?

content/general/authentication.textile
```diff
@@ -139,22 +141,91 @@ ably.requestToken({ clientId: 'client@example.com' }, function(err, token) {
...
+
+* The Ably-compatible token is an "Ably JWT":#ably-jwt-process, or an "Ably Token":#token-process
+* The Ably-compatible token is included under the @x-ably-token@ key in the "JOSE Header":https://tools.ietf.org/html/rfc7519
+* OR (if using JWS) the Ably-compatible token is included using the @x-ably-token@ Claim in the payload
+* The expiry time of the Ably-compatible token must not be earlier than the outer JWT’s expiry time (@exp@ Claim). Ably will reject any JWT if it is unencrypted and its @exp@ Claim is later than the expiry of the enclosed Ably-compatible token. This helps to ensure that tokens are renewed prior to expiry
+
+<a href="/images/diagrams/jwt-embed-auth.png" target="_blank">
+  <img src="/images/diagrams/jwt-embed-auth.png" style="width: 100%" alt="Token embedded in a JWT auth method">
```
mattheworiordan: If we had to choose one of the two ways presented to users, I am not sure we should really encourage people to get Ably tokens on their servers and embed them in the JWT.  Would they not be better embedding an Ably JWT token in their token as a) they have the JWT library already, b) it saves an unnecessary request to our servers from theirs.  @paddybyers @tomczoink WDYT?

mattheworiordan: I realise too many options is a bad thing, but could we consider explaining the pros & cons of each at least?

content/general/authentication.textile
```diff
@@ -308,10 +352,30 @@ The following Javascript example demonstrates how to issue a token with an expli
...
+
+An Ably JWT is not strictly an Ably construct, rather it is a "JWT":https://jwt.io/ which has been constructed to be compatible with Ably. The JWT must adhere to the following to ensure compatability:
+
+* *The JOSE header must include:*
+** @kid@ - Key name, such that an API key of @{{API_KEY}}@ will have key name @{{API_KEY_NAME}}@
+* *The JWT claim set must include:*
+** @iat@ - time of issue in seconds
+** @exp@ - expiry time in seconds
+* *The JWT claim set may include:*
+** @x-ably-capability@ - JSON text encoding of the "capability":https://www.ably.io/documentation/general/authentication#tokens
```
mattheworiordan: Does it matter what style of encoding is used i.e. with / without whitespace?  I assume not given we're not constructing doing a checksum in the same way as we do for our tokens.

content/types/_rest_client_options.textile
```diff
@@ -1,3 +1,3 @@
...
 
-- <span lang="default">defaultTokenParams</span><span lang="python">default_token_params</span><span lang="ruby">:default_token_params</span><span lang="csharp">DefaultTokenParams</span> := When a "TokenParams":/rest/types#token-params object is provided, it will override the client library defaults when issuing new tokens or token requests<br>__Type: "@TokenParams@":/rest/types#token-params__
+- <span lang="default">defaultTokenParams</span><span lang="python">default_token_params</span><span lang="ruby">:default_token_params</span><span lang="csharp">DefaultTokenParams</span> := When a "TokenParams":/rest/types#token-params object is provided, it will override the client library defaults when issuing new Ably Tokens or Ably TokenRequests<br>__Type: "@TokenParams@":/rest/types#token-params__
```
mattheworiordan: I think generally where we refer to Ably Tokens and Ably TokenRequests, we should take that opportunity to link to suitable documentation.  I don't think we should expect developers to know where to find information on  terms / keywords / acronyms etc.

content/general/authentication.textile
paddybyers: The library canonicalises the JSON in any case so the definition should be the same.

tomczoink:
@paddybyers @mattheworiordan @Srushtika any further changes wanted on the current state of this?

PR #423, opened 2018-04-26
Spec should describe emitting ConnectionEvent and not ConnectionState

withakay:
RTN4 tells us a ConnectionEvent should be emitted but RTN4a and RTN4h both mentioned emitting ConnectionState, this PR fixes that.

PR #424, opened 2018-04-26
Spec should describe emitting ConnectionEvent not ConnectionState

withakay:
RTN4 tells us a ConnectionEvent should be emitted but RTN4a and RTN4h both mentioned emitting ConnectionState, this PR fixes that.

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -320,8 +320,8 @@ h3(#realtime-connection). Connection
 ** @(RTN2g)@ Library and version param @lib@ should include the header value described in "RSC7b":#RSC7b. For example, the 1.0.0 version of the Javascript library would use the param @lib=js-1.0.0@
 * @(RTN3)@ If connection option @autoConnect@ is true, a connection is initiated immediately; otherwise a connection is only initiated following an explicit call to @connect()@
 * @(RTN4)@ The @Connection@ implements @EventEmitter@ and emits @ConnectionEvent@ events
-** @(RTN4a)@ It emits a @ConnectionState@ event (one of @INITIALIZED@, @CONNECTING@, @CONNECTED@, @DISCONNECTED@, @SUSPENDED@, @CLOSING@, @CLOSED@, @FAILED@) for every connection state change
-** @(RTN4h)@ It can emit an @UPDATE@ event (this is the only @ConnectionEvent@ which is not a @ConnectionState@). This is used for changes to connection conditions for which the @ConnectionState@ (e.g. @CONNECTED@) does not change. (The library must never emit a @ConnectionState@ event for a state equal to the previous state).
+** @(RTN4a)@ It emits a @ConnectionEvent@ event (one of @INITIALIZED@, @CONNECTING@, @CONNECTED@, @DISCONNECTED@, @SUSPENDED@, @CLOSING@, @CLOSED@, @FAILED@) for every connection state change
+** @(RTN4h)@ It can emit an @UPDATE@ event (this is the only @ConnectionEvent@ which is not a @ConnectionState@). This is used for changes to connection conditions for which the @ConnectionState@ (e.g. @CONNECTED@) does not change. (The library must never emit a @ConnectionEvent@ event for a state equal to the previous state).
 ** @(RTN4b)@ A new connection will emit the following events in order when connecting: @CONNECTING@, then @CONNECTED@
 ** @(RTN4c)@ A connection will emit the following events when closing the connection: @CLOSING@, then @CLOSED@
 ** @(RTN4d)@ @Connection#state@ attribute is the current state of the connection


withakay:
@funkyboy this replaces #423

content/client-lib-development-guide/features.textile
```diff
@@ -320,8 +320,8 @@ h3(#realtime-connection). Connection
...
 * @(RTN4)@ The @Connection@ implements @EventEmitter@ and emits @ConnectionEvent@ events
-** @(RTN4a)@ It emits a @ConnectionState@ event (one of @INITIALIZED@, @CONNECTING@, @CONNECTED@, @DISCONNECTED@, @SUSPENDED@, @CLOSING@, @CLOSED@, @FAILED@) for every connection state change
-** @(RTN4h)@ It can emit an @UPDATE@ event (this is the only @ConnectionEvent@ which is not a @ConnectionState@). This is used for changes to connection conditions for which the @ConnectionState@ (e.g. @CONNECTED@) does not change. (The library must never emit a @ConnectionState@ event for a state equal to the previous state).
+** @(RTN4a)@ It emits a @ConnectionEvent@ event (one of @INITIALIZED@, @CONNECTING@, @CONNECTED@, @DISCONNECTED@, @SUSPENDED@, @CLOSING@, @CLOSED@, @FAILED@) for every connection state change
+** @(RTN4h)@ It can emit an @UPDATE@ event (this is the only @ConnectionEvent@ which is not a @ConnectionState@). This is used for changes to connection conditions for which the @ConnectionState@ (e.g. @CONNECTED@) does not change. (The library must never emit a @ConnectionEvent@ event for a state equal to the previous state).
```
SimonWoolf: > (The library must never emit a @ConnectionEvent@ event for a state equal to the previous state).

But this isn't true. The whole point of RTN4h is that the library *­can­* emit a `ConnectionEvent` when the connection state hasn't changed: the `update` event -- but that is only possible for the one `ConnectionEvent` that is not also a `ConnectionState`; the library will never emit a `ConnectionState` event for a state equal to the previous state. With this change the last two sentences of RTN4h contradicts each other.

withakay: I think it is a little ambiguous, my initial understanding was that `a @ConnectionEvent@ event for a state` would apply to a situation where the state and event types match e.g. `ConnectionEvent.CONNECTED == ConnectionState.CONNECTED`, but I suppose it could be said that a `ConnectionEvent.UPDATE` is emitted *for* a given ConnectionState too.

Arguably, the last sentence of RTN4h is superfluous and could simply be removed, so that it read: 

`
@(RTN4h)@ It can emit an @UPDATE@ event (this is the only @ConnectionEvent@ which is not a @ConnectionState@). This is used for changes to connection conditions for which the @ConnectionState@ (e.g. @CONNECTED@) does not change.
`

SimonWoolf: I see what you're saying, but with this change it's now quite confusing IMO.

I guess it's technically redundant to the word 'change' at the end of RTN4a, but I think it's still useful to have it be explicit. Could move it from RTN4h to RTN4a I guess, but why?

I kindof object to the other change too -- `It emits a @ConnectionEvent@ event (one of @INITIALIZED@, @CONNECTING@, @CONNECTED@, @DISCONNECTED@, @SUSPENDED@, @CLOSING@, @CLOSED@, @FAILED@)`. That sounds like it's defining ConnectionEvent to be one of those 8 things. But that's not correct (or at least, not complete) -- it's the definition of `ConnectionState`; `ConnectionEvent` is one of 9 things (or depending on implementation, one of two things, `ConnectionStateEvent ConnectionState | Update`)

SimonWoolf: How about something like the following:

`* @(RTN4)@ The @Connection@ implements @EventEmitter@ and emits @ConnectionEvent@ events, where a @ConnectionEvent@ is either a @ConnectionState@ or @UPDATE@, and a @ConnectionState@ is either @INITIALIZED@, @CONNECTING@, @CONNECTED@, @DISCONNECTED@, @SUSPENDED@, @CLOSING@, @CLOSED@, or @FAILED@`
`** @(RTN4a)@ It emits a @ConnectionState@ @ConnectionEvent@ for every connection state change`
`** @(RTN4h)@ It emits an @UPDATE@ @ConnectionEvent@ for changes to connection conditions for which the @ConnectionState@ (e.g. @CONNECTED@) does not change. (The library must never emit a @ConnectionState@ @ConnectionEvent@ for a state equal to the previous state)`

WDYT?

withakay: @SimonWoolf that is better.

SimonWoolf: @paddybyers / @funkyboy ^ opinions?

paddybyers: sgtm

PR #432, opened 2018-05-10
Clarified that JWT times are in seconds

tomczoink:
For https://github.com/ably/website/issues/1504

### DIFF
--- a/content/general/authentication.textile
+++ b/content/general/authentication.textile
@@ -159,8 +159,8 @@ bc[javascript]. var header = {
     "kid": "{{API_KEY_NAME}}"
   }
   var claims = {
-    "iat": currentTime,
-    "exp": currentTime + 3600,
+    "iat": currentTime, /* current time in seconds */
+    "exp": currentTime + 3600, /* time of expiration in seconds */
     "x-ably-capability": "{\"*\":[\"*\"]}"
   }
   var base64Header = btoa(header);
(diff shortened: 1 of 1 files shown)


content/general/authentication.textile
```diff
@@ -159,8 +159,8 @@ bc[javascript]. var header = {
     "kid": "{{API_KEY_NAME}}"
   }
   var claims = {
-    "iat": currentTime,
-    "exp": currentTime + 3600,
+    "iat": currentTime, /* current time in seconds */
+    "exp": currentTime + 3600, /* time of expiration in seconds */
```
mattheworiordan: worth adding `var currentTime = Math.round(Date.now()/1000);`?

mattheworiordan: Also, I wonder if we should explicitly change "An example of creating an Ably JWT can be seen below:" to "An example of creating an Ably JWT manually can be seen below. In most scenarios however, we would recommend you use one of the [many JWT libraries available for simplicity](https://jwt.io/)".  We could also add the link to the text "Any compliant third-party JWT library", and update the text below to reflect these changes.

mattheworiordan: Perhaps finally in https://www.ably.io/documentation/general/authentication#ably-jwt, should we add a note to say we recommend the use of a JWT compliant library as well?

tomczoink:
Updated per feedback @mattheworiordan.

Issue #429, opened 2018-05-04
RTN15h: token errors don't cause a connection to fail


paddybyers:
RTN15h says:

>  if the token is renewable, a single attempt to create a new token should be made and a new connection attempt initiated using the new token. If the token creation fails, the next connection attempt fails due to a token error, or the token was not renewable, the connection will transition to the FAILED state ...

It think it is not correct that the connection fails in the case that the token is renewable; this is also inconsistent with RTN14b.

I think this needs to cover two cases separately:

- if the token is not renewable, then the connection transitions to `FAILED`;

- if the token is renewable, then the client immediately attempts to obtain a new token and reconnect; if that attempt fails (either the token renewal fails with a recoverable error, or the connection attempt fails with a token error) then the connection transitions to `DISCONNECTED` and may retry after the appropriate retry period.

PR #258, opened 2017-01-20
API 1.1 ongoing review PR

mattheworiordan:
We will use this PR to track ongoing proposed changes for the 1.1 specification.  Commits merged in are generally accepted to be included in 1.1

PR #439, opened 2018-06-05
RTN15h: token errors, when the library can renew, result in DISCONNEC…

paddybyers:
Fixes: https://github.com/ably/docs/issues/429


SimonWoolf:
> integration-spec-1-1

(btw, we already have `master-1-1` acting as an integration branch for the 1.1 spec, why do we have `integration-spec-1-1` too? neither is a parent of the other)

paddybyers:
> (btw, we already have master-1-1 acting as an integration branch for the 1.1 spec, why do we have integration-spec-1-1 too? neither is a parent of the other)

`integration-spec-1-1` was created as the authoritative integration branch because `master-1-1` had things in it that weren't actually spec changes (eg https://github.com/ably/docs/commit/b276d760702f1e3db01fca2311d0f732e260828e). I do remember a discussion about that with Matt at the time but I can't remember if there was also a plan to merge those non-spec commits.

However, we should tidy things up and decide if there's anything in there that's got lost and needs to be dealt with.

mattheworiordan:
> integration-spec-1-1 was created as the authoritative integration branch because master-1-1 had things in it that weren't actually spec changes (eg b276d76).

@paddybyers the only reason it has non spec 1.1 commits is because master-1-1 has not been rebased, see https://github.com/ably/docs/commit/2d11e3eee595d684f8028d196660a0b7dee50c6f#diff-04c6e90faac2675aa89e2176d2eec7d8 which is on master.

I think we should rebase and not have two separate branches for the same thing.

paddybyers:
> I think we should rebase and not have two separate branches for the same thing.

I think that means rebase `integration-spec-1-1`, delete `master-1-1` and ensure that all non-spec commits are in `master`. You're ok for me to do that?

mattheworiordan:
Well I personally would have simply rebased `master-1-1` given it's used by the [running 1.1 PR](https://github.com/ably/docs/pull/258) we've had open for a very long time.

paddybyers:
Closing (superseded by https://github.com/ably/docs/pull/447)

PR #452, opened 2018-06-26
Clarify that user needs to re-enter presence after a recover

SimonWoolf:
In the v1.1 api presence model, final responsibility for ensuring presence members are in the presence set when they should be falls on the client library. In particular, after a recover, it's up to the client to re-enter them (if it still wants them), just like it's up to the client to re-attach to the channel if it still wants it. This clarifies that.


mattheworiordan:
@paddybyers  you happy for this to be merged?

Issue #444, opened 2018-06-16
Environment default is not an empty string


mattheworiordan:
@tomczoink see below:

![screen_capture 2018-06-15_11-34-56_pm](https://user-images.githubusercontent.com/43789/41500430-f2bdf598-7188-11e8-80f1-162b1f3ccf8e.png)

I believe this is incorrect. The default environment is in fact `null` or if you must put a string value in, `production`.  I would prefer `null` though given it would be `null` in most languages, see https://github.com/ably/ably-java/blob/develop/lib/src/main/java/io/ably/lib/types/ClientOptions.java#L126 for a typed lang for example.

PR #458, opened 2018-07-03
Correct env default from "" to null

tomczoink:
For #444


tomczoink:
@mattheworiordan changed it to have specifics for each lib based off of example you provided. Seem correct?

mattheworiordan:
@tomczoink (and @Srushtika so that you know), please always check how the page renders before you commit changes to the documentation pages.  The Textile formatting is complicated and what you implement may not render as you expect.  See below from this recent change:

screen_capture 2018-07-04_10-17-35_am

tomczoink:
Not merged accidentally closed it @mattheworiordan. Formatting should all be fine now.

mattheworiordan:
@tomczoink does CSharp not support `null` or an equivalent as opposed to an empty string?

tomczoink:
You're right, fixed @mattheworiordan. Good to merge?

PR #460, opened 2018-07-04
Presence page added

Srushtika:


content/general/presence.textile
```diff
@@ -0,0 +1,21 @@
+---
+title: Presence
+section: general
+index: 27
+---
+
+Presence allows you to know the connection status of a client or a device. Apart from letting you know who is online/offline at a particular point in time, you can also subscribe to presence updates( Note: This is only available with Ably's "Realtime Library":/realtime ). This means that Ably will send a message to all it's presence subscribers whenever a client on the same presence channel that they are subscribed to, has left or a new client has joined.
+
+<a href="/images/diagrams/presence.png" target="_blank">
+  <img src="/images/diagrams/presence.png" style="width: 100%" alt="Presence representation">
```
mattheworiordan: What is this diagram representing?

content/general/presence.textile
```diff
@@ -0,0 +1,21 @@
+---
+title: Presence
+section: general
+index: 27
+---
+
+Presence allows you to know the connection status of a client or a device. Apart from letting you know who is online/offline at a particular point in time, you can also subscribe to presence updates( Note: This is only available with Ably's "Realtime Library":/realtime ). This means that Ably will send a message to all it's presence subscribers whenever a client on the same presence channel that they are subscribed to, has left or a new client has joined.
```
mattheworiordan: I think replace this with:

Ably's presence feature allows devices to announce their presence on a channel.  Other devices or services may then subscribe to these presence events (such as entering, updating their state, or leaving the channel) in real time using our realtime SDKs, or via the Reactor service.  You can also request a list of who is online/offline on a channel at a particular point in time via the REST API.

content/general/presence.textile
```diff
@@ -0,0 +1,21 @@
...
+
+<a href="/images/diagrams/presence.png" target="_blank">
+  <img src="/images/diagrams/presence.png" style="width: 100%" alt="Presence representation">
+</a>
+
+In addition to these updates about clients entering and leaving, the Presence API also allows you to retrieve a list of members present. This can be done using both Ably's "Realtime":/realtime and "REST":/rest libraries.
+
+Furthermore, if persistence is enabled on the presence channel, you can also retrieve "presence history":/rest/history#presence-history for the channel, i.e, static data about historical presence states of your clients/devices. This operation also can be done using both Ably's "Realtime":/realtime and "REST":/rest libraries.
```
mattheworiordan: Remove superfluous also

content/general/presence.textile
```diff
@@ -0,0 +1,21 @@
...
+
```
mattheworiordan: Extra white space...

content/general/presence.textile
paddybyers: s/devices/clients I would suggest

content/general/presence.textile
Srushtika: Just a representative diagram of clients and/or devices being online or offline. Any suggestions?

Srushtika:
@mattheworiordan Updated based on your feedback.

Srushtika:
@paddybyers Updated as per your feedback.

content/general/presence.textile
```diff
@@ -0,0 +1,13 @@
+---
+title: Presence
+section: general
+index: 27
+---
+
+Ably's presence feature allows clients or devices to announce their presence on a channel. Other devices or services may then subscribe to these presence events (such as entering, updating their state, or leaving the channel) in real time using our "realtime SDKs":/realtime, or via the "Reactor service":https://www.ably.io/reactor. You can also request a list of clients or devices that are online/offline on a channel at a particular point in time via the "REST API":/rest-api#presence.
```
tomczoink: For the last line, "You can also request...", I'd say we should stick to the initial third person descriptive style. So for example "In addition, presence allows for a list of clients or devices that are online/offline on a channel at a particular point in time via the "REST API":/rest-api#presence to be obtained."

content/general/presence.textile
```diff
@@ -0,0 +1,13 @@
...
+
+<a href="/images/diagrams/presence.png" target="_blank">
+  <img src="/images/diagrams/presence.png" style="width: 100%" alt="Presence representation">
```
tomczoink: For the image, by itself it doesn't really explain anything, looking at it without knowing what it's meant to mean I can't tell. A caption at the least would be good to give some context to the image.

If we're trying to indicate that you can see who is offline/online, perhaps an image helping to demonstrate the fact there's communication between Ably and connected devices?

An extremely rough sketch would be:

presencelol

Not really sure the best way to indicate it, if you can think of anything @Srushtika that'd be great 😄

content/general/presence.textile
Srushtika: I'd disagree, Tom, since we are talking about a product/feature, as a narrator. wdyt?

content/general/presence.textile
Srushtika: That's a great suggestion, will try something very similar to your sketch! :)

content/general/presence.textile
tomczoink: Sure then, when I read it I read it differently, as a description of the service rather than describing the service. From the describing perspective makes sense.

Srushtika:
@tomczoink diagram updated. If it looks good to you this is good to be merged as per your website work.

Srushtika:
Closing this as this presence page is added in this commit: https://github.com/ably/docs/commit/cb78dd5ca71fdfad6ef271e1a6bff3b0cc906807 to avoid rebasing after website and other structural changes.

PR #462, opened 2018-07-06
RDDP changed to DSN

tomczoink:
For website issue https://github.com/ably/website/issues/1577.
Only found a single case of RDDP. Many references exist of 'Ably platform', would we want those changed to 'Ably network'?


mattheworiordan:
@tomczoink it's a good point whether we keep referring to ourselves as a platform.  However, I think changing that is going to have a pretty big impact and is probably unnecessary given we still have a platform that delivers the DSN.  @paddybyers WDYT?

paddybyers:
> probably unnecessary given we still have a platform that delivers the DSN. @paddybyers WDYT?

I agree. There's no problem with just using "platform" as a noun.

Issue #468, opened 2018-07-20
publish method on the Realtime client object


mattheworiordan:
We now support [transient publishes on channels](https://github.com/ably/docs/pull/323), however this API is somewhat counterintuitive if a user simply wants to stream lots of messages across lots of arbitrary channels to Ably.  Whilst it's fair to say this is going to get complicated when users need to set channel options, I expect the vast majority of users don't use channel options, and instead have to do something like this:

```
client.channels.get('foo').publish(name, data);
client.channels.get('bar').publish(name, data);
```

Unless I am mistaken, that approach will result in channels `foo` and `bar` remaining attached (for publishing) using up resources within the Ably platform, as well as in the client.

A more natural and less resource intensive API may look something like:

```
client.publish(channel, name, data, extras, channelOptions)
```

It should be trivial to add this given we already support transient publishes.

JoHuang:
Hi, I think the wasted memory allocation should be released if it became inactive after some time. Or right now it's an issue to us that the memory keeps increasing just like memory leak on our node.js server.

mattheworiordan:
> Hi, I think the wasted memory allocation should be released if it became inactive after some time. 

I am afraid we can't do that.  We don't know what references people have to the object, and we also could potentially cause serious issues if a user has configured a channel with options that are then lost next time the channel is instanced.

> Or right now it's an issue to us that the memory keeps increasing just like memory leak on our node.js server.

So why not release the channels you no longer need explicitly?

mattheworiordan:
Building on the idea that you can publish without instancing a channel should also be added to the REST lib in IMHO.  Customers have asked for this in the past, and more recently again, and it makes sense when in most cases, developers just need to publish to a channel with no channel options, either over REST or Realtime.

@paddybyers @SimonWoolf I would like to consider adding this into the 1.2 release.

mattheworiordan:
@QuintinWillison @paddybyers @SimonWoolf any further thoughts about this going into the 1.2 spec or not?

QuintinWillison:
In the case of both REST and Realtime this feels like a client-side convenience wrapper around existing functionality. However, I'm not familiar enough with the protocol yet to know whether there's a hidden implication I'm missing.

paddybyers:
> In the case of both REST and Realtime this feels like a client-side convenience wrapper around existing functionality.

Yes, that's correct.

mattheworiordan:
[...]

PR #470, opened 2018-07-22
spec: Fix inconsistency with transient publishes functionality (RTL11)

mattheworiordan:
See https://github.com/ably/docs/pull/323

Messages should no longer be NACK’ed when a channel becomes detached given we can continue to publish in the detached state

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -458,8 +458,8 @@ h3(#realtime-channel). Channel
 ** @(RTL3a)@ If the connection state enters the @FAILED@ state, then an @ATTACHING@ or @ATTACHED@ channel state will transition to @FAILED@ and set the @Channel#errorReason@
 ** @(RTL3b)@ If the connection state enters the @CLOSED@ state, then an @ATTACHING@ or @ATTACHED@ channel state will transition to @DETACHED@
 ** @(RTL3c)@ If the connection state enters the @SUSPENDED@ state, then an @ATTACHING@ or @ATTACHED@ channel state will transition to @SUSPENDED@
-** @(RTL3d)@ If the connection state enters the @CONNECTED@ state, then a @SUSPENDED@ channel will initiate an attach operation and transition to @ATTACHING@. If the attach operation times out, the channel should return to the @SUSPENDED@ state (see "RTL4f":#RTL4f).
-* @(RTL11)@ If a channel enters the @DETACHED@, @SUSPENDED@ or @FAILED@ state, then all messages that are still queued for send on that channel should be deleted from the queue triggering a failure for the publish or presence methods invoked for those messages
+** @(RTL3d)@ If the connection state enters the @CONNECTED@ state, then a @SUSPENDED@ channel will initiate an attach operation and transition to @ATTACHING@. If the attach operation times out, the channel should return to the @SUSPENDED@ state (see "RTL4f":#RTL4f)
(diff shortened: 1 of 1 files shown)


SimonWoolf:
> all messages that are still queued for send on that channel

Given transient publishes, is "messages queued for send on that channel" still a useful concept? Why would a client library need a channel-specific queue now, rather than only have the connection-wide queue?

(Especially if we want to have behavioural consistency with the proposed [client#publish](https://github.com/ably/docs/issues/468) functionality, we need as little per-channel state as possible)

content/client-lib-development-guide/features.textile
```diff
@@ -459,7 +459,7 @@ h3(#realtime-channel). Channel
 ** @(RTL3b)@ If the connection state enters the @CLOSED@ state, then an @ATTACHING@ or @ATTACHED@ channel state will transition to @DETACHED@
 ** @(RTL3c)@ If the connection state enters the @SUSPENDED@ state, then an @ATTACHING@ or @ATTACHED@ channel state will transition to @SUSPENDED@
 ** @(RTL3d)@ If the connection state enters the @CONNECTED@ state, then a @SUSPENDED@ channel will initiate an attach operation and transition to @ATTACHING@. If the attach operation times out, the channel should return to the @SUSPENDED@ state (see "RTL4f":#RTL4f).
-* @(RTL11)@ If a channel enters the @DETACHED@, @SUSPENDED@ or @FAILED@ state, then all messages that are still queued for send on that channel should be deleted from the queue triggering a failure for the publish or presence methods invoked for those messages
+* @(RTL11)@ If a channel enters the @SUSPENDED@ or @FAILED@ state, then all messages that are still queued for send on that channel should be deleted from the queue triggering a failure for the publish or presence methods invoked for those messages
```
SimonWoolf: So: this spec item is now only useful for PresenceMessages, as Messages are no-longer per-channel queued. (Don't think we need it for connection-queued messages -- a channel isn't going enter the `suspended` or `failed` state when the connection isn't `connected` except as a result of the connection entering the `suspended` or `failed` states, at which point connection-queued messages will be failed anyway). But for PresenceMessages this change is wrong, they should be failed when the channel is detached.

How about:
`* @(RTL11)@ If a channel enters the @DETACHED@, @SUSPENDED@ or @FAILED@ state, then all presence actions that are still queued for send on that channel per "RTP16b":#RTP16b should be deleted from the queue, and any callback passed to the corresponding presence method invocation should be called with an @ErrorInfo@ indicating the failure`

mattheworiordan:
> Given transient publishes, is "messages queued for send on that channel" still a useful concept? Why would a client library need a channel-specific queue now, rather than only have the connection-wide queue?

It doesn't.  But unless I am mistaken, we don't detail that in our spec as that's an "internals" issue anyway?

> (Especially if we want to have behavioural consistency with the proposed client#publish functionality, we need as little per-channel state as possible)
> Edit: answering my own question, it's still needed for presence actions per RTP16b

Yup, OK.

content/client-lib-development-guide/features.textile
mattheworiordan: > So: this spec item is now only useful for PresenceMessages, as Messages are no-longer per-channel queued. 

Yes, I suppose that is correct.  

Odd that we now have two mechanisms for delivering messages on channels, but it's necessary.

@paddybyers you happy with @SimonWoolf's suggested change?

paddybyers: > @paddybyers you happy with @SimonWoolf's suggested change?

lgtm

Issue #469, opened 2018-07-22
Idempotent REST publishing


mattheworiordan:
Given we support idempotent REST publishing now, we should:

* Update the spec to ensure that all client libs serialise `Message.id` when publishing messages
* Add a requirement for a test that publishes the same message twice, and ensures only one message is published
* Consider a two phase approach to idempotent publishing being enabled always:
  * Option to enable idempotent publishing on message publishes that don't have an explicit ID, i.e. the client lib generates an ID
  * In a future release, perhaps 1.2, enable this option by default

PR #473, opened 2018-07-25
Idempotent REST publishing spec

SimonWoolf:
Fixes https://github.com/ably/docs/issues/469 . 

Takes https://github.com/ably/wiki/issues/191#issuecomment-397741807 as a starting point, with modifications from discussions on [the realtime pr](https://github.com/ably/realtime/pull/1811#discussion_r204420926) and [in slack](https://ably-real-time.slack.com/archives/CB285667N/p1532362702000417).

So: Paddy's original proposal had that client-lib-set ids were structured `<randomStr>:<index>`, with the same randomStr for all Messages in a ProtocolMessage. I wasn't a big fan of that as if you have a ProtocolMessage going around with an id of the form, say, `sIAwterTfU:0`, the `0` might be either a msgSerial (in the case of a realtime publish) or an index (in the case of a rest idempotent publish where it's the index of the first Message and the ProtocolMessage takes its id from the first Message), and you can't know which without looking to see if `messages[0].id` exists.

My proposal is that, if we're not going to enforce a consistent relationship between Message and ProtocolMessage ids (because the PM id is just going to take its id verbatim from the first Message), why not just have it be free text for client-set ids? No structure, no colon, no index, just a new random string for every Message, where the ProtocolMessage picks the first of those for its id. Then the ProtocolMessage id wouldn't get confused with connectionid:msgSerial, and it wouldn't look like it's requiring/enforcing a consistency that it actually doesn't/isn't respectively.

Then we can also drop the restriction that when using user-supplied ids you can only publish in batches -- they would work exactly the same as for client-lib-set ids. (The only caveat is that the user is responsible for setting some id either in all messages or none of them, i.e. if they only set some ids then the behaviour is formally undefined, which allows the client lib to assume that if the first message doesn't have an id then none of them do and vice versa)

wdyt @mattheworiordan @paddybyers


mattheworiordan:
@SimonWoolf, @paddybyers and I sat down earlier and talked through idempotent publishing.  Unfortunately we don't agree with your proposal.  Rough notes below, and I will add relevant comments beneath:

---

Idempotent publishing is technically part of the 1.1 spec, however we can add this functionality now as a patch release into 1.0 so long as automatic idempotent publishing is disabled by default.

Use cases we wish to support:

- Implicit idempotency without client IDs
  - For realtime & REST this works
  - The limit is 2 minutes though for idempotency, whether realtime or REST i.e. if you reconnect a realtime connection after 2 minutes and republish, we can't offer idempotent publishes.  Same applies for REST if you retry the REST publish after 2 minutes.

- Client provided IDs with idempotency
  - Does not work with realtime (we have no API for this)
  - Works with REST, but only one message a time. Whilst we could explore publishes with batches of idempotent publishing, we believe it's OK to leave this out for now to keep simplicity down.  We will however have a support article on how to "hack" the IDs so that it follows the implicit idempotent ID format for customers who really want to work around this.  However, the key thing here is that as a publisher, you need to understand in detail how the protocol works, and the fact that idempotent publishing only works as a batch, and not for individual messages within the batch.  For example, even if we supported arbitrary IDs, a user could publish `[a, b, c]` as a single batch, and then a second later publish `b`, which would publish again i.e. it's not idempotent at a single message level. Equally, they could publish `[a, b, c, d]` and with @SimonWoolf' proposal I believe that too would introduce issues in that we now need to reject that batched message outright.

However, for customers who do need to publish a high volume of idempotent messages, they have the following options:

- Our bulk publish API (https://github.com/ably/wiki/issues/37), but this is not yet supported in client libs, experimental API
- HTTP/2 (https://github.com/ably/wiki/issues/309)

Example for an implicit idempotent publish from a client (i.e. no message ID specified in any messages)
```
POST /channel/messages/
[id: 'is2mDOHSn30fUDFhY/Bnxw==:0',
id: 'is2mDOHSn30fUDFhY/Bnxw==:1',
id: 'is2mDOHSn30fUDFhY/Bnxw==:2']

# response => 201
```

Whilst it's a big ugly to inspect every id of the messages and check they match the expected format, we decided this was the best solution.  We considered sending a transaction ID as a param / header for publish requests and leaving the messages without IDs, but then we'd end up having different approaches for different types of idempotent publishes.  It adds complexity in the protocol and for the user that is arguably a bit unnecessary.

Given above, if a user then tried to publish a single message:

```
POST /channel/messages/
[id: '/publish: is2mDOHSn30fUDFhY/Bnxw==:1']

# response => 201
```

The ID would include the `:1` part because a single message publish can have any arbitrary ID.  We only treat the `:\d+` part differently when a publish is an array of messages.

---

@SimonWoolf I will comment now on the actual spec in line based on above.

content/client-lib-development-guide/features.textile
```diff
@@ -222,6 +222,10 @@ h3(#rest-channel). Channel
...
 ** @(RSL1h)@ Where the library language permits, the @Channel#publish(name, data)@ method should provide an optional argument that allows the @clientId@ value to be specified such as @Channel#publish('event', 'data', { clientId: 'John' })@, and/or the @extras@ field to be specified such as @Channel#publish('event', 'data', { extras: { push: { 'key': 'value' } })@
 ** @(RSL1i)@ If the total size of the message or (if publishing an array) messages, calculated per "TO3l8":#TO3l8, exceeds the @maxMessageSize@, then the client library should reject the publish and indicate an error
+** @(RSL1j)@ When @Message@ objects are provided, any valid @Message@ attribute (that is, an attribute specified in "TM2":#TM2) that is supplied by the caller must be included in the encoded message. (This does not mean it must be included _unaltered_; for example the @data@ and @encoding@ will be subject to processing per "RSL4":#RSL4)
+*** @(RSL1j1)@ A test should exist which publishes a @Message@ which includes a fixed @id@ twice, and checks that it is received only once
```
mattheworiordan: perhaps not "received" given this is about REST. How about ".. and checks that it is published on the channel only once"

content/client-lib-development-guide/features.textile
```diff
@@ -1221,6 +1226,7 @@ h4. ClientOptions
 ** @(TO3g)@ @queueMessages@ boolean - defaults to true. If false, suppresses the default queueing of messages when connection states that anticipate imminent connection (connecting and disconnected). Instead, publish and presence state changes will fail immediately if not in the connected state
 ** @(TO3h)@ @echoMessages@ boolean - defaults to true. If false, suppresses messages originating from this connection being echoed back on the same connection
 ** @(TO3i)@ @recover@ string - A connection recovery string, specified with the intention of inheriting the state of an earlier connection
+** @(TO3n)@ @idempotentRestPublishing@ boolean - defaults to false. If true, "RSL1k":#RSL1k is engaged
```
mattheworiordan: I think this should be "defaults to false for clients with version < 1.1, otherwise true".  I realise that's arguably unnecessary given this is the 1.1 spec, but I think it's useful for developers when implementing now to know that they can bundle into 1.0 so long as it defaults to false.

content/client-lib-development-guide/features.textile
```diff
@@ -222,6 +222,10 @@ h3(#rest-channel). Channel
...
 ** @(RSL1i)@ If the total size of the message or (if publishing an array) messages, calculated per "TO3l8":#TO3l8, exceeds the @maxMessageSize@, then the client library should reject the publish and indicate an error
+** @(RSL1j)@ When @Message@ objects are provided, any valid @Message@ attribute (that is, an attribute specified in "TM2":#TM2) that is supplied by the caller must be included in the encoded message. (This does not mean it must be included _unaltered_; for example the @data@ and @encoding@ will be subject to processing per "RSL4":#RSL4)
+*** @(RSL1j1)@ A test should exist which publishes a @Message@ which includes a fixed @id@ twice, and checks that it is received only once
+** @(RSL1k)@ If @idempotentRestPublishing@ (see "TO3n":#TO3n) is enabled, if the @Message@s do not include caller-supplied @ids@, for each @Message@ in the array the library should randomly generate get 9 bytes from a source of randomness, encode them as base64, and set it as the @Message@ @id@
```
mattheworiordan: Odd that you chose 9 bytes, any thinking?

mattheworiordan: We are going to go with instead that a random ID is generated, and every message is given the ID + `:` + index within the array of messages being published.

content/client-lib-development-guide/features.textile
```diff
@@ -222,6 +222,10 @@ h3(#rest-channel). Channel
...
+*** @(RSL1k1)@ For the purposes of deciding whether the @Message@s include caller-supplied @ids@, the library may assume that the first @Message@ does not have an @id@ then none of them do
```
mattheworiordan: Why?

content/client-lib-development-guide/features.textile
mattheworiordan: Perhaps also add "and contains the ID assigned to the published message"?

content/client-lib-development-guide/features.textile
mattheworiordan: We need to specify that a test should exist that publishes an array of messages to an endpoint that fails the first time, and succeeds upon the second request.  Each attempt must contain the exact same payload with the same generated IDs.

SimonWoolf: >Odd that you chose 9 bytes, any thinking?

base64 is 6 bits per character, so number of bits needs to be divisible by 6 to have it be 'clean', else you get variable numbers of `=`s at the end for padding. So 72 bits (9 bytes; 12 characters) works, as does 96 (12 bytes; 16 characters); Paddy commented that 16 characters was a bit long, so 9 bytes it was.

SimonWoolf:
> For example, even if we supported arbitrary IDs, a user could publish [a, b, c] as a single batch, and then a second later publish b, which would publish again i.e. it's not idempotent at a single message level. Equally, they could publish [a, b, c, d] and with @SimonWoolf's proposal I believe that too would introduce issues in that we now need to reject that batched message outright.

But using `<id>:<index>` doesn't actually do anything to help either of those cases. The ProtocolMessage ID is still just taken from the first Message, so that's still the only one that matters for the purposes of realtime deduping. Realtime isn't planning to actually parse the structure AIUI, it's just being treated as an opaque string, so what's the actual advantage of enforcing that format? What am I missing?

(Besides, ISTM that  'Idempotency is done at the level of the whole publish body' is easy enough to explain and not exactly unintuitive)

content/client-lib-development-guide/features.textile
SimonWoolf: see the bit in parens in last paragraph of original comment

content/client-lib-development-guide/features.textile
mattheworiordan: Ok, this makes perfect sense, :+1:

mattheworiordan:
> But using <id>:<index> doesn't actually do anything to help either of those cases.

Sure, that's true technically.  However, from a DX perspective in regards to the API, unless I read the documentation in detail, I don't think I would expect that publishing a message with `publish([msgAwithId, msgBwithId])` would only be idempotent based on `msgAwithId`, and republishing with `publish([msgAwithId, msgCwithId])` would in fact result in that message not being published.  Then later publishing `publish(msgBwithId)` would not be idempotent even though `msgBwithId` has already been published as part of the A&B pair before.

Sure, if someone goes into the detail of the docs, they will find this answer, but they'll find this answer only AFTER they've realised their assumption around idempotency is broken.

With our proposed approach, they'll find out immediately when they try and publish their messages in batches.  They then have the choice of publishing messages individually (which is almost certainly what they will want in 99% of the cases, especially if we support HTTP/2), using our bulk API, or perhaps coercing their messages to us an ID format which indicates they've understood the shortfalls in a batch API for idempotent message publishing.

So in short, the issue I have is not in how it works technically, but in how it will very likely lead to developers finding out the hard way that our API doesn't work as they expected.

Added to all of this, I still think most people won't **need** batch publishing all that often, and if they do, they should rarely have to understand under the hood how a `ProtocolMessage` is different from a `Message`.  From a developer's perspective, they are just publishing a message, or a bunch of messages.  With our proposed API, they will hit an issue when providing IDs in a batch, and in most cases we'd probably recommend they publish one at a time, or use the bulk API if they need that type of capability.  Rarely, we'll explain how they can publish their messages as a single `ProtocolMessage`.

As an aside, @paddybyers, I assume we have a unique error code for the rejection of batch published messages with different IDs?  If so, with our 1.1 release, we could take users easily to a support article that explains the issue and gives them other options.

content/client-lib-development-guide/features.textile
```diff
@@ -222,6 +222,11 @@ h3(#rest-channel). Channel
...
+** @(RSL1k)@ Idempotent publishing via REST is supported by populating the @id@ attribute of @Message@ instances passed to @publish()@:
+*** @(RSL1k1)@ Idempotent publishing via library-generated @Message@ @id@s is supported if @idempotentRestPublishing@ (see "TO3n":#TO3n) is enabled and one or more @Message@ instances are passed to @publish()@ and all @Message@s have an empty @id@ attribute. The library generates a base @id@ string by base64-encoding  a sequence of at least 9 bytes obtained from a source of randomness. Each individual @Message@ in the set of messages to be published is assigned a unique @id@ of the form &lt;base id&gt;:&lt;serial&gt; (where @serial@ is the zero-based index into the set).
+*** @(RSL1k2)@ Idempotent publishing via client-supplied @Message@ @id@s is supported where a single @Message@ is passed to @publish()@ and it contains a non-empty @id@. The @id@ is perserved on sending the message.
+*** @(RSL1k3)@ If more than one @Message@ is passed to @publish()@ and one or more of those messages contains a non-empty @id@ attribute, then this results in an error being indicated back to the caller.
```
mattheworiordan: I thought this is allowed by the client library, but not allowed by the realtime service.  Reading this, I think it would be fair for a library to reject this publish, whereas it should in fact allow the publish and rely on realtime to reject it if it does not match the pattern expected?  This would allow us to offer a "hack" to customers if they absolutely need to batch REST idempotent publishes.

mattheworiordan:
> As an aside, @paddybyers, I assume we have a unique error code for the rejection of batch published messages with different IDs? If so, with our 1.1 release, we could take users easily to a support article that explains the issue and gives them other options.

@paddybyers ^?

paddybyers:
> No mention of tests needed, but perhaps that is unnecessary in this spec. WDYT?

All of the requirements need to be tested. I don't think the spec needs to state that separately for each item.

content/client-lib-development-guide/features.textile
mattheworiordan: Yes / no?

mattheworiordan:
> All of the requirements need to be tested. I don't think the spec needs to state that separately for each item.

Of course.  However, where a test is not obvious, I think a mention is a good idea.  Take https://docs.ably.io/client-lib-development-guide/features/#RSC1c for example.

Case in point for idempotent publishing, should we say a test should when idempotent publishing is enabled that simulates a failure in the response from an HTTP request, but that allows the HTTP request to reach the Ably service.  With idempotency enabled, this should result in two attempts, and one message, and with idempotency disabled, it should result in one.  

I realise this type of detail is perhaps not suitable technically for a spec, however how do we agree on what is deemed suitable test coverage without this?  Nothing in the spec makes an attempt to explain what is considered acceptable or not from a test coverage perspective.

content/client-lib-development-guide/features.textile
paddybyers: Yes; see https://github.com/ably/docs/commit/473cec8e3d5c5fd21301c76f262bbd97c684b9ee

paddybyers:
> I realise this type of detail is perhaps not suitable technically for a spec, however how do we agree on what is deemed suitable test coverage without this? Nothing in the spec makes an attempt to explain what is considered acceptable or not from a test coverage perspective.

Generally, for unit tests for specific requirements, we have a good understanding of what constitutes a test, and we apply that understanding fairly consistently in PR reviews.

However, what you're describing are integration tests for idempotency, rather than tests for the specific library behaviour specified by this PR. I've added explicit requirements at https://github.com/ably/docs/commit/473cec8e3d5c5fd21301c76f262bbd97c684b9ee

paddybyers:
> I assume we have a unique error code for the rejection of batch published messages with different IDs?

See https://github.com/ably/ably-common/pull/30

PR #491, opened 2018-08-10
Stats results: correct granularity explanation

SimonWoolf:
Documentation currently claims the intervalId implies the interval length. This is not currently the case; month and day intervals use hour granularity, you need to use the `unit` to know the actual interval length. Slack discussion:

> simon [10:52]
out of interest, what's the reasoning for day and month stats metrics using an intervalId with hour-level granularity, just set to the first hour of the day/month (e.g. 2018-08-01:00)? e.g. https://peoplefun-rest.ably.io:8090/metrics/app%3An1dDrw/by/month/intervals/2018-08-01%3A00?format=html . Hour stats intervalId truncate the minutes and minute stats truncate the seconds, but that principle isn't carried through to days and months

> paddy [11:01]
Because in principle a month or day interval start doesn't align with the calendar, or UTC
although in practice we've never had non-zero offsets
but you're supposed to be able to set those offsets as attributes on the account, and that would give rise to a different aggregate
so that hour in the `intervalId` fixes which hour rows contribute to that aggregate


SimonWoolf:
cf https://github.com/ably/realtime/issues/531

mattheworiordan:
Ok, whilst I agree `unit` should be the right attribute, we don't have that defined in the spec at present.  So either we introduce a breaking change now and remove `intervalGranularity`, or we add `unit` to the `Stats` object and add an alias `intervalGranularity` until we do a 2.0 release. Shouldn't we be addressing that though in this PR?

paddybyers:
You're getting dangerously close with your `month` example of having to explain how we decide when the end of the month is. Of course we could explain this, but it's a bit academic since we always align with the start of the calendar month right now. Perhaps a `day` example would avoid that.

SimonWoolf:
>  So either we introduce a breaking change now and remove intervalGranularity, or we add unit to the Stats object and add an alias intervalGranularity until we do a 2.0 release

Sure, any libraries which currently have an `intervalGranularity` should keep it for the remainder of 1.x as an alias of `unit` to avoid breaking backwards compatibility. Ones which don't, don't need to add it.

I'll add all that stuff (plus spec changes) to this PR.

> Perhaps a day example would avoid that

ok, will switch to day

SimonWoolf:
@mattheworiordan @paddybyers pr updated

content/types/_stats.textile
```diff
@@ -8,14 +8,19 @@ h4.
   ruby:    Attributes
   python:  Keyword arguments
 
+- <span lang="default">intervalId</span><span lang="ruby,python">interval_id</span><span lang="csharp">IntervalId</span> := the time at which the interval covered by this @Stats@ object starts. For example, an interval ID value of "2016-03-01:10:02" represents a single minute interval, in UTC. The granularity does not necessarily indicate the period of time the interval covers — for that you should consult the @unit@; for example an interval ID of "2016-03-01:00" with a @unit@ of @day@ would represent the whole day of the 1st March 2016 (UTC, midnight to midnight), but the same interval ID with a @unit@ of @hour@ would represent only the first hour of that day.<br>__Type: @String@__
```
paddybyers: > The granularity does not necessarily indicate the period of time the interval covers

Surely you meant to say "The `intervalId` does not necessarily indicate the period of time the interval covers" ?

I think this is confusing in any case. I think the definition of `unit` would help if it comes before this, and here it should say

> the UTC time at which the time period covered by this @Stats@ object starts. For example, an `intervalId` value of "2018-03-01:10", in a `Stats` object whose `unit` is `day`, would indicate that the period covered is "2018-03-01:10 .. 2018-03-01:11". All `Stats` objects, except those whose `unit` is `minute` have an `intervalId` with resolution of one hour and the time period covered will always begin and end at a UTC hour boundary. For this reason it is not possible to infer the `unit` by looking at the resolution of the `intervalId`. `Stats` objects covering an individual minute will have an `intervalId` indicating that time; for example "2018-03-01:10:02".

SimonWoolf: Updated in 3424176

PR #496, opened 2018-08-16
WebRTC tutorial updates as per Matt's suggestions

Srushtika:


mattheworiordan:
@tomczoink do you know why did a preview site did not get deployed?

mattheworiordan:
This is looking awesome @Srushtika.  Thanks to you and the author.

Few comments:

* Restricting WebRTC to IP, IMHO, is a bad idea.  It means no one can test in different devices easily, and my ad blocker stopped that script from working anyway, see below.  Why can't we use the same approach I took at https://www.ably.io/tutorials/publish-subscribe#lang-javascript, which also has a cool QR code to make it easy for mobile?   All that happens is a query param code is maintained for the session and passed in the links.
* Please change "Sending and receiving calls" to "Making and receiving calls".  Same applies to all other call headings like this.  We send files and messages, but make calls.
* I am not quite sure how the call demo works.  I see two users online, but call now does nothing?  I have a similar problem for the filetransfer demo.
* Copy about which browsers support screen sharing is going to get out of date very quickly.  Could you perhaps state the date and provide a link to find the latest support, perhaps with caniuse?  We should do that throughout.
* Let us know should be a link.

Once again, this is good work!

capto_capture 2018-09-03_02-49-41_am

PR #497, opened 2018-08-16
Channel lifecycle api

paddybyers:
See discussion on commit:

https://github.com/ably/docs/commit/829ca206b8acca4129449c41b4bc6ec9dbe8c6fb

@SimonWoolf continue discussion here. Note that commit isn't in the branch any more because I force-pushed a change :( Perhaps we need to tag it in order to preserve that discussion.

### DIFF
--- a/content/general/channel-lifecycle-api.md
+++ b/content/general/channel-lifecycle-api.md
@@ -1,0 +1,126 @@
+# Channel lifecycle API (beta)
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


Srushtika:
@paddybyers @SimonWoolf @mattheworiordan @tomczoink 
Closing this PR as the docs and tutorials related to the channel lifecycle API are now merged into master.

PR #505, opened 2018-09-05
changes

samuelayo:
- changes needed to use qr code with demo


samuelayo:
@Srushtika please review

Issue #522, opened 2018-09-26
Best practice for presence


mattheworiordan:
@Srushtika whilst talking to a customer today, I remember we have two other common gotchas that we should include in our [best practice guide](https://www.ably.io/documentation/best-practice-guide).

* [x] Remember that a single clientId can be present one or more times on a channel. Therefore, if you are subscribing to leave events, don't assume that `clientId` is gone, as they could be present more than once before that. Check the presence set using `get`
* [x] We have a lot of code to handle the complexities of keeping the members present on a channel in sync in our realtime client lib.  So don't try and build your own code to keep track of who is present or not a channel by subscribing to enter/leave events, and instead rely on our `get` method on the realtime lib, that is always in sync and kept up to date.

Srushtika:
Added these in [here](https://github.com/ably/docs/pull/524)

PR #524, opened 2018-09-28
New points in best practice guide

Srushtika:
Addresses #522


content/root/best-practice-guide.textile
```diff
@@ -148,4 +148,19 @@ Whilst we recommend that you use channels to distribute work more evenly across
 
 h3(#embedding-client-id-in-token). Embedding client ID in the token to ensure trust
 
-A clientID serves as an identification for your clients when they are present on a channel. Your clients can set their own client ID, but in that case, there's a possibility that some of your clients can pretend to be someone else. To prevent this, you can embed a client ID in the token issued to your clients from your auth server. Doing so ensures that the client ID is set by your auth server, thus eliminating the chance of the end clients emulating as others.
\ No newline at end of file
+A clientID serves as an identification for your clients when they are present on a channel. Your clients can set their own client ID, but in that case, there's a possibility that some of your clients can pretend to be someone else. To prevent this, you can embed a client ID in the token issued to your clients from your auth server. Doing so ensures that the client ID is set by your auth server, thus eliminating the chance of the end clients emulating as others.
```
mattheworiordan: Why client ID and clientID.  Should it not always be `clientId`?

content/root/best-practice-guide.textile
```diff
@@ -148,4 +148,19 @@ Whilst we recommend that you use channels to distribute work more evenly across
...
+
+h3(#use-existing-presence-set). Using the synced presence set provided by the realtime library
+
+Building a mechanism that keeps a tab on presence events can be quite complex. Hence, the @Presence@ object in the realtime library exposes a "get method":https://www.ably.io/documentation/realtime/presence#get allowing a client to retrieve an array of all members present on the channel. The Ably client is responsible for keeping track of the presence set from the time that the channel is attached. An up to date presence set is pushed to the client following attach and this set is updated on each subsequent presence event. Thus get returns the already-known presence set retained in memory and does not trigger a new request to the Ably service. You can simply use this method as shown below, instead of venturing into the complexity of building one of your own.
```
mattheworiordan: `get` should be `@get@` given it's a code keyword

content/root/best-practice-guide.textile
```diff
@@ -148,4 +148,19 @@ Whilst we recommend that you use channels to distribute work more evenly across
...
+  console.log('There are ' + members.length + ' members on this channel');
+  console.log('The first member has client ID: ' + members[0].clientId);
+});
+```
+
+h3(#using-presence-set-as-single-source-of-truth). Using the presence set as a single source of truth to see which clients are present
+
+A single @clientId@ may be present multiple times on the same channel via different client connections. As far as Ably is concerned, these are different members of the presence set for the channel, however they will be differentiated by their unique @connectionId@. For example, if a client with ID “Sarah” is connected to a chat channel on both a desktop and a mobile device simultaneously, “Sarah” will be present twice in the presence member set with the same client ID, yet will have two unique connection IDs. This means that if "Sarah" leaves the channel on her mobile, it doesn't necessarily mean she's gone as she might still be connected via desktop. Hence, you should always use the presence set, which you can retrieve using the "get method":https://www.ably.io/documentation/realtime/presence#get, as the single source of truth to see which clients are connected.
```
mattheworiordan: This is good, although perhaps also mention multiple tabs in a browser as this is a common misconception...

content/root/best-practice-guide.textile
```diff
@@ -148,11 +148,11 @@ Whilst we recommend that you use channels to distribute work more evenly across
 
 h3(#embedding-client-id-in-token). Embedding client ID in the token to ensure trust
 
-A clientID serves as an identification for your clients when they are present on a channel. Your clients can set their own client ID, but in that case, there's a possibility that some of your clients can pretend to be someone else. To prevent this, you can embed a client ID in the token issued to your clients from your auth server. Doing so ensures that the client ID is set by your auth server, thus eliminating the chance of the end clients emulating as others.
+A clientID serves as an identification for your clients when they are present on a channel. Your clients can set their own clientID, but in that case, there's a possibility that some of your clients can pretend to be someone else. To prevent this, you can embed a clientID in the token issued to your clients from your auth server. Doing so ensures that the clientID is set by your auth server, thus eliminating the chance of the end clients emulating as others.
```
mattheworiordan: If you're using `clientId` as opposed to `client ID` then it's an attribute / property so should be `@clientId@`. As you will see in https://www.ably.io/documentation/realtime/presence, `clientId` is in that format as opposed to `clientID`.

Srushtika: Got it. Fixed [here](https://github.com/ably/docs/commit/6cab87f145513c7cf3551dd6e759c43ff2ed88d8), thanks!

Issue #518, opened 2018-09-24
Fix docs on createtokenrequest for the .net library


SimonWoolf:
now returns a string not a TokenRequest object, see https://github.com/ably/ably-dotnet/issues/110#issuecomment-310402257

PR #528, opened 2018-10-04
Fixes to various C# code usage and documentation

tomczoink:
Goes through various code samples and API documentation for C#, correcting non-existent variables.

Issues this resolves: 
- #518

Also TODO:

- [ ] Check through API of each page to ensure everything is documented correctly (only partially done).
- [x] Check tutorials


tomczoink:
Parts I'm uncertain on:

- https://www.ably.io/documentation/realtime/versions/v0.8/usage#stats - Also accepts no arguments, should we document this?

- https://www.ably.io/documentation/realtime/versions/v0.8/usage#time - Various titles, such as `time` here, are actually something else in C#, such as `TimeAsync`. Should we update such cases to be the actual variable name, or keep as generalized titles?

- https://www.ably.io/documentation/realtime/versions/v0.8/usage#client-options - Seems to be more options, such as `TlsPort`, should we document them as well? Also should LoggerSink => LogHandler ?

- https://www.ably.io/documentation/realtime/versions/v0.8/connection#on - Seems to also have 'Func<ConnectionStateChange, Task> action' and 'ConnectionState state, Func<ConnectionStateChange, Task> action'. Same for others, have Task option at end. Should we include these as options, or is this a quirk of C# we shouldn't mention?

- https://www.ably.io/documentation/realtime/versions/v0.8/connection#off - Should we include `empty` as viable, given it's possible to do?

- https://www.ably.io/documentation/realtime/versions/v0.8/presence#history - Override function of just (bool untilAttach) is possible too, is this intentional and thus should be documented?

mattheworiordan:
@tomczoink please ask @withakay to review as well

content/realtime/authentication.textile
```diff
@@ -159,7 +159,7 @@ minimize. See example
   ```[csharp]
     AblyRealtime realtime = new AblyRealtime("{{API_KEY}}");
     TokenParams tokenParams = new TokenParams {ClientId = "Bob"};
-    TokenRequest tokenRequest = await realtime.Auth.CreateTokenRequestAsync(tokenParams);
+    String tokenRequest = await realtime.Auth.CreateTokenRequestAsync(tokenParams);
```
withakay: lower case `string` is the preferred keyword (although `string` is actually and alias for `System.String` and as the `System` namespace is imported by default `String` is still valid)

withakay:
Looks good, nicely done @tomczoink

PR #537, opened 2018-10-15
Article concepts longpolling

axefrog:
@Srushtika old pr was closed, hence the missing notification.


Srushtika:
Thanks @axefrog . As you requested, I've just swiped the missing info about libraries from here.

PR #540, opened 2018-10-23
spec: TI5 Clarify when a URL should be appended to log messages

mattheworiordan:
Rather belated, but here's my attempt at clarifying when a URL is included in log entries shown to users.  Specifically addressing https://github.com/ably/ably-ruby/pull/171#issuecomment-426831624 and https://github.com/ably/ably-ruby/pull/171/commits/ea425be4a855d68649a81f161c4483ce6a38ab99.


SimonWoolf:
given that the 'not already present' requirement is only temporary (we can remove it in the 1.1 spec, since for v=1.1 realtime doesn't append an href to the message, so it's only really applicable for particularly keen client libs that want to implement this spec item before 1.1), probably not worth being too particular about it

mattheworiordan:
> probably not worth being too particular about it

I'd be happy with that 👍

Issue #546, opened 2018-11-01
Docs repo links & nav broken


mattheworiordan:
Issue 1:

* Go to https://docs.ably.io/realtime/
* Click "Authentication" in the nav under core features.
* Note you are at the v0.8 docs.  This is broken.

Issue 2:

* Go to https://docs.ably.io/rest/authentication/
* Click on Ably JWT under Token Authentication -> gives a 404 at https://docs.ably.io/general/authentication#ably-jwt

I don't know if these issues exist in the main website, but we definitely need to fix ASAP

PR #547, opened 2018-11-02
Fix docs nav + incorrect auth links

tomczoink:
#546 fixes. Makes core-features versionable, updates auth titles, and corrects links to general/authentication to core-features/authentication.


mattheworiordan:
@tomczoink any idea why the deploy failed?

PR #556, opened 2018-11-06
Updated to new wording for channel lifecycles

tomczoink:


content/realtime/channel-metadata.textile
```diff
@@ -19,19 +19,19 @@ h2(#overview). Overview
 
 Every channel in Ably's Data Stream Network has "lifecycle events":#lifecycle-events. The "channel lifecycle events":#lifecycle-events are triggered whenever a channel changes its "lifecycle state":#lifecycle-events. The "occupancy":#occupancy events are triggered whenever there is a change in the number occupants in a channel, indicating how many of each type of occupant there are. Both the "channel lifecycle":#lifecycle-events and the "occupancy":#occupancy events are published onto special channels called "metachannels":#metachannels, that can be subscribed to with our "realtime libraries":/realtime. This is often used in conjunction with the "Ably Reactor":/reactor, which allows you to send this data to your "serverless endpoints":/general/functions, "queues":general/queues, and "streams":general/firehose.
```
Srushtika: The complete overview will need to be updated, unfortunately so the readers understand what the API is for, the nomenclature and the levels of this metadata as per what we discussed.

content/realtime/channel-metadata.textile
```diff
@@ -17,9 +17,15 @@ jump_to:
...
+The @Channel Metadata API@ allows for information about channels, such as the current state changes of a channel, or the changes in its occupants, to be accessed. This is done through three primary mechanisms:
 
-You can also make use of the "REST API":/rest-api, to "enumerate all active channels":/rest/channel-status#enumeration-rest, or obtain the status of an "individual channel":#metadata-rest. At present, REST "requests":/rest/usage#request for lifecycle events will not count towards your message limit. If you are making use of "Reactor":/reactor however, then each message sent over Reactor *will* count towards your message limits.
+- "Channel lifecycle events":#lifecycle-events provide information on when a channel changes its "lifecycle state":#lifecycle-events.
```
Srushtika: To keep this consistent throughout, perhaps say 
"Channel lifecycle events":#lifecycle-events are triggered when a channel changes its "lifecycle state":#lifecycle-events.

content/realtime/channel-metadata.textile
```diff
@@ -17,9 +17,15 @@ jump_to:
...
 
-You can also make use of the "REST API":/rest-api, to "enumerate all active channels":/rest/channel-status#enumeration-rest, or obtain the status of an "individual channel":#metadata-rest. At present, REST "requests":/rest/usage#request for lifecycle events will not count towards your message limit. If you are making use of "Reactor":/reactor however, then each message sent over Reactor *will* count towards your message limits.
+- "Channel lifecycle events":#lifecycle-events provide information on when a channel changes its "lifecycle state":#lifecycle-events. 
+- "Occupancy":#occupancy events are triggered whenever there is a change in the number occupants in a channel, indicating how many of each type of occupant there are.
+- The "Channel Status API":/rest/channel-status allows for users to "enumerate":/rest/channel-status#enumeration-rest through all active channels, along with providing "REST":/rest-api alternatives to the above two points for individual channels. Our recommendation is to make use of the realtime library where possible.
```
Srushtika: The "Channel Status API":/rest/channel-status allows for users to query a one-off request with regards to status and occupancy data of channels using a REST endpoint. In addition to that, it also allows you  "enumerate":/rest/channel-status#enumeration-rest all the active channels in a particular app. The details of Channel Status API are covered in out "REST library documentation":todo-link.

Srushtika: In the next paragraph after above...

Please note that since the metadata of various channels is prone to change very frequently, unless you have a special use case within your app, we recommend you to subscribe to the Channel Metadata API's realtime events rather than querying this data through REST, as it might become stale as soon as you have received it.

wdyt?

content/realtime/channel-metadata.textile
```diff
@@ -17,9 +17,15 @@ jump_to:
...
+
+Both the "channel lifecycle":#lifecycle-events and the "occupancy":#occupancy events are published onto special channels called "metachannels":#metachannels, that can be subscribed to with our "realtime libraries":/realtime. This is often used in conjunction with the "Ably Reactor":/reactor, which allows you to send this data to your "serverless endpoints":/general/functions, "queues":general/queues, and "streams":general/firehose.
```
Srushtika: Maybe add a new section heading here since this is no longer an overview? Something like...
'Subscribing to Channel Metadata API's realtime events'?

content/realtime/channel-metadata.textile
```diff
@@ -17,9 +17,15 @@ jump_to:
...
+
+At present, REST "requests":/rest/usage#request for the "channel status API":/rest/channel-status  will not count towards your message limit. If you are making use of "Reactor":/reactor however, then each message sent over Reactor *will* count towards your message limits.
```
Srushtika: I think we can skip the REST details here as we have described it in the rest docs. However, we must retain teh reactor info and specify why people might be needing to use reactor. (Coz there are so many messages :P)

PR #564, opened 2018-11-12
Inserts warning about API keys where they are used

tomczoink:
Will insert a warning wherever an Ably key is used in a code snippet.


mattheworiordan:
This didn't deploy fro some reason?  Can you see why and provide a link to this in action?

tomczoink:
I can't actually find where temporary docs should be deployed, does it still exist @mattheworiordan ?

mattheworiordan:
> I can't actually find where temporary docs should be deployed

Yes, it's all part of our Heroku pipelines.

tomczoink:
As far as I could tell docs didn't have a pipeline any more for some reason, so I've created a new pipeline now where this is now deployed @mattheworiordan .

PR #573, opened 2018-11-29
DSX TFL added base

tomczoink:
Currently JS needs to be fixed, doesn't appear as code ¯\\_(ツ)_/¯.  Also good to add bitcoin version of this + update text. Srushtika has code to do it.


mattheworiordan:
@tomczoink please see my commit in regards to why it was not working.

Also, please do try and step back when working on demos/docs/tutorials and think about the user arriving on the page and potentially not having any context.  With your example, you expected people to know what a NAPTAN code is, and didn't explain how the channel name was constructed etc.  How it is now is not right either, but it's hopefully good enough until we create more examples shortly.

P.S. I think the live arrivals board should really be a table, not a console type box.
P.P.S. Your chose a new station was completely broken as you changed the channel object, but didn't unsubscribe/detach, or subscribe the new channel.

PR #581, opened 2018-12-11
Various fixes for push documents

paddybyers:
This updates the push-related docs to fix some errors and update to bring into line with the latest API. The majority of the updates relate to Android in particular.


mattheworiordan:
@paddybyers these changes are great.  I have updated the docs a little to address the old beta notices.

@tomczoink @remotekieran @Srushtika some notes I kept whilst reviewing in terms of actions needed for the push docs:

* [ ] Kieran & Eve: Mock up landing page for push notifications like our other product pages. We can then have a more "traditional" product launch for push.
* [ ] Srushtika: Push documentation home page should have a diagram after the intro, with GCM and APNS shown in the diagram, and arguably a W3C Web Push icon (if available) with "beta" / "experimental" tag on it.
* [ ] Kieran: The push page /general/push should contain a concise introduction to push notifications
* [ ] Tom: Client library SDK docs for REST and Realtime need to include API reference documentation for iOS and Android.  These docs exist to some degree already in the push documentation. In order to keep the push API documentation task manageable for now, I would recommend we add a single Push section into REaltime & REST docs, which contains an API reference to the complete set of API methods on the push object for admin and other methods. Also, there are some new methods now on the channel object's push object (only for devices that support push, iOS and Android). See http://localhost:4000/client-lib-development-guide/features/#push-notifications for the IDL.  On platforms that do not support push activation and other features, we should simply hide the objects and methods completely.  For platforms where support is platform dependent, such as Java, we need a small note for each method that states that this method is only supported on Android for example. We should include basic documentation in SDK API docs, with links to detail on how to use these methods in the existing push docs.
* [ ] Tom: you can perhaps ask Kieran for help for intro docs (if you wish) for the SDK push API docs

PR #582, opened 2018-12-11
Document that untilAttach does not work with direction=forwards

SimonWoolf:
cf https://github.com/ably/realtime/issues/2198#issuecomment-446204336


tomczoink:
@mattheworiordan can you give this a quick glance to confirm this is correct please?

Issue #574, opened 2018-11-30
Nanoc upgrade + performance


mattheworiordan:
Nanoc is now painfully slow as any changes require the entire site to be rebuilt.  We should upgrade to Nanoc 4, and then look at performance as we need to be able to work more quickly locally.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/DOC-120) by [Unito](https://www.unito.io/learn-more)

mattheworiordan:
Hey @tomczoink, did you make any progress with this by any chance?

tomczoink:
A PR for Nanoc upgrade is up now at #599 @mattheworiordan .

mattheworiordan:
@tomczoink can we close this, or is some of this still outstanding?

PR #583, opened 2018-12-11
V1.0 to V1.1 docs update

tomczoink:
This adds a new version, v1.1, by moving v1.0 into its own section.

This also includes the capability to auto-detect and render from partials the correct partial for a version without explicitly stating the version (`rest/version/v1.0/_auth` for example can now just be `partial_version('rest/_auth')`). Hopefully this functionality should simplify upkeep of version refs and avoid potential un-updated refs to pages as we get more versions.

This will have a website change to accompany it, which needs to be merged together with this.


mattheworiordan:
@tomczoink thanks for this.  Here are some general notes on what is needed to do this release:

- [ ] Diff from v1.0 to v1.1 is updated https://github.com/ably/docs/commit/7d19182e0f42721ac9f4a23cc81f77bf548b5769. We need to tag the release as v1.1 and add release notes (see https://github.com/ably/docs/releases, we will link to a blog post, will get Eve/Kieran/Srushtika to help on that)
- [ ] Need to look at performance issues https://github.com/ably/docs/issues/574
- [ ] Add new methods to the API docs.  Where a method / attribute is not supported by the platform, by design, then we should remove that item completely when that language is selected. However, if a method / attribute should be supported, but is simply not due to missing implementation (see https://docs.google.com/spreadsheets/d/1zzzbAFtTKAnrk0kYyMs55QSHKbtJld1PDF-riRsNleg/edit#gid=0, although need clarification if this is up to date), then we need to introduce a new badge / UI element that indicates a method is not implemented. Let's look at what options we have available that make it clear the method should be available, but are not for that language only, with a link to the README in the repo listing out the missing feature (or perhaps matrix in https://github.com/ably/docs/issues/382)
- [ ] Review changes in v1.0 to v1.1 https://github.com/ably/docs/blob/7d19182e0f42721ac9f4a23cc81f77bf548b5769/content/client-lib-development-guide/versions/features-1-0__1-1.diff, and add method / attribute / object changes to the docs.  Note that most of v1.1 has no impact on the customer-facing APIs, and therefore only internal changes.  Also, there are some new types (for push) and new attributes of types that need to be added.
- [ ] Please focus on Javascript, iOS and Java first for us to review. We can then get help from respective client lib SDK maintainers to give us API details in each language so that we can add.

Note: 

- Idempotent publishing can be ommitted from the docs for now, until it's officially supported in the platform

content/realtime/channels.textile
```diff
@@ -255,6 +255,26 @@ channel.publish("action", data: "boom!")
 
 Normally, errors in attaching to a channel are communicated through the attach callback. For implicit attaches (and other cases where a channel is attached or reattached automatically, e.g. following the library reconnecting after a period in the @suspended@ state), there is no callback, so if you want to know what happens, you'll need to listen for channel state changes.
 
+<div lang="jsall,ruby,swift">
+
+h3(#transient-publish). Transient publishing
+
+When attempting to only publish to a client, it is possible to publish without attaching to the channel. This can be beneficial if you intend to publish into many channels, removing the need to attach to a channel each time you wish to publish.
```
mattheworiordan: I think we're missing a key aspect of transient publishing being that by not attaching, the client is not subscribing to messages too, so this avoids messages being sent to the client. See https://docs.google.com/document/d/1CSJ3L6Mi2lkw_C26cG_QtdSwcdV8SaiQfEY5MBuENEc/edit#.

content/realtime/channels.textile
```diff
@@ -213,7 +213,7 @@ Ably provides a "REST API":/realtime/channel-metadata to query your app for meta
...
+<span lang="default">Although the attach operation can be initiated explicitly by a client, it is more common for the client to perform a publish or subscribe operation, and the client library will initiate the attach if the channel is not already attached. The client library allows clients to begin publishing "messages":/realtime/messages to a channel as soon as the channel has been created, and messages are queued until such a time as the attach has succeeded or failed.</span><span lang="jsall,ruby,swift">Although the attach operation can be initiated explicitly by a client, it is more common for the client to subscribe, which will initiate the attach if the channel is not already attached. This client library allows clients to begin publishing "messages":/realtime/messages without attaching to the channel with "transient publishing":#transient-publish.</span>
```
mattheworiordan: Does Android/Java not support transient publishing?  What about .NET too?

content/realtime/channels.textile
mattheworiordan: I am a little concerned that if we do the docs like this, how will we know when to update this as those client SDKs add support?  Perhaps if we go down this route, we should at least add issues in those unsupported repos to update the docs once these features are fully supported. WDYT?

content/realtime/channels.textile
```diff
@@ -575,7 +595,7 @@ bq(definition#publish-data).
...
 
-Publish a single message on this channel based on a given event name and payload. A <span lang="default">callback</span><span lang="java">listener</span> may optionally be passed in to this call to be notified of success <span lang="default">or failure</span><span lang="ruby"></span> of the operation. If the channel is @initialized@ (i.e. no attempt to attach has yet been made for this channel), then calling @publish@ will implicitly attach the channel.
+Publish a single message on this channel based on a given event name and payload. A <span lang="default">callback</span><span lang="java">listener</span> may optionally be passed in to this call to be notified of success <span lang="default">or failure</span><span lang="ruby"></span> of the operation. <span lang="default">If the channel is @initialized@ (i.e. no attempt to attach has yet been made for this channel), then calling @publish@ will implicitly attach the channel.</span><span lang="jsall,ruby,swift">When publish is called with this client library, it "won't attempt to implicitly atach to the channel":#transient-publish.</span>
```
mattheworiordan: I realise you've done a great job at keeping track of which libs have implemented these features, but I am concerned that this is going to lead to a lot of issues down the line.  I would almost prefer that we instead document the behaviour by default as this is 1.1 docs, and have a caveat (perhaps a p.note or some unique styling) that indicates this feature is not yet supported in lib X.

tomczoink:
For upkeep's sake I've removed the various checks of language, and replaced it with links to downloads, which is where the matrix will go I believe. @mattheworiordan .

content/realtime/channels.textile
```diff
@@ -205,15 +205,15 @@ bc[swift]. realtime.channels.get("chatroom").attach { error in
   print("'chatroom' exists and is now available globally in every data center")
 }
 
-Clients attach to a channel in order to participate on that channel in any way (either to publish, subscribe, or be present on the channel).
+Clients attach to a channel in order to participate on that channel in any way (either to publish, subscribe, or be present on the channel). It is worth noting that some libraries allow for "publishing to a channel without attaching to it":#transient-publish. "See if your chosen SDK supports this":/https://www.ably.io/download.
```
mattheworiordan: This link is broken, it has a `/` prefix

content/realtime/channels.textile
```diff
@@ -584,7 +610,7 @@ bq(definition#publish-msg-array).
...
 
-Publish several messages on this channel. A <span lang="default">callback</span><span lang="java">listener</span><span lang="ruby">block</span> may optionally be passed in to this call to be notified of success <span lang="default">or failure</span><span lang="ruby"></span> of the operation. If the channel is @initialized@ (i.e. no attempt to attach has yet been made for this channel), then calling @publish@ will implicitly attach the channel.
+Publish several messages on this channel. A <span lang="default">callback</span><span lang="java">listener</span><span lang="ruby">block</span> may optionally be passed in to this call to be notified of success <span lang="default">or failure</span><span lang="ruby"></span> of the operation. <span lang="default">If the channel is @initialized@ (i.e. no attempt to attach has yet been made for this channel), then calling @publish@ will implicitly attach the channel.</span><span lang="jsall,ruby,swift">When publish is called with this client library, it "won't attempt to implicitly atach to the channel":#transient-publish.</span>
```
mattheworiordan: Typo, `atach` > `attach`

content/realtime/channels.textile
```diff
@@ -805,7 +831,7 @@ bq(definition).
...
+As a convenience, <span lang="default">@attach()@</span><span lang="csharp">@Attach()@</span><span lang="ruby">@attach@</span> will be called implicitly if either "<span lang="default">@publish()@</span><span lang="csharp">@Publish()@</span><span lang="jsall,ruby,swift"></span>":#publish <span lang="default">or</span> <span lang="jsall,ruby,swift"></span> "<span lang="default">@subscribe()@</span><span lang="csharp">@Subscribe()@</span><span lang="ruby">@subscribe@</span>":#subscribe for the @Channel@ is called, or "<span lang="default">@enter()@</span><span lang="csharp">@Enter()@</span><span lang="ruby">@enter@</span>":/realtime/presence#enter or "<span lang="default">@subscribe()@</span><span lang="csharp">@Subscribe()@</span><span lang="ruby">@subscribe@</span>":/realtime/presence#subscribe is called on the "@Presence@":/realtime/presence for this @Channel@.
```
mattheworiordan: This is wrong.  It won't be called implicitly now if publish is called?

content/rest-api/beta.textile
```diff
@@ -0,0 +1,187 @@
...
+
+The REST API has been extended to include batch-mode publish operations. The API concerned is:
+
+h6. POST /messages
+
+This is the new bulk publish API which is implicitly a batch-mode API. This performs a publish in parallel of messages to channels specified in the @request@ body. Check out the code example below on how to do this with the "@request@":/rest/usage#request method in the REST client library.
+
+```[javascript](code-editor:rest/batch-publish)
+var ablyRest = new Ably.Rest({ key: '{{API_KEY}}' })
+ablyRest.request('post', '/messages', null, { "channels": [ "test1", "test2" ], "messages": { "data": 'myData' } },
```
mattheworiordan: Can we wrap this payload or make it a separate var as this won't display in the code editor due to width

tomczoink:
Updated per feedback @mattheworiordan . Originally these were 5 different PRs which have been reviewed separately but have now been merged together for the release today, thus the size of this final PR.

I've created a task on asana for myself to go through all the docs once we're deployed.

PR #587, opened 2018-12-17
JS: Document that exceptions in listeners are caught

SimonWoolf:


tomczoink:
@mattheworiordan does this look good to you to merge?

tomczoink:
@mattheworiordan

SimonWoolf:
I'm going to go ahead and merge this before it falls too far behind

content/realtime/channels.textile
```diff
@@ -680,6 +680,7 @@ h4. Considerations
...
 * The registered <span lang="default">listener</span><span lang="ruby">block</span><span lang="csharp">handler</span> remains active on the channel regardless of the current channel state. For example, if you call @subscribe@ when a channel is @attached@ and it later becomes @detached@ or even @failed@, when the channel is reattached and a message is received, the <span lang="default">listeners</span><span lang="ruby">blocks</span><span lang="csharp">handlers</span> originally registered will still be invoked. Listeners are only removed when calling "@unsubscribe@":#unsubscribe or when a channel is @released@ using the <span lang="default">@Realtime.channels.release(name)@</span><span lang="csharp">@Realtime.Channels.Release(name)@</span> method
+<div lang="jsall">* If an exception is thrown in the subscribe listener and bubbles up to the event emitter, it will be caught and logged at @error@ level, so as not to affect other listeners for the same event</div>
```
mattheworiordan: Why is this only for JS?  Shouldn't all SDKs do this @SimonWoolf?

SimonWoolf: I've no idea whether that statement is true in other languages, different languages have different concurrency models and different exception-handling idioms. E.g. in Elixir the issue wouldn't apply; they'd be delivered to the registered subscriber process Pid, and it'd be up to that process to do something with them and handle any errors that arise in that. Probably something similar in Go, messages are sent down a channel. I do know that it's the case in js and that it can be surprising, since we had a customer using the js lib on intercom complain that they were surprised by it, so I added it for that.

mattheworiordan: I think in all cases this should happen according to https://docs.ably.io/client-lib-development-guide/features/#RTE6.  So if Go is an exception, then fine, but the norm is that every SDK catches exceptions and emits errors.  If this is useful for JS devs, it's useful for all devs.

SimonWoolf: https://github.com/ably/docs/commit/de4d939be6af258e66deb9eb1ef5d19218161d78

PR #593, opened 2019-01-07
android push tutorial textile

unnivm:
I have just committed my initial version of android push development tutorial. I am doing this because to save my work.


unnivm:
Hello @Srushtika ,

Since I could not set up the server locally for viewing  tutorials, I am not sure how it appears at my end. But based on your feedback and after going through the other tutorials, I have corrected most of the issues. Also, I have added the required images as well and made a commit. These are the summary of changes I made:

- Corrected most of the feedback 
- Added images
- Updated tutorial document

Please verify it and let me know.

PR #596, opened 2019-01-14
Directly register ios push

cjbatin:


Srushtika:
@cjbatin just a few other comments/queries:

- Have you added a visual example for every category of notification as per point 12 in the brief?
- Please revisit the 'Further Reading' section from the brief and add those points in the same section of the tutorial. Currently it seems to only have a link to the GitHub repo with the complete code.

content/tutorials/ios-push-tutorial-server-registration.textile
```diff
@@ -0,0 +1,64 @@
...
+
+<%= partial 'shared/tutorials/_setup-ngrok' %>
+
+h2(#step8). Step 8 - Testing the app
+
+<%= partial 'shared/tutorials/_testing_push_ios' %>
+
+<a href="https://github.com/ably/tutorials/pull/54/commits/fecb121e4b66cbb6e3a8ce632f393f18179bf328">See this step in Github</a>
+
+h2(#furtherReading). Further Reading
```
Srushtika: @cjbatin please check the brief document. It has specific details of what you need to include in this section. Please update it with the info.

content/shared/tutorials/_testing_push_ios.textile
```diff
@@ -0,0 +1,103 @@
...
+      console.log('Unable to publish message with push notification; err = ' + err.message);
+      return;
+    }
+    console.log('Message with push notification published');
+    res.send("Push Sent");
+  });
+})
+```
+
+Restart the server and go to http://localhost:3000/push/channel, you should receive a notification on your device. Notice how this notification has both a title and a body. You can do this on all three types of push notifications mentioned earlier, if you wish.
```
Srushtika: @cjbatin you've mentioned that users can implement the visual part in the previous types as well, please explain what's causing the title and the body to show up in the third example - that they could include in the previous ones as well.

PR #598, opened 2019-01-15
Remove docs for PushChannelSubscriptions.get

SimonWoolf:
Not sure where this came from. It's nowhere to be found in the features spec (https://docs.ably.io/client-lib-development-guide/features/#RSH1c), none of the client libraries implement it, realtime doesn't implement an endpoint for it, and since PushChannelSubscriptions have no id, there's nothing 'get' can do except search by params with the sub properties, which you can do with `PushChannelSubscriptions#list(sub)`.


mattheworiordan:
@SimonWoolf thanks, I think we can merge this.  @tomczoink can you merge into 1.1 release?

PR #601, opened 2019-01-21
V1.1 spec additions

tomczoink:
Adds V1.1 spec items to the docs.


content/types/_client_options.textile
```diff
@@ -83,3 +85,5 @@ blang[php].
...
+
+<!-- - <div lang="jsall,java,ruby"><span lang="default">logExceptionReportingUrl</span><span lang="ruby">:use_binary_protocol</span></div> := Defaults to a string value for an Ably error reporting Data Source Name. Typically of the form @https://[KEY]:[SECRET]@errors.ably.io/[ID]@. TODO: Once implemented, insert which values disable this.<br>__Type: @String@__ -->
```
SimonWoolf: - `<span lang="ruby">:use_binary_protocol</span>`?
- Not sure why you've gated this to js, java, and ruby; none of those three implement this. AFAIK the only library that does is ably-ios. Gate to that (i.e. swift,objc) and have a look at the implementation of that to see what it needs to be set to to disable it

content/types/_error_info.textile
```diff
@@ -10,3 +10,5 @@ h4.
 - <span lang="default">statusCode</span><span lang="ruby">status_code</span><span lang="csharp">StatusCode</span> := HTTP Status Code corresponding to this error, where applicable<br>__Type: @Integer@__
 
 - <span lang="default">message</span><span lang="csharp">Message</span> :=  Additional message information, where available<br>__Type: @String@__
+
+- <span lang="default">cause</span><span lang="csharp">Cause</span> :=  Information pertaining to what caused the error<br>__Type: @String@__
```
SimonWoolf: `cause` if present is another ErrorInfo; see the IDL

content/types/_message.textile
```diff
@@ -17,6 +17,15 @@ h4.
...
 - <div lang="swift">data</div> := The message payload, if provided<br>__Type: @String@, @NSData@, @Dictionary@, @Array@__
 
+- <div lang="jsall">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @JSON Object@__
+- <div lang="java">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @JSONObject@, @JSONArray@__
+- <div lang="csharp">Extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: plain C# object that can be converted to Json__
+- <div lang="ruby">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @Hash@, @Array@__
+- <div lang="python">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @Dict@, @List@__
+- <div lang="swift">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @Dictionary@, @Array@__
+- <div lang="objc">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @NSDictionary *@, @NSArray *@__
+- <div lang="php">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @Associative Array@, @Array@__
```
SimonWoolf: Same description copied 8 times? This can be all one line, something like
`<span lang="csharp">Extras</span><span lang="default">extras</span := Metadata and/or ancillary payloads, if provided<br>__Type: <span lang="jsall">@Object@</span><span lang="java">... etc.`

content/types/_presence_message.textile
```diff
@@ -22,6 +22,15 @@ h4.
...
 
+- <div lang="java">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @JSONObject@, @JSONArray@__
+- <div lang="csharp">Extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: plain C# object that can be converted to Json__
+- <div lang="jsall">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @JSON Object@__
+- <div lang="ruby">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @Hash@, @Array@__
+- <div lang="python">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @Dict@, @List@__
+- <div lang="swift">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @Dictionary@, @Array@__
+- <div lang="objc">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @NSDictionary *@, @NSArray *@__
+- <div lang="php">extras</div> := Metadata and/or ancillary payloads, if provided<br>__Type: @Associative Array@, @Array@__
+
```
SimonWoolf: similarly

content/types/_client_options.textile
tomczoink: From what I can tell from our spec sheet, it's not currently implemented for any libraries, but is meant to be enabled for ruby, java and js. If this is wrong happy to change it, I'll check our ios lib 
 now.

SimonWoolf: OK but the user-facing documentation should reflect what's actually implemented, not what we might aspire to implement at some point

(btw I wouldn't read too much into the difference between `-` and `no` on the spreadsheet, in many cases it'll be more accident than a deliberate distinction between an item that should be implemented and one that shouldn't for that client lib)

content/realtime/channels.textile
```diff
@@ -205,7 +205,7 @@ bc[swift]. realtime.channels.get("chatroom").attach { error in
   print("'chatroom' exists and is now available globally in every data center")
 }
 
-Clients attach to a channel in order to participate on that channel in any way (either to publish, subscribe, or be present on the channel).
+Clients attach to a channel in order to <span lang="default">participate on that channel in any way (either to publish, subscribe, or be present on the channel)</span><span lang="jsall,ruby,swift">subscribe or be present on a channel. It is still possible to "publish without attaching":#transient-publish to a channel however</span>.
```
mattheworiordan: I fear this is "buggy" in that we'll have to remember to update this as we support transient publishing.  Could transient publishing not be added to our feature matrix, and this text could be generic enough instead to state where transient publishing is supported, an attach will not necessarily subscribe etc.  I think we could then link "See if your chosen SDK supports transient publishing" for example.  Once we upgrade to 1.2, we could probably then remove that completely and assume all support it in full.

content/realtime/connection.textile
```diff
@@ -108,13 +108,15 @@ A series of connection states is defined as follows:
...
 
+- failed := This state is entered if the client library encounters a failure condition that it cannot recover from. This may be a fatal connection error received from the Ably service (e.g. an attempt to connect with an incorrect API key), or some local terminal error (e.g. the token in use has expired and the library does not have any way to renew it).<br><br>In the failed state, no reconnection attempts are made automatically by the library, and clients may not publish messages. A new connection attempt can be triggered by an explicit call to "<span lang="default">@connect()@</span><span lang="ruby">@connect@</span><span lang="csharp">@Connect()@:#connect</span>":#connect on the @Connection@ object.
+
```
mattheworiordan: This is duplicated below.

content/realtime/messages.textile
```diff
@@ -422,6 +422,12 @@ h6(#data).
 
 The message payload, if provided.<br>__Type: <span lang="default">@String@, @StringBuffer@, @JSON Object@</span><span lang="java">@String@, @ByteArray@, @JSONObject@, @JSONArray@</span><span lang="csharp">@String@, @byte[]@, @plain C# object that can be serialized to Json@</span><span lang="ruby">@String@, @Binary@ (ASCII-8BIT String), @Hash@, @Array@</span><span lang="python">@String@, @Bytearray@, @Dict@, @List@</span><span lang="php">@String@, @Binary String@, @Associative Array@, @Array@</span><span lang="objc">@NSString *@, @NSData *@, @NSDictionary *@, @NSArray *@</span><span lang="swift">@String@, @NSData@, @Dictionary@, @Array@</span>__
 
+h6(#extras).
+  default: extras
+  csharp: Extras
+
+Metadata and/or ancillary payloads, if provided.<br>__Type: <span lang="java">@JSONObject@, @JSONArray@</span><span lang="csharp">plain C# object that can be converted to Json</span><span lang="jsall">@JSON Object@</span><span lang="ruby">@Hash@, @Array@</span><span lang="python">@Dict@, @List@</span><span lang="swift">@Dictionary@, @Array@</span><span lang="objc">@NSDictionary *@, @NSArray *@</span><span lang="php">@Associative Array@, @Array@</span>__
```
mattheworiordan: JSON not Json

content/rest/messages.textile
```diff
@@ -275,6 +275,12 @@ h6(#data).
 
 The message payload, if provided.<br>__Type: <span lang="default">@String@, @StringBuffer@, @JSON Object@</span><span lang="java">@String@, @ByteArray@, @JSONObject@, @JSONArray@</span><span lang="csharp">@String@, @byte[]@, @plain C# object that can be serialized to Json@</span><span lang="ruby">@String@, @Binary@ (ASCII-8BIT String), @Hash@, @Array@</span><span lang="python">@String@, @Bytearray@, @Dict@, @List@</span><span lang="php">@String@, @Binary String@, @Associative Array@, @Array@</span><span lang="objc">@NSString *@, @NSData *@, @NSDictionary *@, @NSArray *@</span><span lang="swift">@String@, @NSData@, @Dictionary@, @Array@</span>__
 
+h6(#extras).
+  default: extras
+  csharp: Extras
+
+Metadata and/or ancillary payloads, if provided.<br>__Type: <span lang="java">@JSONObject@, @JSONArray@</span><span lang="csharp">plain C# object that can be converted to Json</span><span lang="jsall">@JSON Object@</span><span lang="ruby">@Hash@, @Array@</span><span lang="python">@Dict@, @List@</span><span lang="swift">@Dictionary@, @Array@</span><span lang="objc">@NSDictionary *@, @NSArray *@</span><span lang="php">@Associative Array@, @Array@</span>__
```
mattheworiordan: JSOn not JSon

content/realtime/messages.textile
```diff
@@ -420,7 +420,13 @@ h6(#data).
...
+The message payload, if provided.<br>__Type: <span lang="default">@String@, @StringBuffer@, @JSON Object@</span><span lang="java">@String@, @ByteArray@, @JSONObject@, @JSONArray@</span><span lang="csharp">@String@, @byte[]@, @plain C# object that can be serialized to JSON@</span><span lang="ruby">@String@, @Binary@ (ASCII-8BIT String), @Hash@, @Array@</span><span lang="python">@String@, @Bytearray@, @Dict@, @List@</span><span lang="php">@String@, @Binary String@, @Associative Array@, @Array@</span><span lang="objc">@NSString *@, @NSData *@, @NSDictionary *@, @NSArray *@</span><span lang="swift">@String@, @NSData@, @Dictionary@, @Array@</span>__
+
+h6(#extras).
+  default: extras
+  csharp: Extras
+
+Metadata and/or ancillary payloads, if provided.<br>__Type: <span lang="java">@JSONObject@, @JSONArray@</span><span lang="csharp">plain C# object that can be converted to JSON</span><span lang="jsall">@JSON Object@</span><span lang="ruby">@Hash@, @Array@</span><span lang="python">@Dict@, @List@</span><span lang="swift">@Dictionary@, @Array@</span><span lang="objc">@NSDictionary *@, @NSArray *@</span><span lang="php">@Associative Array@, @Array@</span>__
```
mattheworiordan: @SimonWoolf @paddybyers aren't there restrictions in terms of what is allowed in `extras`?  Also, @tomczoink I think we should mention that extras is used for push notifications and link to that.

SimonWoolf: Yeah, this kinda makes it sound like it's just for any random arbitrary metadata that the user might want to send. We should make it clear that it's all reserved and should only be used for specific purposes and with specific fields that we define, of which currently the only one is `push`.

(This is the danger of taking the client lib spec and trying to expand it into user documentation -- from a client lib developer's POV, `extras` is arbitrary metadata, because client libs shouldn't be touching or validating it; but that doesn't mean anything's allowed from the user's POV)

content/client-lib-development-guide/features.textile
```diff
@@ -50,7 +50,7 @@ h2(#test-guidelines). Test guidelines
...
-* @(G4)@ All REST requests and "WebSocket":/concepts/websockets connections to Ably must include the current API version @1.0@. Should any new API version with breaking changes be released, the client library will continue to use the API version explicitly requested
+* @(G4)@ This spec defines API version 1.1. A client library must identify to Ably the version of the spec it uses in all requests and connections, per "RSC7a":#RSC7a and "RTN2F":#RTN2F. The spec it uses is defined as the latest API version for which the library implements all spec items relating to the wire protocol
```
SimonWoolf: RSC7a and RTN2f also refer to 1.0 need to be changed (maybe just change them to say 'the API version per `G4`' or something so it's only specified in one place)

(Also whatever needs changing to make the version dropdown in the top right say 1.1, and link back to 1.0 as an old versions)

content/types/_error_info.textile
```diff
@@ -10,3 +10,5 @@ h4.
 - <span lang="default">statusCode</span><span lang="ruby">status_code</span><span lang="csharp">StatusCode</span> := HTTP Status Code corresponding to this error, where applicable<br>__Type: @Integer@__
 
 - <span lang="default">message</span><span lang="csharp">Message</span> :=  Additional message information, where available<br>__Type: @String@__
+
+- <span lang="default">cause</span><span lang="csharp">Cause</span> :=  Information pertaining to what caused the error<br>__Type: @ErrorInfo@__
```
SimonWoolf: ", where available"

PR #608, opened 2019-02-06
Add push vs realtime

tomczoink:
I've thrown together a simple version of the Push vs Realtime. We can create nicer styling on the website for this, however this should give you a rough idea of how it'll come out.

Seen in https://ably-docs-pr-608.herokuapp.com/compare/realtime-vs-push/.


remotekieran:
Looks good. Few comments on styling:

* Can we make the main headings anchors so we can link to them from the initial list at the start of the comparison
* If possible, let's mimic the table styling of the Ably v competitor pages. [For example](https://www.ably.io/compare/ably-vs-realtime). If not, then I'm happy with standard styling. Note: the current table under 'How mature is the feature set?' is broken.

Apart from that I'm happy with everything else.

tomczoink:
Thanks for quick points @remotekieran . Anchors should be in, I've added them to the contents list now.

Table is now fixed, I'll see what I can do with styling the table similarly to the existing ones on the website, should be simple enough.

PR #616, opened 2019-02-13
Linked feature matrix page from docs usage pages

Srushtika:


content/rest/usage.textile
```diff
@@ -178,6 +178,8 @@ blang[objc,swift].
   let rest = ARTRest(key: apiKey)
   ```
 
+Visit the "Feature Support Matrix":https://www.ably.io/feature-support-matrix to see the list of all the available features.
```
mattheworiordan: These links are wrong

mattheworiordan: Also, please reconsider the text based on my comment on the download page, aaand consider why we're using capitalisation here.

PR #631, opened 2019-02-28
ATM websocket Disconnected 50x handling 

mattheworiordan:
Make it clearer that a DISCONNECTED response with 50x should necessitate the use of a fallback host 

Note this is needed for ATM at the point the fallback hosts are pointing to the new cluster, and the placement constraint has been set on the old cluster.


mattheworiordan:
@paddybyers can we merge?

PR #636, opened 2019-03-06
Small lifecycle updates

mattheworiordan:
Whilst looking at lifecycle docs, I noticed a lot of small issues.

@Srushtika as a lot of this work was yours, can you look at each commit please and consider this for newer docs?

Also, some of these changes should be applied to the existing other tutorials.  

I've not done an exhaustive review btw.


mattheworiordan:
> Regarding the index number for tutorials or any textile files in general, is there a strategy to be followed? 

Just look at the nav on the left hand side.  The index numbers govern where they are positioned, so you can use these numbers to control placement of the pages in the nav.

mattheworiordan:
> Thanks, I've looked at all the commits, will follow this on in the future.

@Srushtika OK, will you merge these changes in then as I believe some of this applies to your push release anyway?

PR #663, opened 2019-04-29
update history docs to use PaginatedRequestParams

withakay:
update history docs to use PaginatedRequestParams for History calls (breaking change made in v1.1

See https://github.com/ably/ably-dotnet/pull/323 for more details


withakay:
@mattheworiordan sorry I missed those files, I have pushed 2 commits to cover everything now (I hope).

Regarding StatsRequestParams, that class still exists as a subclass of PaginatedRequestParams, probably best to keep discussion of that on the original PR here https://github.com/ably/ably-dotnet/pull/323

PR #666, opened 2019-05-02
Add http streaming api documentation

SimonWoolf:
Initial rough-and-ready version of SSE / HTTP streaming api documentation.

### DIFF
--- a/content/sse/index.textile
+++ b/content/sse/index.textile
@@ -26,12 +26,13 @@ h2(#getting-started). Getting Started
 SSE is incredibly simple to get started with. The code sample below provides an example of how to use it with Ably.
 
 ```[javascript](code-editor:sse/sse)
-var url = "https://realtime.ably.io/event-stream?v=1.1&key={{API_KEY}}&channels=myChannel";
+var key ='{{API_KEY}';
+var url ='https://realtime.ably.io/event-stream?channels=myChannel&v=1.1&key' + key;
 var eventSource = new EventSource(url);
 
 eventSource.onmessage = function(event) {
   var message = JSON.parse(event.data);
-  console.log("Message: " + message.name + " - " + message.data);
+  console.log('Message: ' + message.name + ' - ' + message.data);
 };
 ```
 
@@ -73,56 +74,69 @@ See an example of a "plain event stream":#event-stream below, except instead of 
 
 Keepalive packets are sent as SSE comments (@:keepalive@).
 
-h5. Example using curl
-
-```[sh]
-curl "https://rest.ably.io/sse?key={{API_KEY}}&channel=example&v=1.1"
-⏎
-id: cbfKayrzgAXDWM:1556806691343-0
-event: message
-data: {"id":"YqigX7VFsR:0:0","name":"foo","timestamp":1556806691341,"encoding":"json","channel":"channel","data":"{\"foo\":1}"}
-⏎
-:keepalive
-⏎
-event: error
-data:{"message":"Token expired. (See https://help.ably.io/error/40142 for help.)","code":40142,"statusCode":401,"href":"https://help.ably.io/error/40142"}
-```
-
-h5. Example using a library
+h5. Code example
 
 ```[javascript](code-editor:sse/sse)
-var url = "https://realtime.ably.io/event-stream?v=1.1&key={{API_KEY}}&channels=myChannel";
+var apiKey = '{{API_KEY}}';
+var url = 'https://realtime.ably.io/event-stream?channels=myChannel&v=1.1&key=' + apikey;
 var eventSource = new EventSource(url);
 
 eventSource.onmessage = function(event) {
   var message = JSON.parse(event.data);
-  console.log("Message: " + message.name + " - " + message.data);
+  console.log('Message: ' + message.name + " - " + message.data);
 };
 ```
+
 ```[python]
 import json
 import sseclient
 
-channels = 'date-channel,time-channel'
-url = 'https://realtime.ably.io/sse?v=1.1&key={{API_KEY}}&channels=yourChannel'
+api_key='{{API_KEY}}'
+url = "https://realtime.ably.io/sse?channels=myChannel&v=1.1&key=%s" % (api_key)
 
 def with_urllib3(url):
-    import urllib3
-    http = urllib3.PoolManager()
-    return http.request('GET', url, preload_content=False)
+  import urllib3
+  http = urllib3.PoolManager()
+  return http.request('GET', url, preload_content=False)
 
-response = with_urllib3(url)  
+response = with_urllib3(url)
 client = sseclient.SSEClient(response)
 for event in client.events():
-    message = json.loads(event.data)
-    print("Channel: %s  - Message: %s  - %s " % (message['channel'], message['name'], message['data']))
+  message = json.loads(event.data)
+  print("Channel: %s - Message: %s - %s " % (message['channel'], message['name'], message['data']))
 ```
 
-h3(#event-stream). Plain event stream
+```[curl]
+curl "https://rest.ably.io/sse?channel=example&v=1.1" \
+  --user "{{API_KEY}}"
+⏎
+id: cbfKayrzgAXDWM:1556806691343-0
+event: message
+data: {
+  "id":"YqigX7VFsR:0:0",
+  "name":"foo",
+  "timestamp":1556806691341,
+  "encoding":"json",
+  "channel":"channel",
+  "data":"{\"foo\":1}"
+}
+⏎
+:keepalive
+⏎
+event: error
+data:{
+  "message":"Token expired. (See https://help.ably.io/error/40142 for help.)",
+  "code":40142,
+  "statusCode":401,
+  "href":"https://help.ably.io/error/40142"
+}
+```
+
+h3(#event-stream). Plain HTTP event stream
 
 h6. GET realtime.ably.io/event-stream
 
-Start a streaming HTTP request.
+Start a streaming HTTP request, for ease of consuming with any HTTP library that supports streaming. When possible, we recommend using an SSE library as opposed to a raw HTTP stream as SSE libraries automatically handle reconnecting and resuming from the last received ID.
 
 h5. Request parameters
 
@@ -161,18 +175,47 @@ Non-string payloads will be encoded as before, but without enveloping you will n
 
 Note that failures on opening the connection (for example, invalid authentication details) may be sent as a non-streamed http response (with a response body of the form @{"error": <ErrorInfo>}@), not an @error@ event in a streamed response.
 
-h5. Curl example
-
-bc[sh](code-editor:sse/eventstream). curl "https://rest.ably.io/event-stream?key={{API_KEY}}&channel=example&v=1.1"
-  {"id":"cbfKayrzgAXDWM:1556804156735-0","event":"message","data":{"id":"oZs6XaGYx8:0:0","name":"message-name","timestamp":1556804156730,"encoding":"json","channel":"example","data":"{\"foo\":1}"}}
-  {"event":"error","data":{"message":"Token expired. (See https://help.ably.io/error/40142 for help.)","code":40142,"statusCode":401,"href":"https://help.ably.io/error/40142"}}
-
 h5. Code example
 
-```[javascript](code-editor:sse/eventstream)
-xhttp.onreadystatechange = function() {
-  console.log(xhttp.responseText);
-};
-xhttp.open("GET", "https://realtime.ably.io/event-stream?v=1.1&key={{API_KEY}}&channels=yourChannel");
-xhttp.send();
+```[curl]
+curl "https://rest.ably.io/event-stream?channel=example&v=1.1" \
+  --user "{{API_KEY}}"
+
+{
+  "id":"cbfKayrzgAXDWM:1556804156735-0",
+  "event":"message",
+  "data":{
+    "id":"oZs6XaGYx8:0:0",
+    "name":"message-name",
+    "timestamp":1556804156730,
+    "encoding":"json",
+    "channel":"example",
+    "data":"{\"foo\":1}"
+  }
+}
+⏎
+{
+  "event":"error",
+  "data":{
+    "message":"Token expired. (See https://help.ably.io/error/40142 for help.)",
+    "code":40142,
+    "statusCode":401,
+    "href":"https://help.ably.io/error/40142"
+  }
+}
+```
+
+```[nodejs]
+const request = require('request');
+const apiKey = '{{API_KEY}}';
+const url = 'https://realtime.ably.io/event-stream?channels=myChannel&v=1.1&key=' + apikey;
+
+request
+  .get(url)
+  .on('response', function(response) {
+    console.log(response.statusCode) // stream started, 200
+  })
+  .on('data', function(data) {
+    console.log('Envelope:', data.toString())
+  });
 ```
--- a/content/sse/index.textile
+++ b/content/sse/index.textile
@@ -13,7 +13,7 @@ jump_to:
 
 The Ably SSE and raw HTTP streaming API provides a way to get a realtime stream of events from Ably in circumstances where using a full Ably Realtime client library, or even an MQTT library, is impractical.
 
-HTTP streaming allows for a request from a client to be held by a server, allowing it to push data to the client without further requests. This, much like WebSockets, help avoid the overhead involved in normal HTTP requests. "Server-sent events":https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events (SSE) provide a thin layer on top of HTTP streaming. A common use of SSE is through the use of "EventSource":https://developer.mozilla.org/en-US/docs/Web/API/EventSource on web pages.
+HTTP streaming allows for a request from a client to be held by a server, allowing it to push data to the client without further requests. This, much like WebSockets, help avoid the overhead involved in normal HTTP requests. "Server-sent events":https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events (SSE) provide a thin layer on top of HTTP streaming. A common use of SSE is through the use of "the EventSource API":https://developer.mozilla.org/en-US/docs/Web/API/EventSource in all modern web browsers.
 
 p(tip). This feature is currently in beta, however is ready for production loads. If you're interested in using this feature, or have suggestions on how we can improve it, we would like you to "get in touch":https://www.ably.io/contact.
 
@@ -37,7 +37,7 @@ eventSource.onmessage = function(event) {
 
 h2(#authentication). Authentication
 
-Typically token authentication should be used, using a "token issued from your server":/core-features/authentication#token-request-process using an Ably SDK and an API key. This can then be passed to Ably using @token@ or an @Authorization: Bearer <base64-encoded token>@ header. 
+Typically token authentication should be used, using a "token issued from your server":/core-features/authentication#token-request-process using an Ably SDK and an API key. This can then be passed to Ably using a querystring @token@ parameter or an @Authorization: Bearer <base64-encoded token>@ header.
 
 It is possible to use an API key as authentication, however for "security reasons":https://support.ably.io/support/solutions/articles/3000038607 that is rarely advisable. If you do wish to use an API key however, you can use a querystring parameter of @key@ or an @Authorization: Basic <base64-encoded key>@ header. See "REST API authentication":/rest-api#authentication for more information.
 
@@ -71,7 +71,7 @@ h5(#envelope-format). Envelope format
 
 See an example of a "plain event stream":#event-stream below, except instead of a JSON object with @id@, @event@, @data@ members, you get an SSE event.
 
-Keepalive packets are sent as sse comments (@:keepalive@).
+Keepalive packets are sent as SSE comments (@:keepalive@).
 
 h5. Example using curl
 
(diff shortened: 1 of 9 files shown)


SimonWoolf:
@mattheworiordan tagged you as this is a good time if you have any feedback about the format, key names etc.

content/streaming-api/index.textile
```diff
@@ -0,0 +1,115 @@
...
+h3(#event-stream). Plain event stream
+
+h6. GET realtime.ably.io/event-stream
+
+Start a streaming HTTP request.
+
+h5. Request parameters
+
+- channels := (mandatory) one or more channel names, separated by commas (or the @separator@ if specified). Non-url-safe characters should be URL-encoded (for example, @?channels=foo%3Fbar@ will subscribe to the channel @foo?bar@). Alias: @channel@
+- separator := (optional) a separator, to enable easy subscriptions to channels with commas in their name. For example, @?channels=?separator=|&channel=fo,o|ba,r@ will subcribe to the two channels @fo,o@ and @ba,r@.
```
mattheworiordan: Typo:

` @?channels=?separator=|&channel=fo,o|ba,r@`

should be the following I think

`@?separator=|&channel=fo,o|ba,r@`

content/streaming-api/index.textile
```diff
@@ -0,0 +1,115 @@
...
+- v := (mandatory) the version of the api you are requesting. The current (and only) version of the API is 1.1, so only @v=1.1@ is accepted
+- key := (optional) an Ably API key to use, if using basic auth
+- accessToken := (optional) an Ably auth token to use, if using token auth
```
mattheworiordan: Can this not be `token`?  Would JWT work, I assume so?

content/streaming-api/index.textile
```diff
@@ -0,0 +1,115 @@
...
+- lastEvent := (optional) an @id@ to resume from, see "connection state recovery":#recovery below
+- enveloped := (optional, default true) if true, the @data@ from each event envelope for a @message@ event will be a "Message":/realtime/types#message object. If false, it will be the payload from the message directly. See "Envelope format":#envelope-format below
+- heartbeats := (optional, default false) if @true@ will use an explicit heartbeat event rather than a newline as a keepalive packet
```
mattheworiordan: I thought we were going to support fields so that subscribers could select only the fields they want sent.

content/streaming-api/index.textile
```diff
@@ -0,0 +1,115 @@
...
+h5(#envelope-format). Envelope format
+
+Once a streaming response is established, every line (other than empty lines sent as keepalive packets) will be a simple JSON object of the following form:
+
+bc[json]. {
+  event: <string, the event type, such as "message", "presence", "error", or "heartbeat">,
+  data: <string, the message data, see below. Not present for some event types,
+         e.g. heartbeats>,
+  id: <string, the id to use to resume from this point, see 'connection state
+       recovery' below. Only presence for message and presence events>
```
mattheworiordan: Only **present**

content/streaming-api/index.textile
SimonWoolf: Yeah, I just didn't do it in the MVP (wasn't essential).  I'll add it at some point, maybe as part of the work I do next week

content/streaming-api/index.textile
SimonWoolf: Auth is processed by the same code (AuthMgr) whether for rest, websocket, or this. So JWT will work fine, and we can add `token` as an alias of `accessToken` (and `access_token`) if we want as long as nothing else uses that param (and I don't imagine anything does), sure

mattheworiordan: > So JWT will work fine, and we can add token as an alias of accessToken

Ok, sounds good.  Given we have `key` as opposed to `accessKey`, I think using `token` is preferable

tomczoink:
We should update our adapters page https://www.ably.io/adapters once this is merged.

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
+---
+title: SSE and Raw HTTP Streaming API
+section: streaming-api
```
mattheworiordan: Does this appear in the nav anywhere?  I can't see it in https://ably-docs-pr-666.herokuapp.com/

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+  API Routes:
+    - Server-sent events#sse
+    - Plain event stream#event-stream
+---
+
+The Ably SSE and raw HTTP streaming API provides a way to get a realtime stream of events from Ably in circumstances where using a full Ably Realtime client library, or even an MQTT library, is impractical.
+
+HTTP streaming allows for a request from a client to be held by a server, allowing it to push data to the client without further requests. This, much like WebSockets, help avoid the overhead involved in normal HTTP requests. "Server-sent events":https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events (SSE) is another method of achieving this for web pages, usually done through the use of "EventSource":https://developer.mozilla.org/en-US/docs/Web/API/EventSource.
+
+This feature is currently in beta. If you're interested in using this feature, or have suggestions on how we can improve it, please "get in touch":https://www.ably.io/contact.
```
mattheworiordan: Perhaps:

This feature is currently in beta, however is ready for production loads. If you're interested in using this feature, or have suggestions on how we can improve it, we would like you to "get in touch":https://www.ably.io/contact.

mattheworiordan: I also think this should appear in a note i.e. on the right.  I don't think it belongs in the middle of the content.

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+
+It is subscribe-only: you can not interact with the channel, including to publish, enter presence, query the presence set, attach and detach from channels (without closing and re-opening the stream), or anything else.
+
+Where possible, we strongly recommend the use of one of our "Realtime client libraries":https://www.ably.io/download, which provide "more features and higher reliability":https://support.ably.io/solution/articles/3000061848, and the full use of our "normal realtime messaging API":/realtime.
```
mattheworiordan: Should we not say something like.  

"Customers who do not want to use a client library on platforms that support SSE, and only require simple subscribe-only streams, may choose to use SSE because it's an open standard, simple, and requires no SDKs on the client-side.  However, where possible, we strongly recommend..."

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+
+h2(#authentication). Authentication
+
+The same authentication options as used in the REST API also apply here. So for basic auth, a querystring parameter of @key@ or an @Authorization: Basic <base64-encoded key>@ header; for token auth, a querystring parameter of @accessToken@ containing an access token or an @Authorization: Bearer <base64-encoded token>@ header. See "REST API authentication":/rest-api#authentication for more information.
```
mattheworiordan: I don't think we should be recommending that customers deal with authentication via our raw REST API.  I think we should say that customers typically use a token that is issued from their servers using an Ably SDK and an API key.  It is also possible to simply use an API, however for security reasons that is rarely advisable. Find out why token authentication is preferred in almost any case.  In addition, Ably supports JWT tokens which an be included in the @token@ query param.  I would recommend we then link to a JSBin with examples of each please.  ALso, we decided on using @token@ as opposed to @accessToken@ I believe.

mattheworiordan: > The same authentication options as used in the REST API also apply here

Surely it's the same authentication options used by any clients connecting to Ably over any protocol apply here.  

Then we should link to find out more about Ably authentication, not linking to the raw REST API where customers are going to have to try and figure out how to generate tokens themselves, which is almost never correct.

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+
+
```
mattheworiordan: Extra line break

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+h2(#connection-state-recovery). Connection state recovery
+
+If you are disconnected, you can supply the @id@ of the last event you received as the @lastEvent@ querystring parameter of a new connection, to resume from that point.
```
mattheworiordan: This does not apply to SSE, at least in that SSE automatically does this.  This is very misleading as it makes a developer think they need to do this themselves.  This only applies when using XHR

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+
+Connection state is only retained for two minutes. See "Connection state explained":/realtime/connection#connection-state-explained for full documentation.
+
+Due to this feature being in beta, it is currently subject to other restrictions. For example, at the moment you can not resume from part-way through a backlog of messages that was being sent as part of a previous resume; you can only resume from a message (so you can't resume unless you've received at least one message on the connection); and you can only resume from the id of a message that was sent in the last two minutes (so it will not work if the last message sent on the connection was more than two minutes ago). We expect to fix all of these by the time this feature graduates out of beta.
```
mattheworiordan: Perhaps add "Get in touch" if you'd like to find out when these shortcomings will be fixed in producvtion.

content/sse/index.textile
mattheworiordan: I think at this point, we should tell people that SSE is incredibly simple, open, and here is a code sample showing how easy it is.  With a link to JSBin to try it out live.

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+
+The @/event-stream@ endpoint will give an SSE response if the @Accept@ header is set to @text/event-stream@. The @/sse@ endpoint is also provided as an easier way of forcing an SSE response.
+
+h5. Request parameters
+
+Identical to the request parameters for a plain event stream above.
+
+h5(#envelope-format). Envelope format
+
+See for a plain event stream above, except instead of a JSON object with @id@, @event@, @data@ members, you get an SSE event.
```
mattheworiordan: I don't follow this.  What do people nee to see above?

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+
+Keepalive packets are sent as sse comments (@:keepalive@).
```
mattheworiordan: Is this relevant to end users?  Not one for me to comment on, but do SSE clients even emit comments?

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+
+h6. GET realtime.ably.io/sse
+
+Start a streaming HTTP request that conforms to the "Server-Sent Events":https://www.w3.org/TR/eventsource/ spec, for ease of consuming with an SSE library.
+
+The @/event-stream@ endpoint will give an SSE response if the @Accept@ header is set to @text/event-stream@. The @/sse@ endpoint is also provided as an easier way of forcing an SSE response.
+
+h5. Request parameters
+
+Identical to the request parameters for a plain event stream above.
```
mattheworiordan: There is no plain event stream above?  Also, why should a developer need to look at another reference.  Just show the arguments here.  Let our customers be lazy, we should do the hard work of making these docs readable and easy to sue.

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+h5(#envelope-format). Envelope format
+
+See for a plain event stream above, except instead of a JSON object with @id@, @event@, @data@ members, you get an SSE event.
+
+Keepalive packets are sent as sse comments (@:keepalive@).
+
+h5. Example
+
+```[sh]
+curl "https://rest.ably.io/sse?key={{API KEY}}&channel=example&v=1.1"
```
mattheworiordan: This is a curl example, but people will be using clients and not accessing raw streams.  Please provide examples in other languages.

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+
+Start a streaming HTTP request.
+
+h5. Request parameters
+
+- channels := (mandatory) one or more channel names, separated by commas (or the @separator@ if specified). Non-url-safe characters should be URL-encoded (for example, @?channels=foo%3Fbar@ will subscribe to the channel @foo?bar@). Alias: @channel@.
+- separator := (optional) a separator, to enable easy subscriptions to channels with commas in their name. For example, @?separator=|&channel=fo,o|ba,r@ will subcribe to the two channels @fo,o@ and @ba,r@.
+- v := (mandatory) the version of the api you are requesting. The current (and only) version of the API is 1.1, so only @v=1.1@ is accepted.
+- key := (optional) an Ably API key to use, if using basic auth.
+- accessToken := (optional) an Ably auth token to use, if using token auth.
```
mattheworiordan: I think we agreed to use the shortcut @token@

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+
+h6. GET realtime.ably.io/event-stream
+
+Start a streaming HTTP request.
+
+h5. Request parameters
+
+- channels := (mandatory) one or more channel names, separated by commas (or the @separator@ if specified). Non-url-safe characters should be URL-encoded (for example, @?channels=foo%3Fbar@ will subscribe to the channel @foo?bar@). Alias: @channel@.
+- separator := (optional) a separator, to enable easy subscriptions to channels with commas in their name. For example, @?separator=|&channel=fo,o|ba,r@ will subcribe to the two channels @fo,o@ and @ba,r@.
+- v := (mandatory) the version of the api you are requesting. The current (and only) version of the API is 1.1, so only @v=1.1@ is accepted.
```
mattheworiordan: Could we have mandatory stuff at the top please

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+- key := (optional) an Ably API key to use, if using basic auth.
+- accessToken := (optional) an Ably auth token to use, if using token auth.
+- lastEvent := (optional) an @id@ to resume from, see "connection state recovery":#recovery below.
```
mattheworiordan: @SimonWoolf do we additionally support `rewind` now?

mattheworiordan: Is `lastEvent` supported for SSE, or more accurately, should we tell people it's not needed except when starting a new SSE connection that resumes from a previous connection?

mattheworiordan: This recovery link does not exist, and refers to a link below, which does not exist (it's above).

content/sse/index.textile
```diff
@@ -0,0 +1,119 @@
...
+
+If @enveloped@ is false, the @data@ will for a @message@ or @presence@ event be the @data@ payload from the @Message@ or @PresenceMessage@. Other events are unaffected.
+
+Non-string payloads will be encoded as before, but without enveloping you will not have the benefit of the @encoding@ field to tell you what encoding has been done.
+
+Note that failures on opening the connection (for example, invalid authentication details) may be sent as a non-streamed http response (with a response body of the form @{"error": <ErrorInfo>}@), not an @error@ event in a streamed response.
+
+h5. Example
+
+bc[sh]. curl "https://rest.ably.io/event-stream?key={{API KEY}}&channel=example&v=1.1"
```
mattheworiordan: We could easily provide a trivial example in a JSBin and inline that subscribes via XHR, and resumes automatically setting `lastEvent`

layouts/_nav.html.erb
```diff
@@ -4,6 +4,7 @@
   <%= nav_items 'realtime', :title => 'Realtime client lib' %>
   <%= nav_items 'rest', :title => 'REST client lib' %>
   <%= nav_items 'rest-api', :title => 'REST API' %>
+  <%= nav_items 'sse', :title => 'SSE' %>
```
mattheworiordan: I don't think this is appearing.

content/sse/index.textile
SimonWoolf: > Surely it's the same authentication options used by any clients connecting to Ably over any protocol apply here. Then we should link to find out more about Ably authentication

I don't follow you here. It's the same underlying protocol as our client libraries use sure, but that's not much help for someone who isn't a client library developer... Our normal authentication docs are written assuming you're using a client library, the only places the actual protocol is documented are the client library development guide, and the raw REST api docs

(Unless you're just saying we should copy & paste (or better transclude) the relevant bits from the rest api docs here, rather than linking?)

> , not linking to the raw REST API where customers are going to have to try and figure out how to generate tokens themselves, which is almost never correct.

Nothing about using the raw REST api means customers have to generate tokens by hand. The [rest api docs](https://www.ably.io/documentation/rest-api#token-authentication) just explain, once you've obtained a token (from a client lib on your auth server or otherwise), how to include it in the REST request you make so that we make (`Authorization` header or querystring param). Which is the the same as what someone using SSE or light-http-streaming will have to do

content/sse/index.textile
SimonWoolf: > do we additionally support rewind now?

Yes (currently just with a number though)

> Is lastEvent supported for SSE, or more accurately, should we tell people it's not needed except when starting a new SSE connection that resumes from a previous connection?

sgtm (if supplied it takes precedence over the SSE `headers['last-event-id']`)

content/streaming-api/index.textile
```diff
@@ -12,7 +12,9 @@ jump_to:
     - Server-sent events#sse
 ---
 
-The Ably raw HTTP streaming API provides a way to get a realtime stream of events from Ably in circumstances where using a full Ably Realtime client library, or even an MQTT library, is impractical.
+The Ably SSE + raw HTTP streaming API provides a way to get a realtime stream of events from Ably in circumstances where using a full Ably Realtime client library, or even an MQTT library, is impractical.
+
+HTTP streaming allows for a request from a client to be held by a server, allowing it to push data to the client without further requests. This, much like WebSockets, help avoid the overhead involved in normal HTTP requests. "Server-sent events":https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events (SSE) is another method of achieving this for web pages, usually done through the use of "EventSource":https://developer.mozilla.org/en-US/docs/Web/API/EventSource.
```
SimonWoolf: > Server-sent events (SSE) is another method of achieving this for web pages

@tomasmcm It's not really a different method, it's a thin layer on top of HTTP streaming. It's not limited to web pages either, eg we have SSE python and elixir examples in the tutorials repo

SimonWoolf:
@tomczoink do you want to update per latest round of feedback or would you like me to? (or any specific bits you want me to look at?)

tomczoink:
I'm happy to implement the feedback @SimonWoolf , I'd appreciate you giving it a review once I'm done though.

accesstoken:
Hi
I shouldn’t be involved in this conversation. Wrong person I think.

mattheworiordan:
Sorry @accesstoken, see https://github.com/ably/docs/pull/666#discussion_r292225722, should have wrapped that in a `code block`!

content/sse/index.textile
tomczoink: How is `rewind` to be defined @SimonWoolf ?

tomczoink:
@mattheworiordan @SimonWoolf I've implemented changes, a review on current state would be much appreciated.

Also, for the point:

> We don't explain why developers would use the raw HTTP Streaming API, and we should. Why would they use that over SSE?

What would be a reason to use HTTP Streaming over SSE? SSE not being available on the platform, or any other points to make?

SimonWoolf:
> What would be a reason to use HTTP Streaming over SSE? SSE not being available on the platform

basically that yup, if you have to do it from scratch yourself then parsing json is probs easier than implementing an sse client

content/sse/index.textile
SimonWoolf: `- rewind := (optional integer) if specified, will send a backlog of this many messages once the connection opens. For example, @rewind=1@ will give you the most recent message sent on the channel. This is best-effort — only messages from within the last two minutes will be available, and only if the channel has been continuously active since the message was sent; it is not a replacement for the "history API":/rest/history. It only has an effect for new connections; when "resuming a previous connection":#connection-state-recovery, it is ignored in favour of sending you the messages you missed since you were last connected`

mattheworiordan:
I've made quite a few small changes to these docs, please review individual commits and notes.

@tomczoink @remotekieran @SimonWoolf can you review please?

content/sse/index.textile
```diff
@@ -140,15 +140,15 @@ Start a streaming HTTP request, for ease of consuming with any HTTP library that
...
+- lastEvent := **optional**. An @id@ to resume from. Only required when starting a new SSE connection which resumes from a previous connection.
+- rewind := **optional**. An integer which, if specified, will send a backlog of the number of messages specified once the connection opens. For example, @rewind=1@ will give you the most recent message sent on the channel. This is best-effort — only messages from within the last two minutes will be available, and only if the channel has been continuously active since the message was sent; it is not a replacement for the "history API":/rest/history. It only has an effect for new connections; when resuming a previous connection using @lastEvent@, it is ignored in favour of sending you the messages you missed since you were last connected.
+- enveloped := **optional**. Default is **true**. If true, the @data@ from each event envelope for a @message@ event will be a "Message":/realtime/types#message object. If false, it will be the payload from the message directly. See "Envelope format":#envelope-format below.
+- heartbeats := ≈ if @true@ will use an explicit heartbeat event rather than a newline as a keepalive packet.
```
SimonWoolf: @tomczoink `≈`?

mattheworiordan:
@tomczoink FYI, I added `curl` as a new code type, so we need to make sure this works in the website when this is merged in.  It may not, I can't recall how the language mappings are shared (or not) between the website & the docs repo.

content/sse/index.textile
mattheworiordan: Why is this one not flagged as optional?

tomczoink: Fixed with https://github.com/ably/docs/pull/666/commits/2e53135582c4897a94af7a2cc966669c7dbf20c8.

mattheworiordan:
Minor request, I noticed in https://www.ably.io/tutorials/sse-and-http-streaming, that we refer to url incorrectly, it's an acronym, `URL`.  I notice we also sometimes refer to mqtt in lower case too.  Mind just doing a grep to fix those?

mattheworiordan:
@tomczoink don't worry about it, done in dc7e784

content/streaming-api/index.textile
paddybyers: Except that "access token" is widely known and accepted terminology, among the various other token types used in oauth schemes (client token, user token, refresh token etc). I don't see the advantage in adding an ambiguous `token` param.

mattheworiordan: >  I don't see the advantage in adding an ambiguous token param.

Brevity and consistency with our client SDKs which only refer to tokens as `token` or `tokenDetails`

paddybyers: So if we have both `accessToken` and `token`, which is the preferred or primary name, and which is the alias? In other situations we've argued against having multiple ways of doing the same thing, because it only creates confusion.

mattheworiordan: > In other situations we've argued against having multiple ways of doing the same thing, because it only creates confusion.

My preference is just `token`

Issue #671, opened 2019-05-07
Add public `Auth.currentTokenDetails`


paddybyers:
Some existing libraries expose the current `TokenDetails` publically whereas they are private in others. 

It is proposed that we add a public field `currentTokenDetails` to expose the `TokenDetails`, if any, that the library is using and considers to be valid.

mattheworiordan:
@paddybyers should it not just be `currentToken` to allow for JWT, token strings, and token objects?

SimonWoolf:
fwiw ably-js exposes token strings by wrapping them as a minimal tokenDetails object, i.e. `{token: <token>}`. (and jwt similarly since the client lib can't tell a token string from a jwt token). Probably a bit nice than having a property that might be a `tokenDetails` or a `string`, esp for strongly-typed langs

mattheworiordan:
Good point, probably better to have a `TokenDetails` object then./

SimonWoolf:
Fixed by https://github.com/ably/docs/pull/673

PR #673, opened 2019-05-20
Spec: add requirements for Auth.currentTokenDetails

paddybyers:
See https://github.com/ably/docs/issues/671


paddybyers:
> (I'd prefer `auth.tokenDetails` over `auth.currentTokenDetails` just because then ably-js is already compliant, if renaming it would need to support both to avoid a breaking change. But shrug)

@mattheworiordan any opinion either way?

mattheworiordan:
I would be happy with `tokenDetails` as opposed to `currentTokenDetails`.  `current` generally is probably unnecessary in most cases across the lib.

mattheworiordan:
Should we update this PR so it can be merged?  If we do merge it, are we expecting all client libs to implement, or is this going to form part of a new spec version?  If the latter, this PR is based on `master` which is wrong.

mattheworiordan:
Ping @paddybyers

SimonWoolf:
I updated the pr. no strong opinion on whether it should be modifying the current version or against a new version, other than a general inclination to default to modifing the current version given a change which doesn't break compatibility given we have no idea if/when we might want to go to 1.2

PR #674, opened 2019-05-21
Base fanout vs pubnub compare page

tomczoink:
For https://github.com/ably/website/issues/2427

Left a few comments in the google doc to be addressed, but otherwise this should be ready to for review.


EveHarris:
Just two things - 

For consistency there should be a full stop after 'unknown' 6th line down on first table - Is it globally fast, reliable etc 

Also as it is being published in May Please change 'March' to 'May' in first line of disclaimer: 

Disclaimer: All information is as accurate as possible as of publishing in March 2019.

PR #678, opened 2019-05-31
PHP list to list_ for push, remove () from functions

tomczoink:
For issues brought up in https://github.com/ably/ably-php/pull/76#discussion_r289413576. Main point is in the code it seems `Hash` should be `array` so I've changed it here, though if I've misunderstood that please let me know.


content/general/push/admin.textile
```diff
@@ -230,7 +230,7 @@ bq(definition).
   default: list(Object params, callback("ErrorInfo":/realtime/types#error-info err, "PaginatedResult":#paginated-result<"DeviceDetails":#device-details device> resultPage))
   ruby:    "Deferrable":/realtime/types#deferrable list(Hash params) -> yields "PaginatedResult":#paginated-result<"DeviceDetails":#device-details>
   python: "PaginatedResult":#paginated-result list(params=Object)
-  php: "PaginatedResult":#paginated-result list(Hash params)
+  php: "PaginatedResult":#paginated-result list_ (array params)
```
mattheworiordan: Why is there a space here?  Should it not be `list_(array params)` for consistency?

Issue #684, opened 2019-06-04
Batch / Transient publishing docs are inadequate


mattheworiordan:
Customers have asked about batch publishing a few times (see https://support.ably.io/a/tickets/100659 for example), and what I've noticed is that:

* We make almost no mention of it anywhere a developer may look for this. See the REST publish docs - https://www.ably.io/documentation/rest/messages#message-publish.  No mention of batch, bulk or transient publishing.  Same applies to realtime publish docs https://www.ably.io/documentation/realtime/messages#message-publish. 
* The channel docs https://www.ably.io/documentation/rest/channels at least mention the batch API, but in the getting started section.  So a user who finds the publish method, has no way of knowing how they can do batch/bulk operations, or how indeed they could consider using transient publishing at all.  The realtime channel docs are similar, mentioning transient publishing, but no mention of the batch API which has some advantages over transient publishing (one message to many channels for example).
* Going to the REST API docs at https://www.ably.io/documentation/rest-api makes no mention of batch/bulk publishing.  There is a small note about a beta API existing, but a) it's not obvious, b) we can't expect people to explore all our docs to find important info, c) we should mention the bulk/batch API near to where people may be looking/learning.
* Support FAQs are inadequate (non-existent largely) - see below

Whilst reviewing all of this, I saw lots of issues that I marked up:

<img width="876" alt="screenshot_2019-06-04_07-19-24_pm" src="https://user-images.githubusercontent.com/43789/58903797-29a4eb80-86fe-11e9-9e36-43eaa5cd9030.png">
<img width="862" alt="screenshot_2019-06-04_07-16-48_pm" src="https://user-images.githubusercontent.com/43789/58903798-29a4eb80-86fe-11e9-807b-c5ed60aa8a55.png">
<img width="938" alt="screenshot_2019-06-04_07-06-55_pm" src="https://user-images.githubusercontent.com/43789/58903799-2a3d8200-86fe-11e9-90a1-86e35ae309fd.png">
<img width="1038" alt="screenshot_2019-06-04_07-03-31_pm" src="https://user-images.githubusercontent.com/43789/58903800-2a3d8200-86fe-11e9-9cdb-f47d5845aaf0.png">
<img width="1037" alt="screenshot_2019-06-04_07-02-34_pm" src="https://user-images.githubusercontent.com/43789/58903801-2a3d8200-86fe-11e9-9404-16b32bc64e53.png">
[...]

PR #685, opened 2019-06-06
Extend batch docs

tomczoink:
For #684, largely involves mentioning batch publishing more extensively throughout the docs, as well as cleaning up some of the batch publish docs themselves.

Includes the addition of subscribe/publish sections in realtime/channels and history/publish in rest/channels to make the functionality even more obvious per feedback in the initial issue.

### DIFF
--- a/content/realtime/channels.textile
+++ b/content/realtime/channels.textile
@@ -14,12 +14,17 @@ api_separator:
 jump_to:
   Help with:
     - Getting started#getting-started
-    - Channels & lifecycle#channels
+    - Channels#channels
+    - Subscribing to a channel#subscribing
+    - Publishing to a channel#publishing
+    - Channel lifecycle#channel-lifecycle
     - Channel metadata#channel-metadata
-    - Channel states
-    - Handling failures
-    - Channel namespaces
-    - Presence#presence
+    - Implicit attach#implicit-attach
+    - Transient Publishing#transient-publish
+    - Channel states#channel-states
+    - Handling failures#handling-failures
+    - Channel namespaces#channel-namespaces
+    - Presence#presence-api
   Channel API properties:
     - state
     - errorReason#error-reason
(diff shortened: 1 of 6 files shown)


content/rest/channels.textile
```diff
@@ -18,6 +18,8 @@ jump_to:
   Help with:
     - Getting started#getting-started
     - Channels#channels
+    - Subscribing to a channel#subscribing
```
mattheworiordan: @tomczoink is this intentional?  REST lib with subscribing?  We could arguably have a note that states if you want to subscribe, use the realtime lib, useful... but this leads to nowhere I think?

content/realtime/channels.textile
```diff
@@ -255,11 +372,15 @@ channel.publish("action", data: "boom!")
 
 Normally, errors in attaching to a channel are communicated through the attach callback. For implicit attaches (and other cases where a channel is attached or reattached automatically, e.g. following the library reconnecting after a period in the @suspended@ state), there is no callback, so if you want to know what happens, you'll need to listen for channel state changes.
 
+h3(#batch-publish). Batch publishing
+
+It is common for a single message to be intended for multiple channels. With a realtime connection, you can effectively send a message to multiple channels at once by allowing multiple concurrent publish operations. If you wish to send a message to multiple channels within a single operation, you can make use of the "REST batch API":/rest-api/beta#batch.
```
paddybyers: Since realtime doesn't support batch publishing I do think having this heading is misleading. I realise you wanted to keep the headings consistent, but the APIs are different, and I don't think it's appropriate in this case.

Perhaps this section should literally be called "Publishing to multiple channels".

The the text could be:

Often it is necessary to publish a single message in multiple channels at the same time. In the realtime API, this is achieved simply by making multiple separate publish() requests, once in each of the channels in question, and the realtime protocol allows for those concurrent requests to be in-flight simultaneously, so the publish on one channel is not delayed waiting for completion of operations in other channels.

It is also possible to publish one or more messages into multiple channels in a single operation using the "REST batch API":/rest-api/beta#batch.

content/rest/channels.textile
```diff
@@ -197,7 +200,153 @@ bc[swift]. let key = ARTCrypto.generateRandomKey()
...
+```[go]
+  rest, err := ably.NewRestClient(ably.NewClientOptions("{{API_KEY}}"))
+  channel := rest.Channels.Get("{{RANDOM_CHANNEL_NAME}}", nil)
+  page, err := channel.History(nil)
+  fmt.Println("Last published message: %s\n", page.Messages[0].Data)
+```
+
+h3(#batch-publish). Batch publishing
+
+It is common for a single message to be intended for multiple channels. If you wish to send a message to multiple channels within a single operation, you can make use of the "REST batch API":/rest-api/beta#batch. With a "realtime":/realtime connection, you can also effectively send a message to multiple channels at once by allowing multiple concurrent publish operations.
```
paddybyers: I suggest removing the last sentence about realtime here.

content/rest/channels.textile
```diff
@@ -18,6 +18,8 @@ jump_to:
   Help with:
     - Getting started#getting-started
     - Channels#channels
+    - Obtaining history of a channel#channel-history
```
mattheworiordan: @tomczoink I am afraid you need to add `#publishing` here as it's missing

mattheworiordan: Also ,why is this obtaining history and getting history below?

tomczoink: Fixed both with https://github.com/ably/docs/commit/f0f9e2a19363fcb157ec7fde54dea03b1244f3ef.

PR #695, opened 2019-07-09
Add docs for unenveloped publish

SimonWoolf:


mattheworiordan:
LGTM, I agree a simple `curl` example would make sense here TBH.

SimonWoolf:
Added curl example https://github.com/ably/docs/pull/695/commits/7da5e86e77caa6ccd02d4de445899c80d6c74eae

Issue #697, opened 2019-07-10
Push docs method definition in API ref is wrong


Srushtika:
https://www.ably.io/documentation/general/push/activate-subscribe#push-channel

In push Activate/Subscribe docs (the Swift version), the method definition for `subscribeClient()` (correct) is mentioned as `subscribeClientId()`. This needs to be fixed. 

Moreover, the example shows 
```
realtime.channels.get("pushenabled:foo").subscribeClient { error
    // Check error.
}
```
However, the `subscribeClient method is available on the push object of a channel, so it really should be 

```
realtime.channels.get("pushenabled:foo").push.subscribeClient { error
    // Check error.
}
```

@tomczoink I can do this change quickly but assigning to you as you've done this before and may need to know/track this kind of info. Please do another round of verification before updating. Thanks!

Issue #696, opened 2019-07-09
Push 'Publishing notifications' page's API ref index is out of sync


Srushtika:
<img width="968" alt="Screenshot 2019-07-10 at 00 15 03" src="https://user-images.githubusercontent.com/5900152/60929109-dcd9a500-a2a7-11e9-84b4-b4a19c387b44.png">

See screenshot above, the index is missing methods like `activate()`

https://www.ably.io/documentation/general/push/activate-subscribe

PR #698, opened 2019-07-11
Correction on subscribeClientId usage

tomczoink:
This PR updates the 1.1 spec to make use of only `subscribeClient` over `subscribeClientId`, and also fixes some documentation which incorrectly makes use of `subscribeClientId`. For #697.

Also fixes `activate` method not appearing in TOC. This was due to a clashing href, which this PR fixes with https://github.com/ably/docs/pull/698/commits/7746a7823f3d07f31d3da7c1fbc6b6e123587310. For #696.

Fixes spelling error per #699.

### DIFF
--- a/content/types/_push_channel.textile
+++ b/content/types/_push_channel.textile
@@ -13,10 +13,10 @@ bq(definition).
 Subscribe your device to the channel's push notifications.
 
 h6(#subscribe-client-id).
-  default: subscribeClientId
+  default: subscribeClient
 
 bq(definition).
-  default: subscribeClientId()
+  default: subscribeClient()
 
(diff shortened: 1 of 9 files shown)


content/types/_push_channel.textile
```diff
@@ -13,10 +13,10 @@ bq(definition).
 Subscribe your device to the channel's push notifications.
 
 h6(#subscribe-client-id).
-  default: subscribeClientId
+  default: subscribeClient
```
paddybyers: should we update the anchors as well? ie `#subscribe-client-id` -> `#subscribe-client`

PR #702, opened 2019-07-15
Added site_title as metadata option

tomczoink:
This allows for the HTML title on the main website `www.ably.io` to be forced from docs pages. This allows for an intermediate point between using the full title of a page (often too verbose for this purpose) and simply using the url (not complex enough, can't pick up on correct capitalisation such as for WebRTC).

### DIFF
--- a/content/tutorials/ios-push-server-registration.textile
+++ b/content/tutorials/ios-push-server-registration.textile
@@ -1,12 +1,13 @@
 ---
 title: iOS Push Notifications Tutorial - Device registration via server
+site_title: iOS Push Notification Registration via Server
 section: tutorials
 index: 69
 ---
 
 Ably can deliver native Push Notifications to iOS devices using "Apple’s Push Notification service":https://developer.apple.com/notifications/. Native Push Notifications, unlike our "channel-based Pub/Sub messaging":/realtime/channels/, do not require the device to maintain a connection to Ably, as the underlying platform or OS is responsible for maintaining its own battery efficient transport to receive Push Notifications. Therefore, native Push Notifications are commonly used to display visual notifications to users or launch a background process for an app in a battery efficient manner.
 
-In this tutorial, we'll see how to set up and send Push Notifications to your iOS app using "Ably's Push Notification service":/general/push. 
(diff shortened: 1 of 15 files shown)


content/tutorials/ios-push-server-registration.textile
```diff
@@ -1,12 +1,13 @@
 ---
 title: iOS Push Notifications Tutorial - Device registration via server
+site_title: iOS Push Notification Registration via Server
```
mattheworiordan: Why not Tutorial?

tomczoink: On the site it automatically inserts `Tutorial / TITLE`, so felt it was overkill to include `Tutorial` twice.

content/tutorials/webhook-chuck-norris.textile
```diff
@@ -1,5 +1,6 @@
 ---
 title: WebHooks and Chuck Norris
+site_title: "WebHooks and Chuck Norris"
```
mattheworiordan: I see this is duplicated, which feels unnecessary.

Why don't we just encourage people to define a title once that works for all cases?  This approach is error prone as every new page created needs a site title defined, yet the author of this page has no way of knowing how that works with the website.

mattheworiordan:
I have added my comments, but I feel this is the wrong approach.  Why not instead encourage people to standardise how titles are defined in the docs so that they are used correctly on the website by default.  By having two different titles, without any explanation of why two titles are needed, we're going to end up having problems, especially as the team grows.

content/tutorials/ios-push-server-registration.textile
mattheworiordan: > so felt it was overkill to include Tutorial twice.

Ok, so how will an author of a page in the docs repo know what to expect on the website?  That is my concern with this approach.  I think we need to a) have a similar approach to titling here and on the website, b) have a simple guide in our docs on how to name pages.

tomczoink:
Sure makes sense, I'll finish my changes from the website side and make those make use of the normal `title` metadata, close this PR and open a new one when any titles which needed shortening/changing to better match the website styling 👍 .

mattheworiordan:
>  open a new one when any titles which needed shortening/changing to better match the website styling 

Sounds sensible.

PR #710, opened 2019-07-18
Push: ensure all LocalDevice details are deleted on deregistration

paddybyers:
Also: check deviceToken validity on each launch; handle failures to obtain a deviceToken.

Partially fixes: https://github.com/ably/realtime/issues/2544


tomczoink:
@paddybyers do you want to go with the version of RSH2d in https://github.com/ably/docs/pull/694/files or here?

tomczoink:
Confirmed with @paddybyers this is the version we'd prefer for RSH2d.

PR #711, opened 2019-07-22
Spec: fix situations in which presence auto-re-enter needs to happen

SimonWoolf:


SimonWoolf:
fyi @withakay

content/client-lib-development-guide/features.textile
```diff
@@ -571,18 +571,19 @@ h3(#realtime-presence). Presence
...
-* @(RTP17)@ The Presence object is also responsible for keeping a separate copy of an object logically equivalent to the @PresenceMap@ containing only members that match the current @connectionId@. Any incoming presence message that satisfies @RTP17b@ should be applied to this object in the same way as for the normal @PresenceMap@. This object should be private and is used to maintain a list of members that need to be automatically re-entered by the @Presence@ object when required to by @RTP5c2@.
+* @(RTP17)@ The Presence object should also keep a second @PresenceMap@ containing only members that match the current @connectionId@. Any incoming presence message that satisfies @RTP17b@ should be applied to this object in the same way as for the normal @PresenceMap@. This object should be private and is used to maintain a list of members that need to be automatically re-entered by the @Presence@ object when required to by @RTP17c@.
```
paddybyers: "incoming presence message" .. what does "incoming mean" ? Does it refer to messages received by the client on the realtime connection, or messages passed as arguments to an `enter()` or `update()` call?

content/client-lib-development-guide/features.textile
```diff
@@ -571,18 +571,19 @@ h3(#realtime-presence). Presence
...
+** @(RTP17c)@ The Presence object should perform automatic re-entry in the following situations:
+*** @(RTP17c1)@ After a @SYNC@ operation has completed, per "@RTP18b@":#RTP18b
+*** @(RTP17c2)@ When an @ATTACHED@ message is received with no "@HAS_PRESENCE@":#TR3a flag (so no @SYNC@ is expected as the server does not believe anyone is currently present)
+** @(RTP17d)@ Automatic re-entry consists of, for each member of the @RTP17@ internal @PresenceMap@ whose @memberKey@ is not also a member of the normal @PresenceMap@, publishing a @PresenceMessage@ with an @ENTER@ action using the @clientId@ and @data@ attributes from that member, and removing that member from the internal @PresenceMap@
+** @(RTP17e)@ If any of these automatic @ENTER@ presence messages fail to be published, an @UPDATE@ event should be emitted on the channel with @resumed@ set to true and @reason@ set to an @ErrorInfo@ object with @code@ @91004@, a @message@ indicating that an automatic re-enter has failed and indicating the @clientId@, and @cause@ set to the the reason for the enter failure. The error should also be logged at @warn@ level or higher.
```
paddybyers: "fail to be published" - it's not clear if this is referring to a failure of the publish attempt referred to in `RTP17d`, or something else (eg a failure to receive the message subsequently on the realtime connection.

Perhaps "If the publish attempt fails for any of these automatic @ENTER@ presence messages, ...."

content/client-lib-development-guide/features.textile
SimonWoolf: https://github.com/ably/docs/pull/711/commits/b1e8c0cc6874fbe0ff60b2616557220b5cbcfcbf

content/client-lib-development-guide/features.textile
SimonWoolf: https://github.com/ably/docs/pull/711/commits/5b538f1d1a4f611f8e5d69ed4d242197b7678411

mattheworiordan:
Thanks.

Before this is merged, please can you:

* [ ] Create an issue in all libraries that implement realtime presence to correct this
* [ ] Describe what test may be useful to verify the spec change is fixed (making it easier for all client lib devs)
* [ ] Update the spec sheet (or rename existing spec items) with these new spec items as I assume, because numbering has been changed, some of these previously implemented spec items may be partially implemented anyway i.e. if we don't do this now in the sheet, later it will be much harder for each client lib developer to understand the spec items are renumbered and moved and quite similar to previous implementations.

paddybyers:
@SimonWoolf please confirm that the tests that need to exist for this should cover the following.

Construction of these tests will require the use of a mock transport that is capable of sending specific messages to the backend, suppressing the sending of specific messages to the backend, and suppressing the processing of specific messages received from the backend. 

1) Server-sent `ATTACHED` indicates `hasPresence = false` but local PresenceMap has members that must be re-entered.

The sequence of events would be:
- connect;
- attach;
- enter a member (via the API), but suppress the `ENTER` sent by the client to the backend;
- simulate the receipt of that member from the backend, so it is entered by the library into the internal `PresenceMap`;
- disconnect, and wait for < `connectionStateTtl`;
- trigger a resume;
- await the `CONNECTED` and then `ATTACHED`. The received `ATTACHED` message should have `hasPresence=false`;
- verify that the library attempts to re-enter the member;
- verify that the subsequent expected sequence of events occurs (ie the member is entered in the backend, and subsequently reflected back to the client).

2) Server-sent `ATTACHED` indicates `hasPresence = true` followed by `SYNC`; local PresenceMap has members that must be reentered.

The sequence of events would be:

- connect;
- attach;
- enter a member, `memberA`, normally;
- enter a member, `memberB` via the API, but suppress the `ENTER` sent by the client to the backend;
- simulate the receipt of `memberB` from the backend, so it is entered by the library into the internal `PresenceMap`;
- disconnect, and wait for < `connectionStateTtl`;
- trigger a resume;
- await the `CONNECTED` and then `ATTACHED`. The received `ATTACHED` message should have `hasPresence=true`;
- await the `SYNC`;
- verify that the library attempts to re-enter `memberB`;
- verify that the subsequent expected sequence of events occurs (ie `memberB` is entered in the backend, and subsequently reflected back to the client).

SimonWoolf:
>  please confirm that the tests that need to exist for this should cover the following.
lgtm

PR #713, opened 2019-08-02
Misc updates to SSE docs (in particular no longer in beta)

SimonWoolf:
[SSE resume phase 2](https://github.com/ably/realtime/pull/2542) (and its [hefty preparatory stuff](https://github.com/ably/realtime/pull/2523)) are now deployed, so we can remove the wording about sse being in beta.

Changed `token` to `accessToken`, since afaict we're [no closer to a decision on `token` than we were a month ago](https://github.com/ably/realtime/pull/2507), and in the meantime the docs have been wrong.

Also toned down the pro-token-auth wording a bit -- it comes across as schizophrenic to have docs that say "Typically token authentication should be used" and that it's "rarely advisable" to use an api key when literally every one of the SSE code examples and tutorials we provide use basic auth. (And presumably they do so because it's a PITA to do token auth properly with SSE without a client library -- but to the extent that people use sse because it's easy, lightweight, and needs no setup, those reasons will apply to them to).

### DIFF
--- a/content/sse/index.textile
+++ b/content/sse/index.textile
@@ -15,8 +15,6 @@ The Ably SSE and raw HTTP streaming API provides a way to get a realtime stream 
 
 HTTP streaming allows for a request from a client to be held by a server, allowing it to push data to the client without further requests. This, much like WebSockets, help avoid the overhead involved in normal HTTP requests. "Server-sent events":https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events (SSE) provide a thin layer on top of HTTP streaming. A common use of SSE is through the use of "the EventSource API":https://developer.mozilla.org/en-US/docs/Web/API/EventSource in all modern web browsers.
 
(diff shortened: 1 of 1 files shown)


content/sse/index.textile
```diff
@@ -136,7 +132,7 @@ h3(#event-stream). Plain HTTP event stream
 
 h6. GET realtime.ably.io/event-stream
 
-Start a streaming HTTP request, for ease of consuming with any HTTP library that supports streaming. When possible, we recommend using an SSE library as opposed to a raw HTTP stream as SSE libraries automatically handle reconnecting and resuming from the last received ID.
+Start a streaming HTTP request, for ease of consuming with any HTTP library that supports streaming. Instead of SSE events, uses JSON envelopes. When available, we recommend using an SSE library as opposed to a raw HTTP stream as SSE libraries automatically handle reconnecting and resuming from the last received ID.
```
tomczoink: I'd suggest something as follows for the first two sentences:

`Starts a streaming HTTP request. This allows for messages to be easily consumed with any HTTP library that supports streaming. This uses JSON envelopes instead of SSE events....`

SimonWoolf:
Updated per comments

PR #720, opened 2019-08-20
Added batch presence documentation

tomczoink:
Adds beta documentation for using the batch presence functionality which is being introduced.

### DIFF
--- a/content/rest-api/beta.textile
+++ b/content/rest-api/beta.textile
@@ -213,9 +213,9 @@ curl -X GET https://rest.ably.io/presence?channel=<channel names> \
 ```
 
 where:
-* @<channel names>@ is a @String@ of channel names seperated by commas, or a @String@ of channel names separated by a given @separator@.
+* @<channel names>@ is a @String@ of channel names separated by commas, or a @String@ of channel names separated by a given @separator@.
 
-Below is an example of a curl request which uses a @separator@ to seperate channels @channel,1@ and @channel,2@:
+Below is an example of a curl request which uses a @separator@ to separate channels @channel,1@ and @channel,2@:
 
 ```[json](code-editor:rest/batch-presence)
 curl -X GET https://rest.ably.io/presence?channel,1!channel,2=&separator=! \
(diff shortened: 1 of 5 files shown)


mattheworiordan:
I am intrigued, will the partial success be compatible with the `request` object's handling of the response, and be accessible to customers?  @tomczoink can you check that as I fear it won't work as expected.

The method definition in JS at lesat is:

`request(String method, String path, Object params, Object body, Object headers, callback(ErrorInfo err, HttpPaginatedResponse results))`

How will `ErrorInfo` contain the partial response in a way that the user can meaningfully use?

paddybyers:
> How will `ErrorInfo` contain the partial response in a way that the user can meaningfully use?

@mattheworiordan It won't because, if you remember, in `request()` any error response body is returned as a response, not an error. The error callback argument is only for the case that there was no response.

See the realtime tests for how to access the various elements of the response from javascript via the `request()` API:
https://github.com/ably/realtime/blob/97e9e805546b356e95575595af85530a6fcc5688/test/realtime/presence.js#L1660
https://github.com/ably/realtime/blob/97e9e805546b356e95575595af85530a6fcc5688/test/realtime/presence.js#L1802

@tomczoink I did suggest (https://github.com/ably/docs/pull/720#pullrequestreview-278227454) that we do need to have documentation somewhere that shows how you do this.

mattheworiordan:
> It won't because, if you remember, in request() any error response body is returned as a response, not an error. 

Ok, I forgot that we stated in HP3 "If the response is a single JSON object then items returns an array with one JsonObject. If the response is empty, then items returns an empty array.".  Therefore, the client can get the error JSON response.

Would be good to see a code example of a success, failure, and partial failure with some basic handling.  For example, how will a use know that it's a partial success vs full success when processing a success callback?  Easily done I am sure, but we should provide an example.

content/rest-api/beta.textile
```diff
@@ -216,7 +216,7 @@ curl -X GET https://rest.ably.io/presence?channel=<channel names> \
 where:
 * @<channel names>@ is a @String@ of channel names seperated by commas, or a @String@ of channel names separated by a given @separator@.
 
-Below is an example of a curl request which uses a @separator@:
+Below is an example of a curl request which uses a @separator@ to seperate channels @channel,1@ and @channel,2@:
```
paddybyers: typo: separate

content/rest-api/beta.textile
```diff
@@ -311,3 +310,61 @@ response body
...
+            "channel": "channel1",
+            "presence": [
+              { "action": 1, "clientId": "CLIENT1" },
+              { "action": 1, "clientId": "CLIENT2" }
+            ]
+          }
+        */
+      }
+    } else {
+      if(response.items[0].batchResponse) {
```
paddybyers: The definitive way to check for a partial success is to check for the `40020` batch error code. You should do this before looking in `batchResponse`

PR #727, opened 2019-09-09
Spec: log level of Message decode errors should only be errors if encoding known but unsuccessful

SimonWoolf:
as discussed


content/client-lib-development-guide/features.textile
```diff
@@ -264,7 +264,8 @@ h3(#rest-channel). Channel
...
+** @(RSL6b)@ If a given encoding (i.e. a @/@-delimited section of the @encoding@ field) is understood by the library but the library is unable to successfully apply the corresponding decoding (for example, incompatible encryption details are provided or invalid Base64 is detected in the message payload), a message should logged at @error@ level, and the message delivered with all completed decodings (that could be successfully applied) applied, with the remaining encodings in the @encoding@ field. For example, if a message had a decoding of "utf-8/cipher+aes-128-cbc/base64", and the payload was successfully Base64 decoded but the payload could not be encrypted because the @CipherParam@ details were not configured, the message would be delivered with a binary payload and an @encoding@ of "utf-8/cipher+aes-128-cbc"
```
paddybyers: but the payload could not be encrypted ->  but the payload could not be decrypted

mattheworiordan:
> as discussed

Before I reply on this PR (as I am at present against this), can someone elaborate on the rationale behind this?

SimonWoolf:
> Before I reply on this PR (as I am at present against this), can someone elaborate on the rationale behind this?

`vcdiff` (and likely more things in the future) is an encoding that is not known by client libraries (at least until we add support for them). So currently, if someone uses deltas with a client library that has not yet got explicit delta support added (passing the messages to an external vcdiff library), it will work, but they will get an `RealtimeChannel.onMessage(): Error processing the vcdiff encoding, decoder returned "unknown encoding"` error-level log in their console for every message. Which is not fatal, but is quite annoying. This change just modifies the log level so warnings for completely unknown encodings are at info level, but failures to decode known encodings are still error level.

As usual we can make this apparently better by introducing more complexity (client or channel options to control error levels or suppress logs for particular specified encodings, etc), but given we've been trying to avoid adding more complexity to client libs than is absolutely necessary, this seemed like a reasonable compromise.  (And is apparently what the java lib does already)

mattheworiordan:
Thanks for this explanation, and this is what I assumed was the rationale.  Sadly, I am very against this approach then.

In almost every case, our SDKs intentionally hide the complexity of encoding and decoding from end users, specifically so that they don't need to worry about this.  Our SDKs are here to simplify how developers interact with our APIs, and provide guarantees about behaviour.  The reason I have an issue with this PR is:

* Hiding errors because they are annoying is a strange approach to errors. The reason there is an error is because we're sending payloads that our client SDKs do not understand, so that is an error.  
* We already have an issue with our client SDKs where decoding errors can be hidden from users without them knowing, giving them unexpected results.  I am a firm believer our promise to customers is we provide a predictable service, with binary outcomes of success / failure, and nothing in between.  If a decoding error happens, and a customer receives a half-decoded message, we're already failing on our "promise" to customers.  We have previously discussed a subscribe handler with an error option, or an error callback, and I would advocate reconsidering introducing that in fact in 2.0.  Hiding this error just makes this problem even worse, and in practice we should now recommend to customers that everyone has an additional code path that checks if every message has an `encoding` left in the message. For example `if (message.encoding) { throw("Shit, I have no idea what to do with an encoding system I have not designed"); } else { happyPath(message) }`  Why should they do this?  Our client SDKs should be handling all encoding / decoding issues for the protocol it supports.  Sending something in the protocol that we know the SDKs cannot currently deal with that will now silently fail is a mistake.
* `vcdiff` is a new feature that we happen to be introducing, but is not (yet) part of the Ably protocol or supported by the SDKs.  We should either support `vcdiff`, or not "hack" the protocol to introduce a new encoding that is not supported by the SDKs.  The entire point of introducing [simple deltas](https://github.com/ably/wiki/issues/324) was so that we a) didn't change the SDKs, b) it was portable across transports.

I think this PR is a knee-jerk reaction to introducing a breaking change that is shouldn't have been done in the first place.

tomczoink:
@SimonWoolf is this good to be closed?

paddybyers:
Yes, we can close this.

PR #728, opened 2019-09-10
Touch commit to trigger workflow

mattheworiordan:


mattheworiordan:
I think we can close this, it was just an experiment to use Github actions.  
If we decide to move to actions from Travis, then we can refer back to this.

PR #729, opened 2019-09-20
Channel Parameters: add first draft of documentation, including rewind

paddybyers:
This is a first draft of some documentation for channel parameters, and `rewind`.

@SimonWoolf @codemerx please review for accuracy, and please add examples for SSE and MQTT.

I don't feel strongly about where this goes in the hierarchy but this seems to make sense, alongside channel-metadata.

### DIFF
--- a/content/realtime/channel-params.textile
+++ b/content/realtime/channel-params.textile
@@ -1,0 +1,91 @@
+---
+title: Channel Parameters
+section: realtime
+index: 32
+jump_to:
+  Help with:
+    - Overview#overview
+    - Using params with the Ably library#using-params-with-library
+    - Using params with non-Ably transports#using-params-with-other-transports
+  rewind parameter#rewind
+    - Examples#examples
+---
+
+h2(#overview). Overview
+
+Channel parameters are a general mechanism by which a client can express properties of a channel, or of its attachment to a channel. A set of channel parameters is simply a set of key/value pairs; the keys correspond to specific features that Ably defines. Having a general mechanism for specifying such parameters allows the feature set to be expanded over time without requiring any API change and, in most cases, without requiring any additional library functionality.
+
+The methods provided for specifying channel parameters, and the currently available features, are outlined below.
+
+h2(#using-params-with-library). Using params with the Ably library
+
+The current Ably libraries, at version 1.1, do not expose an API for expressing channel parameters. This means that it is necessary to specify parameters in a way that is opaque to the library. Library version 1.2, to be released shortly, will provide direct API support for expressing channel parameters.
+
+A set of params is expressed by including a query string, using standard URL query syntax and encoding, within the qualifier part of a channel name. The qualifier part is in square brackets at the start of the channel name.
+
+For example, to specify the parameter @foo@ with value @bar@ on channel @baz@, the qualified channel name would be @[?foo=bar]baz@. If the channel name already has a qualifier, such as @[meta]log@, then the query string follows the existing qualifier, as in @[meta?foor=bar]log@.
+
+Using this syntax with the Ably library means that channel parameters are specified for the lifetime of the @Channel@ instance; in order to reference the same channel, but with different channel parameters, it is necessary to get a new @Channel@ instance, using a qualified name that includes the new channel parameters.
+
+Details of the explicit channel parameters API for use with later library versions, will be made available in due course.
+
+h2(#using-params-with-other-transports). Using params with non-Ably transports
+
+It is possible to interact with Ably channels using transports that do not involve using an Ably library; for example using SSE without any library, or using a supported non-Ably protocol such as MQTT. In these cases, it is also necessary to use a qualified channel name.
+
+A set of params is expressed by including a query string, using standard URL query syntax and encoding, within the qualifier part of a channel name. The qualifier part is in square brackets at the start of the channel name.
+
+For example, to specify the parameter @foo@ with value @bar@ on channel @baz@, the qualified channel name would be @[?foo=bar]baz@. If the channel name already has a qualifier, such as @[meta]log@, then the query string follows the existing qualifier, as in @[meta?foo=bar]log@.
+
+In an SSE connection, it is also possible to specify channel parameters as a query string in the connection URL, instead of as a qualifier on an individual channel name. In this case, the given channel parameters apply to all channel attachments associated with that connection.
+
+h2(#rewind). rewind parameter
+
+The @rewind@ channel parameter relates to the initial attachment of a connection to a channel, and expresses the intent to attach to the channel at a position, or a point in time, in the past (that is, effectively "rewinding" the channel for the purposes of the present attachment).
+
+A @rewind@ parameter can express a channel position in terms of a number of messages, or a time interval.
+
+A @rewind@ value that is simply a number @n@ (eg @rewind=1@) is a request to attach to the channel at a position @n@ messages before the present position. If that attachment is successful, and one or more messages exist on the channel prior to the present position, then those messages will be delivered to the subscriber immediately after the attachment has completed, and before any subsequent messages that arise in real time.
+
+If fewer than the requested number of messages exists on the channel (including the case that there are no prior messages), then the available messages are sent; this does not constitute an error.
+
+A @rewind@ value can also be a string that is a time interval specifier. Supported specifier values express an integral number of seconds (eg @15s@), minutes (eg @5m@) or hours (eg @4h@). If that attachment is successful, and one or more messages exist on the channel in the given time interval prior to the present time, then those messages will be delivered to the subscriber immediately after the attachment has completed, and before any subsequent messages that arise in real time.
+
+If fewer than the requested number of messages exists on the channel in that interval (including the case that there are no messages), then the available messages are sent; this does not constitute an error.
+
+If the number of messages within the specified interval is greater than the attachment @limit@ (which defaults to 100), then only the most recent messages up to that limit are sent. The attachment succeeds, but truncation of the message backlog is indicated as a non-fatal error in the attachment response.
+
+It is also possible to specify a different limit via a @limit@ channel parameter. So, for example, to request up to 10 messages, in a window 5m before the present time, specify a channel parameter string of @[?rewind=5m&limit=10]@.
+
+By default, a maximum of two minutes of channel history is available when attaching. This means that a rewind time specifier of greater than two minutes will only be able to rewind by two minutes. If a channel has persistence enabled, then it is possible to rewind back in time by up to the persistence TTL on the channel. 
+
+The channel position expressed by a @rewind@ parameter has an effect only on an initial channel attachment. Any subsequent reattachment of the same channel on the same connection, in order to resume the connection, will attempt to resume with continuity from the point at which the connection dropped.
+
+Any @rewind@ parameter value that cannot be parsed either as a number or a time specifier represents an error, and any attachment request will fail with an error.
+
+h2(#examples). Examples
+
+h3(#examples-ably). Using channel parameters with the Ably library
+
+To subscribe to a channel, getting the most recent message if available:
+
+```
+  const subscriber = new Realtime(API_KEY);
+  subscriber.channels.get('[?rewind=1]example-channel').subscribe((msg) => {
+    console.log('Received message: ', msg);
+  });
+```
+
+To subscribe to a channel, getting up to 10 messages in the last 5 minutes:
+
+```
+  const subscriber = new Realtime(API_KEY);
+  subscriber.channels.get('[?rewind=5m&limit=10]example-channel').subscribe((msg) => {
+    console.log('Received message: ', msg);
+  });
+```
+
+h3(#examples-sse). Using channel parameters with SSE
+
+h3(#examples-mqtt). Using channel parameters with MQTT
+
--- a/content/realtime/channel-params.textile
+++ b/content/realtime/channel-params.textile
@@ -5,21 +5,44 @@ index: 32
 jump_to:
   Help with:
     - Overview#overview
-    - Using params with the Ably library#using-params-with-library
+    - Using params with v1.2+ Ably libraries#using-params-with-lib-v12
+    - Using params with v1.1 or earlier Ably libraries#using-params-with-lib-v11
     - Using params with non-Ably transports#using-params-with-other-transports
-  rewind parameter#rewind
     - Examples#examples
+  Parameters:
+    - rewind#rewind
 ---
 
 h2(#overview). Overview
 
-Channel parameters are a general mechanism by which a client can express properties of a channel, or of its attachment to a channel. A set of channel parameters is simply a set of key/value pairs; the keys correspond to specific features that Ably defines. Having a general mechanism for specifying such parameters allows the feature set to be expanded over time without requiring any API change and, in most cases, without requiring any additional library functionality.
+Channel parameters are a general mechanism by which a client can express properties of a channel, or of its attachment to a channel. A set of channel parameters is simply a set of key/value pairs, where both keys and values are strings; the keys correspond to specific features that Ably defines.
 
 The methods provided for specifying channel parameters, and the currently available features, are outlined below.
 
-h2(#using-params-with-library). Using params with the Ably library
+h2(#supported-params). Currently supported channel params
 
-The current Ably libraries, at version 1.1, do not expose an API for expressing channel parameters. This means that it is necessary to specify parameters in a way that is opaque to the library. Library version 1.2, to be released shortly, will provide direct API support for expressing channel parameters.
+- rewind := Used to request that an attachment start from some number of messages or point in time in the past. See "rewind":#rewind for more information
+- delta := **In beta**. Used to request that data payloads should be sent as deltas to the previous payload. "Contact us":https://www.ably.io/contact for more information and supported values
+
+h2(#using-params-with-lib-v12). Using channel params with v1.2+ Ably client libraries
+
+With a client library that is version 1.2 or higher, you can specify channel params as part of the @params@ property of the "channel options":/realtime/channels#channel-options.
+
+h4. Example
+
+For example, to specify the @rewind@ channel param with the value @"1"@:
+
+```[jsall]
+  // only with ably-js v1.2 or higher; currently in beta
+  const realtime = new Ably.Realtime('{{API_KEY}}');
+  const channel = realtime.channels.get('{{RANDOM_CHANNEL_NAME}}', {
+    rewind: '1'
+  });
+```
+
+h2(#using-params-with-lib-v11). Using channel params with v1.1 or earlier Ably client libraries
+
+The current Ably libraries, at version 1.1, do not expose an API for expressing channel parameters. This means that it is necessary to specify parameters in a way that is opaque to the library.
 
 A set of params is expressed by including a query string, using standard URL query syntax and encoding, within the qualifier part of a channel name. The qualifier part is in square brackets at the start of the channel name.
 
@@ -27,7 +50,14 @@ For example, to specify the parameter @foo@ with value @bar@ on channel @baz@, t
 
 Using this syntax with the Ably library means that channel parameters are specified for the lifetime of the @Channel@ instance; in order to reference the same channel, but with different channel parameters, it is necessary to get a new @Channel@ instance, using a qualified name that includes the new channel parameters.
 
-Details of the explicit channel parameters API for use with later library versions, will be made available in due course.
+h4. Example
+
+For example, to specify the @rewind@ channel param with the value @"1"@:
+
+```
+  const realtime = new Ably.Realtime('{{API_KEY}}');
+  const channel = realtime.channels.get('[?rewind=1]{{RANDOM_CHANNEL_NAME}}');
+```
 
 h2(#using-params-with-other-transports). Using params with non-Ably transports
 
@@ -39,7 +69,26 @@ For example, to specify the parameter @foo@ with value @bar@ on channel @baz@, t
 
 In an SSE connection, it is also possible to specify channel parameters as a query string in the connection URL, instead of as a qualifier on an individual channel name. In this case, the given channel parameters apply to all channel attachments associated with that connection.
 
-h2(#rewind). rewind parameter
+h4. SSE example
+
+For example, to specify the @rewind@ channel param with the value @"1"@ using a querystring parameter, where it will apply to all channels:
+
+```[javascript]
+var querystring = 'v=1.1&channels={{RANDOM_CHANNEL_NAME}}&rewind=1&key={{API_KEY}}';
+var eventSource = new EventSource('https://realtime.ably.io/event-stream?' + querystring);
+```
+
+Or to specify the same parameter but only applying to one channel of two, using a qualified channel name:
+
+```[javascript]
+  var channelOne = encodeURIComponent('[?rewind=1]{{RANDOM_CHANNEL_NAME}}');
+  var channelTwo = '{{RANDOM_CHANNEL_NAME}}';
+  var channels = channelOne + ',' + channelTwo;
+  var querystring = 'v=1.1&key={{API_KEY}}&channels=' + channels';
+  var eventSource = new EventSource('https://realtime.ably.io/event-stream?' + querystring);
+```
+
+h2(#rewind). Rewind parameter
 
 The @rewind@ channel parameter relates to the initial attachment of a connection to a channel, and expresses the intent to attach to the channel at a position, or a point in time, in the past (that is, effectively "rewinding" the channel for the purposes of the present attachment).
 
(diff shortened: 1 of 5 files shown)


content/realtime/channel-params.textile
```diff
@@ -0,0 +1,91 @@
...
+
+Any @rewind@ parameter value that cannot be parsed either as a number or a time specifier represents an error, and any attachment request will fail with an error.
+
+h2(#examples). Examples
+
+h3(#examples-ably). Using channel parameters with the Ably library
+
+To subscribe to a channel, getting the most recent message if available:
+
+```
```
mattheworiordan: Given we support a) interactive code samples, b) injection of keys, c) multiple languages for code samples, may I ask that we at least have one set of code samples for each transport before we hand over to devrel team?

For example, see Javascript example below from https://github.com/ably/docs/blob/master/content/realtime/channels.textile:

```[nodejs](code-editor:realtime/channel-publish)
var Ably = require(‘ably’);
var realtime = new Ably.Realtime(‘{{API_KEY}}’);
var channel = realtime.channels.get(‘{{RANDOM_CHANNEL_NAME}}’);
channel.subscribe(function(message) {
console.log("Received: " message.data);
});
channel.publish(“example”, “message data”);
```

Note that `code-editor:realtime/channel-publish` then links to https://github.com/ably/docs/blob/master/content/code/realtime/channel-publish.code, which is a full working version of the code that runs in JSBin.

Note the workflow may be easier for now to just create a JSBin directly at https://jsbin.ably.io, and include a link for Devrel to pick up to that JSBin.  For example, if you go to https://jsbin.ably.io/erobeq/1/edit which is any random JSBin which automatically injects an API key that works, you can just append `.textile` to the URL https://jsbin.ably.io/erobeq/1.textile to get the code block that goes into the docs repo.

content/realtime/channel-params.textile
```diff
@@ -0,0 +1,91 @@
...
+    - Overview#overview
+    - Using params with the Ably library#using-params-with-library
+    - Using params with non-Ably transports#using-params-with-other-transports
+  rewind parameter#rewind
+    - Examples#examples
+---
+
+h2(#overview). Overview
+
+Channel parameters are a general mechanism by which a client can express properties of a channel, or of its attachment to a channel. A set of channel parameters is simply a set of key/value pairs; the keys correspond to specific features that Ably defines. Having a general mechanism for specifying such parameters allows the feature set to be expanded over time without requiring any API change and, in most cases, without requiring any additional library functionality.
```
mattheworiordan: This is written from the perspective of us, not the developer.  They don't care if it's extensible or not, they just want to know how to use channel params.  Anyway, this is very useful and I think whoever picks this up and figures out how it applies to the docs should worry about this point.

mattheworiordan: I think we should have a section very early on which states supported channel params, with perhaps a list of upcoming (alpha) params too alluding to deltas for example.

content/realtime/channel-params.textile
```diff
@@ -0,0 +1,91 @@
...
+
+If the number of messages within the specified interval is greater than the attachment @limit@ (which defaults to 100), then only the most recent messages up to that limit are sent. The attachment succeeds, but truncation of the message backlog is indicated as a non-fatal error in the attachment response.
+
+It is also possible to specify a different limit via a @limit@ channel parameter. So, for example, to request up to 10 messages, in a window 5m before the present time, specify a channel parameter string of @[?rewind=5m&limit=10]@.
+
+By default, a maximum of two minutes of channel history is available when attaching. This means that a rewind time specifier of greater than two minutes will only be able to rewind by two minutes. If a channel has persistence enabled, then it is possible to rewind back in time by up to the persistence TTL on the channel. 
+
+The channel position expressed by a @rewind@ parameter has an effect only on an initial channel attachment. Any subsequent reattachment of the same channel on the same connection, in order to resume the connection, will attempt to resume with continuity from the point at which the connection dropped.
```
mattheworiordan: What happens if the re-attachment cannot provide continuity i.e. connection has gone from suspended to connected?

paddybyers: If continuity is not possible, then the response to the attach request indicates the loss of continuity and the reason as a non-fatal error, in the usual way.

mattheworiordan: > then the response to the attach request indicates the loss of continuity and the reason as a non-fatal error, in the usual way.

Ok, so then for my understanding, are we saying that:

* Customers should listen for channel continuity failures, and use the history API if they want to recover messages at that point, as usual.
* The client SDK knows which channel attributes should be sent or not when calling ATTACH because the realtime system may in fact not know which ATTACH requests are as a result of state lost vs a new attach.  For example, user connects, attaches to `foo`, then becomes disconnected for 10 minutes, resumes (and fails), so automatic re-ATTACH is sent on `foo` without the `rewind` param, whilst the user happens to call attach on `bar` which includes rewind, and the `rewind` param is sent at that time.  If this is the case, this introduces some complexity into the client SDK and means arbitrary key/val pairs are no longer arbitrary.  Perhaps if we have to have options that mean different things on reattach, they could be suffixed with `!` to mean always send for example (not suggesting that is the solutin, just an example).

content/realtime/channel-params.textile
```diff
@@ -0,0 +1,91 @@
...
+
+A @rewind@ value that is simply a number @n@ (eg @rewind=1@) is a request to attach to the channel at a position @n@ messages before the present position. If that attachment is successful, and one or more messages exist on the channel prior to the present position, then those messages will be delivered to the subscriber immediately after the attachment has completed, and before any subsequent messages that arise in real time.
+
+If fewer than the requested number of messages exists on the channel (including the case that there are no prior messages), then the available messages are sent; this does not constitute an error.
+
+A @rewind@ value can also be a string that is a time interval specifier. Supported specifier values express an integral number of seconds (eg @15s@), minutes (eg @5m@) or hours (eg @4h@). If that attachment is successful, and one or more messages exist on the channel in the given time interval prior to the present time, then those messages will be delivered to the subscriber immediately after the attachment has completed, and before any subsequent messages that arise in real time.
```
SimonWoolf: > `minutes (eg @5m@) or hours (eg @4h@)`

We don't currently support 'h'.

content/realtime/channel-params.textile
```diff
@@ -0,0 +1,91 @@
...
+
+If fewer than the requested number of messages exists on the channel in that interval (including the case that there are no messages), then the available messages are sent; this does not constitute an error.
+
+If the number of messages within the specified interval is greater than the attachment @limit@ (which defaults to 100), then only the most recent messages up to that limit are sent. The attachment succeeds, but truncation of the message backlog is indicated as a non-fatal error in the attachment response.
```
SimonWoolf: > (which defaults to 100)
we bumped it up to 1000

content/realtime/channel-params.textile
```diff
@@ -0,0 +1,91 @@
...
+
+It is also possible to specify a different limit via a @limit@ channel parameter. So, for example, to request up to 10 messages, in a window 5m before the present time, specify a channel parameter string of @[?rewind=5m&limit=10]@.
```
SimonWoolf: no it isn't

content/realtime/channel-params.textile
```diff
@@ -0,0 +1,91 @@
...
+
+By default, a maximum of two minutes of channel history is available when attaching. This means that a rewind time specifier of greater than two minutes will only be able to rewind by two minutes. If a channel has persistence enabled, then it is possible to rewind back in time by up to the persistence TTL on the channel.
```
SimonWoolf: > If a channel has persistence enabled, then it is possible to rewind back in time by up to the persistence TTL on the channel. 

Not currently it isn't.

The rewind-reaching-into-persisted-history work was done with the goal of supporting rewind=1 even if it has to reach into persistence; arbitrary time-based rewind beyond the two-minute window is currently neither tested nor enabled. It might work without many further changes, but if we want it to, we would still need to do the work to implement that. (And even with rewind=1 it isn't currently enabled by default -- we still need to decide when/how to do that)

content/realtime/channel-params.textile
SimonWoolf: > mean different things on reattach, they could be suffixed with ! to mean always send for example

I'd be against adding that complexity to the client libs (definitely when using channel-paramterers-in-the-name, but also when using client libs that support them natively).

In the case of a resume attempt that fails, realtime knows that that's the case and can itself decide to not apply the rewind option. So the only problem case is a clean attachment as a result of the client lib removing stale connection state after two minutes -- but in that case, is it actually a problem having the rewind param apply again? As long as time-based rewind is only live history, then it's only the last two minutes meaning that it's a subset of the messages the user would have missed. That seems like perfectly reasonable behaviour.

Admittedly if rewind can silently reach into persisted history with time-based options (eg the user can specify rewind=1h) then it becomes more of a problem. But I didn't realised we'd decided to do that -- see other comment.

mattheworiordan: > But in that case, is it actually a problem having the rewind param apply again? 

Well potentially yes because it's a change in behaviour.  I don't like telling customers rewind only applies once, except when the client decides to drop the state and you don't know when that happens, so just don't rely on this behaviour.

> As long as time-based rewind is only live history, then it's only the last two minutes meaning that it's a subset of the messages the user would have missed. That seems like perfectly reasonable behaviour.

I think it's unreasonable to state sometimes one thing happens, and sometimes something else happen.  

I think a channel attach flag should be included now from client libs to indicate that the attach is a reattach vs an explicit attach with params.  This way realtime can always decide on the right behaviour, and could, for example treat different params differently without having to bake in any logic into the SDKs.

content/realtime/channel-params.textile
mattheworiordan: Default for history queries generally is 100, why was this bumped?

content/realtime/channel-params.textile
mattheworiordan: Why not?

content/realtime/channel-params.textile
mattheworiordan: >  It might work without many further changes, but if we want it to, we would still need to do the work to implement that. (And even with rewind=1 it isn't currently enabled by default -- we still need to decide when/how to do that)

That seems like an arbitrary limitation.  Why would we limit history to two minute queries for persisted history?  That makes absolutely no sense IMHO.

content/realtime/channel-params.textile
SimonWoolf: The default REST api page limit of 100 isn't relevant -- rewind isn't a REST query, it's a resume parameter, and that's always had a limit of 1000. It was inconsistent that rewinding back to a channelSerial would have a limit of 1000, but rewinding to the exact same point with a time or a number param would give you a limit of 100.

(I've previously [done a PR to lower the resume limit be lowered given that the maxOutboundRate limits it anyway](https://github.com/ably/realtime/pull/2197), but we didn't get consensus on that)

content/realtime/channel-params.textile
SimonWoolf: We've never supported a separate `limit` channel param. I don't recall it ever being mentioned before now. It's certainly not in any of the issues, https://github.com/ably/realtime/issues/2464 or otherwise. I think Paddy may have just gotten the user API confused with the internal ChannelRecoverOptions struct or something.

If we want to support something like that then we'd need to discuss what the api should look like, then implement it. Not entirely obvious, since two different parameters both of which can be a limit might be a bit confusing - eg what would `rewind=10&limit=5` mean?

content/realtime/channel-params.textile
SimonWoolf: > That seems like an arbitrary limitation. Why would we limit history to two minute queries for persisted history? That makes absolutely no sense IMHO.

Rewind is a parameter that requests a specific attach point, like resume; that's all it's ever been. We added a kludge where rewind=1 can reach back into persisted history to support data-object use-cases, but we don't currently have a feature where resuming (whether from a channelSerial, a timestamp, or just a limit) on a persisted channel can go back to an arbitrary point in time.

We have previously considered allowing that -- arbitrary resuming on persisted channels, no more two minute window, yay. We even came up with some concrete plans for how to do so (e.g. switching timelinequery to index by the channelTimeserial rather than the frontend timestamp). But we decided against that, in favour of putting our effort into replacing both live and persisted history with NATS/Liftbridge. Did you change your mind about that?

paddybyers: > we don't currently support resuming (whether from a channelSerial, a timestamp, or just a limit) back to an arbitrary point in time on persisted channels.

I understood that we were supporting that for a numeric limit (ie `rewind=10`), although there is a disparity about what the number means (ie messages vs protocol messages). Rewind into persisted history via a time specifier I agree isn't supportable without a change to `TimelineQuery` to apply time bounds based on channel timeserial (which is possible, but not highest priority).

content/realtime/channel-params.textile
SimonWoolf: @mattheworiordan discussed this with @paddybyers just now. Conclusions:
- In most cases when the user specifies, say, rewind=2m and then does a reattach after eg 3 minutes due to internet connection loss, the most useful behaviour is probably to give them the last 2 minutes of messages, rather than nothing
- We should have an api that both works for users now and can nicely transition to a future when we can do arbitrary time resumes (whether with liftbridge or our current persisted history)
- The case that I was worrying about (rewind=72h, reattach after three minutes, get 72h of history) I should not worry about: any future in which we allow more than rewind=2m / rewind=1 (fulfilled from persisted history using timeserials etc) is a future in which the client library would not clear its connection state after 2 minutes, so realtime would know it's a resume and be able to fulfil it from the requested resume point.
- So we think an API of the form 'the rewind param applies whenever the channel could not resume from its previous point' is probably the best option right now: easy to explain, doesn't require any rewind-specific client library changes, easy to extend to a future where we can resume a lot more than we do now, but also works well for what we can do now (ie mostly backfill from live history)

mattheworiordan: > In most cases when the user specifies, say, rewind=2m and then does a reattach after eg 3 minutes due to internet connection loss, the most useful behaviour is probably to give them the last 2 minutes of messages, rather than nothing

I disagree.  If a user loses continuity on a channel, we have a very well documented approach and set of recommendations about letting the developer decide what is most appropriate.  It is not our job to decide that, and the reason we provide that. See https://www.ably.io/documentation/realtime/channels#nonfatal-errors and specifically "If channel continuity is lost in the process, the library will notify you though a resumed flag in the ATTACHED or UPDATE event, so you can decide how to handle that failure.".  This approach contravenes that principle.

> We should have an api that both works for users now and can nicely transition to a future when we can do arbitrary time resumes (whether with liftbridge or our current persisted history)

Agreed

> The case that I was worrying about (rewind=72h, reattach after three minutes, get 72h of history) I should not worry about: any future in which we allow more than rewind=2m / rewind=1 (fulfilled from persisted history using timeserials etc) is a future in which the client library would not clear its connection state after 2 minutes, so realtime would know it's a resume and be able to fulfil it from the requested resume point.

Ok, but there will still be a connection state / channel persistence history limit.  There will always be a limit to how far back one can go.

> So we think an API of the form 'the rewind param applies whenever the channel could not resume from its previous point' is probably the best option right now: 

> easy to explain
No, it's not.

> , doesn't require any rewind-specific client library changes,
We are making changes now.  They don't need to be rewind specific, they are reattach specific so that realtime can decide now and in the future for any other param we may choose to support.  We're making a change now in the spec that could help us in future in treating new undefined params differently based on the explicit vs implicit attach in future.

> easy to extend to a future where we can resume a lot more than we do now
As above, there is always going to be a limit to state.

> but also works well for what we can do now (ie mostly backfill from live history)
Perhaps, but still wrong IMHO.

mattheworiordan:
Can someone additionally clarify, does this feature reach into persisted history now, or do we plan on making that work now?

content/realtime/channel-params.textile
mattheworiordan: > Not entirely obvious, since two different parameters both of which can be a limit might be a bit confusing 

Sure, but that's clear to document.  Rewind integers are incompatible with limit params, whereas time based ones can have a limit param.  I don't see why this is confusing and I think it makes a lot of sense.

content/realtime/channel-params.textile
mattheworiordan: >  Did you change your mind about that?

I was not made aware of those issues.

SimonWoolf:
We had a call on some of the undecided aspects of this this morning. Conclusions:
- Behaviour on resume: we accepted that the client library would set a flag if it was knowingly performing an unclean attach; if realtime gets this flag on an attachment that looks clean, it will not apply rewind to it.
  - We didn't decide a name. We rejected `M_ATTACH_IMPLICIT` since an implicit attach already means something different. `M_ATTACH_UNCLEAN` sounds weirdly moralistic
  - In the case of a recover (where you have the problem that neither the client nor realtime knows whether the attach is clean if the recover fails) we decided we can just document that rewind would apply and not worry about the inconsistency, since recover is rarely used
  - If someone does want rewind on failed resume behaviour we can always add a `rewindOnFailedResume` channel param later
- Limit: we decided we should have a fairly low rewind limit (50 or 100); this would simultaneously solve the ['too many messages for the outbound rate limit'](https://github.com/ably/realtime/pull/2197) problem, and also allow reasonable caching behaviour by having the messagecache unconditionally cache that many messages, even from older than the connectionstatettl ago.
  - I'm not sure if we decided that limit should apply only to rewind or also to normal resumes. I don't see any justification for normal resumes to have a greater limit than rewind tbh, I'd be inclined to apply it to both. I might also revive https://github.com/ably/realtime/pull/2197 as part of this.
- We decided that the mode that was most important to rewind -- in particular for the purposes of reaching into persisted history -- was limit-based, not time-based, and that if it makes things easier (e.g. for the purposes of messagecache behaviour) we were happy to not support time-based rewind for now.
  - Or possibly only support time-based rewind for the duration of the connectionStateTtl, which would fit fine with the messagecache.

mattheworiordan:
> means you should take your own recovery actions' if we never sent any messages if resumed=false

I know you didn't ask me, but yes, that's my preference :)

paddybyers:
If the only reason for failing to resume completely is the limit, then I think it's a natural/expected/helpful thing for the developer that they get the messages that we are able to send, which is the `$limit` most recent messages, and we signal the discontinuity.

SimonWoolf:
Updates and M_ATTACH_RESUME flag: https://github.com/ably/docs/pull/729/commits/eca0933d3331ee4dd2929e6a0cd03c14883f9f23

content/realtime/channels.textile
```diff
@@ -236,6 +236,149 @@ channel.subscribe { message in
...
+
+minimize. View details
+  "Rewind":/realtime/channel-params has two methods of obtaining messages from history. You can retreive a specified number of messages from history or messages from a set period of time into the past.
+
+  In Ably libraries older than @v1.2@, you can specify the "rewind":/realtime/channel-params parameter in the "metadata":/realtime/channel-metadata of the channel name, appending it at the end of any metadata otherwise specified. For example, if you have no metadata and wish to subscribe to channel @my_channel@ and fetch the most recent message from "history":/realtime/history, you would specify the channel as @[?rewind=1]my_channel@. If the channel had some metadata, @[some_metadata]my_channel@, you would apply "rewind":/realtime/channel-params with @[some_metadata?rewind=1]my_channel@. For more details, look at our "channel paramater":/realtime/channel-params documentation.
```
paddybyers: I'm not really a fan of expressing this as "libraries older than 1.2". There isn't a 1.2 release anywhere, or even announced, yet we're referring to it here.

Personally I think here we should be saying that you can specify the rewind param using the channel params mechanism. In current Ably libraries, or when using the service without a library, this is by qualifying the channel name. There is a future library release planned in which it will be possible to specify channel params directly via the API.

Then, when we do release 1.2, we can update the text to say "In Ably library versions 1.2 or later, you can specify the rewind parameter as a channel param [ref]. In library versions earlier than 1.2, channel params are expressed by qualifying the channel name [ref] ....".

> "metadata":/realtime/channel-metadata of the channel name

This name qualifier is not the same as channel metadata. (Channel Metadata refers to the information such as activity and occupancy.) This text needs to use the correct term (refer back to that text) but it's going to be something like "you can specify the "rewind":/realtime/channel-params parameter by encoding it as a param in the qualifier part of the channel name ...".

content/realtime/channels.textile
```diff
@@ -236,6 +236,149 @@ channel.subscribe { message in
...
+  let realtime = ARTRealtime(key: "{{API_KEY}}")
+  let channel = realtime.channels.get("[?rewind=1]{{RANDOM_CHANNEL_NAME}}")
+  channel.subscribe { message in
+      print("Received: \(message.data)")
+  }
+  ```
+
+  *Rewind by period of time*
+
+  To rewind by a set period of time, add @?rewind=TIME@ to your metadata, where @TIME@ is how far back in time you wish to rewind. For example, to obtain 10 seconds of messages from history as you subscribe, you would do the following:
```
paddybyers: In the initial documentation there is reference to a "time specifier", which is a way of expressing a time interval. We need to use the same language here. Saying "where @TIME@ is how far back in time you wish to rewind" is too imprecise.

content/realtime/history.textile
```diff
@@ -134,14 +134,86 @@ Both the "@Channel@":/realtime/channels and "@Presence@":/realtime/presence obje
...
 
 h3(#continuous-history). Continuous history
 
+By using "rewind":/realtime/channel-params#rewind or History's @untilAttach@, it is possible to obtain message history that is continuous with the realtime messages received on an attached channel.
+
+h4(#rewind). Rewind
+
+If you wish to obtain history as part of attaching to a channel, you can use the "rewind channel parameter":/realtime/channel-params#rewind. This will act as though you had attached to a channel from a certain message or time in the past, and play through all messages since that point. Check out the main "rewind documentation":/realtime/channel-params#rewind for further details.
+
+In versions of Ably prior to v1.2, rewind is specified via the metadata. The parameter is appeneded to any existing metadata, so if the channel were @[some_metadata]my_channel@, you would specify the use of rewind via @[some_metadata?rewind=1]my_channel@. You can specify either a number of messages up to 100 to rewind to (for example @10@), or a period of time in either seconds (@10s@), or minutes (@2m@) up to 2 minutes.
```
paddybyers: Please see my comments above. I don't think we should be referring to 1.2 before it's a thing, and it's not channel metadata.

content/realtime/channel-params.textile
```diff
@@ -0,0 +1,158 @@
...
+h2(#overview). Overview
+
+Channel parameters are a general mechanism by which a client can express properties of a channel, or of its attachment to a channel. A set of channel parameters is simply a set of key/value pairs, where both keys and values are strings; the keys correspond to specific features that Ably defines.
+
+The methods provided for specifying channel parameters, and the currently available features, are outlined below.
+
+h2(#supported-params). Currently supported channel params
+
+- rewind := Used to request that an attachment start from some number of messages or point in time in the past. See "rewind":#rewind for more information
+- delta := **In beta**. Used to request that data payloads should be sent as deltas to the previous payload. "Contact us":https://www.ably.io/contact for more information and supported values
```
paddybyers: I don't think we should be saying that deltas are in beta. Beta implies that it's feature-complete, and it isn't yet (because the library functionality isn't released yet). I think we should be describing this as experimental for now.

content/realtime/channel-params.textile
```diff
@@ -0,0 +1,158 @@
...
+
+h2(#using-params-with-lib-v12). Using channel params with v1.2+ Ably client libraries
+
+With a client library that is version 1.2 or higher, you can specify channel params as part of the @params@ property of the "channel options":/realtime/channels#channel-options.
```
paddybyers: Terms that compare library versions I think should be "earlier" and "later". In this PR there's a mix of terms used ("older", "higher" etc).

content/realtime/channel-params.textile
```diff
@@ -0,0 +1,158 @@
...
+
+If fewer than the requested number of messages exists on the channel (including the case that there are no prior messages), then the available messages are sent; this does not constitute an error.
+
+A @rewind@ value can also be a string that is a time interval specifier. Supported specifier values express an integral number of seconds (eg @15s@) or minutes (eg @2m@). If that attachment is successful, and one or more messages exist on the channel in the given time interval prior to the present time, then those messages will be delivered to the subscriber immediately after the attachment has completed, and before any subsequent messages that arise in real time.
+
+If you wish to use a time interval rewind but with a custom limit on the number of messages to be returned, you can use the @rewindLimit@ channel param. For example, to request up to 10 messages in a window 5m before the present time, specify a channel parameter string of @rewind=5m&rewindLimit=10@. If fewer than the requested number of messages exists on the channel in that interval (including the case that there are no messages), then the available messages are sent; this does not constitute an error.
```
paddybyers: "but with a custom limit" -> "but additionally specify a limit"

Issue #737, opened 2019-10-03
Spec for params option to rest channel.publish()


SimonWoolf:

SimonWoolf:
The problem here is RSL1h, and the various interpretations of it, which are a bit of a mess.

>  `(RSL1h)` Where the library language permits, the `Channel#publish(name, data)` method should provide an optional argument that allows the `clientId` value to be specified such as `Channel#publish('event', 'data', { clientId: 'John' })`, and/or the `extras` field to be specified such as `Channel#publish('event', 'data', { extras: { push: { 'key': 'value' } })`

AFAICT the only library that implements it as-written (where the publish takes an optional map that is mixed in to the message) is ably-ruby (not surprising since it looks like Matt added it in 2015, encoding what ably-ruby already did). Ably-js and ably-java do not implement it at all; if you want to include a clientId or extras you have to use the single-message-object form of publish (eg in ably-js, `publish({name: 'event', data: 'data', extras: {...}, clientId: 'John'})`). Ably-dotnet accepts a clientId as an optional extra argument, but if you want extras you need to use the Message form. Ably-ios and ably-php both accept both clientId and extras as optional arguments using the type to disambiguate -- which means that any json-encodable object (ios) / php array (php)  is interpreted as an `extras`.

Which means the approach I've taken in ably-js (adding an extra argument for qs params) will only work nicely in ably-js, ably-java, and ably-dotnet. In ably-ruby it'd get confused with the attributes argument; in ably-ios and ably-php it'd get confused with the extras argument. 

One possible conclusion is that the separate-argument forms (name, data, ...) are now hopelessly different across the different client libs, and we should allow the params argument only for the `publish(Message)` / `publish(Message[])` forms...

@mattheworiordan @paddybyers wdyt?

paddybyers:
> we should allow the params argument only for the `publish(Message)` / `publish(Message[])` forms...

agreed.

And the form that doesn't take a `Message`, but `name` and `data`, should IMO be limited to just that. Otherwise it's just an arbitrary set of other fields, and an arbitrary set of precedence rules when disambiguating the possibilities when looking at what's actually been passed.

Issue #644, opened 2019-03-23
Publish with a single Message object


mattheworiordan:
I've noticed in the push documentation, we are using an undocumented form of the publish method.

Here's the docs on the method with overloaded versions:

<img width="768" alt="Capto_Capture 2019-03-23_12-19-08_pm" src="https://user-images.githubusercontent.com/43789/54865950-e0cbd080-4d65-11e9-9467-59613dc344da.png">

Yet here is what is being called from the push publishing docs:

<img width="760" alt="Capto_Capture 2019-03-23_12-20-19_pm" src="https://user-images.githubusercontent.com/43789/54865965-0a84f780-4d66-11e9-9c1f-bfed44a6f2e9.png">

@Srushtika @tomczoink why are we using an undocumented method definition?

@paddybyers @SimonWoolf should we perhaps modify the spec anyway to support three overloaded versions:

* `publish(String name, Object data)`
* `publish(Message message)`
* `publish(Message[] messages)`

In addition, @tomczoink I think the docs are wrong anyway.

`publish(String name, Object data)` should in fact be `publish(String name, Object data, String clientId, Object extras)`.   Now in languages like Javascript, which does not support named params, we may not want to support that, or perhaps we'd support it slightly differently as follows: `publish(String name, Object data, Object extraMessageAttributes)` would could then contain an arbitrary set of key/val pairs such as `publish('name', 'data', { clientId: 'foo', extras: { .. });`

My concern either way is that we have to continue to support new arguments / params for the publish method to:

* Support query params being added to the request to support, amongst other things, features like [wait for channel to become active](https://github.com/ably/wiki/issues/310).  Where do those params go?
* New message attributes, that in the first version of the method, will start to get unwieldy if they're defined in the method itself such as `publish(String name, Object data, String clientId, Object extras, String id, String someNewParam)`

Therefore, perhaps instead, we should only support three forms as follows:

* `publish(String name, Object data, Object params)`
* `publish(Message message, Object params)`
* `publish(Message[] messages, Object params)`

Thus if a user wants to do more than send a simple message with a name & data value, they need to instead send one or more message objects, and specify params separately.

WDYT @paddybyers @SimonWoolf?

SimonWoolf:
> Thus if a user wants to do more than send a simple message with a name & data value, they need to instead send one or more message objects, and specify params separately.

[...]

PR #742, opened 2019-10-07
Support channel#publish(Message); remove RSL1h; add Channel#publish(Message, params)

SimonWoolf:
Fixes https://github.com/ably/docs/issues/737
Fixes https://github.com/ably/docs/issues/644

Implements what we agreed in March with https://github.com/ably/docs/issues/644 , with the minor tweak of only supporting rest params in the Message and Message[] forms, not the (name, data) form, since as discovered in https://github.com/ably/docs/issues/737 many client libs currently take a dict as an additional in that form and interpret it in different ways (as a message mixin, as `extras`, ...), which can't be changed without a backwards compatibility break until version 2.0.


content/client-lib-development-guide/features.textile
```diff
@@ -237,6 +237,7 @@ h3(#rest-channel). Channel
 *** @(RSL1k3)@ If more than one @Message@ is passed to @publish()@ and one or more of those messages contains a non-empty @id@ attribute, then all message ids (present or absent) are preserved on sending the batch of messages.
 *** @(RSL1k4)@ An explicit test for idempotency of publishes with library-generated ids shall exist that simulates an error response to a successful publish of a batch of messages, expects an automatic retry by the library, and verifies that the net outcome is that the batch is published only once.
 *** @(RSL1k5)@ An explicit test for idempotency of publishes with client-supplied ids shall exist that involves multiple explicit publish requests for a given message and verifies that the net outcome is that the message is published only once.
+** @(RSL1l)@ The @publish(Message)@ and @publish(Message[])@ forms of the method should take an extra @Dict<String, Stringifiable>@ argument. These parameters should be encoded using normal querystring-encoding and sent as part of the query string of the REST publish. (@Stringifiable@ is defined in @RTC1f@)
```
mattheworiordan: Can we suggest a test here please so that a developer knows the params are workign?  Otherwise it's hard to know if the implementation works as expected for every dev who implements this feature without us checking it.

paddybyers: You mean simply state that a test must exist that proves the params are included the request (eg by intercepting the request the library makes, a mock http layer or echo server); or suggest a test that actually passes the params to realtime, so that the effect of that is verifiable in the result?

mattheworiordan: > or suggest a test that actually passes the params to realtime, so that the effect of that is verifiable in the result?

The latter, that is what we have always aimed to do with SDK tests no? We build our SDKs to test against sandbox for a reason, and stubbing out incorrect params or a bad implementation doesn't prove the SDK does, it just gives the developer a happy feeling.

SimonWoolf: How about an undocumented `_forceNack=true` param that forces the publish to be rejected with a particular error code, that the client lib developer can check for?

SimonWoolf: ^ https://github.com/ably/docs/pull/742/commits/eec87554204df72f846c474f467d598224c39392

PR #743, opened 2019-10-14
Update showing the rewind functionality.

fakeharxy:
As per my email to Jodie: I've updated the docs in three places: core-features/history, realtime/channels and realtime/history. I think this makes sense for now, although as more channel parameters are added and the API updated they may need to be moved to a more prominent position.


mattheworiordan:
Thanks very much for your input @fakeharxy.

I believe you have been provided feedback on this PR, so I am closing this now.

PR #758, opened 2019-11-05
Update RSC15f to reference RTN17e

marto83:
Added a clarification in RSC15f that the Rest library should not store the fallback host when there is a realtime connection with a fallback host.


content/client-lib-development-guide/features.textile
```diff
@@ -90,7 +90,7 @@ h3(#restclient). RestClient
...
-** @(RSC15f)@ Once/if a given fallback host succeeds, the client should store that successful fallback host for @ClientOptions.fallbackRetryTimeout@. Future HTTP requests during that period should use that host. If during this period a "qualifying errors":#RSC15d occurs on that host, or after @fallbackRetryTimeout@ has expired, it should be un-stored, and the fallback sequence begun again from scratch, starting with the default primary host (@rest.ably.io@ or @ClientOptions#restHost@)
+** @(RSC15f)@ Once/if a given fallback host succeeds, the client should store that successful fallback host for @ClientOptions.fallbackRetryTimeout@. Future HTTP requests during that period should use that host. If during this period a "qualifying errors":#RSC15d occurs on that host, or after @fallbackRetryTimeout@ has expired, it should be un-stored, and the fallback sequence begun again from scratch, starting with the default primary host (@rest.ably.io@ or @ClientOptions#restHost@). This does NOT apply when the realtime connection is using a fallback host as per (RTN17e). In that case we should always prefer the host used for the realtime connection.
```
SimonWoolf: Thing is that this structure makes it sound like the "this does NOT apply" applies to the whole of the preceding paragraph (storing a successful fallback etc), when actually it should only apply to the parenthetical immediately preceding.

How about: `...starting with the default primary host per @RSC15e@ (or in the case of a connected realtime client, the connected host per @RTN17e@)`

mattheworiordan: True, that is my understanding too.  So perhaps instead:

```
** @(RSC15f)@ Once/if a given fallback host succeeds, the client should store that successful fallback host for @ClientOptions.fallbackRetryTimeout@. Future HTTP requests during that period should use that host first, except in the case of a connected realtime client, where the connected host should be used per @RTN17e@. If during this period a "qualifying errors":#RSC15d occurs on that host, or after @fallbackRetryTimeout@ has expired, it should be un-stored, and the fallback sequence begun again from scratch, starting with the default primary host (@rest.ably.io@ or @ClientOptions#restHost@), or in the case of @RTN17e@, the connect host again.
```

WDYT?  Same as @marto83's suggestion logically, but just worded so that the exception is baked in from the start.

marto83: Thanks Matt. This is makes it clear that fallbackRetryTimeout is not to be used when the realtime host has fallen back. If everyone is happy I'll update the spec item.

marto83: I've update the branch with the copy Matt provided and will start implementing the spec item.

SimonWoolf: > I thought the case was if there is a realtime connection fallback we do not store a successful fallback. In the case of an error we attempt a fallback host but with the next request we go back to the Realtime connected host.

I don't think that's sensible behaviour. The effect of this new wording is to make the stored fallback completely ineffective for a connected realtime client. The client library still stores it but then ignores it.

Surely that's not the behaviour we want. The point of storing a successful host is to avoid further hammering a region that is rejecting rest requests for whatever reason -- perhaps it's overloaded, perhaps its instanceRateCounters have exceeded, perhaps the rest elb or the router is broken and not routing rest requests, whatever. Yes, having an ongoing realtime connection to a host makes some of those less likely to be the case, but still perfectly possible. Do we really want to say that we want to throw away all the benefits of storing the rest host just because a client has a realtime connection?

(There's subtleties like untilAttach requests, but there the router in the healthy region can make a decision about whether to forward to the serial-region or not depending on whether it can successfully connect to it, we don't need the client lib to have to make that determination)

paddybyers: > The point of storing a successful host is to avoid further hammering a region that is rejecting rest requests for whatever reason -- perhaps it's overloaded ....

Yes; remember the incident we had where the TCP ELBs were fine in a region, but HTTP ELBs under-provisioned, so HTTP requests were failing.

marto83: So from what @SimonWoolf is saying we want to first try using the realtimeConnectedHost but if that fails we will store the fallbackHost as per normal rest library functionality and then once `fallbackRetryTimeout` has passed we will go back to the realtimeConnectedHost if it is still different than the default host. Am I understanding it correctly?

SimonWoolf: Yes. My proposal was what I wrote in https://ably-real-time.slack.com/archives/C030C5YLY/p1572968260200900 (copied below), which is also the effect of my suggestion https://github.com/ably/docs/pull/758#discussion_r342658932 .

> when doing a rest request:
> - By default, rest requests go to the realtime-connected host (if any, else options.restHost)
> - If an error requiring the use of a fallback occurs, then the usual fallback cascade rsc15a applies; once a request has succeeded on a fallback then that host is persisted for the fallbackRetryTimeout
> - During that period, rest requests go to the stored host. Once the fallbackRetryTImeout expires (or some other thing clears it, like a qualifying error on that host), it is cleared, so you go back to step 1 (i.e. you default to the realtime-connected host, if any, else options.restHost).

marto83: I updated the copy and implemented it this way here: https://github.com/ably/ably-dotnet/blob/c7fad2402c30001f8931ef759688b781f9f01864/src/IO.Ably.Shared/Http/AblyHttpClient.cs#L203

PR #763, opened 2019-11-06
Add addRequestIds client option

SimonWoolf:
Currently this just encodes what ably-ruby does, https://github.com/ably/ably-ruby/pull/133 .

A few obvious questions:
- the ably-ruby behaviour is to use a query string param, which Paddy points out was so it would end up in router logs. Do we want to keep that, or make it a header instead (a bit neater, but needs router changes to log it)? (Or allow both? In ably-js in particular if it's in the query string param we could replace the `rnd` xhr cachebuster with it)
- Do we want to enabled it by default? in ably-ruby it's disabled by default

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -76,6 +76,7 @@ h3(#restclient). RestClient
 ** @(RSC7a)@ The header @X-Ably-Version: 1.1@ must be included in all REST requests to the Ably endpoint
 ** @(RSC7b)@ The header @X-Ably-Lib: [lib][.optional variant]?-[version]@ should be included in all REST requests to the Ably endpoint where @[lib]@ is the name of the library such as @js@ for @ably-js@, @[.optional variant]@ is an optional library variant, such as @laravel@ for the @php@ library, which is always delimited with a period such as @php.laravel@, and where @[version]@ is the full client library version using "Semver":http://semver.org/ such as @1.0.2@. For example, the 1.0.0 version of the Javascript library would use the header @X-Ably-Lib: js-1.0.0@
 *** @(RSC7b1)@ When it is not possible to send the @X-Ably-Lib@ header, such as for @JSONP@ requests, the library version should be sent as a query param such as @lib=js-1.0.0@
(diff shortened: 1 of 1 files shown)


mattheworiordan:
> Do we want to keep that, or make it a header instead (a bit neater, but needs router changes to log it)?

No, the request ID is useful in load balancer and edge proxy logs too, including Cloudflare & Cloudfront.  We need that request ID.

> Do we want to enabled it by default? in ably-ruby it's disabled by default

Feels overkill to enable by default, but I'm not that bothered to be honest.  We could shorten it perhaps so that it's not overly unique or too many bytes, and concatenated to `rid=ABCDABCDAC` for example.

content/client-lib-development-guide/features.textile
```diff
@@ -76,6 +76,7 @@ h3(#restclient). RestClient
...
 *** @(RSC7b1)@ When it is not possible to send the @X-Ably-Lib@ header, such as for @JSONP@ requests, the library version should be sent as a query param such as @lib=js-1.0.0@
+** @(RSC7c)@ If the @addRequestIds@ client option is enabled, every REST request to Ably should include in a @request_id@ query string parameter a random string obtained by url-safe base64-encoding a sequence of at least 9 bytes obtained from a source of randomness. This request ID must remain the same if a request is retried to a fallback host per @RSC15@. If the request fails, the request ID must be included in the @ErrorInfo@ returned to the user, and printed as part of any corresponding error log
```
mattheworiordan: Should we not output this id if a request fails as well, even if just the first time, so that customers can see these issues in their logs if they are interested. They would be warn only though.  I think that is what we do with Ruby.

SimonWoolf: Isn't that achieved by `and printed as part of any corresponding error log`?

mattheworiordan: Well it’s a bit ambiguous because an error is only returned if all attempts fail, so one would naturally assume the .. and .. part refers to the corresponding error log to the error returned, which does not happen if the first attempt fails

SimonWoolf: fair point, I'll change to `Any log messages associated with the request should include the request ID. If the request fails, the request ID must be included in the @ErrorInfo@ returned to the user.`

tomczoink:
@SimonWoolf is this needing any further changes or good to merge in?

SimonWoolf:
> is this needing any further changes or good to merge in?

At this point, given we're close to releasing 1.2, I'm tempted to merge this into `master-1-2` and make it part of the 1.2 release, rather than have it as an interstitial 1.1 feature. @QuintinWillison wdyt?

QuintinWillison:
@SimonWoolf it doesn't sound like an urgent change so making it part of 1.2 sounds right.

PR #766, opened 2019-11-12
1.2 spec: add channel params, modes, M_ATTACH_RESUME flag

SimonWoolf:
Rather than add this to https://github.com/ably/docs/pull/729 , did a new PR, since the spec updates are for `master-1-2` only, wheras the user-facing bit in #729 is mostly applicable already (due to the querystring syntax) so can be targeted at `master`.

This only contains the general channel params spec items. The codecs and delta-specific bits I believe @codemerx is going to do.

Made one small change from what we discussed in https://github.com/ably/docs/pull/729 , which is to make `channelOptions.modes` (and `channel.modes`) non-optional (for client lib devs to implement, of course it's still optional in the sense that a user doesn't have to specify it). The main reason we made it optional initially is because we wanted to have a very minimal, backwards-compatible implementation. In the end we gave that up and decided to just make it a v1.2 spec, so there's no particular reason it should be optional, and it would make the documentation rather confusing if some client libraries implement it and others don't. (Of course we will still support comma-separated modes as a channel param for SSE, MQTT etc. to specify using the in-channel-name-querystring, but we can document that in the SSE section, and not clutter up the ably client lib documentation with it)


content/client-lib-development-guide/features.textile
```diff
@@ -268,6 +268,7 @@ h3(#rest-channel). Channel
...
 ** @(RSL6b)@ If, for example, incompatible encryption details are provided or invalid Base64 is detected in the message payload, an error message will be sent to the logger, but the message will still be delivered with last successful decoding and the @encoding@ field. For example, if a message had a decoding of "utf-8/cipher+aes-128-cbc/base64", and the payload was successfully Base64 decoded but the payload could not be encrypted because the @CipherParam@ details were not configured, the message would be delivered with a binary payload and an @encoding@ with the value "utf-8/cipher+aes-128-cbc"
+* @(RSL7)@ @Channel#setOptions@ takes a @ChannelOptions@ object and sets or updates the stored channel options, then indicates success. (For libraries that are not REST-only, completion indication should be done in the same way as @RealtimeChannel#setOptions@, for uniformity; for example, if @RealtimeChannel#setOptions@ takes a callback, @Channel#setOptions@ should as well)
```
mattheworiordan: I think we're being too specific about whether the callback style should be used or not.  For example, in Ruby, REST is sync, and Realtime is evented.  So the sync version cannot really provide a callback without it getting weird.

content/client-lib-development-guide/features.textile
```diff
@@ -488,6 +490,11 @@ h3(#realtime-channel). Channel
 ** @(RTL4f)@ Once an @ATTACH@ @ProtocolMessage@ is sent, if an @ATTACHED@ @ProtocolMessage@ is not received within the "default realtime request timeout":#defaults, the attach request should be treated as though it has failed and the channel should transition to the @SUSPENDED@ state. The channel will then be subsequently automatically re-attached as described in "RTL13":#RTL13
 ** @(RTL4d)@ If the language permits, a callback can be provided that is called when the channel is attached successfully or the attach fails and the @ErrorInfo@ error is passed as an argument to the callback
 ** @(RTL4e)@ If the user does not have sufficient permissions to attach to the channel, the channel will transition to @FAILED@ and set the @Channel#errorReason@
+** @(RTL4j)@ If the attach is not a clean attach (where a 'clean attach' is an attach attempt where the channel has either not previously been attached or has been explicitly detached since the last time it was attached); for example an automatic reattach triggered by "@RTN15c3@":#RTN15c3 or "@RTL13a@":#RTL13a (non-exhaustive); the library should set the "@ATTACH_RESUME@":#TR3f flag in the @ATTACH@ message
```
mattheworiordan: Can we provide a hint to a developer in regards to how they can test this behaviour with the realtime system?

content/client-lib-development-guide/features.textile
```diff
@@ -488,6 +490,11 @@ h3(#realtime-channel). Channel
...
+** @(RTL4k)@ If the user has specified a non-empty @params@ object in the @ChannelOptions@ ("@TB2c@":#TB2c), it must be included in a @params@ field of the @ATTACH@ @ProtocolMessage@
+*** @(RTL4k1)@ If @params@ are sent in the @ATTACH@ message, the corresponding @ATTACHED@ (and any subsequent @ATTACHED@s) will contain the subset of those params that the server recognises and validates. This should be exposed in a read-only @params@ field of the @Channel@ (or a @getParams()@ method where that is more idiomatic)
+** @(RTL4l)@ If the user has specified a @modes@ array in the @ChannelOptions@ ("@TB2d@":#TB2d), it must be encoded as a bitfield per "@TR3@":#TR3 and set as the @flags@ field of the @ATTACH@ @ProtocolMessage@. (For the avoidance of doubt, when multiple different spec items require flags to be set in the @ATTACH@, the final @flags@ field should be the bitwise OR of them all)
```
mattheworiordan: What would happen in a language where `modes` is an empty array as opposed to a `null` array.  Is an empty array or `null` value ignored in both these cases?  Should we be specific about that?

content/client-lib-development-guide/features.textile
```diff
@@ -488,6 +490,11 @@ h3(#realtime-channel). Channel
...
+** @(RTL4m)@ On receipt of an @ATTACHED@, the client library should decode the @flags@ into an array of @ChannelMode@s (that is, the same format as @ChannelOptions.modes@) and expose it as a read-only @modes@ field of the @Channel@ (or a @getModes()@ method where that is more idiomatic). This should only contain @ChannelMode@s: it should not contain flags which are not modes (see "@TB2d@":#TB2d)
```
mattheworiordan: P.S. I don't think `@ChannelMode@s` renders correctly in Textile

mattheworiordan: Where in other libs do we do this `get` variation on an attribute such as `getModes`?

content/client-lib-development-guide/features.textile
```diff
@@ -1332,6 +1343,8 @@ h4. ChannelOptions
 ** @(TB2b)@ @cipher@, which is either:
 *** @(TB2b1)@ A @CipherParams@ instance, or
 *** @(TB2b2)@ an options hash (or language equivalent) consisting of any subset of @CipherParams@ fields that includes a @key@. In this case, the client library should call @getDefaultParams@, passing it the options hash, to obtain a @CipherParams@ instance
+** @(TB2c)@ @params@ (for realtime client libraries only) a @Dict<string,string>@ of key/value pairs
+** @(TB2d)@ @modes@ (for realtime client libraries only) a typed array of modes, which are the subset of flags defined in "@TR3@":#TR3 whose value is ≥16
```
mattheworiordan: I am a bit unclear how `modes ` works (excuse my stupidity). Are we saying they are typed bit flags, and the end user (consumer of these APIs) needs to be able to do some basic bitwise & operations to work out which flags are set?

content/client-lib-development-guide/features.textile
SimonWoolf: Makes no difference -- encoding an empty array into a bitfield is equivalent to bitwise-ORing the bitfield with 0, which is a no-op.

Realtime behaviour after https://github.com/ably/realtime/pull/2741/files#diff-0757a77e03530e3b881b82269acaa354R2305 is merged will be to apply the default set of modes if the flags field does not have any modes set (a change we had to make as a consequence of adding the new M_ATTACH_RESUME flag since it's then no longer true that the client only includes the flags field if it wants to specify modes)

content/client-lib-development-guide/features.textile
SimonWoolf: my impression was that ably-java does that when it wants to enforce read-only access, eg https://github.com/ably/ably-java/blob/b3bd021154f8da6d0389683ba60788d19c31b7a9/lib/src/main/java/io/ably/lib/rest/Auth.java#L916

content/client-lib-development-guide/features.textile
SimonWoolf: > Are we saying they are typed bit flags, and the end user (consumer of these APIs) needs to be able to do some basic bitwise & operations to work out which flags are set?

No -- intention is for the client library to decode it into an array of elements of the ChannelMode enum. (Under the hood those'll be, shrug, whatever enum elements actually are the language in question). E.g. `[ChannelMode.Publish, ChannelMode.Subscribe]`.

The wording in TB2d isn't very clear, I'll clarify it. Clearer in the IDL (`modes?: [ChannelMode]`) and in RTL4m ("...the client library should decode the `flags` into an array of `ChannelMode`s...")

content/client-lib-development-guide/features.textile
SimonWoolf: OK, sure. I'll just remove that sentence and trust client lib devs to do the sensible thing in their language

content/client-lib-development-guide/features.textile
SimonWoolf: https://github.com/ably/docs/pull/766/files#diff-11900b395df266bc2bbfc1d11bdfc7a0R494

SimonWoolf:
(Updated per feedback: https://github.com/ably/docs/pull/766/commits/f254a999f7217288edbf8f3d1daa528c0f849e93 , https://github.com/ably/docs/pull/766/commits/bf19ea9dbca0910098376ac7e89328eea15e5793)

mattheworiordan:
> (Updated per feedback: f254a99

@SimonWoolf did you squish some commits together and force push?  I can't see the changes now :/

SimonWoolf:
> did you squish some commits together and force push? I can't see the changes now :/

@mattheworiordan sorry, yes. Unsquashed them again for easier review: https://github.com/ably/docs/pull/766/commits/ef3c4c28c816f146b901f556dad6c43ab24dfe4a , https://github.com/ably/docs/pull/766/commits/ef84d9632abac501be8a3204b2b423f1afb8f5a1

content/client-lib-development-guide/features.textile
```diff
@@ -488,6 +490,12 @@ h3(#realtime-channel). Channel
 ** @(RTL4f)@ Once an @ATTACH@ @ProtocolMessage@ is sent, if an @ATTACHED@ @ProtocolMessage@ is not received within the "default realtime request timeout":#defaults, the attach request should be treated as though it has failed and the channel should transition to the @SUSPENDED@ state. The channel will then be subsequently automatically re-attached as described in "RTL13":#RTL13
 ** @(RTL4d)@ If the language permits, a callback can be provided that is called when the channel is attached successfully or the attach fails and the @ErrorInfo@ error is passed as an argument to the callback
 ** @(RTL4e)@ If the user does not have sufficient permissions to attach to the channel, the channel will transition to @FAILED@ and set the @Channel#errorReason@
+** @(RTL4j)@ If the attach is not a clean attach (where a 'clean attach' is an attach attempt where the channel has either not previously been attached or has been explicitly detached since the last time it was attached), for example an automatic reattach triggered by "@RTN15c3@":#RTN15c3 or "@RTL13a@":#RTL13a (non-exhaustive), the library should set the "@ATTACH_RESUME@":#TR3f flag in the @ATTACH@ message
```
paddybyers: Can we not state more explicitly, in terms of channel state, when the flag is set?

SimonWoolf: @paddybyers Not really, as unfortunately it's not just a function of the current and previous channel state. Consider the following two sequences of states:

1. initialized, `attach()` called -> attaching -> suspended -> attaching -> attached

2. attached, server-initiated detach received -> attaching -> suspended -> attaching -> attached

in both cases the last ATTACH message is sent when going from `suspended` to `attaching`, but in the first case ATTACH_RESUME should be set, and in the second case it should not be.

So I'm not sure there is a nicer way to express the condition than `a 'clean attach' is an attach attempt where the channel has either not previously been attached or has been explicitly detached since the last time it was attached`, you do need a separate boolean to keep track of it. (Unless we add eg `attaching_unclean` and `suspended_unclean` states, which is horrible)

SimonWoolf: I could be more explicit about needing a boolean variable and what exact conditions should set it to what though, if you think that'd be better?

SimonWoolf: @paddybyers https://github.com/ably/docs/pull/766/commits/d6a3cbd5a0c9cc334a11c47ac3dab52ad5bb5c11

PR #767, opened 2019-11-13
Rewind draft tutorial

Srushtika:


content/tutorials/channel-rewind.textile
```diff
@@ -0,0 +1,217 @@
...
+title: Historical data with Channel Rewind
+section: tutorials
+index: 32
+---
+
+The Ably Realtime service organizes the message traffic within applications into named channels. Channels are the “unit” of message distribution; clients attach to any number of channels to subscribe to messages, and every message published to a channel is broadcasted to all its subscribers. 
+
+h2(#context). Context
+
+If a subscriber was to ever go offline, Ably by default, will retain all messages published and intended for that subscriber for a maximum of two min. During these two minutes, Ably will continously try to re-establish connection with that subscriber client, and if successful it'll relay all the retained messages to the subscriber in the order of their arrival. However, some clients may expect longer disconnections and hence may need message retention for more than the default two minute duration. For such scenarios Ably provides a feature called "History":/core-features/history using which one can retrieve missed historical messages that are a maximum of 72 hours old.
```
tomczoink: continously

tomczoink: "re-establish connection" a connection or the connection maybe?

content/tutorials/channel-rewind.textile
```diff
@@ -0,0 +1,217 @@
...
+
+History works perfectly well in scenarios where messages are being shared in realtime between various clients and some sort of message delivery guarantees are required by the application. However, in a data streaming scenario where a data provider continously streams data in realtime as it is generated, a new consumer just tuning in to get that data may have to sit idle until the next available update arrives. If they are interested in previously published messages in the recent past, they have to do a separate history request as soon as they join. To deal with such scenarios more efficiently, we recently launched a new feature called "Channel Rewind":/realtime/channels.
+
+One of the main motivating factors for building the rewind functionality within Ably was the recent launch of our new product called the "API Streamer":/api-streamer that lets data producers stream it to any number of consumers. As part of the API Streamer, we also launched the "Ably Hub":/api-streamer/hub that serves as a marketplace for open, free or commercial data streams most of which come with a self-service option to start receiving the data updates.
```
tomczoink: `lets data producers stream it` stream data, or what is it?

content/tutorials/channel-rewind.textile
```diff
@@ -0,0 +1,217 @@
...
+
+Rewind is simply a parameter you add when attaching to channels on Ably. If you are on a version less than v1.2, you can add the rewind parameter as part of the channel scope. Also, rewind being just another parameter on Ably's existing channels offering, this new feature can be seamlessly used with any of "the protocols that Ably currently supports":https://www.ably.io/adapters.
+
+In this tutorial we'll look at implementing the rewind feature for a weather monitoring app and see examples using Ably's native protocol, SSE and MQTT.
+
+<%= partial 'tutorials/_step-1-setup-free-account' %>
+
+h2(#subscribe-to-weather-data-stream). Step 2 - Subscribe to weather data stream on the Ably Hub
+
+As mentioned before, the "Ably Hub":https://www.ably.io/hub is a marketplace for free, open or commercial data streams. The "Weather Data":https://www.ably.io/hub/products/19 product on it is open and free to consume. It's also self-subscribe, meaning if you are interested to use it you can simply navigate to this product and click on the subscribe button and you'll instantly have access to this data stream. We'll make use of this data stream to see how channel rewind works.
```
tomczoink: `interested to use it you` => `interesting in using it you`

content/tutorials/channel-rewind.textile
```diff
@@ -0,0 +1,217 @@
...
+
+<a href="/images/tutorials/rewind/hub-self-subscribe.png" target="_blank">
+    <img src="/images/tutorials/rewind/hub-self-subscribe.png" style="width: 60%" alt="Self subscribe option on Ably Hub">
+  </a>
```
tomczoink: Indentation

content/tutorials/channel-rewind.textile
```diff
@@ -0,0 +1,217 @@
...
+        <h4 id="result-open-weather"></h4>
+        </div>
+    </td>
+    </tr>
+    </table>
+</body>
+</html>
+```
+
+Next, we'll instantiate the Ably library and attach to the channels with the rewind param enabled. With Ably version lower than 1.2, the rewind param needs to be added as part of the channel qualifier within the square brackets. If the channel already has a qualifier like @[meta]*@ or like in our case @[product:ably-openweathermap/weather]*@, we can simply append the existing qualifier name with @?rewind=<position>@ or @?rewind=<time>@
```
tomczoink: `With Ably version lower than...` to either `With an Ably version lower than` or `With Ably versions lower than` I think.

content/tutorials/channel-rewind.textile
```diff
@@ -0,0 +1,217 @@
...
+      var ably = new Ably.Realtime("<YOUR-API-KEY>"),
+          $resultWeather = $('#result-open-weather')
+          channelWeather;
+  )};
+</script>
+```
+
+Make sure to add in your own Ably API Key in the above code snippet.
+
+What we need now to get the data from the relevant channels is to retrieve the user selection from the drop down menu and use the correlated code to subscribe to that particular channel. We'll add the rewind functionality here and set value to 1, signifying the number of previously published messages we'd like to retrieve from this data stream, on attaching to a channel.
```
tomczoink: I think could be simpler first line,  `We now need to retrieve the user selection...` or something.

tomczoink: `and set value to 1` => `and set its value to 1`

tomczoink: or `and set it to 1` or something.

content/tutorials/channel-rewind.textile
```diff
@@ -0,0 +1,217 @@
...
+
+You can use any MQTT library to publish and receive messages to and from Ably. For this tutorial, we'll make use of the "MQTT NPM library":https://www.npmjs.com/package/mqtt, but feel free to use any that you like.
+
+Create a new javascript file, say @main.js@. We'll begin by requiring the MQTT library and setting the correct options for our client to use Ably:
+
+```[javascript]
+var mqtt = require('mqtt');
+var options = {
+    keepalive: 30,
+    username: '<FIRST-HALF-OF-YOUR-ABLY-API-KEY>',
```
tomczoink: Can't remember the official names for first half and second half, but whatever they are we should use those + maybe a link which explains how a key is broken up? Maybe also a link somewhere around here to the Ably MQTT docs

content/tutorials/channel-rewind.textile
```diff
@@ -0,0 +1,217 @@
...
+});
+```
+
+h2(#live-demo). Live demo
+
+<iframe src="https://fourth-estimate.glitch.me/" height="200px" width="100%" frameBorder="0" ></iframe>
+
+h2(#conclusion). Conclusion
+
+In this tutorial, we saw the usage of rewind channel parameter in a real-world data streaming application. We saw three different ways of its usage, i.e., using Ably's native client SDKs, Ably's SSE endpoint and Ably's MQTT endpoint.
```
tomczoink: `we saw the usage of rewind channel parameter` => `we saw the usage of the rewind channel parameter`

tomczoink: `three different ways of its usage` => `three different ways to use it`

content/tutorials/channel-rewind.textile
```diff
@@ -0,0 +1,226 @@
...
+                </div>
+            </td>
+        </tr>
+    </table>
+</body>
+
+</html>
+```
+
+Next, we'll instantiate the Ably library and attach to the channels with the rewind param enabled. With Ably versions lower than 1.2, the rewind param needs to be added as part of the channel qualifier within the square brackets. If the channel already has a qualifier like @[meta]*@ or like in our case @[product:ably-openweathermap/weather]*@, we can simply append the existing qualifier name with @?rewind=<position>@ or @?rewind=<time>@
```
paddybyers: Same comments apply here as for the other docs PR. I think we should for now be saying in the present Ably libraries it is possible to specify the rewind param by qualifying the channel name. Later library releases will allow it to be specified explicitly in the API.

content/tutorials/channel-rewind.textile
```diff
@@ -0,0 +1,226 @@
...
+title: Historical data with Channel Rewind
+section: tutorials
+index: 32
+---
+
+The Ably Realtime service organizes the message traffic within applications into named channels. Channels are the “unit” of message distribution; clients attach to any number of channels to subscribe to messages, and every message published to a channel is broadcasted to all its subscribers. 
+
+h2(#context). Context
+
+If a subscriber was to ever go offline, Ably by default, will retain all messages published and intended for that subscriber for a maximum of two min. During these two minutes, Ably will continuously try to re-establish the connection with that subscriber client, and if successful it'll relay all the retained messages to the subscriber in the order of their arrival. However, some clients may expect longer disconnections and hence may need message retention for more than the default two minute duration. For such scenarios Ably provides a feature called "History":/core-features/history using which one can retrieve missed historical messages that are a maximum of 72 hours old.
```
paddybyers: I had quite a few detailed comments on this section but, before dealing with them, I think we need to fix the story.

Resume is peripherally relevant to rewind, but I think this section needs to start by motivating the rewind functionality. The discussion about resume and history is not relevant.

So I think the narrative for this needs to be something like:

- usually in pub/sub applications a client is interested in receiving messages as they occur in real time. Ably's API reflects this, in that an application can `subscribe()` to a channel; the client attaches to the channel, and receives messages from that point forward;

- however, in some applications it is useful to have access also to messages that occurred in the past. Ably supports a `history()` API whereby, for channels that are persisted, it is possible to query explicitly for messages on a channel within any specified time bounds;

- for some applications, the requirement for historical access is simply to be able to obtain the most recent message or messages on a channel. This can apply, for example, if <example ...>

- it would be possible for an application to achieve that by subscribing to a channel and also using the history API. However, Ably now provides a much simpler way to subscribe to a channel from a specified point in the past; that is, the subscriber will be sent all historical messages on the channel from the specified point forwards, and will receive all subsequent messages in realtime.

content/tutorials/channel-rewind.textile
Srushtika: Fixed, thanks.

content/tutorials/channel-rewind.textile
Srushtika: Fixed, thanks.

content/tutorials/channel-rewind.textile
```diff
@@ -0,0 +1,226 @@
...
+        eventSource;
+    var apiKey = "<YOUR-API-KEY>";
+    $('select#city-id').on('change', function () {
+        eventSource.close();
+        subscribeOpenWeather($('select#city-id').val());
+    });
+    subscribeOpenWeather($('select#city-id').val());
+    function subscribeOpenWeather(id) {
+        var channelWeather = "[product:ably-openweathermap/weather?rewind=1]weather:" + id;
+        var URL = `https://realtime.ably.io/sse?v=1.1&key=${apiKey}&channels=${channelWeather}`;
```
SimonWoolf: If you want to include a channel name with params in the name, you need to uri-encode it, which you're not doing here. A much easier way is just to add `rewind=1` as another url param. This is explained in the documentation pr:
https://github.com/ably/docs/pull/729/files#diff-7b572e68f8441588d6eacc180bc8f989R76-R89

SimonWoolf: (to explain this a bit further: the reason this is especially necessary here is that question marks have special meaning in urls, they start the query string ("/sse?v=..."). Maybe in some browsers having subsequent question marks inside the query string will work, but it's not guaranteed to work in all of them. The correct/safe thing to do that will work in all browsers is uri-encode them. (Really this should be done for all api streamer sse examples -- the brackets and forwardslash in `[product:foo/bar]baz` are technically also disallowed, though again, will probably work in some browsers))

PR #768, opened 2019-11-19
Rewind docs

tomczoink:
Adds rewind details to `/realtime/history` and `/realtime/channels`.

### DIFF
--- a/content/realtime/channels.textile
+++ b/content/realtime/channels.textile
@@ -236,6 +236,147 @@ channel.subscribe { message in
 }
 ```
 
+h4(#rewind). Subscribing to a channel and fetching history with rewind
+
+If you wish to both subscribe to a channel, as well as obtain some messages from said channel's "history":/realtime/history seamlessly, you can make use of the "rewind":/realtime/channel-params feature. This will result in the specified number of messages from "history":/realtime/history being returned as part of the subscription process.
+
+minimize. View details
+  "Rewind":/realtime/channel-params has two methods of obtaining messages from history, either by getting a specified number of messages from history, or messages from a set period of time into the past.
+
+  In Ably libraries older than @v1.2@, you can specify the "rewind":/realtime/channel-params parameter in the "metadata":/realtime/channel-metadata of the channel name, tacking it onto the end of any metadata otherwise specified. For example, if you have no metadata and wish to subscribe to channel @my_channel@ and fetch the most recent message from "history":/realtime/history, you would specify the channel as @[?rewind=1]my_channel@. If the channel had some metadata, @[some_metadata]my_channel@, you would apply "rewind":/realtime/channel-params with @[some_metadata?rewind=1]my_channel@. For more details, look at our "channel paramater":/realtime/channel-params documentation.
+
+  *Note* that "rewind":/realtime/channel-params is limited to a maximum of @100 messages@, and can only get messages no older than 2 minutes. In addition, rewind will only apply upon *attaching* to a channel, so any subsequent subscriptions post-attach will not fetch old messages.
+
+  *Rewind by number of messages*
+
+  To rewind by a set number of messages, add @?rewind=NUM_MSG@ to your metadata, where @NUM_MSG@ is the number of messages you wish to get at most. For example, to obtain 1 messages from history as you subscribe, you would do the following:
+
+  ```[javascript](code-editor:realtime/rewind)
+    var realtime = new Ably.Realtime('{{API_KEY}}');
+    var channel = realtime.channels.get('[?rewind=1]{{RANDOM_CHANNEL_NAME}}');
+    channel.subscribe(function(message) {
+      alert('Received: ' + message.data);
+    });
+  ```
+
+  ```[nodejs](code-editor:realtime/rewind)
+    var Ably = require('ably');
+    var realtime = new Ably.Realtime('{{API_KEY}}');
+    var channel = realtime.channels.get('[?rewind=1]{{RANDOM_CHANNEL_NAME}}');
+    channel.subscribe(function(message) {
+      console.log("Received: "  message.data);
+    });
+  ```
+
+  ```[ruby]
+    realtime = Ably::Realtime.new('{{API_KEY}}')
+    channel = realtime.channels.get('[?rewind=1]{{RANDOM_CHANNEL_NAME}}')
+    channel.subscribe do |message|
+      puts "Received: #{message.data}"
+    end
+  ```
+
+  ```[java]
+    AblyRealtime realtime = new AblyRealtime("{{API_KEY}}");
+    Channel channel = realtime.channels.get("[?rewind=1]{{RANDOM_CHANNEL_NAME}}");
+    channel.subscribe(new MessageListener() {
+      @Override
+      public void onMessage(Message message) {
+        System.out.println("New messages arrived. " + message.name);
+      }
+    });
+  ```
+
+  ```[csharp]
+    AblyRealtime realtime = new AblyRealtime("{{API_KEY}}");
+    IRealtimeChannel channel = realtime.Channels.Get("[?rewind=1]{{RANDOM_CHANNEL_NAME}}");
+    channel.Subscribe(message => {
+      Console.WriteLine($"Message: {message.Name}:{message.Data} received");
+    });
+  ```
+
+  ```[objc]
+  ARTRealtime *realtime = [[ARTRealtime alloc] initWithKey:@"{{API_KEY}}"];
+  ARTRealtimeChannel *channel = [realtime.channels get:@"[?rewind=1]{{RANDOM_CHANNEL_NAME}}"];
+  [channel subscribe:^(ARTMessage *message) {
+      NSLog(@"Received: %@", message.data);
+  }];
+  ```
+
+  ```[swift]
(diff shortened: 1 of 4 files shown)


content/realtime/channels.textile
```diff
@@ -236,6 +236,147 @@ channel.subscribe { message in
 }
 ```
 
+h4(#rewind). Subscribing to a channel and fetching history with rewind
```
Srushtika: IMHO, the title feels like it's two actions like subscribe to channel and make a call to the history api. Perhaps, 'Subscribing to a channel with rewind option enabled'?

content/realtime/channels.textile
```diff
@@ -236,6 +236,147 @@ channel.subscribe { message in
...
+
+If you wish to both subscribe to a channel, as well as obtain some messages from said channel's "history":/realtime/history seamlessly, you can make use of the "rewind":/realtime/channel-params feature. This will result in the specified number of messages from "history":/realtime/history being returned as part of the subscription process.
+
+minimize. View details
```
Srushtika: I think it's ok to keep in maximized. What are your thoughts?

content/realtime/channels.textile
```diff
@@ -236,6 +236,147 @@ channel.subscribe { message in
...
+  "Rewind":/realtime/channel-params has two methods of obtaining messages from history, either by getting a specified number of messages from history, or messages from a set period of time into the past.
+
+  In Ably libraries older than @v1.2@, you can specify the "rewind":/realtime/channel-params parameter in the "metadata":/realtime/channel-metadata of the channel name, tacking it onto the end of any metadata otherwise specified. For example, if you have no metadata and wish to subscribe to channel @my_channel@ and fetch the most recent message from "history":/realtime/history, you would specify the channel as @[?rewind=1]my_channel@. If the channel had some metadata, @[some_metadata]my_channel@, you would apply "rewind":/realtime/channel-params with @[some_metadata?rewind=1]my_channel@. For more details, look at our "channel paramater":/realtime/channel-params documentation.
```
Srushtika: `tacking it onto the end` -> `appending it at the end` feels more general tech jargon IMO.

content/realtime/channels.textile
```diff
@@ -236,6 +236,147 @@ channel.subscribe { message in
...
+
+  *Note* that "rewind":/realtime/channel-params is limited to a maximum of @100 messages@, and can only get messages no older than 2 minutes. In addition, rewind will only apply upon *attaching* to a channel, so any subsequent subscriptions post-attach will not fetch old messages.
```
Srushtika: This sounds a bit misleading as there's no mention of persistence.

content/realtime/channels.textile
```diff
@@ -236,6 +236,147 @@ channel.subscribe { message in
...
+
+  *Rewind by number of messages*
+
+  To rewind by a set number of messages, add @?rewind=NUM_MSG@ to your metadata, where @NUM_MSG@ is the number of messages you wish to get at most. For example, to obtain 1 messages from history as you subscribe, you would do the following:
```
Srushtika: `messages` -> `message`

content/realtime/channels.textile
```diff
@@ -236,6 +236,147 @@ channel.subscribe { message in
...
+  let realtime = ARTRealtime(key: "{{API_KEY}}")
+  let channel = realtime.channels.get("[?rewind=1]{{RANDOM_CHANNEL_NAME}}")
+  channel.subscribe { message in
+      print("Received: \(message.data)")
+  }
+  ```
+
+  *Rewind by period of time*
+
+  To rewind by a set period of time, add @?rewind=TIME@ to your metadata, where @TIME@ is the period of time you wish to get messages for. For example, to obtain 10 seconds of messages from history as you subscribe, you would do the following:
```
Srushtika: It sounds like the time is in the future. Perhaps, ` is the period of time you wish to rewind back and the fetch historical messages`

content/realtime/history.textile
```diff
@@ -134,14 +134,84 @@ Both the "@Channel@":/realtime/channels and "@Presence@":/realtime/presence obje
...
 
 h3(#continuous-history). Continuous history
 
+It is possible to obtain message history that is continuous with the realtime messages received on an attached channel, through the use of either "rewind":/realtime/messages, or the use of @untilAttach@ with History.
+
+h4(#rewind). Rewind
+
+If you wish to obtain history as part of attaching to a channel, you can make use of the "rewind channel parameter":/realtime/channel-params#rewind. This will act as though you had attached to a channel from a certain message or time in the past, and play through all messages since that point. Check out the main "rewind documentation":/realtime/channel-params#rewind for further details.
+
+In versions of Ably prior to v1.2, rewind is specified via the metadata. The parameter is appeneded to any existing metadata, so if the channel were @[some_metadata]my_channel@, you would specify the use of rewind via @[some_metadata?rewind=1]my_channel@. You can specify either a number of messages up to 100 to rewind to (for example @10@), or a period of time in either seconds (@10s@), or minutes (@2m@) up to 2 minutes.
```
Srushtika: I think what you've specified is in versions less than 1.2. For 1.2 or later, it is specified via channel params, see: https://github.com/ably/docs/pull/729/files#diff-7b572e68f8441588d6eacc180bc8f989R27 (apply changes, if any, in the other file too)

content/realtime/channels.textile
tomczoink: My thinking was that as channel is a very introductory page, it'd been good to avoid pushing bits such as `publishing` too far down the page. Happy to go either way though, do you think the size is an issue?

Srushtika: Fair point, let's leave the minimize then.

content/realtime/history.textile
tomczoink: Sure, this is our v1.1 docs though, we'll need to implement v1.2 references when we increment everything I imagine? Or am I misunderstanding?

content/realtime/channels.textile
```diff
@@ -236,6 +236,149 @@ channel.subscribe { message in
 }
 ```
 
+h4(#rewind). Subscribing to a channel with rewind option enabled
```
remotekieran: "wtih the rewind option enabled"

content/realtime/channels.textile
```diff
@@ -236,6 +236,149 @@ channel.subscribe { message in
...
+
+If you wish to both subscribe to a channel, as well as obtain some messages from said channel's "history":/realtime/history seamlessly, you can make use of the "rewind":/realtime/channel-params feature. This will result in the specified number of messages from "history":/realtime/history being returned as part of the subscription process.
```
remotekieran: Suggest changing the first sentence to:

'It is possible to subscribe to a channel and obtain messages from that channel's "history":/realtime history with a single API request. To do so you would use the "rewind":/realtime/channel-params feature.'

content/realtime/channels.textile
```diff
@@ -236,6 +236,149 @@ channel.subscribe { message in
...
+
+minimize. View details
+  "Rewind":/realtime/channel-params has two methods of obtaining messages from history, either by getting a specified number of messages from history, or messages from a set period of time into the past.
```
remotekieran: I would split these into two clear distinctions, probably with bullet points rather than a single sentence. Or simple break it into two sentences. e.g.

Rewind":/realtime/channel-params has two methods of obtaining messages from history. You can retreive a specified number of messages from history or messages from a set period of time into the past.

content/realtime/history.textile
```diff
@@ -134,14 +134,84 @@ Both the "@Channel@":/realtime/channels and "@Presence@":/realtime/presence obje
...
 
-To enable history on a channel, it is necessary to add a channel rule in the settings of your "application dashboard":https://support.ably.io/solution/articles/3000030053-how-do-i-access-my-app-dashboard. See the "documentation on channel rules":https://support.ably.io/solution/articles/3000030057-what-are-channel-rules-and-how-can-i-use-them-in-my-app for further information on what they are and how to configure them.
+To enable history on a channel, it is necessary to add a channel rule in the settings of your "application dashboard":https://support.ably.io/solution/articles/3000030053. See the "documentation on channel rules":https://support.ably.io/solution/articles/3000030057 for further information on what they are and how to configure them.
 
 h3(#continuous-history). Continuous history
 
+It is possible to obtain message history that is continuous with the realtime messages received on an attached channel, through the use of either "rewind":/realtime/messages, or the use of @untilAttach@ with History.
```
remotekieran: This is kind of awkward to read, I think because of passive voice. Suggest rephrasing:

By using "rewind":/realtime/messages or History's @untilAttach@, it is possible to obtain message history that is continuous with the realtime messages received on an attached channel.

Also, I could be misunderstanding, but is continuous the right word to use here?

content/realtime/history.textile
```diff
@@ -134,14 +134,84 @@ Both the "@Channel@":/realtime/channels and "@Presence@":/realtime/presence obje
...
+
+h4(#rewind). Rewind
+
+If you wish to obtain history as part of attaching to a channel, you can make use of the "rewind channel parameter":/realtime/channel-params#rewind. This will act as though you had attached to a channel from a certain message or time in the past, and play through all messages since that point. Check out the main "rewind documentation":/realtime/channel-params#rewind for further details.
```
remotekieran: Would again suggest dropping "you can make use of the..." for "you can use the..."

PR #777, opened 2019-11-26
Tutorials metadata update for new website tutorials design

renatodeleao:
TL;DR: 
- This PR adapts tutorials metadata with information required for wesbite indexing and new design view.
- Writes guidelines for future tutorials

### Notes
Before https://github.com/ably/website/pull/2866, tutorials indexing was hardcoded into the website, since the display required extra metadata that wasn't available here at the time. This required a manual update everytime a new tutorial was added. An automatic indexing mechanism was implemented at https://github.com/ably/website/pull/2866 but that requires new metadata fields to properly categorize tutorials.

#### New metadata fields are required for proper indexing and filtering at ably.io/tutorials

- `category*` — field is now required to put the tutorial in the correct section at 
- `platform*` — field that allow us to filter tutorials for each tab of new design: `(browser| mobile | mixed)`
- `alt_title` — field, because titles at the website can be slightly different from docs
- `authors*`— an array of multiple author hashes
  - expects an hash `{author_name: "name", author_bio: "optional", author_profile_url: "link", author_image: "gravatar url or relative path to assets/images/ folder on the website"}`
- `reading_time:` a number in minutes of the expected reading time
- `level`: a string (easy|medium|hard)
- `tags`: an Array of strings with (ably-realtime, ably-features, ex)
- `libraries`: (if code is uses any framework or library `(react js | vue js | angular... )`

<sup>* denotes required fields</sup>

#### Advised structure

With the new stepped view mode, each `h2` will be used as marker for a step/chapter of the tutorial. This requires extra careful when building tutorials content: specially when live examples includes markup as the following

```html

h2(#test). Live demo

<div id="app">
  <!-- please avoid using this, prefer and lower heading -->
  <h1>Demove title</h1>
</div>

```

A new helpbox component, this will created collapsible sections at single-page view mode, and will expand automatically at multi-page view mode

```markdown
div(docs-helpbox). 
 *HelpboxTitle wrapped in strong tag* 
 %Content wrapped in span tag, but can contain other inline markdown elements%
```

### DIFF
--- a/app/assets/stylesheets/documentation.scss
+++ b/app/assets/stylesheets/documentation.scss
@@ -830,3 +830,23 @@ nav.jump-to-nav {
     }
   }
 }
+
+
+.docs-helpbox {
+  background-color: #F1F1F1;
+  padding: 20px;
+  border-radius: 4px;
+
+  // title
+  strong {
+    display: block;
+    font-size: 18px;
+    margin-bottom: 8px;
+    color: #161616;
+  }
+
+  // body
+  span {
+    color: #555555;
+  }
+}
--- a/content/client-lib-development-guide/documentation-formatting-guide.textile
+++ b/content/client-lib-development-guide/documentation-formatting-guide.textile
@@ -77,6 +77,33 @@ competitors:
   - "competitor2 ID, which should match the ID used on the website"
 ---
 
+h4(#tutorials-metadata). Tutorial pages metadata
+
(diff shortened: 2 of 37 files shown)


mattheworiordan:
Out of interest, is the plan that these tutorials will render in some form in this static site that is similar to how they will render on the website?  I think they should so that devs can see how things will work (functionally)

PR #778, opened 2019-11-27
Changed jsbin jquery refs to ably endpoint

tomczoink:
Per discussion in Slack, changed refs to our version of jQuery.

### DIFF
--- a/content/code/adapters/pubnub-pub-sub.code
+++ b/content/code/adapters/pubnub-pub-sub.code
@@ -29,7 +29,7 @@ $('input#send').on('click', function() {
 [--- /Javascript ---]
 
 [--- HTML ---]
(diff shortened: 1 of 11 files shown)


content/code/adapters/pubnub-pub-sub.code
```diff
@@ -29,7 +29,7 @@ $('input#send').on('click', function() {
 [--- /Javascript ---]
 
 [--- HTML ---]
-<script src="//cdnjs.cloudflare.com/ajax/libs/jquery/2.1.3/jquery.min.js"></script>
+<script src="//jsbin-files.ably.io/js/jquery-1.8.3.min.js"></script>
```
mattheworiordan: FYI, this is absolutely fine to point to Ably.  The reported issue was in fact not that we're using jQuery from Cloudflare as opposed to Ably, but instead that we had upgraded to 2.1.3 which no longer defines `$`, so all examples relying on `$` no longer worked.

PR #779, opened 2019-11-27
Initial attempt on delta related changes specification

codemerx:
Attempts to specify the changes to the client library introduced by the deltas feature. There would include deltas, channel params, bits of rewind, etc.

This PR is based on the discussions and materials at https://github.com/ably/wiki/pull/391 and https://github.com/ably/wiki/issues/380.


content/client-lib-development-guide/features.textile
```diff
@@ -1332,6 +1335,16 @@ h4. ChannelOptions
...
+**** @(TB2c3)@ @rewind@ The rewind channel parameter relates to the initial attachment of a connection to a channel, and expresses the intent to attach to the channel at a position, or a point in time, in the past (that is, effectively “rewinding” the channel for the purposes of the present attachment). A rewind parameter can express a channel position in terms of a number of messages, or a time interval (ending in "s" or "m")
+**** @(TB2c4)@ @rewindLimit@ Sets a limit to the number of messages that will be returned if rewind is specified.
+*** @(TB2d)@ @modes@ array of channel subscription modes, e.g. "publish,subscribe,presence_subscribe". If present, these should overwrite the flags field @(TR3)@ of the ATTACH protocol message. 
+**** @(TB2d1)@ @publish@ Requests message publishing permission.
+**** @(TB2d1)@ @subscribe@ Requests message subscribing permission.
+**** @(TB2d1)@ @presence@ Requests presence message publishing permission.
+**** @(TB2d1)@ @presence_subscribe@ Requests presence message publishing permission.
```
SimonWoolf: I don't understand why you've included this. The ChannelParams are already specified in https://github.com/ably/docs/pull/766/ , which you approved without comment -- if you disagreed with what was done in that PR then why would you not have commented there?

(FWIW most of this does not belong in the client library development spec: we went for a model where params are opaque to the library, so these spec items are mostly not things that are actually implementable by a client library developer. C.f. https://github.com/ably/docs/pull/766/files#diff-11900b395df266bc2bbfc1d11bdfc7a0R494 : `rewind` is only mentioned in the context of describing a method to test that the client library is successfully encoding the params)

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +267,8 @@ h3(#rest-channel). Channel
...
+** @(RSL6c)@ If decoding fails on "vcdiff" encoding that means message delta application failed.Since applying a delta depends on the previous message being present a failure in the delta application on the client side makes it impossible to apply any subsequent deltas received. That would render the channel unusable so a full message needs to be requested from Ably (the server) to replace the failed delta message. That is done by sending a new ATTACH request (the channel state might still be ATTACHED).
```
SimonWoolf: This needs to be a lot more detailed, specific, and prescriptive. For example, think of the steps you're doing in ably-js/ably-java with `decodeFailureRecoveryInProgress` to avoid sending multiple `ATTACH`es while waiting for the previous one; the storing the channelSerial and setting it in the `ATTACH`, and so on. Those should all be spec items.

The idea is that a client library developer should, only using the spec, be able to produce a correct implementation, without reference to any existing libraries. Of course that ideal is not always achievable (and older spec items are often not nearly specific enough), but that's what we are aiming for with any new spec additions.

content/client-lib-development-guide/features.textile
codemerx: Yep, I approved this some time ago and did not really have it in mind when writing this one. Also true that `params` are opaque.

content/client-lib-development-guide/features.textile
codemerx: Elaborated some more. Please, check it out.

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,19 @@ h3(#rest-channel). Channel
...
+** @(RSL6c)@ If decoding fails on "vcdiff" encoding that means message delta application failed.Since applying a delta depends on the previous message being present a failure in the delta application on the client side makes it impossible to apply any subsequent deltas received. That would render the channel unusable so a full message needs to be requested from Ably (the server) to replace the failed delta message. That is done by:
+*** @(RSL6c1)@ Following the procedure described in @(RTL4c)@ under the caveats of @(RTL4d)@, @(RTL4e)@ and @(RTL4f)@. The ProtocolMessage.channelSerial of the ATTACH ProtocolMessage sent as per @(RTL4c)@ should be set to the channel serial of the message which "vcdiff" decoding failed.
```
SimonWoolf: > should be set to the channel serial of the message which "vcdiff" decoding failed.

Not quite: it should be set to the channelSerial of the previous, base message (which was either a full message or the last delta message whose decode succeeded).

(also: we should add ", if present" at the end, to allow for the possibility that a message may not have a channelSerial)

content/client-lib-development-guide/features.textile
```diff
@@ -1304,6 +1317,7 @@ h4. ClientOptions
 **** @(TO3l8e)@ The size of a @null@ or omitted property is zero
 *** @(TO3l9)@ @maxFrameSize@ integer - default 524288 (512KiB). The maximum size of a single POST body or "WebSocket":/concepts/websockets frame. This is mostly only relevant for `Rest#request` (e.g. for batch publishes), since publishes will hit the @maxMessageSize@ limit before this
 *** @(TO3l10)@ @fallbackRetryTimeout@ integer - default 600000 (10 minutes). (After a failed request to the default endpoint, followed by a successful request to a fallback endpoint), the period in milliseconds before HTTP requests are retried  against the default endpoint
+** @(TO3o)@ @codecs@ Dict<string:PluggableCodec> A map between an encoding string and a PluggableCodec object that could encode/decode the specified encoding. The client library might downcast a @PluggableCodec@ to particular codec type. Any codec provided in @codecs@ should be invoked by the client library when decoding/encoding a message utilizng the relevant encoding.
```
SimonWoolf: *utilizing

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,19 @@ h3(#rest-channel). Channel
...
+** @(RSL6b)@ If, for example, incompatible encryption details are provided or invalid Base64 is detected in the message payload, an error message will be sent to the logger, but the message will still be delivered with last successful decoding and the @encoding@ field. For example, if a message had a decoding of "utf-8/cipher+aes-128-cbc/base64", and the payload was successfully Base64 decoded but the payload could not be encrypted because the @CipherParam@ details were not configured, the message would be delivered with a binary payload and an @encoding@ with the value "utf-8/cipher+aes-128-cbc. Additional steps need to be taken if decoding failed on "vcdiff" encoding, see @(RSL6c)@ 
+** @(RSL6c)@ If decoding fails on "vcdiff" encoding that means message delta application failed.Since applying a delta depends on the previous message being present a failure in the delta application on the client side makes it impossible to apply any subsequent deltas received. That would render the channel unusable so a full message needs to be requested from Ably (the server) to replace the failed delta message. That is done by:
```
SimonWoolf: *`failed. Since`

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,19 @@ h3(#rest-channel). Channel
...
+*** @(RSL6c1)@ Following the procedure described in @(RTL4c)@ under the caveats of @(RTL4d)@, @(RTL4e)@ and @(RTL4f)@. The ProtocolMessage.channelSerial of the ATTACH ProtocolMessage sent as per @(RTL4c)@ should be set to the channel serial of the message which "vcdiff" decoding failed.
+*** @(RSL6c2)@ Any subsequent "vcdiff" decoding failures as per "RSL6c":#RSL6c while the recovery procedure from a previous failure as per "RSL6c1":#RSL6c1 is still in progress, i.e. the channel is in ATTACHING state, should not trigger a new recovery procedure as per "RSL6c1":#RSL6c1.
+** @(RSL6d)@ The last message on each channel has to be stored at all times since it might be needed to decode the "vcdiff" encoding. The last message bookkeping rules are:
+*** @(RSL6d1)@ Any transport layer introducing encoding (e.g. base64 introduced by sending binary on text transport) should be decoded before storing the last message
```
SimonWoolf: We know what a transport-introduced encoding is, but it isn't defined in the spec anywhere, so that may be ambiguous to a client lib developer. For simplicity you could just hardcode the knowledge in the spec item: `If the message encoding string ends in @base64@, the message should be base64-decoded (and the encoding string modified accordingly per @RSL6@) before storing the last message`

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,19 @@ h3(#rest-channel). Channel
...
+** @(RSL6c)@ If decoding fails on "vcdiff" encoding that means message delta application failed.Since applying a delta depends on the previous message being present a failure in the delta application on the client side makes it impossible to apply any subsequent deltas received. That would render the channel unusable so a full message needs to be requested from Ably (the server) to replace the failed delta message. That is done by:
+*** @(RSL6c1)@ Following the procedure described in @(RTL4c)@ under the caveats of @(RTL4d)@, @(RTL4e)@ and @(RTL4f)@. The ProtocolMessage.channelSerial of the ATTACH ProtocolMessage sent as per @(RTL4c)@ should be set to the channel serial of the message which "vcdiff" decoding failed.
+*** @(RSL6c2)@ Any subsequent "vcdiff" decoding failures as per "RSL6c":#RSL6c while the recovery procedure from a previous failure as per "RSL6c1":#RSL6c1 is still in progress, i.e. the channel is in ATTACHING state, should not trigger a new recovery procedure as per "RSL6c1":#RSL6c1.
+** @(RSL6d)@ The last message on each channel has to be stored at all times since it might be needed to decode the "vcdiff" encoding. The last message bookkeping rules are:
```
SimonWoolf: Not clear from this that we're only storing the `Message.data` payload, not the entire `Message`

*bookkeeping

content/client-lib-development-guide/features.textile
```diff
@@ -13,6 +13,7 @@ jump_to:
     - Auth#rest-auth
     - Channels#rest-channels
     - Channel#rest-channel
+    - Pluggable Codecs#pluggalbe-codecs
```
QuintinWillison: Typo in the second instance ('pluggalbe' > 'pluggable').

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,19 @@ h3(#rest-channel). Channel
...
+** @(RSL6b)@ If, for example, incompatible encryption details are provided or invalid Base64 is detected in the message payload, an error message will be sent to the logger, but the message will still be delivered with last successful decoding and the @encoding@ field. For example, if a message had a decoding of "utf-8/cipher+aes-128-cbc/base64", and the payload was successfully Base64 decoded but the payload could not be encrypted because the @CipherParam@ details were not configured, the message would be delivered with a binary payload and an @encoding@ with the value "utf-8/cipher+aes-128-cbc. Additional steps need to be taken if decoding failed on "vcdiff" encoding, see @(RSL6c)@
```
QuintinWillison: I think that should be "could not be **decrypted**".

content/client-lib-development-guide/features.textile
QuintinWillison: Perhaps some `@code@` formatting too (e.g. `ATTACH` and the method and class names).

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,19 @@ h3(#rest-channel). Channel
...
+*** @(RSL6c1)@ Following the procedure described in @(RTL4c)@ under the caveats of @(RTL4d)@, @(RTL4e)@ and @(RTL4f)@. The ProtocolMessage.channelSerial of the ATTACH ProtocolMessage sent as per @(RTL4c)@ should be set to the channel serial of the message which "vcdiff" decoding failed.
+*** @(RSL6c2)@ Any subsequent "vcdiff" decoding failures as per "RSL6c":#RSL6c while the recovery procedure from a previous failure as per "RSL6c1":#RSL6c1 is still in progress, i.e. the channel is in ATTACHING state, should not trigger a new recovery procedure as per "RSL6c1":#RSL6c1.
+** @(RSL6d)@ The last message on each channel has to be stored at all times since it might be needed to decode the "vcdiff" encoding. The last message bookkeping rules are:
+*** @(RSL6d1)@ Any transport layer introducing encoding (e.g. base64 introduced by sending binary on text transport) should be decoded before storing the last message
+*** @(RSL6d2)@ If "vcdiff" encoding is present the direct output of decoding the "vcdiff" encoding is stored as a last message, no matter that more encodings might be present
+
+h3(#pluggable-codecs). Pluggable Codecs
```
QuintinWillison: Pending agreement as to whether this is a 'codec' or a 'plugin'.

see https://github.com/ably/ably-java/pull/532#discussion_r350645112

Your thoughts, please, @paddybyers and @mattheworiordan.

content/client-lib-development-guide/features.textile
```diff
@@ -1332,7 +1346,6 @@ h4. ChannelOptions
 ** @(TB2b)@ @cipher@, which is either:
 *** @(TB2b1)@ A @CipherParams@ instance, or
 *** @(TB2b2)@ an options hash (or language equivalent) consisting of any subset of @CipherParams@ fields that includes a @key@. In this case, the client library should call @getDefaultParams@, passing it the options hash, to obtain a @CipherParams@ instance
-* @(TB3)@ The client lib may optionally provide an alternative constructor @withCipherKey@ for ChannelOptions that takes a @key@ only. (This must be differentiated from the normal constructor such that it is clear that the value being passed in is a key). (This is intended for languages where requiring a hash map is unidiomatic)
```
QuintinWillison: Why has this been removed? It feels unrelated, though I may be missing something.

content/client-lib-development-guide/features.textile
```diff
@@ -1901,6 +1914,13 @@ class HttpPaginatedResponse // RSC19b
   errorCode: Int // HP6
   errorMessage: String // HP7
   headers: Dict<String, String> // HP8
+  
+class PluggableCodec // PC2
+  // Empty class/interface. Pluggable codecs are not expected to share any common interface.
+  // Still it is better to have a common base type for pluggable codecs in statically typed languages.
+  
+class VCDiffDecoder
+  decode([byte] delta, [byte] base) -> [byte]
```
QuintinWillison: Pending agreement as to whether these args should be reversed.

see https://github.com/ably/ably-java/pull/532#discussion_r350663132

content/client-lib-development-guide/features.textile
codemerx: I think you are right but I didn't actually author this; the diff tool says I did. @SimonWoolf are we right about this or is it something we do not understand?

content/client-lib-development-guide/features.textile
codemerx: Should not be removed for sure.

content/client-lib-development-guide/features.textile
SimonWoolf: I agree it should be `decrypted`. Looks like that mistake [dates back to 2015](https://github.com/ably/docs/commit/c5e8a3407#diff-11900b395df266bc2bbfc1d11bdfc7a0R115) 🙂

content/client-lib-development-guide/features.textile
codemerx: Could think of no stopper so made this generic plugins.

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,20 @@ h3(#rest-channel). Channel
...
+** @(RSL6b)@ If, for example, incompatible encryption details are provided or invalid Base64 is detected in the message payload, an error message will be sent to the logger, but the message will still be delivered with last successful decoding and the @encoding@ field. For example, if a message had a decoding of "utf-8/cipher+aes-128-cbc/base64", and the payload was successfully Base64 decoded but the payload could not be decrypted because the @CipherParam@ details were not configured, the message would be delivered with a binary payload and an @encoding@ with the value "utf-8/cipher+aes-128-cbc. Additional steps need to be taken if decoding failed on "vcdiff" encoding, see @(RSL6c)@
```
QuintinWillison: Missing closing quotation marks on the second encoding you quote.

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,20 @@ h3(#rest-channel). Channel
...
+** @(RSL6c)@ If decoding fails on "vcdiff" encoding that means message delta application failed. Since applying a delta depends on the previous message being present a failure in the delta application on the client side makes it impossible to apply any subsequent deltas received. That would render the channel unusable so a full message needs to be requested from Ably (the server) to replace the failed delta message. That is done by:
+*** @(RSL6c1)@ Following the procedure described in "RTL4c":#RTL4c under the caveats of "RTL4d":#RTL4d, "RTL4e":#RTL4e and "RTL4f":#RTL4f. The @ProtocolMessage.channelSerial@ of the @ATTACH@ @ProtocolMessage@ sent as per "RTL4c":#RTL4c  should be set to the channel serial of the previous message to the message for which "vcdiff" decoding failed.
```
QuintinWillison: Instead of "the previous message to the message", perhaps "the message previous to the message" is more readable.

content/client-lib-development-guide/features.textile
```diff
@@ -479,7 +480,7 @@ h3(#realtime-channel). Channel
 * @(RTL11)@ If a channel enters the @DETACHED@, @SUSPENDED@ or @FAILED@ state, then all presence actions that are still queued for send on that channel per "RTP16b":#RTP16b should be deleted from the queue, and any callback passed to the corresponding presence method invocation should be called with an @ErrorInfo@ indicating the failure
 ** @(RTL11a)@ For clarity, any messages awaiting an @ACK@ or @NACK@ are unaffected by channel state changes i.e. a channel that becomes detached following an explicit request to detach may still receive an @ACK@ or @NACK@ for messages published on that channel later
 * @(RTL4)@ @Channel#attach@ function:
-** @(RTL4a)@ If already @ATTACHED@ nothing is done
+** @(RTL4a)@ If already @ATTACHED@ @ATTACH@ ProtocolMessage is generated and sent using the most recent @ChannelOptions@. Ably (the server) should send the first message after the atach point in full and not as delta message if delta subscription is active.
```
mattheworiordan: Typo, `atach` is `attach`

codemerx: @mattheworiordan you might be looking at older commits. This text has been gone for a while.

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,20 @@ h3(#rest-channel). Channel
...
+
+h3(#plugins). Plugins
+* @(PC1)@ Specific client library features that are not commonly used could be extracted to their own libraries as plugins in order to reduce the size of the client library. While residing in their own library such plugins are considered to be logical part of the client library code and as such are tightly coupled with the client library implementation. The client library is  aware of the plugin specific type and capabilities. I.e. each such plugin is by design coupled to a specific version of the client library.
+* @(PC2)@ The plugins are not expected to share any common methods. Yet there exists the convenience interface @Plugin@ that should be used in the strongly typed languages as the type of the @ClientOptions.plugins@ collection as per "TO3o":#TO3o.
+* @(PC3)@ A plugin provided under the type @PluginType.vcdiff@ should be capable of decoding "vcdiff" encoded messages. It must implement the the @VCDiffDecoder@ interface and the client library must be able to use it by casting it to this interface.
```
QuintinWillison: I'm not sure that the workings of the vcdiff plugin need documenting in the protocol as they will be down to client implementations, some of which will not offer them at all as it'll be baked in (i.e. not optional). Couldn't we just drop PC3 entirely?

content/client-lib-development-guide/features.textile
SimonWoolf: we should add ", if present" at the end, to allow for the possibility that a message may not have a channelSerial

content/client-lib-development-guide/features.textile
```diff
@@ -1901,6 +1917,20 @@ class HttpPaginatedResponse // RSC19b
   errorCode: Int // HP6
   errorMessage: String // HP7
   headers: Dict<String, String> // HP8
+  
+class Plugin // PC2
+  // Empty class/interface. Plugins are not expected to share any common interface.
+  // Still it is better to have a common base type for plugins in statically typed languages.
+  
+  enum PluginType
```
QuintinWillison: I think this may need indenting back.

content/client-lib-development-guide/features.textile
```diff
@@ -1332,6 +1335,16 @@ h4. ChannelOptions
 ** @(TB2b)@ @cipher@, which is either:
 *** @(TB2b1)@ A @CipherParams@ instance, or
 *** @(TB2b2)@ an options hash (or language equivalent) consisting of any subset of @CipherParams@ fields that includes a @key@. In this case, the client library should call @getDefaultParams@, passing it the options hash, to obtain a @CipherParams@ instance
+*** @(TB2c)@ @params@ If provided these are sent to Ably in a new params property @(TR4n)@ of the ATTACH ProtocolMessage. Realtime will echo back the applied params in the ATTACHED if that field was specified in the ATTACH. Params could be:
```
mattheworiordan: >  If provided 

If provided by what.  This is a data structure.

> these are sent to Ably in a new params property @(TR4n)@ of the ATTACH ProtocolMessage

I would think "...a new params property" will soon just become "the @params@ property".  We should not be referencing things as new in the spec.

> Realtime will echo back the applied params in the ATTACHED if that field was specified in the ATTACH.

This spec should not in `ChannelOptions`, whihc is just a datatype.

Under what conditions is an `ATTACH` message sent, and when will it be sent back. Find where that happens in the spec, and then developers will understand under what condition this happens, and how to test it. 

Please note in Textile we use `@` symbol to surround code blocks. This is all covered in the [documentation style guide](https://docs.ably.io/client-lib-development-guide/documentation-formatting-guide/) in examples through this document.

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,20 @@ h3(#rest-channel). Channel
...
+*** @(RSL6c1)@ Following the procedure described in "RTL4c":#RTL4c under the caveats of "RTL4d":#RTL4d, "RTL4e":#RTL4e and "RTL4f":#RTL4f. The @ProtocolMessage.channelSerial@ of the @ATTACH@ @ProtocolMessage@ sent as per "RTL4c":#RTL4c  should be set to the channel serial of the previous message to the message for which "vcdiff" decoding failed.
+*** @(RSL6c2)@ Any subsequent "vcdiff" decoding failures as per "RSL6c":#RSL6c while the recovery procedure from a previous failure as per "RSL6c1":#RSL6c1 is still in progress, i.e. the channel is in @ATTACHING@ state, should not trigger a new recovery procedure as per "RSL6c1":#RSL6c1.
+** @(RSL6d)@ The data payload of the last message on each channel, i.e. @Message.data@ has to be stored at all times since it might be needed to decode the "vcdiff" encoding. The last message bookkeeping rules are:
```
SimonWoolf: > `The data payload of the last message on each channel, i.e. @Message.data@`

This is still technically ambiguous with respect to protocol messages with multiple messages. Maybe `The data payload from the last @Message@ in the @messages@ array of the most recent @ProtocolMessage@ to arrive on the channel (which will have been the most recent @Message@ to be decoded) should be stored at all times, since ...`

(Relatedly: do you think if it might be worth adding a new child of `RSL6` to specify that the messages array must be decoded serially, first-to-last? I can't see anything that actually requires that at the moment, so a client library could theoretically currently be in compliance by decoding them last-to-first, or in parallel)

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,20 @@ h3(#rest-channel). Channel
...
+*** @(RSL6d1)@ If the message encoding string ends in @base64@, the message should be base64-decoded (and the encoding string modified accordingly per "RSL6":#RSL6) before storing the last message
```
SimonWoolf: > before storing the last message

Given we're only storing the payload (and id), not the whole `Message`, I wonder if it might be clearer to refer to storing `the last payload`, in case someone thinks that the use of "message" implies they should be storing the `Message` object

content/client-lib-development-guide/features.textile
```diff
@@ -1594,7 +1610,7 @@ class ChannelStateChange:
 class ChannelOptions:
   +withCipherKey(key: Binary | String)? -> ChannelOptions // TB3
   cipher: (CipherParams | Params)? // RSL5a, TB2b
-
+
```
mattheworiordan: White space.

content/client-lib-development-guide/features.textile
mattheworiordan: > If decoding fails on "vcdiff" encoding that means message delta application failed.

I don't think "that means message delta application failed" is clear enough.  Perhaps:

`Given deltas are applied to the previous message published on a channel, when a "vcdiff" encoding fails to be decoded, it makes it impossible for a client to apply subsequent deltas received from that point of failure forward.  As such, the client must automatically do the following to retrieve the last complete message published:`

...

> Following the procedure described in "RTL4c":#RTL4c under the caveats of "RTL4d":#RTL4d, "RTL4e":#RTL4e and "RTL4f":#RTL4f. The @ProtocolMessage.channelSerial@ of the @ATTACH@ @ProtocolMessage@ sent as per "RTL4c":#RTL4c  should be set to the channel serial of the previous message to the message for which "vcdiff" decoding failed.

I am not clear how `RTL4d` applies in this context, the callback cannot be provided as that is provided as part of `Channel#attach`, and this error arose as part of a message decoding failure.

In regards to `RTL4e`, I don't follow why this is mentioned because a) this is unlikely, b) it's covered in full by https://docs.ably.io/client-lib-development-guide/features/#RTL14.

I think this is more complicated than it needs to be, and could perhaps be simplified as follows:

`Sending an @ATTACH@ @ProtocolMessage@ with the channel serial set to the previous message to the message for which "vcdiff" decoding failed to the server, transitioning the channel state to @ATTACHING@, and waiting for a confirmation @ATTACHED@, as per @RTL4c@ and @RTL4f@`

I have a few other questions relating to discussions we previously had:

* What happens if subsequent messages arrive, with deltas, on the channel before the `ATTACH` is sent and received, specifically in regards to emitting messages.  I understand we don't try and recovery again as per `RSL6c2`, but will we keep emitting broken messages?
* How will a user know that the message they received is broken?  AFAICT, subscribing to messages and receiving a broken delta is never going to be the end user's fault, yet will now result in:
  * A message emitted that a silent failure and a "useless" binary (unusable by the end user, and with no error)
  * Possibility that more messages being sent from the server before the `ATTACH` arrives, i.e. more broken deltas emitted
  * Duplicate messages emitted i.e. we guarantee exactly-once-delivery, yet if message 1 and 2 are received OK, 3 arrives and is broken, so the recovery process starts, and in the mean time 4 and 5 are sent, the user will receive messages 3, 4 and 5 with binary delta (effectively unusable and unexpected broken) payloads, and 3, 4 and 5 again as correct payloads.

I very much dislike everything about the silent failures, broken messages and duplicate messages, in spite of a user not doing anything wrong.

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,20 @@ h3(#rest-channel). Channel
...
+** @(RSL6c)@ If decoding fails on "vcdiff" encoding that means message delta application failed. Since applying a delta depends on the previous message being present a failure in the delta application on the client side makes it impossible to apply any subsequent deltas received. That would render the channel unusable so a full message needs to be requested from Ably (the server) to replace the failed delta message. That is done by:
+*** @(RSL6c1)@ Following the procedure described in "RTL4c":#RTL4c under the caveats of "RTL4d":#RTL4d, "RTL4e":#RTL4e and "RTL4f":#RTL4f. The @ProtocolMessage.channelSerial@ of the @ATTACH@ @ProtocolMessage@ sent as per "RTL4c":#RTL4c  should be set to the channel serial of the previous message to the message for which "vcdiff" decoding failed.
+*** @(RSL6c2)@ Any subsequent "vcdiff" decoding failures as per "RSL6c":#RSL6c while the recovery procedure from a previous failure as per "RSL6c1":#RSL6c1 is still in progress, i.e. the channel is in @ATTACHING@ state, should not trigger a new recovery procedure as per "RSL6c1":#RSL6c1.
```
mattheworiordan: As above, what happens to messages being emitted?

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,20 @@ h3(#rest-channel). Channel
...
+** @(RSL6d)@ The data payload of the last message on each channel, i.e. @Message.data@ has to be stored at all times since it might be needed to decode the "vcdiff" encoding. The last message bookkeeping rules are:
+*** @(RSL6d1)@ If the message encoding string ends in @base64@, the message should be base64-decoded (and the encoding string modified accordingly per "RSL6":#RSL6) before storing the last message
+*** @(RSL6d2)@ If "vcdiff" encoding is present the direct output of decoding the "vcdiff" encoding is stored as a last message, no matter that more encodings might be present
+** @(RSL6e)@ The @id@ of the last received message on each channel has to be stored. In case the next message received is delta message (i.e. contains @vcdiff@ encoding) the stored last message id has to be compared against the @Message.extras.delta.from@ of the delta message. If these are different the delta application fails and the recovery procedure from "RSL6c":#RSL6c must be executed.
```
mattheworiordan: No error is emitted? No log message?  Is the (broken) message still emitted to subscribers of the channel?

SimonWoolf: I'm wondering whether all this really does belong under RSL6 (which is supposed to just be "Message decoding"). Tempted to say everything except the `RSL6b` modification might fit better in the RealtimeChannel spec, as RTL16 -- after all most of the work this spec is talking about will be done by the channel, not by the decode() function. @paddybyers any opinion?

mattheworiordan: Inconsistent use of `@id@` and `id`.

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,20 @@ h3(#rest-channel). Channel
...
+
+h3(#plugins). Plugins
+* @(PC1)@ Specific client library features that are not commonly used could be extracted to their own libraries as plugins in order to reduce the size of the client library. While residing in their own library such plugins are considered to be logical part of the client library code and as such are tightly coupled with the client library implementation. The client library is  aware of the plugin specific type and capabilities. I.e. each such plugin is by design coupled to a specific version of the client library.
```
mattheworiordan: > `The client library is  aware`

Extra space between is and aware.

content/client-lib-development-guide/features.textile
```diff
@@ -1901,6 +1917,20 @@ class HttpPaginatedResponse // RSC19b
...
+  
+  enum PluginType
+  "vcdiff"
+
+class VCDiffDecoder
+  decode([byte] delta, [byte] base) -> [byte]
+  
+class DeltaExtras
+  from: String // the id of the message the delta was generated from
+  format: String //the delta format. For 1.2  only vcdiff is supported
```
mattheworiordan: Inconsistent spacing.  
Why mention 1.2 only?  This is the 1.2 spec.  Perhaps just "Only "vcdiff" is currently supported"

content/client-lib-development-guide/features.textile
mattheworiordan: >  A plugin provided under the type @PluginType.vcdiff@

I had to search around for a bit to understand what you meant here because I assumed that was a type, whereas it's an enum value.  So perhaps:

`A plugin provided with the @PluginType@ enum key value of @vcdiff@`

content/client-lib-development-guide/features.textile
codemerx: Broken messages should not get delivered. As far as I can remember we explicitly had a discussion about duplicates/out of order messages being delivered in this scenario with @paddybyers and @SimonWoolf. It was based on [this PR comment](https://github.com/ably/ably-js/pull/615#discussion_r337599849). A decision was made that this is the behaviour we want but I am not really sure I remember all the reasoning in detail. Could you, guys, elaborate on this?

mattheworiordan: That comment appears to be more about the channel state than what happens with messages.

SimonWoolf: > receiving a broken delta is never going to be the end user's fault, yet will now result in: A message emitted that a silent failure and a "useless" binary (unusable by the end user, and with no error)

In what Tsviatko has implemented in JS and Java, it does not result in that; instead, if the id fails to match or it fails to vcdiff-decode, the message is discarded after logging an error, it is not delivered to the user. Eg see https://github.com/ably/ably-js/pull/615/files#diff-ee8e6cae22c16e2972f2b341a26d2ed0R451 . Further deltas on the channel will suffer the same issue, so will also be discarded.

(admittedly that is currently not made explicit in the spec, which it should be).

So in your scenario, the user will see messages 1 and 2, then a bunch of error messages to the console, then messages 3, 4, and 5, as it should be. (That's if the recovery succeeds -- if it doesn't then the `attached` will have `resumed=false` so the user will get an `update` notifying them of the discontinuity as usual).

There is one scenario where that breaks down, which is if before the reattach succeeds, the server happens to send a full message anyway, which is then processed by the client fine and emitted -- and will then be emitted again after the auto-reattach and recover. We discussed that this morning and decided that we can solve it by specifying that the client library discards messages when not in the `attached` state, and also that it goes into the `attaching` state after initiating the recovery process. That way you get no duplicates and no missing messaging.

That'd be quite a significant change, but we can't think of any reason it'd be invalid -- wdyt @mattheworiordan ?

(The other point, which @paddybyers is keen to emphasize, is that if the server correctly sticks to the protocol then these issues will never happen -- it's never 'expected'. Our (and everyone else's) guarantees are always implicitly qualified by 'assuming the server doesn't have a bug which breaks the guarantee'. So we could have decided to have just had no recovery procedure at all -- it's not like it'd be the only place where the client assumes the server is going to keep to the protocol. But here we're taking extra effort to try and keep to some subset of our guarantees even if the server has a bug (for this particular category of bugs), because we think it's a bug we're particularly likely to run into)

content/client-lib-development-guide/features.textile
codemerx: I am not sure I understand your point. PC3 defines the interface between the vcdiff plugin and the client library not the internal workings of the vcdiff plugin. If PluginType.vcdiff exists we have to be defining how it can be used I guess.

content/client-lib-development-guide/features.textile
codemerx: At the time of your review this should not have been part of this PR any more. I removed this spec completely and left it to https://github.com/ably/docs/pull/766 to document channel params. You must have been looking at some older version of this PR.

content/client-lib-development-guide/features.textile
codemerx: Added an extra point specifying message decoding order.

content/client-lib-development-guide/features.textile
mattheworiordan: > That'd be quite a significant change, but we can't think of any reason it'd be invalid -- wdyt

Agreed, can't think of an issue.

content/client-lib-development-guide/features.textile
marto83: I agree about moving the spec items under RealtimeChannel. If you implementing the Rest library only then all the above points are not relevant.

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,23 @@ h3(#rest-channel). Channel
...
+*** @(RSL6c3)@ Send an @ATTACH@ @ProtocolMessage@ with the channel serial set to the previous message to the message for which "vcdiff" decoding failed to the server, transitioning the channel state to @ATTACHING@, and waiting for a confirmation @ATTACHED@, as per @RTL4c@ and @RTL4f@
+*** @(RSL6c4)@ Any subsequent "vcdiff" decoding failures as per "RSL6c":#RSL6c while the recovery procedure from a previous failure as per "RSL6c3":#RSL6c3 is still in progress, i.e. the channel is in @ATTACHING@ state, should not trigger a new recovery procedure as per "RSL6c3":#RSL6c3.
+** @(RSL6d)@ The data payload of the last message on each channel, i.e. @Message.data@ has to be stored at all times since it might be needed to decode the "vcdiff" encoding. The last message bookkeeping rules are:
+*** @(RSL6d1)@ If the message encoding string ends in @base64@, the message should be base64-decoded (and the encoding string modified accordingly per "RSL6":#RSL6) before storing the last payload.
+*** @(RSL6d2)@ If "vcdiff" encoding is present the direct output of decoding the "vcdiff" encoding is stored as a last message, no matter that more encodings might be present
```
QuintinWillison: Perhaps "the last message" rather than "a last message" as otherwise this suggests there could be more than one last message stored.

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,23 @@ h3(#rest-channel). Channel
...
+** @(RSL6e)@ The @id@ of the last received message on each channel has to be stored. In case the next message received is delta message (i.e. contains @vcdiff@ encoding) the stored last message @id@ has to be compared against the @Message.extras.delta.from@ of the delta message. If these are different the delta application and hence the message decoding fail. The recovery procedure from "RSL6c":#RSL6c must be executed.
```
QuintinWillison: "is delta message" reads better as "is a delta message".
"hence the message decoding fail" needs a "should" before the word "fail".

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,23 @@ h3(#rest-channel). Channel
...
+* @(PC2)@ The plugins are not expected to share any common methods. Yet there exists the convenience interface @Plugin@ that should be used in the strongly typed languages as the type of the @ClientOptions.plugins@ collection as per "TO3o":#TO3o.
+* @(PC3)@ A plugin provided with the @PluginType@ enum key value of @vcdiff@ should be capable of decoding "vcdiff" encoded messages. It must implement the the @VCDiffDecoder@ interface and the client library must be able to use it by casting it to this interface.
+** @(PC3a)@ The base argument of the @VCDiffDecoder.decode@ method should receive the last message on a channel as specified by "RSL6d":#RSL6d. If the last message is a string it should be converted to utf-8 before being passed as base argument of the @VCDiffDecoder.decode@ method.
```
QuintinWillison: UTF-8 should be in capital letters as it's unquoted.

content/client-lib-development-guide/features.textile
```diff
@@ -554,6 +571,7 @@ h3(#realtime-channel). Channel
...
 ** @(RTL13c)@ If the connection is no longer @CONNECTED@, then the automatic attempts to re-attach the channel described in "RTL13b":#RTL13b must be cancelled as any implicit channel state changes subsequently will be covered by "RTL3":#RTL3
 * @(RTL14)@ If an @ERROR ProtocolMessage@ is received for this channel (the channel attribute matches this channel's name), then the channel should immediately transition to the FAILED state, and the @Channel.errorReason@ should be set
+* @(RTL16)@ The first message that will be delivered to the channel subscribers after the channel enters @ATTACHING@ state is the first message received after the channel enters @ATTACHED@ state. Any messages originating from any previous attachment that were not delivered to the channel subscribers at teh moment the channel entered the @ATTACHING@ state shall be ignored.
```
QuintinWillison: typo "teh" > "the"

content/client-lib-development-guide/features.textile
QuintinWillison: My preference remains to reverse these args but it feels like I'm alone in this. :cry:

marto83:
I’m looking at the spec items about delta recovery. The spec makes it clear about what I need to do in the happy path where the recovery succeeds. However it doesn't cover what happens if the recovery fails. 
Let’s stay we start a recovery and send a new ATTACH message but that times out - the default behaviour will kick in and the channel will become SUSPENDED. What happens when we re-attach it (no matter who triggered that). Are we still in recovery mode and send the serial of the last message? Or do we forget about it and you just get attach from that moment on.

content/client-lib-development-guide/features.textile
codemerx: This should be addressed in the new edit of RSL6c

codemerx:
> 
> 
> I’m looking at the spec items about delta recovery. The spec makes it clear about what I need to do in the happy path where the recovery succeeds. However it doesn't cover what happens if the recovery fails.
> Let’s stay we start a recovery and send a new ATTACH message but that times out - the default behaviour will kick in and the channel will become SUSPENDED. What happens when we re-attach it (no matter who triggered that). Are we still in recovery mode and send the serial of the last message? Or do we forget about it and you just get attach from that moment on.

We do not specify recovery mode for the channel. We specify recovery actions that should be taken. If the ATTACH never results in ATTACHED reply then one just follows the spec, i.e. the channel becomes SUSPENDED and so on. Whatever happens to it later on should not depend on the failed delta decoding recovery attempt.

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,22 @@ h3(#rest-channel). Channel
...
+** @(RSL6c)@ Given "vcdiff" encoded deltas are applied to the previous message published on a channel, when a "vcdiff" encoding fails to be decoded, it makes it impossible for a client to apply subsequent deltas received from that point of failure forward. As such, the client must automatically execute the following recovery procedure in lieu of "RTL7e":#"RTL7e":
+*** @(RSL6c1)@ Log error with code 40018
+*** @(RSL6c2)@ Discard the message
+*** @(RSL6c3)@ Send an @ATTACH@ @ProtocolMessage@ with the channel serial set to the previous message to the message for which "vcdiff" decoding failed to the server, transitioning the channel state to @ATTACHING@, and waiting for a confirmation @ATTACHED@, as per @RTL4c@ and @RTL4f@
+** @(RSL6d)@ The data payload of the last message on each channel, i.e. @Message.data@ has to be stored at all times since it might be needed to decode the "vcdiff" encoding. The last message bookkeeping rules are:
+*** @(RSL6d1)@ If the message encoding string ends in @base64@, the message should be base64-decoded (and the encoding string modified accordingly per "RSL6":#RSL6) before storing the last payload.
```
paddybyers: Is the removal of `base64`, on the assumption that it is an encoding step required by the transport, required unconditionally, or only in the case of a text transport?

content/client-lib-development-guide/features.textile
```diff
@@ -267,7 +268,23 @@ h3(#rest-channel). Channel
...
+** @(RSL6c)@ Given "vcdiff"-encoded deltas are applied to the previous message published on a channel, when a "vcdiff" encoding fails to be decoded, it makes it impossible for a client to apply subsequent deltas received from that point of failure forward. As such, the client must automatically execute the following recovery procedure in lieu of "RTL7e":#"RTL7e":
+*** @(RSL6c1)@ Log error with code 40018
+*** @(RSL6c2)@ Discard the message
+*** @(RSL6c3)@ Send an @ATTACH@ @ProtocolMessage@ with the @channelSerial@ set to the previous message to the message for which "vcdiff" decoding failed to the server, transitioning the channel state to @ATTACHING@, and waiting for a confirmation @ATTACHED@, as per @RTL4c@ and @RTL4f@
```
marto83: When we transition the channel back to Attaching. Should we emit an Error with Code 40018 on the channel as well?

content/client-lib-development-guide/features.textile
codemerx: This has been discussed a couple of times and @SimonWoolf provided a good summary of these discussions in Slack:

"I've given my opinion on this before, but will reiterate here for posterity: I would be against this because it adds complexity and is unnecessary.
In particular, in ably-js, the 'obvious' way to do it (looking at whether the currently active transport is text or binary) is subtly buggy, because during the upgrade period there is a period where the active transport is still the comet transport but an incoming message can validly arrive on the upgrade (websocket) transport.
Of course if this were something that was genuinely necessary, there are ways around that: e.g. annotating the message with the format while at the transport level, or decoding base64 if-text-transport as part of deserializing the ProtocolMessage, before it's delivered to the Channel. But it isn't necessary. There is no possibility of a message coming down a binary transport with an encoding string ending in /base64, because even if someone did do some base64-encoding themself and publish a message with the encoding field already set to base64, then (as noted in the call) realtime will have assumed that any base64 encoding in an inbound message was added by the transport and removed it on entering the system. So this would be extra complexity that serves no purpose. (edited) 
simon 5:52 PM
(also I suspect that no customer has or will ever publish a message with an encoding string pre-set. It's one of those things that the system technically supports but no-one ever does because if they want to have app-level encoding info there's no reason for them to use our encoding system rather than have their own (as part of the data object) which they know the client libs aren't going to mess with)"

content/client-lib-development-guide/features.textile
SimonWoolf: > Should we emit an Error with Code 40018 on the channel as well?

@marto83 that isn't a thing, there is no `Error` event on a channel. Channels should only emit ChannelEvents, which are enumerated at https://www.ably.io/documentation/realtime/types#channel-event , and do not include `Error`. Is ably-dotnet emitting `Error` events on the channel at the moment?

content/client-lib-development-guide/features.textile
```diff
@@ -565,6 +571,18 @@ h3(#realtime-channel). Channel
 * @(RTL14)@ If an @ERROR ProtocolMessage@ is received for this channel (the channel attribute matches this channel's name), then the channel should immediately transition to the FAILED state, and the @Channel.errorReason@ should be set
 * @(RTL16)@ @Channel#setOptions@ takes a @ChannelOptions@ object and sets or updates the stored channel options.
 ** @(RTL16a)@ If the user has provided either @ChannelOptions.params@ or @ChannelOptions.modes@ and the channel is in either the @attached@ or @attaching@ state, @Channel#setOptions@ sends an @ATTACH@ message to the server with the params & modes encoded per "@RTL4@":#RTL4, and indicates success once the server has replied with an @ATTACHED@ (or indicates failure if the channel becomes detached or failed before that happens, as with @Channel#attach@); else it indicates success immediately
+* @(RTL17)@ From the time that the channel enters @ATTACHING@ state, no further received messages are passed to subscribers until after the channel next enters the @ATTACHED@ state. Any messages received after the event that triggered the transition to @ATTACHING@, and prior to the transition to @ATTACHED@, must be discarded.
```
SimonWoolf: It seems odd to special-case the `attaching` state in the spec. I know that that was our motivation for making the change, but I don't think the general client library API should special-case that state over all the other states that the user doesn't expect to receive messages. I'd argue that a better form of this spec would be something like `* @(RTL17)@ Message listeners should only be called if the channel is in the @ATTACHED@ state. Any messages received after when the channel is in any state other than @ATTACHED@ should be discarded`

content/client-lib-development-guide/features.textile
```diff
@@ -267,8 +268,13 @@ h3(#rest-channel). Channel
...
 *** @(RSL6a2)@ A set of tests must exist to ensure that the client library provides interoperability for the @extras@ field which is a JSON-encodable object (ie a value that represents a JSON @object@ value and supports serialization to and from JSON text). The test, at a minimum, should publish a message with an @extras@ object such as @{"push":[{"title":"Testing"}]}@ and ensure it is received with an equivalent JSON-encodable object
-** @(RSL6b)@ If, for example, incompatible encryption details are provided or invalid Base64 is detected in the message payload, an error message will be sent to the logger, but the message will still be delivered with last successful decoding and the @encoding@ field. For example, if a message had a decoding of "utf-8/cipher+aes-128-cbc/base64", and the payload was successfully Base64 decoded but the payload could not be encrypted because the @CipherParam@ details were not configured, the message would be delivered with a binary payload and an @encoding@ with the value "utf-8/cipher+aes-128-cbc"
-* @(RSL7)@ @Channel#setOptions@ takes a @ChannelOptions@ object and sets or updates the stored channel options, then indicates success
```
SimonWoolf: Why are you removing RSL7?

content/client-lib-development-guide/features.textile
```diff
@@ -565,6 +571,18 @@ h3(#realtime-channel). Channel
...
+** @(RTL19c)@ In the case of a delta message with a @vcdiff@ @encoding@ step, the @vcdiff@ decoder must be used to decode the base payload of the of delta message, applying that delta to the stored base payload. The direct result of that vcdiff delta application, before performing any further decoding steps, is stored as the updated base payload.
+* @(RTL20)@ The @id@ of the last received message on each channel must be stored along with the base payload. When processing a delta message (i.e. one whose @encoding@ contains @vcdiff@ step) the stored last message @id@ must be compared against the delta reference @id@, indicated in the @Message.extras.delta.from@ field of the delta message. If the delta reference @id@ of the received delta message does not equal the stored @id@ corresponding to the base payload, the message decoding must fail. The recovery procedure from "RTL18":#RTL18 must be executed. 
+* @(RTL21)@ The messages in the @messages@ array of a @ProtocolMessage@ should each be decoded in ascending order of their index in the array.
+* @(RTL22)@ @Channel#setOptions@ takes a @ChannelOptions@ object and sets or updates the stored channel options, then indicates success
```
SimonWoolf: This is a duplicate of RTL16

content/client-lib-development-guide/features.textile
codemerx: TBH I do not think it makes much of a difference in terms of what has to be implemented. I am fine with your edit, too. However, I think the wording here comes from @paddybyers so maybe he has a comment on this?

Also, is it correct to talk about "messages received"? The messages might be received before entering the `ATTACHING` state and they still will have to be discarded.

content/client-lib-development-guide/features.textile
codemerx: I moved it to RTL22 but that was a mistake actually

content/client-lib-development-guide/features.textile
QuintinWillison: No harm in specifying, I guess, even though it looks like we'll only be needing this for JS.

content/client-lib-development-guide/features.textile
SimonWoolf: > Also, is it correct to talk about "messages received"? The messages might be received before entering the ATTACHING state and they still will have to be discarded.

Fair enough, happy to go with `No messages should be passed to subscribers if the channel is in any state other than @ATTACHED@`.

The main reason I prefer this is it feels weird for the spec to theoretically be fine with a message being emitted while `detached`, `failed`, `suspended` etc., but not while `attaching`. Even if we know that can never happen, it feels wrong that just `attaching` is singled-out for this. (One case where the change will make a difference is the `detaching` state: with your version you can still have messages emitted during `detaching`, which feels inconsistent; in mine you won't.)

codemerx: Ok fixed now.

content/client-lib-development-guide/features.textile
```diff
@@ -565,6 +572,17 @@ h3(#realtime-channel). Channel
...
+* @(RTL17)@ No messages should be passed to subscribers if the channel is in any state other than @ATTACHED@.
+* @(RTL18)@ Given "vcdiff"-encoded deltas are applied to the previous message published on a channel, when a "vcdiff" encoding fails to be decoded, it makes it impossible for a client to apply subsequent deltas received from that point of failure forward. As such, the client must automatically execute the following recovery procedure in lieu of "RTL7e":#"RTL7e":
+** @(RTL18a)@ Log error with code 40018
+** @(RTL18b)@ Discard the message
+** @(RTL18c)@ Send an @ATTACH@ @ProtocolMessage@ with the channel serial set to the previous message to the message for which "vcdiff" decoding failed to the server, transitioning the channel state to @ATTACHING@, and waiting for a confirmation @ATTACHED@, as per @RTL4c@ and @RTL4f@. @ChannelStateChange.reason@ should be set to @ErrorInfo@ object with with code 40018.
```
paddybyers: `channel serial -> @channelSerial@`

content/client-lib-development-guide/features.textile
```diff
@@ -565,6 +572,17 @@ h3(#realtime-channel). Channel
...
+* @(RTL19)@ The data payload of the last message on each channel must be stored at all times since it will be needed to decode any subsequent message that has a "vcdiff" encoding step. The stored value is the "base payload" of the most recent message; this is the @data@ member of the message, in string or binary form, once all application-level encoding steps have been applied. The base payload is derived initially by processing a non-delta message; the processing and bookkeeping rules are as follows:
+** @(RTL19a)@ When processing any message (whether a delta or a full message), if the message @encoding@ string ends in @base64@, the message @data@ should be base64-decoded (and the encoding string modified accordingly per "RSL6":#RSL6).
```
paddybyers: `encoding -> @encoding@`

content/client-lib-development-guide/features.textile
```diff
@@ -1085,7 +1103,7 @@ h4. Message
...
 ** @(TM2d)@ @data@ string, buffer or JSON-encodable object or array
 ** @(TM2e)@ @encoding@ string
-** @(TM2i)@ @extras@ JSON-encodable object, used to contain any arbitrary key value pairs which may also contain other primitive JSON types, JSON-encodable objects or JSON-encodable arrays. The extras field is provided to contain message metadata and/or ancillary payloads in support of specific functionality, e.g. push. Each of these supported extensions is documented separately; for 1.1 the only supported extension is @push@, via the @extras.push@ member. The processing of any other members is undefined
+** @(TM2i)@ @extras@ JSON-encodable object, used to contain any arbitrary key value pairs which may also contain other primitive JSON types, JSON-encodable objects or JSON-encodable arrays. The extras field is provided to contain message metadata and/or ancillary payloads in support of specific functionality, e.g. push. Each of these supported extensions is documented separately; for 1.1 the only supported extension is @push@, via the @extras.push@ member; 1.2 adds the @delta@ extension which is of type @DeltaExtras@. The processing of any other members is undefined
```
paddybyers: `extras -> @extras@`

PR #781, opened 2019-12-05
Clarify auth server time requirements and queryTime

SimonWoolf:
In the end Paddy decided against making the time offset static, so for now (pre-jwt), the only actions are docs clarifications re auth server time syncing and client lib instance sharing with queryTime


mattheworiordan:
> In the end Paddy decided against making the time offset static

Why?

SimonWoolf:
>> In the end Paddy decided against making the time offset static

> Why?

cc @paddybyers for a definitive answer. (I believe something along the lines of: it did used to be static in ably-java, he then made it an instance variable because having it static was screwing up the tests for some reason)

PR #785, opened 2019-12-31
Add more explanation of bulk publishing limits

SimonWoolf:


content/rest-api/beta.textile
```diff
@@ -30,6 +30,8 @@ This means that caller can usually make the call and have access to the success
 
 All batch-mode api calls are limited to addressing at most 100 different channels per request.
 
+All POST/PUT requests (such as bulk publish) are subject to a maximum request body size of 2MB.
```
mattheworiordan: I think @POST@ and @PUT@ should be styled as code blocks IMHO

mattheworiordan: We use `MiB` style for byte sizes unilaterally to avoid confusion in some instances where `MB` can mean 2000 KiB or can also mean 2048 KiB.  We had a big discussion on this before, but sadly I can't find a link to it.

content/rest-api/beta.textile
```diff
@@ -109,6 +111,10 @@ There is an obvious mapping from the request body to the array of individual pub
 
 The 100 channel limit caps the total number of _different_ channels in the request, irrespective of how they are arranged. So if the same channel name occurs in multiple @BatchSpec@ objects in a single requests, it will contribute only one channel towards the limit.
 
+The @messages@ array of each @BatchSpec@ is treated atomically (for each channel), and so subject to the usual "max message size limits":https://support.ably.io/support/solutions/articles/3000035792 — the total size of all @Messages@ in the array must be less than the limit (64kB for paid accounts, 16kB for free accounts). If the total number of messages you want to publish to one channel adds up to more than that, you can just split the messages for that channel up into multiple batchspecs. All you lose is atomicity (that is, it is then possible that the @Messages@ from one batchspec will be accepted, and the other rejected).
```
mattheworiordan: `KiB` as opposed to `kB`

content/rest-api/beta.textile
mattheworiordan: I found what we documented, if you look in our doc ["Ably Realtime Copy Guidelines & Dictionary"](https://docs.google.com/document/d/1elqHBcMoNk3cOkxcJtd-t4v8WW55d3EfBVRYvT88mvQ/edit), there is a section:

Bandwidth and storage uses the [IEC standard](https://en.wikipedia.org/wiki/IEC_60027) always to avoid ambiguity.  Therefore, we use bytes, KiB, MiB, GiB, and TiB.  We never use KB, GB, TB as one KB can mean 1,000 bytes or 1,024 bytes.  See [internal discussion](https://github.com/ably/website/issues/1597).

PR #788, opened 2020-01-07
Add RTL6d7: no bundling if realtime message id set

SimonWoolf:
Fixes https://github.com/ably/docs/issues/787

ably-js implementation: https://github.com/ably/ably-js/pull/623

TODO: open issues for any realtime client lib that does bundling-while-disconnected (not sure which of them do)

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -518,6 +518,7 @@ h3(#realtime-channel). Channel
 *** @(RTL6d4)@ Messages can only be bundled together if they are of the same type (that is, @Message@ versus @PresenceMessage@)
 *** @(RTL6d5)@ Only contiguous messages in the queue can be bundled together. For example, if the user publishes three messages, A, B, and C, of which A and C could be bundled together under @RTL6d1-4@ but B could not, then no bundling should occur
 *** @(RTL6d6)@ The order of messages in the resulting @ProtocolMessage@ Messages must match the publish order. For example, if the user publishes @Message@ D, then the @Message@ array [E, F], then @Message@ G, the final @messages@ array should be [D, E, F, G]
+*** @(RTL6d7)@ Messages must not be bundled if any have had had their @Message.id@ property set
 ** @(RTL6e)@ Unidentified clients using "Basic Auth":https://en.wikipedia.org/wiki/Basic_access_authentication (i.e. any @clientId@ is permitted as no @clientId@ specified):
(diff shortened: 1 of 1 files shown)


mattheworiordan:
I can't help feeling we are adding a whole lot of complexity with bundling that could simply be avoided by disabling bundling completely in the realtime libs, with I expect near on no meaningful impact for our users.  Is this perhaps a rare opportunity to remove bundling and simplify our spec?

SimonWoolf:
> Is this perhaps a rare opportunity to remove bundling and simplify our spec?

It does give a genuine benefit, albeit for a fairly niche use-case: consider someone publishing 10 tiny messages per second (eg regular gps updates) who's disconnected from the internet for 30s. Without bundling, on reconnection they'd try to publish 300 messages at once and get throttled (not to mention cause a bit of an instantaneous load spike on the connection & channel). With bundling, those will all be nicely bundled up into one (or two or three) protocolmessages, which go through easily. Plus 1 protocolmessage with 300 messages is much more bandwidth-friendly on the subscribers than 300 protocolmessages. 

If we were starting fresh now, then I agree, I don't think we'd bother with it. But given it's already there in ably-js it seems a shame to actively remove it. (Unless/until we need to rewrite it or something, but this is a pretty small change - https://github.com/ably/ably-js/pull/623 ).

When writing RTL6d I did deliberately use '`may`', so it's optional for client libraries (`Messages that have been queued may be sent in a single ProtocolMessage by bundling them...`). It probably wouldn't be worthwhile for any client libs other than the most popular ones in intermittent-connectivity environments (ie js, android, and ios) to implement it.

Previously very briefly discussed in https://github.com/ably/realtime/issues/603

mattheworiordan:
> It does give a genuine benefit, albeit for a fairly niche use-case

Well quiet.  That's my point I suppose.  We can imagine lots of use-cases I am sure to justify any feature we provide :)

> If we were starting fresh now, then I agree, I don't think we'd bother with it. But given it's already there in ably-js it seems a shame to actively remove it.

Sure, but that approach doesn't scale.  One has to be bold to simplify 💪

paddybyers:
I'd be ok with removing bundling. In the example you gave, a client shouldn't be generating messages routinely and periodically whilst disconnected. I'm ok in this case advising the customer that they should be taking notice of connection state if their publishing comprises status updates that are always superseded within a short period of time.

SimonWoolf:
After discussion with Paddy and Quintin, resolving this for now with https://github.com/ably/docs/pull/788/commits/5ae9b4937b4a7adffc35781952337692adf046c9 (stating that while the protocol permits bundling (and outlining the requirements if it is to be done), client libs [should not](https://tools.ietf.org/html/rfc2119) in general do it).

PR #794, opened 2020-01-13
Made it clearer that a token needs to be renewed when requesting one

### LINKED ISSUE #775: Customer feedback: make it clearer that authCallback should fetch a fresh tokenRequest each time
A customer was fetching a tokenRequest from the auth server before initing the client, and then specifying an`authCallback` of `(_, cb) => cb(null, tokenRequest)`. After I explained that the authCallback had to fetch a new, fresh token/tokenRequest each time, I asked if we could have made the docs on this clearer. Their response:

> Just for some context, I joined the project rather late and did a partial rewrite of the client.
> Having said that, docs can probably be improved. Under the client library SDK section, authentication seems split between the Authentication and Constructor sections. I just noticed that API reference is listed at the bottom of the page. I don’t see my method of authentication there because I only see the explicit auth methods on the SDK. I can see there’s an authCallback property but I have to wonder if this is the same as the one in ClientOptions on the other page. It’s disjointed.
> More specifically, it says “which is called when a new Ably-compatible token is required” which doesn’t make it clear that this will be called multiple times. This would be a good time to call out why tokens should not be kept in memory on the client. Again, I inherited the code and this option was already used so I am working backwards rather than forwards as the docs kind of assume.

(There may be an element of the current docs assuming that it's obvious just from the fact that it's a callback that it may be called multiple times and shouldn't just return the same thing each time, as if not, it could just take a token request rather than a callback. But things which are obvious to us may well not be to a new customer: https://en.wikipedia.org/wiki/Illusion_of_transparency )

tomczoink:
fixes #775


content/types/_auth_options.textile
```diff
@@ -19,9 +19,9 @@ h4.
...
...back(err, tokenOrTokenRequest))@</span><span lang="java">@TokenCallback@ instance</span><span lang="python">callable (eg a lambda)</span><span lang="ruby">proc / lambda (called synchronously in REST and Realtime but does not block EventMachine in the latter)</span> which is called when a new Ably-compatible token is required. The token returned should be different each time it's called. The role of the callback is to _new_ one of the following: an Ably Token string (in plain text format); a signed "@TokenRequest@":/realtime/types#token-request ; a "@TokenDetails@":/realtime/types#token-details (in JSON format); an "Ably JWT":/core-features/authentication#ably-jwt. These may be submitted automatically by the library to the "Ably REST API @requestToken@":/rest-api#request-token. See "an authentication callback example":<%= JsBins.url_for('authentication/auth-callback') %> or "our authentication documentation":/rest/authentication for details of the Ably TokenRequest format and associated API calls.<br>__Type: <span lang="default">@Callable@</span><span lang="java">@TokenCallback@</span><span lang="ruby">@Proc@</span><span lang="csharp">@Func<TokenParams, Task<TokenDetails>>@</span>__
```
SimonWoolf: > `The role of the callback is to _new_ one of the following`

verb?

SimonWoolf: > `The role of the callback is to _new_ one of the following`

verb?

SimonWoolf:
Added some [more wording tweaks](https://github.com/ably/docs/pull/794/commits/3810508b2b676c1040f7cf787d1137efad3fddb4), and copied the wording across to client_options as well.

(@tomczoink do you want to have a go at DRYing up the shared bits of auth_options and client_options? there's lots of duplication, and this isn't the first time one of them's been updated without the other)

tomczoink:
Good point on the duplication, I'll create an issue on it now and implement it this week.

PR #798, opened 2020-01-16
Recover missing "set up APNs" step in iOS push activation instructions. 

tcard:
This paragraph was gutted at some point and it refers to a "now" after
a "set up APNs" instruction that isn't there anymore. So, bring it back.


tcard:
Can somebody merge this? I'm not authorized.

PR #800, opened 2020-01-17
fixes #621 - Adds jsbin example for connection recovery in browser

### LINKED ISSUE #621: Add example of how to use the browser recover option
I put together a simple demo on how to use the `recover` option with the JS client, when refreshing a page and using presence.  I think it provides a reasonable example of how a developer can utilise this option and explicitly manage entering again.

See https://jsbin.ably.io/ajided/14/edit, and https://jsbin.ably.io/oxesuy/8/edit

tomczoink:
fixes #621 Adds it to the existing connection recovery with no filters code sample. Potential to add more explicit reference to it elsewhere.

### DIFF
--- a/content/code/realtime/connection-recover.code
+++ b/content/code/realtime/connection-recover.code
@@ -1,0 +1,95 @@
+[--- Javascript ---]
+var ably = new Ably.Realtime({
+      key: '{{API_KEY}}',
+      recover: function(_, cb) { cb(true); },
+      clientId: 'CHANGE-THIS-IF-YOU-WANT-MULTIPLE-CLIENTS'
+    }),
+    channel = ably.channels.get('recovery-presence'),
+    $present = $('#present');
+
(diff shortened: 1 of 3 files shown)


content/code/realtime/connection-recover.code
```diff
@@ -0,0 +1,95 @@
...
+[--- /Javascript ---]
+
+[--- HTML ---]
+<html>
+<head>
+  <script src="//cdnjs.cloudflare.com/ajax/libs/jquery/2.2.4/jquery.min.js"></script>
+  <script src="//cdn.ably.io/lib/ably-1.js"></script>
+</head>
+<body>
+  <h1><a href="https://www.ably.io" target="_blank"><img src="/images/favicon.png">Ably connection recovery across reload</a> example</h1>
```
Srushtika: I think it's worth linking this directly connection recovery docs instead of the homepage.

PR #804, opened 2020-01-21
Condensed Token methods from auth_options and client_options into a partial

### LINKED ISSUE #802: Reduce duplication between auth_options and client_options
Currently both AuthOptions and ClientOptions have overlap in content. We should DRY this up.

tomczoink:
fixes #802 , condensed them into a single partial.


Srushtika:
Looks like there's some error deploying though, possibly an issue unrelated to the PR

tomczoink:
Yep, this PR is relying on partial stacking which I'm working on now, I'll merge this in after that. Thanks 👍

PR #810, opened 2020-02-06
Spec: ChannelParams tweak to expose them in the attached for any way of specifying them

SimonWoolf:
As discussed.

You're right, there is indeed a forwards-compatibility requirement, https://docs.ably.io/client-lib-development-guide/features/#RSF1 , which I'd not really noticed -- I assume we added it after the previous trouble we had. That has been in there for a couple years now (pre-1.1). So, ok, perhaps it's time we can start taking advantage of that, and not bother with version sniffing for this.

(Arguably we should still sniff for 1.1 as libs before that wouldn't have had to implement that robustness principle, but shrug, at that point we've got a pretty good chance that no customers would actually be affected, so we could probably just forget it unless/until anyone complains)


SimonWoolf:
Added clarification when attached message contains no params: https://github.com/ably/docs/pull/810/commits/8e52726b0e75f0e05727f019af7cfb79be70b68e (cc @codemerx )

PR #811, opened 2020-02-07
Updated partials structure to work with stacked partials

tomczoink:
Main changes:
- Moves all partials into a base-level `partials` folder. This is useful for avoiding them being compiled, as well as going forward making it easier to version.
- Changes partials from compiling from a file (File.load...) to using compiled content (@items[...].compiled_content). Avoids both partials existing as acessible pages as well as a small speed bump.
- Version check no longer automatically upgrades to latest version of a file if an equivelent version file doesn't exist. This should help with detecting if something's been done incorrectly.


Rules
```diff
@@ -43,6 +43,10 @@ compile "/code/**/*" do
   filter :ably_jsbin_publisher
 end
 
+compile "/partials/**/_*" do
```
mattheworiordan: @tomczoink please see https://wiki.ably.io/pages/Git-best-practice.html in regards to git commit messages.  Anyone (including me) looking at this commit won't understand why this change was needed.  You are only stating in the commit message what we can tell from the code itself.  Why was this changed done, why is it needed?

PR #813, opened 2020-02-11
Added further details on API key permissions

tomczoink:


content/core-features/authentication.textile
```diff
@@ -193,7 +193,9 @@ API keys and their capabilities are "configured using the dashboard":https://sup
...
+An "Ably API key":https://support.ably.io/solution/articles/3000030054 can have single set of permissions, applied to any number of channels or "queues":/general/queues. To create a key with certain permissions, simply go to create a new API key and "select the desired permissions":https://support.ably.io/support/solutions/articles/3000030502.
+
+You can also choose whether to restrict the API key to only channels, only "queues":/general/queues, or to match a set of channel/queue names. If you've chosen to restrict the API key to *selected channels and queues*, you can use a comma seperated list of resources the API key can access, making use of "wildcards":#wildcards to provide access to areas of your app. It is worth noting an API key will provide the same permissions to all resources it has access to.
```
paddybyers: s/seperated/separated/

" It is worth noting an API key will provide the same permissions to all resources it has access to." I'm not sure what this means but, if it means what I think it means, it's not true. The capability map for a key defines a set of permitted operations for each resource, and they can be different for different resources.

tomczoink: My understanding was that an API Key can only have one set of permissions, which will result in the intersection of said permissions and each resource's valid permissions as the overall permissions provided to a resource (resource being a channel/queue). So if I set it as `subscribe, publish` for all queues and channels, queues would end up with just `subscribe` and channels would have both. Is this correct, or have I misunderstood? I can try to clarify it better if correct.

paddybyers: You've misunderstood I think. A capability can say:

```
{
  "foo": ["publish"],
  "bar":["history"],
  "[queue]*":["subscribe"]
}
```

So there can be different permissions enabled explicitly for different resources (queues, channels).

tomczoink: Okay, as far as I can see there's no way for an API key to be given different capabilities however (even if it's possible from our side), the options we provide for a single API key are a single set of permissions for set of potential resources. If there is a way for a user to specify different permissions for different resources within the same API key I'd appreciate knowing how as that'd be very useful information. If not, I could still just remove this caveat, as it's somewhat self-evident from the process itself.

Screenshot 2020-02-11 at 11 11 32

SimonWoolf: I think the wording as you have it is fine. We know that the underlying system allows different ops for each resource, and it's just a website restriction to make the dashboard easier to use, but the effect on the customer is no different from if the system didn't allow it.

(Of course we may end up having to reword it if we add an api key provisioning system, with access to the full flexibility. But we don't have that yet.)

PR #814, opened 2020-02-12
Removes specific pricing from comparisons

remotekieran:
I removed specific pricing comparisons from compare table. We're updating the pricing and this section doesn't add much value anyway.

@tomczoink can you please just give this a once over and then merge into master if all is OK?

I've also updated the comparisons that don't live in this table directly in the website repo.

### DIFF
--- a/data/compare.yaml
+++ b/data/compare.yaml
@@ -629,24 +629,11 @@ AblyCompare:
           ably: >
             *Yes.*<br/><br/>Ably's pricing is simple and transparent. You pay for the messages, peak active channels and peak connections you use for the month. You can either pay for what you've used at the end of the month, or reserve capacity in advance each month and benefit from a discount.
           pubnub: >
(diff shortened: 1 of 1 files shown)


mattheworiordan:
> this section doesn't add much value anyway.

Do you not think we're missing one key metric that does add value which is pricing comparison of a message vs Ably's message price?  It is only one metric, but may help customers understand how our pricing is more competitive on that one dimension at least.  Otherwise there is no way for a customer to understand that as they scale, the pricing will adversely diverge significantly from PubNub's

remotekieran:
The section as it previously was didn't add much value because, in my opinion, all it displayed was that Ably was the much cheaper option. As we're trying to move away from differentiating on price, I think it was the right choice to remove a side-by-side comparison of price where the outcome was a bold price much lower than the competition.

But I do agree that a per message price comparison adds value. I'll add this to my backlog and update soon @mattheworiordan.

PR #819, opened 2020-02-21
Update to Reactor WebHook + Function documentation

tomczoink:
Since various changes have happened surrounding Reactor, the documentation had become out of date.  This PR makes various changes to address this:

- Merges '/general/webhooks' and '/general/functions' into a single page, '/general/events'. Now that we're introducing new integrations alongside functions with the same functionality, with deviation existing between functions (AWS Lambda vs Azure for example) rather than between functions and non-functions, it didn't make sense for a documentation standpoint to continue seperating them. This step makes documentation more in-line with the actual user experience of setting up a Rule
- Adds a new sub page to `/general/events` for each integration (for example `/general/events/zapier`). This allows us to discuss limitations and go into detail on each integration, providing relevant details for each integration in a more consolidated area.

@SimonWoolf I think some of what I've got it here wrong (namely the AWS Lambda page), however I'm not sure on the actual specifics involved. Any help is appreciated. In addition, I have added a section on batch-related transport details, however I'm not certain on non-batch related transport details (https://github.com/ably/docs/compare/zapier-additions?expand=1#diff-4002e67e81d9745387d812796fb23240R74). Any details you can provide on this would be much appreciated so I can flesh it out.

@remotekieran if you can make sure you're happy with the new structuring/copy etc, that'd be great.


content/root/how-ably-works.textile
```diff
@@ -14,7 +14,7 @@ jump_to:
...
+Ably's global realtime service enables Internet enabled devices, such as a browser, phone, server or IoT sensor, to stream data in realtime between to any other Internet connected device in milliseconds. The Ably platform brings enterprise scale messaging to developers by delivering "100% service availability":http://support.ably.io/solution/articles/3000029531, "message delivery guarantees":http://support.ably.io/solution/articles/3000044640 and "low-latencies globally":http://support.ably.io/solution/articles/3000044625 (typically less than 60ms).
```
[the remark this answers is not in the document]
remotekieran: I should've caught this one when I did pricing updates recently but I didn't think to check the docs repo

content/general/events.textile
```diff
@@ -0,0 +1,374 @@
...
+  * The back off for consecutively failing requests will increase until it reaches 60s. All subsequent retries for failed requests will then be made every 60s until a request is successful
+  * The queue of events is retained for 5 minutes. If an event cannot be delivered within 5 minutes, then the events are discarded to prevent the queue from growing indefinitely
+
+h3(#envelope). Envelopes
+
+When you configure a Reactor Event rule using "single requests":#batching, you are given the option to envelope messages, which is enabled by default. In most cases, we believe an enveloped message provides more flexibility as it contains additional metadata in a portable format that can be useful such as the @clientId@ of the publisher, or the @channel@ name the message originated from.
+
+However, where performance is a primary concern, you may choose not to envelope messages and instead have only the message payload (@data@ element) published. This has the advantage of requiring one less parsing step, however decoding of the raw payload in the published message will be your responsibility.
```
SimonWoolf: > However, where performance is a primary concern, you may choose not to envelope messages and instead have only the message payload (@data@ element) published.

Performance will rarely if ever be the reason someone will turn enveloping off. Much more common will be if convenience -- they don't need anything in the envelope, or the target expects a specific format that the payload conforms to but the envelope does not

content/general/events.textile
```diff
@@ -0,0 +1,374 @@
...
+
+Note that messages published to queues are by default encoded as JSON (a text format), however you can choose to have messages encoded with "MsgPack":msgpack.org (a binary format) in your Reactor rules.
```
SimonWoolf: This paragraph is odd (why is it talking about queues in the page for reactor events?) and unnecessary (there's a section on payload encoding just below)

content/general/events.textile
```diff
@@ -0,0 +1,374 @@
...
+
+Check out examples of "enveloped":#envelope-examples and "non-enveloped":#no-envelope-examples examples down below.
+
+h3(#encoding). Payload Encoding
+
+The encoding of payloads sent is defined when setting up a Reactor Rule in the "Reactor tab of your app":https://support.ably.io/solution/articles/3000074406. You can have the message sent in JSON format, or as a "MessagePack":https://msgpack.org payload.
+
+* "JSON":https://www.json.org (JavaScript Object Notation): An efficient data-interchange format which is fairly standard and provides simple text based encoding.
+* "MessagePack":https://msgpack.org: An efficient binary serialization format that is similar to JSON, but faster and smaller.
```
SimonWoolf: Not necessary faster. But will be smaller. Worth highlighting that it's most useful for binary payloads, because if you have a json envelope with a binary payload, the payload will need to be base64-encoded.

content/general/events.textile
```diff
@@ -0,0 +1,374 @@
...
+
+However, where performance is a primary concern, you may choose not to envelope messages and instead have only the message payload (@data@ element) published. This has the advantage of requiring one less parsing step, however decoding of the raw payload in the published message will be your responsibility.
+
+Note that messages published to queues are by default encoded as JSON (a text format), however you can choose to have messages encoded with "MsgPack":msgpack.org (a binary format) in your Reactor rules.
+
+Check out examples of "enveloped":#envelope-examples and "non-enveloped":#no-envelope-examples examples down below.
+
+h3(#encoding). Payload Encoding
+
+The encoding of payloads sent is defined when setting up a Reactor Rule in the "Reactor tab of your app":https://support.ably.io/solution/articles/3000074406. You can have the message sent in JSON format, or as a "MessagePack":https://msgpack.org payload.
```
SimonWoolf: This should be clearer that it's referring to the format of the envelope, and only does anything if enveloping is turned on. (I think the option is hidden if enveloping is off)

content/general/events.textile
```diff
@@ -0,0 +1,374 @@
...
+# Trigger an event using the "Dev Console":https://support.ably.io/support/solutions/articles/3000062195 in your app dashboard which will generate a WebHook. You should then confirm that the WebHook has been received in your RequestBin
+# Check that the @X-Ably-Signature@ header in your WebHook request matches the HMAC SHA-256 you create using our "Javascript HMAC SHA-256 demo":<%= JsBins.url_for('reactor/webhook-hmac-sha-256') %>
+
+h2(#examples). Examples
+
+Given the various potential combinations of @enveloped@, @batched@ and message sources, it can be good to know what to expect given certain combinations of rules.
+
+All messages sent from a Reactor Event will contain the following headers:
+
+- host := the URL of the endpoint this message has been sent to
```
SimonWoolf: this is part of the http spec (and isn't the full url, just the host), not something we set manually, so don't think we need to document it here

content/general/events.textile
```diff
@@ -0,0 +1,374 @@
...
+- content-type := the type of the payload. This will be @text/json@ or @application/x-msgpack@, depending on what Encoding you chose in your Reactor Rule
```
SimonWoolf: - `application/json`, not `text/json`.

- that sentence is true for enveloped rules. For unenveloped, the content-type might be `application/json`, `text/plain`, or `application/octet-stream`, depending on whether the data is json, text, or binary.

content/general/events/aws-lambda.textile
```diff
@@ -0,0 +1,163 @@
...
+```
+
+The @x-ably-message-name@ header would be @publish@, and the payload would be @example@.
+
+h4(#no-envelope-example-presence). Non-enveloped presence events
+
+For @Presence@ events, there will be the additional headers:
+
+- x-ably-message-action := the action performed by the event (@update@, @enter@, @leave)
+- x-ably-message-client-id := the client ID of the connection which sent the presence event
```
SimonWoolf: messages may also have a client-id if the message was published by an identified client

content/general/events/aws-lambda.textile
```diff
@@ -0,0 +1,163 @@
...
+
+h2(#fields). Rule fields
+
+- AWS Region := The region you chose for your "AWS Lambda function":https://docs.aws.amazon.com/general/latest/gr/rande.html#lambda_region
+- Function Name := The name you gave your "AWS Lambda function":https://docs.aws.amazon.com/lambda/latest/dg/getting-started.html
+- AWS Authentication Scheme := the "authentication scheme":https://support.ably.io/support/solutions/articles/3000071908 you wish to use for your function. Either @AWS Credentials@ or @ARN of an assumable role@
+- Qualifier := _optional_ qualifier for your AWS Lambda function
+- "Source":/general/events#sources := Choose which of @Message@ or @Presence@ events on channels should activate this Reactor Event Rule.
+- Channel filter := An optional filter, which allows the Rule to be applied to a restricted set of channels. This can be specified as a regular expression, allowing for swathes of channels to be used
+- Encoding := The encoding to be used by this Rule. This can only be JSON for AWS Lambda Functions
```
SimonWoolf: > The encoding to be used by this Rule. This can only be JSON

why bother even showing & documenting the field if there's only one valid choice?

content/general/events/aws-lambda.textile
```diff
@@ -0,0 +1,163 @@
...
+- site := the Ably datacenter which sent the message
+- timestamp := a timestamp represented as milliseconds since the epoch for the presence event
+
+In addition, it will contain another field which will contain the actual message, which is named according to the message type.
+
+h4(#envelope-example-message). Enveloped message events
+
+For @message@ events, there will be a @messages@ field, which will contain the "@Message@":/rest/messages#properties event.
+
+minimize. batched @message@ payload example:
```
SimonWoolf: > batched @message@ payload example:

This is misnamed

content/general/events/aws-lambda.textile
```diff
@@ -0,0 +1,163 @@
...
+      "name": "my message name"
+    }]
+  }
+  ```
+
+h4(#envelope-example-presence). Enveloped presence events
+
+For @presence@ events, there will be a @presence@ field, which will contain the "@Presence@":/realtime/presence#presence-message event.
+
+minimize. batched @message@ payload example:
```
SimonWoolf: misnamed here too

content/general/events/aws-lambda.textile
```diff
@@ -0,0 +1,163 @@
...
+
+h2(#restrictions). Restrictions
+
+At present, it is not possible to "batch":/general/event#batching messages to AWS Lambda Functions, nor can messages be "encoded":/general/events#encoding in anything besides JSON.
+
+h2(#example). Examples
+
+h3(#envelope-examples). Enveloped event payloads
+
+Enveloped messages do not contain any additional headers beyond the "shared headers":/general/events#examples.
```
SimonWoolf: lambda function rules have no concept of headers. We invoke the function with a payload, that's it

content/general/events/aws-lambda.textile
```diff
@@ -0,0 +1,163 @@
...
+      "timestamp": 1582270137276,
+      "data": "some data in the presence object",
+      "action": 4
+    }]
+  }
+  ```
+
+h3(#no-envelope-examples). Non-enveloped event payloads
+
+Non-enveloped messages contain a few additional headers beyond the "shared headers":/general/events#examples. These are:
```
SimonWoolf: as above, lambda has no concept of headers

content/general/events.textile
```diff
@@ -0,0 +1,374 @@
...
+
+If *Single request* is chosen for a Reactor Event Rule, then a @POST@ request will be sent to your specified endpoint/service each time an event occurs. Although this can be useful for some use-cases where the endpoint can only process one message per request, or needs the event as soon as it's available, it can result in the endpoint being overloaded with requests. To avoid this, it's possible to instead make use of *Batch request* instead, which will batch messages sent within a set timeframe together.
+
+h4(#single-request). Single request details
+
+Single request is best suited for scenarios where you're wanting a 1-to-1 relationship between sent messages and events being called. If you are making use of a serverless system which is expecting a single piece of data each time, and then intends to perform some transformation/event following that, then Single request will likely work well for you. If you're using a single server, which has the potential to be overloaded by requests, and can process multiple events per payload sent, Batch request will be a better choice.
+
+minimize. View single request transport details
+  It just works
```
SimonWoolf: - rate limits: https://support.ably.io/support/solutions/articles/3000053845-do-you-have-any-connection-message-or-rate-limits-on-accounts- under 'reactor events'. 15 free, 30 paid. (Come to think of that doesn't make any sense, we really need to rethink that.)
- request timeout: 15s
- multiple in-flight at once, up to the max concurrency (listed on the limits article). If the number of in-flight requests exceeds the max concurrency, new messages coming in are placed in a short queue (length 10); if that queue length is exceeded, further messages are rejected.
- if the post request is rejected with 5xx or times out, we retry twice more, once after 4s, then if that fails, again after 20s.

content/general/events.textile
```diff
@@ -0,0 +1,390 @@
...
+
+p(tip). Reactor Events are "rate limited":#transport and are suitable for low to medium volumes of updates. If you expect a high volume of events and messages (averaging more than 25 per second), then you should consider using our "message queues":/general/queues or "firehose":/general/firehose as they are more suitable for higher volumes.
+
+Generally, realtime interactions such as subscribing to events and messages are best done using our "realtime client libraries":/realtime, or our "message queues":/general/queues or "firehose":/general/firehose. However, it can sometimes be preferable to be notified over HTTP when events occur such as a user publishing a message or members entering the presence set on a channel. For example, you may want to start pushing content to a channel only when the channel becomes active and there is someone there to receive the data, or they may want to send a welcome message to someone when they first enter a chat channel. Reactor Events are useful for achieving this in both these scenarios.
```
mattheworiordan: "Generally, realtime interactions such as subscribing to events and messages are best done using our" is slightly misleading.

"For clients subscribing to events and messages on-demand, this is often best done using our "realtime client libraries":/realtime or subscribing to Ably using any of the "realtime protocols we support":https://www.ably.io/adapters.  However, when a persistent subscription is required to push data into third party systems, the Reactor is designed for this use case and is available as Reactor Events (for HTTP requests, serverless functions, etc), Reactor Queues (data is pushed into our own hosted message queues that you can subscribe to), or Reactor Firehose (stream events into third party systems like Kafka and AWS Kinesis).  

If you want to be notified as events arise, and would like to trigger serverless functions, or invoke an HTTP request to an endpoint, then Reactor Events is the right choice.  For example, if you want to send a welcome message to someone when they become present on a chat channel, you can use Reactor Events to trigger a serverless function immediately after they enter, which in turn can publish a welcome message back to that user on the chat channel.

---

WDYT?  

You also mention things like "you may want to start pushing content to a channel only when the channel becomes active and there is someone there to receive the data", yet we don't mention channel lifecycle events  and we only support in batched mode.  Happy to include, but previously you only mentioned messages and presence as the source of events.

content/general/events.textile
```diff
@@ -0,0 +1,390 @@
...
+
+<a href="/images/diagrams/reactor-webhooks-intro.png" target="_blank">
+  <img src="/images/diagrams/reactor-webhooks-intro.png" style="width: 100%" alt="Ably Reactor Events diagram">
+</a>
+
+You can "configure events":https://support.ably.io/support/solutions/articles/3000074406 from the "Reactor tab in your app":https://support.ably.io/solution/articles/3000074406 on a per app basis which can apply to one or more channels in that app. Reactor Events can be filtered by channel naming using a regular expression, for example @^click_.*_mouse$@. This would match the string @click_@ followed by a string followed by @_mouse@, for example, @click_left_mouse@.
+
+h3(#integrations). Available integrations
+
+At present, we have explicit integration with the following services, however you can also specify *any endpoint* for the Reactor Event Rule to use:
```
mattheworiordan: Perhaps instead:

At present, in addition to support for any custom HTTP endpoint, we have ready-made integrations with the following services:

content/general/events.textile
```diff
@@ -0,0 +1,390 @@
...
+
+* "AWS Lambda Functions":/general/events/aws-lambda
+* "Azure Functions":/general/events/azure
+* "Google Cloud Functions":/general/events/google-functions
+* "IFTTT":/general/events/ifttt
+* "Cloudflare Workers":/general/events/cloudflare
+* "Zapier":/general/events/zapier
+
+Below are details pertaining to general webhook details, so if you need integration-specific details, go to their relevant pages.
```
mattheworiordan: I am not sure this line is needed.

content/general/events.textile
```diff
@@ -0,0 +1,390 @@
...
+
+h2(#configure). Configuring a webhook
+
+webhooks are configured from the Reactor tab in your "app dashboard":https://support.ably.io/support/solutions/articles/3000030053. The following fields are shared between each webhook, although certain integrations may have restrictions on them:
```
mattheworiordan: Webhook not webhook given it's the start of a sentence.

mattheworiordan: > although certain integrations may have restrictions on them:

Not sure what you mean by that?

content/general/events.textile
```diff
@@ -0,0 +1,390 @@
...
+
+- URL := The URL of your server where messages will be sent. Some integrations may instead ask for fields such as key or function name, and construct the final URL for you from that
```
mattheworiordan: But this is for webhooks?

content/general/events/azure.textile
```diff
@@ -0,0 +1,45 @@
...
+
+<br/>
+
+If the Rule is in the *Batch Request* mode, it will have the following additional options:
+
+- Sign with key := Ably will optionally sign the data with the specified private key. This will be included as an HTTP header @X-Ably-Signature@ in every HTTP post request issued to your server. See "webhook security":/general/events#security for more details.
+
+h2(#example). Examples
+
+Messages to Azure Functions will match the examples provided on the "main Reactor Events page":/general/events#examples.
```
mattheworiordan: Could we not just use partials and thus reduce the need for developers to refer back to general docs each time?

content/general/events/azure.textile
```diff
@@ -0,0 +1,45 @@
...
+  - none
+jump_to:
+  Help with:
+    - Fields#fields
+    - Example#example
+---
+
+As part of "Reactor Events":/general/events, it is possible to integrate with various systems.
+
+"Azure Functions":https://azure.microsoft.com/en-gb/services/functions provide event-driven serverless compute functions which allow users to easily run code whenever events are sent to it.
```
mattheworiordan: Should we not somewhere indicate that all the detail that sits behind how Reactor Events work, is in the Reactor Events docs.  I am concerned someone arriving here, won't realise that.

PR #824, opened 2020-02-26
Push: Fix activate() behavior while syncing device

tcard:
From https://github.com/ably/ably-java/commit/257a6c1e481c572de3d365e3c40c29a7d17632ca:

> This isn't in the spec (yet). The previous behavior causes the
activation callback to be called right away with no error, while a
previous activation is ongoing. The correct behavior is to serialize
the calls, so that the second call is processed after the first has
finished.

> At the same time, remove the CalledActivate enqueue from
NotActivated, which relied on the old behavior but wasn't doing
anything really useful. (This needs to be removed from the spec
too.)

So, this:

```js
rest.push.activate((err) => {
    console.log('First:', err);
});

// Push is now syncing the existing registration.

rest.push.activate((err) => {
    console.log('Second:', err);
});

// Syncing finishes.
```

Before,

```
Second: null  # This gets printed right when the second activate is called
First: <maybe an error>
```

After:

```
First: <maybe an error>
Second: <immediate null if first activate succeeded, otherwise retries and maybe another error>
```

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -1015,7 +1015,7 @@ h3(#activation-state-machine). Activation State Machine
 **** @(RSH3d3c)@ Either way, when the registration is done, a @RegistrationSynced@ or @SyncRegistrationFailed@ event should be fired.
 **** @(RSH3d3d)@ Transitions to @WaitingForRegistrationSync@.
 ** @(RSH3e)@ State @WaitingForRegistrationSync@:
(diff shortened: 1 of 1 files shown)


tomczoink:
@tcard  is this waiting on a change in ably-java before being merged in, or is this good to go? Same for #825.

QuintinWillison:
@tomczoink there should. be no need to wait for Java so I think @tcard may just be awaiting an approval from @paddybyers.

tcard:
@tomczoink Yes, just waiting for review. Also I've been offline for a few days. I just merged both.

PR #835, opened 2020-03-09
Documenting the delta channel param

codemerx:
Adding documentation for the `delta` channel parameter similar to what we have for the `rewind` one.

### DIFF
--- a/content/realtime/channel-params.textile
+++ b/content/realtime/channel-params.textile
@@ -142,13 +142,13 @@ h3(#rewind-examples-mqtt). Rewind example with MQTT
 
 h2(#delta). Delta parameter
 
-The @delta@ channel parameter expresses the intent to receive the delta (diff) between the previous and current message instead of the current message in full on a given channel. This parameter is oblivious to the sender and only affects the subscriber using it. I.e. the sender sends messages in full and Ably Realtime would compute and send the message deltas to any subscriber using this parameter. 
+The @delta@ channel parameter allows subscribers to expresse their desire to receive deltas (diffs) between the previous and current message instead of the current message in full on a given channel. The effects of this parameter are invisible to message senders and only affect subscribers that specify it - i.e. the sender sends messages in full and Ably computes and sends the message deltas to any subscriber which subscribed using this parameter. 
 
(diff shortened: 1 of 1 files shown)


content/realtime/channel-params.textile
```diff
@@ -139,3 +139,111 @@ h3(#rewind-examples-mqtt). Rewind example with MQTT
     ...
   });
 ```
+h2(#delta). Delta parameter
+
+The @delta@ channel parameter expresses the intent to receive the delta (diff) between the previous and current message instead of the current message in full on a given channel. This parameter is oblivious to the sender and only affects the subscriber using it. I.e. the sender sends messages in full and Ably Realtime would compute and send the message deltas to any subscriber using this parameter.
```
SimonWoolf: > Ably Realtime would compute

Just 'Ably' -- referring to ourselves "Ably Realtime" sounds quite formal outside of legal documents.

(There's places where we refer to things like "Ably Realtime client libraries", but that's because those are referring to "{Ably} {Realtime client libraries}", as opposed to our REST client libraries)

content/realtime/channel-params.textile
```diff
@@ -139,3 +139,111 @@ h3(#rewind-examples-mqtt). Rewind example with MQTT
     ...
   });
 ```
+h2(#delta). Delta parameter
+
+The @delta@ channel parameter expresses the intent to receive the delta (diff) between the previous and current message instead of the current message in full on a given channel. This parameter is oblivious to the sender and only affects the subscriber using it. I.e. the sender sends messages in full and Ably Realtime would compute and send the message deltas to any subscriber using this parameter. 
+
+It should also be noted that Ably Realtime does not guarantee every message to be a delta once the @delta@ parameter is specified. Ably Realtime might chose to send some messages in full depending on system load, etc. So a subscriber using the @delta@ parameter should be prepared to handle a random mix of delta and full messages.
```
SimonWoolf: > It should also be noted that Ably Realtime does not guarantee every message to be a delta once the `delta` parameter is specified

the infinitive ("...guarantee every message to be a delta"), while not wrong, is a relatively unusual, stilted construction (especially as most of our documentation is written with a fairly casual tone). Suggested rewording: "Note that requesting deltas does not guarantee that every message will be a delta."

content/realtime/channel-params.textile
QuintinWillison: fbf72fa

content/realtime/channel-params.textile
QuintinWillison: fbf72fa

PR #846, opened 2020-03-25
Tutorials - Changed Ably Team attribution to team members

tomczoink:
Rather than displaying tutorials made by the direct Ably team as 'Ably Team', this makes the attribution go to the team members involved.


fliptopbox:
@tomczoink I really like that idea, people aka real people are better than an institutional entity. So my only request is that we schedule this update. Right now ALL the existing tutorials have out-dated metadata that does not work with the new UI. @jamienewcomb please can we so this after the merge into master. Pulling this now will cause a HUGE conflict.

fliptopbox:
@Srushtika @tomczoink here is the current list of tutorials and the new metadata FYI https://docs.google.com/spreadsheets/d/1eEgrVcJlJQirhW2TRLP4RXK9YwCPQdyXrosxdMgJFD0/edit#gid=423364465

PR #849, opened 2020-04-02
SSE docs - added lastevent info

### LINKED ISSUE #837: Document how to renew tokens and resume with continuity for SSE
If an SSE connection requires a new token, then that's a new URL and therefore a new `EventSource`. Customers need to know how a new `EventSource` can resume with continuity from an existing connection.

We need this process to be written up, ie some explanation together with a code snippet of how to make it work. Then it probably drops quite naturally somewhere into the SSE documentation.

Srushtika:
Resolves #837

### DIFF
--- a/content/sse/index.textile
+++ b/content/sse/index.textile
@@ -53,7 +53,7 @@ Implementing transparent connection resumes when tokens need to be renewed requi
 
 h4(#detecting-token-expiry). Detecting token expiry
 
-When a connection is closed as a result of any error (that is, it's not just a dropped connection), then the @onerror@ event will occur on the EventSource instance, and the data attribute of the event will contain an Ably error body with the information about the nature of error. In the case of a token error - that is an error arising from a problem with the auth token - the code in the error body will indicate that. Token errors have a code in the range @40140 <= code < 40150@. In such cases, the authentication can be retried with a new @accessToken@.
+When a connection is closed as a result of any error (that is, it's not just a dropped connection), then the @error@ event will occur on the EventSource instance, and the data attribute of the event will contain an Ably error body with the information about the nature of error. In the case of a token error - that is an error arising from a problem with the auth token - the code in the error body will indicate that. Token errors have a code in the range @40140 <= code < 40150@. In such cases, the authentication can be retried with a new @accessToken@.
 
 In the future we plan to send an event on the connection that indicates that the token will expire imminently, which will allow a new connection to be established prior to the closure of the previous connection.
 
--- a/content/sse/index.textile
+++ b/content/sse/index.textile
@@ -46,14 +46,13 @@ The SSE protocol and the EventSource API are designed so that a dropped connecti
 
 When a token expires the connection will end. However, the default EventSource behaviour of automated reconnection will not work, because the (expired) credentials are part of the connection URL. What is needed is for a new connection to be established, with an updated @accessToken@. The question then arises as to how to do that with continuity - that is, how to establish a new connection but supply the correct @lastEventId@ so that the new connection resumes from the point that the prior connection became disconnected.
 
-
 h3(#message-continuity-token-auth). Implementing message continuity with Token Auth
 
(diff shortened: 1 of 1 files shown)


content/sse/index.textile
```diff
@@ -42,6 +42,64 @@ If using basic auth, you can use a querystring parameter of @key@ or an @Authori
...
+
+
+h3(#message-continuity-token-auth). Implementing message continuity with Token Auth
+
+Implementing transparent connection resumes when tokens need to be renewed requires a few additional steps - detecting token expiry and resuming the connection from the point of the last delivered message using the @lastEventId@ attribute:
+
+h4(#detecting-token-expiry). Detecting token expiry
+
+When a connection is closed as a result of any error (that is, it's not just a dropped connection), then the @onerror@ event will occur on the EventSource instance, and the data attribute of the event will contain an Ably error body with the information about the nature of error. In the case of a token error - that is an error arising from a problem with the auth token - the code in the error body will indicate that. Token errors have a code in the range @40140 <= code < 40150@. In such cases, the authentication can be retried with a new @accessToken@.
```
SimonWoolf, line 56: > `the @onerror@ event`

the event name is `error`; `onerror` is just a way of adding a listener that fires for an `error` event

Srushtika: You are right. Fixed, thanks

content/sse/index.textile
```diff
@@ -42,6 +42,64 @@ If using basic auth, you can use a querystring parameter of @key@ or an @Authori
 
 Connection state is only retained for two minutes. See "Connection state explained":/realtime/connection#connection-state-explained for full documentation.
 
+The SSE protocol and the EventSource API are designed so that a dropped connection is resumed transparently; the client implementation will reconnect and supply a @lastEventId@ param that ensures that the resuming connection delivers any events that have arisen since the connection was dropped. Ably uses this mechanism to reattach all channels in a new connection to the exact point that had been reached in the prior connection.
+
+When a token expires the connection will end. However, the default EventSource behaviour of automated reconnection will not work, because the (expired) credentials are part of the connection URL. What is needed is for a new connection to be established, with an updated @accessToken@. The question then arises as to how to do that with continuity - that is, how to establish a new connection but supply the correct @lastEventId@ so that the new connection resumes from the point that the prior connection became disconnected.
+
```
paddybyers, line 48: why two lines of whitespace?

content/sse/index.textile
```diff
@@ -42,6 +42,64 @@ If using basic auth, you can use a querystring parameter of @key@ or an @Authori
...
+
+h3(#message-continuity-token-auth). Implementing message continuity with Token Auth
+
+Implementing transparent connection resumes when tokens need to be renewed requires a few additional steps - detecting token expiry and resuming the connection from the point of the last delivered message using the @lastEventId@ attribute:
+
+h4(#detecting-token-expiry). Detecting token expiry
+
+When a connection is closed as a result of any error (that is, it's not just a dropped connection), then the @error@ event will occur on the EventSource instance, and the data attribute of the event will contain an Ably error body with the information about the nature of error. In the case of a token error - that is an error arising from a problem with the auth token - the code in the error body will indicate that. Token errors have a code in the range @40140 <= code < 40150@. In such cases, the authentication can be retried with a new @accessToken@.
```
paddybyers, line 56: EventSource -> @EventSource@

content/sse/index.textile
```diff
@@ -42,6 +42,64 @@ If using basic auth, you can use a querystring parameter of @key@ or an @Authori
...
+    } else {
+      // ... handle other types of error -- for example, retry on 5xxxx, close on 4xxxx
+    }
+  }
+}
+
+connectToAbly();
+```
+
+An important thing to note here is that the EventSource API tries to auto-reconnect and re-subscribe to the SSE endpoint when any error occurs, even the token expiry error like in this case. This means that upon manually re-subscribing to the SSE endpoint with a new token, there will be two active subscriptions to the endpoint - one with the old token which would continue to throw an error due to expired credentials and another with the new token. Hence, it is important to close the previous EventSource subscription using @eventSource.close()@ before re-subscribing with the new token as shown in the snippet above.
```
paddybyers, line 99: @EventSource@

PR #861, opened 2020-04-09
Client libs spec: Optional Rest.stats.start/end.

tcard:
They have defaults in the API, so it shouldn't be mandatory to provide them in the libraries.

`end` was already marked as `api-default` which implicitly makes it optional, but only in the IDL.

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -69,7 +69,7 @@ h3(#restclient). RestClient
 * @(RSC6)@ @RestClient#stats@ function:
 ** @(RSC6a)@ Returns a @PaginatedResult@ page containing @Stats@ objects in the @PaginatedResult#items@ attribute returned from the stats request
 ** @(RSC6b)@ Supports the following params:
-*** @(RSC6b1)@ @start@ and @end@ are timestamp fields represented as milliseconds since epoch, or where suitable to the language, Date or Time objects. @start@ must be equal to or less than @end@ and is unaffected by the request direction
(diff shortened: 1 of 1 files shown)


tcard:
Maybe the older versions should be retroactively fixed too?

QuintinWillison:
In terms of retroactive changes to older versions, that's something I would not know the right approach for. Pinging @tomczoink

content/client-lib-development-guide/features.textile
```diff
@@ -69,7 +69,7 @@ h3(#restclient). RestClient
 * @(RSC6)@ @RestClient#stats@ function:
 ** @(RSC6a)@ Returns a @PaginatedResult@ page containing @Stats@ objects in the @PaginatedResult#items@ attribute returned from the stats request
 ** @(RSC6b)@ Supports the following params:
-*** @(RSC6b1)@ @start@ and @end@ are timestamp fields represented as milliseconds since epoch, or where suitable to the language, Date or Time objects. @start@ must be equal to or less than @end@ and is unaffected by the request direction
+*** @(RSC6b1)@ @start@ and @end@ are optional timestamp fields represented as milliseconds since epoch, or where suitable to the language, Date or Time objects. @start@, if provided, must be equal to or less than @end@ if provided or to the current time otherwise, and is unaffected by the request direction
```
QuintinWillison, line 72: The use of "if provided" suggests they are optional, thus I would expect the types to be `Time?`. Does this need a bit more clarity perhaps?

tcard: For what I've seen, `api-default` implicitly makes the type optional. We can make that explicit, but then we should fix the other occurrences.

mattheworiordan: We use `Time` types when they exist and are idiomatic in the language.  Javascript is one of those languages that idiomatically uses Ints to represent time.  Most other languages have common time type attributes, which makes it far easier for developers to consume our API.

PR #856, opened 2020-04-03
Version 1.2

QuintinWillison:
I'm creating this PR in Draft state so as to try to ensure it doesn't get merged too easily by mistake as this **must not** land on `master` until we're ready for 1.2 to go live on the main website.

See [this internal issue](https://github.com/ably/wiki/issues/419) for discussion around potential changes to implementation or branching strategy going forward.

PR #871, opened 2020-04-16
Documentation version 1.2 changes

tomczoink:
This PR updates our documentation to v1.2, in addition to adding v1.1 versions. This should be merged once v1.2 has been fully released. This is intended to replace #856 in combination with #870, per discussion of seperating doc version changes to spec version changes.


QuintinWillison:
@tomczoink - beyond where I got to with #856, I don't have anything further informed to contribute in respect of these changes. I think the only other person I'm aware of with enough domain knowledge of how docs is built and/or setup to interact with the website downstream is @mattheworiordan - therefore I think only he would be qualified to 'approve' this.

I would imagine both @Ugbot and @kennethkalmer would have an interest too, though equally they will be still accumulating context themselves in this area in order to have an opinion.

I'll remove myself as a reviewer.

PR #872, opened 2020-04-17
change support domain from ably.io to ably.com

dpflucas:
Change any references from support.ably.io to support.ably.com


MarkWoulfeAbly:
@dpflucas Looks like we're still on ably.io. Any idea when this change will be implemented?

kennethkalmer:
I wonder if it would be possible to make the URL configurable in place like we did in the website repo

kennethkalmer:
@dpflucas @MarkWoulfeAbly @tomczoink I rebased the work and updated several more files. If this looks good you folks can pull the trigger on this

kennethkalmer:
⚠️ Urgh, ignore my "pull the trigger" comment ⚠️

Issue #874, opened 2020-04-23
Client SDKs should gracefully handle non-JSON/MsgPack responses


mattheworiordan:
If a client SDK receives an error response from Ably that happens to be in an unexpected format such as HTML or text, it should handle that gracefully.

See https://github.com/ably/ably-go/issues/154 an issue that recently came up.

Whilst the SDKs should handle this anyway, I think it would be sensible to make this explicit so that there is test coverage over it.

Related tickets in proxy layers https://ably.atlassian.net/browse/INF-153, https://ably.atlassian.net/browse/INF-167

QuintinWillison:
Historical breadcrumbs re [ably-go](https://github.com/ably/ably-go):

- fixed (April 2020): https://github.com/ably/ably-go/issues/154
- related (January 2020): https://github.com/ably/ably-go/issues/127

mattheworiordan:
@QuintinWillison can you propose a spec item for this to reflect what was done in Go?

QuintinWillison:
Sure. Internal story [SDK-13](https://ably.atlassian.net/browse/SDK-13) created so it can be managed in the backlog.

PR #888, opened 2020-06-01
Client library spec: state requirements for handling of response bodies having unexpected content type

paddybyers:
Fixes: https://github.com/ably/docs/issues/874


content/client-lib-development-guide/features.textile
```diff
@@ -84,6 +84,11 @@ h3(#restclient). RestClient
 * @(RSC8)@ Supports two protocols:
 ** @(RSC8a)@ "MessagePack":http://msgpack.org/ binary protocol (this is the default for environments having a suitable level or support for binary data)
 ** @(RSC8b)@ JSON text protocol (used when @useBinaryProtocol@ option is false)
+** @(RSC8c)@ The library sets the @Accept@ and @Content-Type@ type headers, to reflect whether the client is configured to use a binary or JSON based protocol, to @application/x-msgpack@ or @application/json@ respectively.
```
QuintinWillison, line 87: Unnecessary "type" plus, despite it being obvious to most  I would argue that it would help to insert the word "request" before headers to make it clear we're talking about the outbound communication from the client library to the server.

```suggestion
** @(RSC8c)@ The library sets the @Accept@ and @Content-Type@ request headers, to reflect whether the client is configured to use a binary or JSON based protocol, to @application/x-msgpack@ or @application/json@ respectively.
```

content/client-lib-development-guide/features.textile
```diff
@@ -84,6 +84,11 @@ h3(#restclient). RestClient
...
+** @(RSC8d)@ Irrespective of the specified @Accept@ header, the library should attempt where possible to process any response body based on the content type indicated in a @Content-Type@ header.
```
QuintinWillison, line 88: Again, I think this could be more specific...

```suggestion
** @(RSC8d)@ Irrespective of the @Accept@ header set by the library (see "RSC8c":#RSC8c), it should attempt where possible to process any response body based on the received @Content-Type@ header.
```

However, now that I even read that back to myself, I'm failing to find something specific that I could implement in code around this. What does this mean in practice? Are we saying that the client library needs to have multiple parsers handling different possibly received content types? Or is it meant to mangle say `text/html` to `application/json` because that's what it specified in `Accept`? **I think this is too ambiguous.**

content/client-lib-development-guide/features.textile
```diff
@@ -84,6 +84,11 @@ h3(#restclient). RestClient
...
+** @(RSC8e)@ If the library is unable to process the specified content type, then the library processes the response as follows.
+*** @(RSC8e1)@ In the case of a @statusCode@ indicating an error response, the error must be propagated back to the caller with as much information as it is possible to obtain from the response, including the @statusCode@ and, where applicable, an error message containing text from the response body to help to diagnose the cause of the error.
```
QuintinWillison, line 90: This is a little vague, isn't it? What code should I write to make my client library compliant? Is there a limit? What if I receive some Textile from the server, for example? Am I meant to have a parser for that? **Needs scope.**

content/client-lib-development-guide/features.textile
```diff
@@ -84,6 +84,11 @@ h3(#restclient). RestClient
...
+*** @(RSC8e2)@ In the case of a @statusCode@ indicating a non-error response, an error must be propagated back to the caller with a @statusCode@ of 400, a @code@ of @40013@ and an error message containing the original @statusCode@ and, where applicable, text from the response body, to help to diagnose the cause of the error.
```
QuintinWillison, line 91: I don't understand this. Where is "status code indicating a non-error response" defined and under what circumstances would this happen? Is this a 200 OK response and in which case are we just describing an inability at the client-side to parse the response body? I'm really confused by this one.

content/client-lib-development-guide/features.textile
paddybyers: If the client says it `Accept`s JSON, but the server responds with msgpack, it should parse it, and vice versa, if the library already contains support for the alternate format. It should not add support for parsing anything (eg html) just in order to be able to process erroneously-formatted responses.

MarkWoulfeAbly:
@paddybyers FYI there is some feedback to address here.

paddybyers:
@QuintinWillison I've made some changes: please see https://github.com/ably/docs/pull/888/commits/bfcb292156fecf01102086c64cd951f7b018c8b5

content/client-lib-development-guide/features.textile
QuintinWillison: fixed https://github.com/ably/docs/pull/888/commits/bfcb292156fecf01102086c64cd951f7b018c8b5

content/client-lib-development-guide/features.textile
QuintinWillison: improved https://github.com/ably/docs/pull/888/commits/bfcb292156fecf01102086c64cd951f7b018c8b5

content/client-lib-development-guide/features.textile
QuintinWillison: refined, tighter spec. https://github.com/ably/docs/pull/888/commits/bfcb292156fecf01102086c64cd951f7b018c8b5 

@paddybyers, presumably the `base64-encoding` has been specified in case the response body is not textual - in which case we expect to see base64 in the debug messages and we then decode that as part of our debugging workflow?

content/client-lib-development-guide/features.textile
QuintinWillison: refined https://github.com/ably/docs/pull/888/commits/bfcb292156fecf01102086c64cd951f7b018c8b5

though [my earlier question](https://github.com/ably/docs/pull/888#discussion_r441447075) also applies here.

QuintinWillison:
@paddybyers just [a question](https://github.com/ably/docs/pull/888#discussion_r441447075) remains to ensure that I've inferred the right reasons behind your choice to base64 error/debug message content.

content/client-lib-development-guide/features.textile
paddybyers: I've suggested base64 unconditionally, yes, because otherwise you'd be having to build knowledge into the library about which content types are binary and which aren't. Nothing is going to inspect these encoded messages other than humans when investigating issues.

MarkWoulfeAbly:
@paddybyers happy for me to land?

PR #890, opened 2020-06-03
Update Docker to use docker-compose & update README

mattheworiordan:
Docker compose is far superior in that it a single command can be used to set everything up in one step.

@MarkWoulfeAbly I saw your doc https://docs.google.com/document/d/1X7hX_A2bQmAAHlLhgGRxhKf8SkDASZX8Quowi0hB8rQ/edit this evening, and realised you've probably spent a load of time just trying to get things running locally, which must be frustrating.

In future, if you feel you are blocked on things, please just ask me, and I'll make sure I unblock you or get someone else to help.  Whilst I could fix this in around 2 hours as I am very familiar with this env and Docker, I can certainly empathise why this was very hard for you to figure out.  I am here if you need me in future if you feel you are spending time on the wrong things.

I hope this works on Windows.  Can you confirm that when you edit a file, the changes are reflected in the local web server running on http://localhost:4000/?  Nanoc is a bit slow, but you should see something in the terminal immediately after you update a file, and then a notice later that states it has finished compiling.  Example...

```
webserver_1  | /usr/lib/ruby/gems/2.7.0/gems/nanoc-core-4.11.15/lib/nanoc/core/compiler_loader.rb:35: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
webserver_1  | /usr/lib/ruby/gems/2.7.0/gems/nanoc-core-4.11.15/lib/nanoc/core/compiler.rb:8: warning: The called method `initialize' is defined here
webserver_1  | /usr/lib/ruby/gems/2.7.0/gems/nanoc-core-4.11.15/lib/nanoc/core/dependency_store.rb:116: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
webserver_1  | /usr/lib/ruby/gems/2.7.0/gems/nanoc-core-4.11.15/lib/nanoc/core/dependency_props.rb:16: warning: The called method `initialize' is defined here
webserver_1  |
webserver_1  | 01:41:11 - INFO - Compilation succeeded.
```


JackUkleja:
Nice. Will try this out later. 

p.s. Is it easy enough to add a smoke test for this CI pipeline into TravisCI?

JackUkleja:
docker-compose.yml seems to achieve the build and compile steps, but we seem to have lost the ability to run the local server (`bundle exec nanoc view`) and `guard` to monitor for changes?

README.md
```diff
@@ -59,24 +59,21 @@ The code that is published to JSBin can be found in
...
+### Running with Docker and Docker compose
 
-If you would prefer to use Docker, then you can set up the env as follows:
+If you would prefer to use Docker, then you can run the web server as follows:
 
-* `./docker-run build` to build the initial Docker image
-* `./docker-run compile` to compile the HTML for the static site (this is a once off command)
-* `./docker-run server` to start up a web server at <http://localhost:4000>
-* `./docker-run guard` to watch for changes and rebuild the site automatically. This is normally run in conjunction with the `./docker-run server` command so that you can see changes immediately
+* `docker compose up`
```
MarkWoulfeAbly, line 74:
```suggestion
* `docker-compose up`
```

MarkWoulfeAbly:
FYI if you cancel and re-run compose the change does appear but I don't imagine that's the ideal workflow here. Also the updated text I added seems to be ignoring CSS, not sure whats going on there.

image

README.md
```diff
@@ -71,7 +71,7 @@ _For those who use [ASDF](https://github.com/asdf-vm/asdf) or compatible tooling
 
 If you would prefer to use Docker, then you can run the web server as follows:
 
-* `docker compose up`
+* `docker-compose up`
```
mattheworiordan, line 74: Oops, thanks 👍

mattheworiordan:
> The fact that guard is watching the image is probably the issue? The old system would detect changes in the repo folder itself an update.

Yup, Nanoc has moved on and Guard is now part of it, so see https://github.com/guard/guard-nanoc, and then https://github.com/nanoc/nanoc/tree/master/guard-nanoc.

So this is the new way...

However, I was wondering if Windows would send signals into Docker containers when files change.  Sadly it seems not.   I will have to do a screenshare with one of you so we can get to the bottom of this :(

> Also the updated text I added seems to be ignoring CSS, not sure whats going on there.

The first line is always bigger than the rest, see all doc pages, that is used generally as the "intro" text.

mattheworiordan:
@JackUkleja @MarkWoulfeAbly I am afraid I am stumped because I cannot get Docker to run on my Windows VM.

See https://github.com/docker/for-win/issues/3597, which I cannot work around.

screenshot_2020-06-03_11-22-05_pm

So I am afraid you Windows folks are going to have to figure out how you can document doing this via Windows Linux please.  Could you do that so at least we know that people on Windows can get up and running quickly and work with this?

MarkWoulfeAbly:
sure no worries @mattheworiordan, this is way better than what we had previously so im going to land it regardless. Me and @JackUkleja can go into more detail on the other side of things and then re-write the readme to match.

PR #897, opened 2020-06-08
[WEB-409] Add / update mentions of "deltas" on Ably website

kavalerov:
Updated mentions of deltas in the comparisons with competitors.


kavalerov:
I couldn't check this locally as can't find comparison pages when I run docs locally - what is the best way to check this? 
@kennethkalmer / @fliptopbox maybe you can help?

kavalerov:
Checked the copy and fixed typos.

PR #904, opened 2020-06-11
Corrected details in rewind documentation on persistence

### LINKED ISSUE #903: Channel Rewind: time-based rewind info is incorrect
We currently state in our Rewind docs:

> By default, a maximum of two minutes of channel history is available when attaching. This means that a rewind time specifier of greater than two minutes will only be able to rewind by two minutes. If a channel has persistence enabled, then it is possible to rewind back in time by up to the persistence TTL on the channel.

This is incorrect. We should update the docs to make it clear that Rewind when used with time-based params like `rewind=3m` will not work as time-based rewinds are only live history (2 minutes). It's only numerical rewinds that can reach into persisted as of now.

For more see this: https://github.com/ably/docs/pull/729#issuecomment-545498109

tomczoink:
fixes #903 . Let me know if you're happy with language.

### DIFF
--- a/content/realtime/channels/channel-parameters/rewind.textile
+++ b/content/realtime/channels/channel-parameters/rewind.textile
@@ -124,7 +124,9 @@ h2(#additional-info). Additional information
 
 At most 100 messages will be sent in a rewind request. If the number of messages within the specified interval is greater than that limit, then only the most recent messages up to that limit are sent. The attachment succeeds, but truncation of the message backlog is indicated as a non-fatal error in the attachment response.
 
-By default, a maximum of two minutes of channel history is available when attaching. This means that a rewind time specifier of greater than two minutes will only be able to rewind by two minutes. If a channel has persistence enabled, then it is possible to rewind back in time by up to the persistence TTL on the channel.
+Rewind by time (@rewind=30s@) can only get messages at most 2 minutes old. So, if you specify @rewind=3m@, you will only get the last 2 minutes of history upon attaching.
+
+Rewind by a certain number of messages (@rewind=20@) is restricted by the channel's persistence period. If you've not enabled "persisted history":/realtime/history#persisted-history, this will be 2 minutes. If you have persistence enabled, this'll be 24 hours for free accounts, and 72 hours for paid accounts.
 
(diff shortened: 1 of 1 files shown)


content/realtime/channels/channel-parameters/rewind.textile
```diff
@@ -124,7 +124,9 @@ h2(#additional-info). Additional information
 
 At most 100 messages will be sent in a rewind request. If the number of messages within the specified interval is greater than that limit, then only the most recent messages up to that limit are sent. The attachment succeeds, but truncation of the message backlog is indicated as a non-fatal error in the attachment response.
 
-By default, a maximum of two minutes of channel history is available when attaching. This means that a rewind time specifier of greater than two minutes will only be able to rewind by two minutes. If a channel has persistence enabled, then it is possible to rewind back in time by up to the persistence TTL on the channel.
+Rewind by time (@rewind=30s@) can only get messages at most 2 minutes old. So, if you specify @rewind=3m@, you will only get the last 2 minutes of history upon attaching.
+
+Rewind by a certain number of messages (@rewind=20@) is restricted by the channel's persistence period. If you've not enabled "persisted history":/realtime/history#persisted-history, this will be 2 minutes. If you have persistence enabled, this'll be 24 hours for free accounts, and 72 hours for paid accounts.
```
remotekieran, line 129: Can we make the disparity between the two clearer? i.e. it's not possible to reach into persisted history using time params.

For what it's worth, this is what I have on the blog currently:

> The number of messages to retrieve. For example rewind=10 would provide the last ten messages published on a channel. The limit is 100. If the request goes back far enough it can reach into persisted history (provided you have it enabled).

> The amount of time in the past. For example rewind=10s would provide all messages sent within the last ten seconds. And rewind=2m would provide all messages sent within the last two minutes. The max can rewind is up to two minutes. For now, it's not possible to reach into persisted history using time-based parameters. If you'd like the ability to do this, please get in touch.

mattheworiordan: > Rewind by time (@rewind=30s@) can only get messages at most 2 minutes old

When you say rewind 30s and then state at most 2 minutes old, I feel it's a bit confusing.  I think something like this is better:

"Rewind by time by using a channel parameter such as @rewind=30s@ will retrieve messages from the last 30 seconds. However, the maximum duration permitted for rewind by time is 2 minutes.  As such, the channel parameter @rewind=3m@, upon attaching, would only retrieve messages from the last 2 minutes"

Then in regards to the next paragraph, I have suggested a few changes below to reflect the fact that history can have a variable TT:.

```
Rewind by a quantity of messages such as the channel parameter @rewind=20@, will retrieve the last 20 messages published on that channel.  However, this is limited by the channel's persistence configuration. If you've not enabled "persisted history":/realtime/history#persisted-history on the channel, then messages will only be retained for 2 minutes. If you have persistence enabled, messages are available for up to 24 hours for free accounts, and 72 hours (by default) for paid accounts.
```

Issue #865, opened 2020-04-14
New version of reactor events docs seem to have dropped the decoding section?


SimonWoolf:
Didn't notice this at the time, but the recent restructuring of reactor events docs seems to have resulted in the decoding sections - https://www.ably.io/documentation/general/versions/v1.0/webhooks#decoding-messages and https://www.ably.io/documentation/general/versions/v1.0/webhooks#decoding-presence , talking about `fromEncodedArray` etc. - being dropped. @tomczoink any particular reason? Or is it now somewhere else and I just can't find it?

MarkWoulfeAbly:
@SimonWoolf I believe the reason that these sections were removed is that in the current event page links to the realtime pages on messages and presence. At least it should, the links are currently broken but I am fixing them in the PR addressing this.

My question is, for decoding messages and presence, are these only related to reactor events or are they applicable elsewhere? Ideally, I would want to add the lost information to the pages that we link to (https://www.ably.io/documentation/realtime/messages and https://www.ably.io/documentation/realtime/presence) however if the information is only relevant to reactor events then maybe I should just update the page itself.

SimonWoolf:
> My question is, for decoding messages and presence, are these only related to reactor events or are they applicable elsewhere? 

The point is that with reactor events you are receiving the events 'out-of-band', not through an ably client library. As the section says, "Messages sent over the realtime service are automatically decoded into Messages object by the Ably client library. With webhooks you need to to do this explicitly, using `Message.fromEncodedArray`".

MarkWoulfeAbly:
I see, so it sounds like the only relevant parts to just webhooks are the decoding sections, so we can add those back in, but not the messages or presence sections as a whole.

MarkWoulfeAbly:
Having reviewed with fresh eyes I felt that the sections should be re-added as-is due to webhook specific information being absent from the pages it linked to. I have kept the links intact for more general information. @SimonWoolf if you could take a look at that PR and make sure nothing is incorrect (due to this information being copied from v1.0) that would be great thanks.

Issue #893, opened 2020-06-05
Migration help article needed for 0.8 -> 1.0 webhook format


paddybyers:
A customer wants to migrate existing webhooks to the new format.

What we've previously said:

> Apologies, it looks like our docs are inadequate on this right now. The information is there, but currently very hard to extract -- you can compare our 0.8 webhook docs https://www.ably.io/documentation/general/versions/v0.8/webhooks#presence with our 1.0 docs https://www.ably.io/documentation/general/events#batch-example-presence , but since our webhook docs were recently restructured, there is no longer any easy way to access the old versions (and as you say, no diff or migration guide). I will raise an issue on this with the docs team.
> 
> In any case: there is precisely one difference between 0.8 and 1.0/1.1 reactor presence events, and that is that for 0.8 rules the presence action was a string (e.g. "action":"enter"); for 1.0+ rules it is an integer (e.g. "action":2).
> 
> The reason for the change is that in 1.0, we moved from a model where the presence event was intended to be 'self-describing' to a model where you pass the event (encoded per our normal wire protocol) to a client library method, PresenceMessage.fromEncoded, to decode it. This then not only decode the presence action to a string, but also gives you a typed PresenceMessage in statically typed languages, decodes any data into the appropriate native data type using the encoding, decrypt the data if it was encrypted using Ably's end-to-end encryption (and the correct key is passed in), etc. 
> 
> https://www.ably.io/documentation/general/versions/v1.0/webhooks#decoding-presence has more information.
> 
> There is no exposed ui setting to make a new rule use the older rule version, it is strongly deprecated. I can change it manually for you if you really need me to though. (edited)

PR #917, opened 2020-06-29
Consolidate functions and webhooks documentation between versions

MarkWoulfeAbly:
This PR addresses #893, that is, aimed at resolving the disconnect between the reactor integration rules documentation before and after V1.0


MarkWoulfeAbly:
Hey @SimonWoolf do you think it is worth also adding a note on the cloudflare, ifttt and zapier pages that they also work with v1.0? Is there any difference when using them with v1.1 and v1.0?

SimonWoolf:
> Hey @SimonWoolf do you think it is worth also adding a note on the cloudflare, ifttt and zapier pages that they also work with v1.0? Is there any difference when using them with v1.1 and v1.0?

The version makes literally no difference to any rule types ever, with one single exception, which is the presence action for batched webhook rules in 0.8 vs higher.

Honestly, I think what we do now — that is, when we restructured and rewrote the rule docs, having that apply only to v1.1/1.2 and keeping the old structure for v0.8/v1.0 — is horribly confusing.

Due to the different page structures a customer can't change the version in the dropdown to manually compare them, so we're reduced to this kludge of editing the body copy to mention the older or newer versions. But even with that, what are we expecting them to do, read the two differently-structured different sets of rule documentations to compare them and get a useful answer about what changed? That just doesn't work IMO.

My suggestion would be: just get rid of the version dropdown and older versions for rule docs. A version dropdown makes sense for client library API docs, but I don't think it does for rules. Instead just have a changelog, or version table or something. (With all of one single item big).

(Also restore the crucial information about how to decode messages received over rules, which got stripped out during the restructure and not been reinstated yet - https://github.com/ably/docs/issues/865 )

MarkWoulfeAbly:
So i dont know why force pushing a reset causes this to close but ill be opening a separate PR with the new structure.

MarkWoulfeAbly:
@SimonWoolf I have reverted the previous changes and copied across the events and function pages from v1.1 to v1.0.

Before working on a similar commit for v0.8 I just want to clarify, beyond the presence action for batched webhook rules, which i can update the copy for, are all the other pages good to just copy over as is?

SimonWoolf:
> beyond the presence action for batched webhook rules, which i can update the copy for, are all the other pages good to just copy over as is?

yes. (I guess that means you decided to keep the dropdown?)

(relatedly, might be a good time to fix https://github.com/ably/docs/issues/865 while you're doing webhook docs?)

mattheworiordan:
@MarkWoulfeAbly please can you refer to https://ably.atlassian.net/wiki/spaces/ENG/pages/312377505/Git-PR-best-practice as this PR has no comment or description.  Everyone needs this to have context of what they're reviewing.

Can you please explain what you're trying to achieve with this PR?  I am not clear from the single commit.

MarkWoulfeAbly:
> yes. (I guess that means you decided to keep the dropdown?)
> 
> (relatedly, might be a good time to fix #865 while you're doing webhook docs?)

@SimonWoolf Yes I will be keeping the API dropdown for now to keep consistency with the rest of our documentation. I will also be addressing #865 but will do that in a separate PR after this one lands.

@mattheworiordan It seems that the Issue this PR addresses has been unlinked for some reason, I will update the description.

MarkWoulfeAbly:
So to recap the changes:

1. WebHooks (v0.8) and Functions (v0.8,1.0) have been removed
2. Events(v0.8) examples have been updated to show presence.action as a string and not an enum.
3.

Issue #437, opened 2018-05-25
Spec suggestion: only allow channels.release() for channels in an inactive( initialized, detached, or failed) state


SimonWoolf:
Since release()ing a channel in an active state results in undefined behaviour.

Context: https://stackoverflow.com/questions/50533925/connectionmanager-onchannelmessage-received-message-with-different-connections

SimonWoolf:
This came up again with [a customer](https://app.intercom.com/a/apps/ua39m1ld/inbox/inbox/conversation/27475913094) (cf https://github.com/rmosolgo/graphql-ruby/pull/2991) who was calling release() without waiting for a detach to finish. This change would be very easy and would close that trap. wdyt @QuintinWillison @paddybyers ?

(we should probably also document the release() method better than we do now, currently it's barely documented at all, cc @MarkWoulfeAbly )

MarkWoulfeAbly:
If we plan on changing how the method works then I would wait until we do that before updating the documentation, unless it's something that would take a while to change.

QuintinWillison:
@MarkWoulfeAbly this one's tagged with the "client lib spec" label which means it will modify just the [Client Library Features spec](https://docs.ably.io/client-lib-development-guide/features/). The process is generally that we try to update the spec first before we then roll out the change across the client libraries (the long tail, as it were).

@SimonWoolf what do we mean by 'only allow'? Would you propose failing early with a synchronous error or would this be an effective NOP with perhaps just a warning emitted to the log?

SimonWoolf:
> what do we mean by 'only allow'? Would you propose failing early with a synchronous error or would this be an effective NOP with perhaps just a warning emitted to the log?

Throw an exception

> If we plan on changing how the method works then I would wait until we do that before updating the documentation, unless it's something that would take a while to change.

To be clear nothing would actually change in what you would be allowed to do; I'm just proposing we throw instead of silently entering an invalid, broken state if you do something you're already not allowed to do. FWIW I think that's a pretty uncontroversial change.

Whether we document the method in our api docs is kindof independent of whether we make that change. We could do either without the other if we want. (I think we should do both).

Argument against documenting it: since it's quite a niche method (important for 1% of usecases, but irrelevant for 99%), its presence clutters up the documented API and might confuse people who would never have needed it anyway.

Argument in favour of documenting it: people (like the customer above) discover it via source code or autocomplete and try and use it whether it's documented or not. Also documenting it helps the 1% of people who _do_ need it.

PR #921, opened 2020-07-08
Only allow channels.release() for channels in an inactive( initialized, detached, or failed) state

SimonWoolf:
Fixes https://github.com/ably/docs/issues/437

Example implementation in https://github.com/ably/ably-js/pull/672


SimonWoolf:
Any objections?

paddybyers (approved):
LGTM, thanks

SimonWoolf:
https://docs.ably.io/client-lib-development-guide/features/#RTS4a , in the realtime-only part of the spec which I didn't notice, already existed, and was a fix for the same issue, in a different and incompatible way.

:/

PR #924, opened 2020-07-21
Spec: make calling attach() on an instantiated library (ie with autoconnect false) raise an error, not wait

SimonWoolf:
See discussion https://ably-real-time.slack.com/archives/CURL4U2FP/p1595339878007300


paddybyers (approved):
LGTM, thanks

PR #926, opened 2020-07-24
Client lib spec: Define "upgrade" unambiguously

tcard:
It was used for different processes in an informal way. In an instance (RTN19), it was even misleading as the reader may think that "upgrade" includes "reauthorization" when it doesn't.


content/client-lib-development-guide/features.textile
```diff
@@ -354,6 +354,7 @@ h3(#realtimeclient). RealtimeClient
 h3(#realtime-connection). Connection
 
 * @(RTN1)@ @Connection@ connects to the Ably service using a "websocket":/concepts/websockets connection. The "ably-js library":https://github.ably.io/ably/ably-js supports additional transports such as Comet and XHR streaming; however non-browser client libraries typically use only a websocket transport
+  * @(RTN1a)@ A connection in ably-js may upgrade its transport. This is left unspecified but may trigger side effects that are specified.
```
SimonWoolf, line 357: the indented bullet is done with two asterisks, no space indent

But don't bother with this, having spec items specific to ably-js is confusing. (and we'll likely be removing upgrading from ably-js at some point anyway). I'd just remove the reference to 'upgrading' from the other two spec items

tcard: Done at 10405d91.

SimonWoolf (approved):

PR #928, opened 2020-07-30
[DOC-7] Add redirect support 

dpflucas:
When pages are renamed and requires a redirect from their old location, the redirects can be configured via the frontmatter of the page. For example, if we renamed the fictional `/dsx/index.textile` file to `/hub/index.textile`, we can add the following frontmatter to handle the redirect:

~~~yaml
# other frontmatter
redirect_from: "/dsx"
---

~~~

When running `nanoc compile`, two things happen:

* `output/dsx/index.html` is created which contains the smallest amount of HTML to perform a "meta redirect" to `/hub`, and
* `data/redirects.yaml` gets updated with the redirect information.

Changes to `data/redirects.yaml` must be tracked in git, this is used by the website to perform redirects.

**IMPORTANT**: After this merge you need to run `docker-compose up --build` to get the dependencies sorted out.

Redirect support is provided by the [nanoc-redirector](https://github.com/gjtorikian/nanoc-redirector) gem


MarkWoulfeAbly:
Just making this a draft for now.

Rules
```diff
@@ -12,6 +12,14 @@
 #   “content/about.html”). To select all children, grandchildren, … of an
 #   item, use the pattern “/about/**/*/”.
 
+require 'nanoc-redirector'
+
+postprocess do
+  @items.each do |item|
+    NanocRedirector::RedirectFrom.process(item, item.identifier.without_ext, output_dir: 'content')
```
kennethkalmer, line 19: The HTML files should not be in the `content` directory, they should be in the `output` directory with the other files.

content/api-streamer/consumer-testing/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
```
kennethkalmer, line 1: 🗑️

lib/redirects.rb
```diff
@@ -0,0 +1,21 @@
+class Redirects
```
kennethkalmer, line 1: Can we call this `AblyDocs::Redirects` to prevent potential conflict with the website?

mattheworiordan:
I am not too comfortable with this PR for a few reasons:

* The PR does not explain why this is being done, only what it does.  Every PR needs a description with the why.

* It won't work AFAICT, this is served up as a static site.  See the buildpacks used (documented in README)

```
$ heroku buildpacks
=== ably-docs Buildpack URLs
1. https://github.com/heroku/heroku-buildpack-ruby
2. https://github.com/ably-forks/heroku-buildpack-nanoc
3. https://github.com/heroku/heroku-buildpack-static
```

* How will we rationalise redirects in the website vs the docs repo?  Which will take precedence? 

* This model will break if we move to any other static site builder like Gatsby as that too generates static content.

MarkWoulfeAbly:
Just removing myself as a reviewer as I'm not confident in my ability to be able to approve or deny the work here. @kennethkalmer can you address the comments left by matt above?

kennethkalmer:
Just to cover the conversation I had with @mattheworiordan about the redirects for docs, it is different to our global website redirects because the docs site lives this dual life: _edge_ & _stable_. 

The way we've setup the redirects in this PR caters for _edge_ where the redirects are handled by near-empty HTML files that contain a meta redirect, and for _stable_ where the redirect data is stored in a YAML file (akin to how the jsbin data is stored) and the website reads that YAML data to handle the redirects.

This can be merged, but I first want to spend time with @MarkWoulfeAbly and show him how it all hands together before we do.

kennethkalmer:
I updated the PR description to be more complete and include an example.

Additionally, with other static site generators people have contributed similar tools to help, like [nsresulta/gatsby-plugin-meta-redirect](https://github.com/nsresulta/gatsby-plugin-meta-redirect), [Hugo Aliases](https://gohugo.io/content-management/urls/#how-hugo-aliases-work), and [Middleman redirect](https://middlemanapp.com/basics/redirects/) to name a few. Our challenge is just how to extract this information into an intermediary format for the website to resolve.

kennethkalmer:
@MarkWoulfeAbly I removed the test data and I updated the `docker-compose up --build` instructions we discovered to the README.

Issue #919, opened 2020-07-06
Evenloped event payloads has content-type included twice


tomczoink:
Issue raised by Paul:

the enveloped event payloads have content-type listed twice https://www.ably.io/documentation/general/events#envelope-examples and I think the second one is correct.

MarkWoulfeAbly:
@SimonWoolf can you confirm which is the correct description here?

SimonWoolf:
The second is correct for enveloped payloads. The first [is for unenveloped](https://github.com/ably/docs/pull/819#discussion_r382602409).

`Batched event payloads` section also incorrectly has both. It's the same as for enveloped (json or msgpack).

`x-ably-envelope-appid` also seems to have wrongly made it into the headers for enveloped & batched rules.

MarkWoulfeAbly:
So just to clarify, the work that needs to be done here is:

1. Remove the first content-type entry from https://www.ably.io/documentation/general/events#envelope-examples and https://www.ably.io/documentation/general/events#batch-examples
2. Remove x-ably-envelope-appid from both https://www.ably.io/documentation/general/events#batch-examples and https://www.ably.io/documentation/general/events#envelope-examples

Is this correct @SimonWoolf ?

SimonWoolf:
yes

Issue #864, opened 2020-04-14
No easy way to see what's changed for different rule versions


SimonWoolf:
A customer recently pointed out that there is no easy way to see what has changed between different versions of reactor rules. In webhooks this is especially true since now that it's been restructured, you can't even change the version in the dropdown to manually compare them, as there is no dropdown. But TBH even if you could, expecting customers to manually diff between two pages is a crap way of figuring out what's changed. Any reason not to have a changelog / version history or sthg section?

PR #929, opened 2020-07-31
Updated reactor events to address issues #864, #865 and #919

MarkWoulfeAbly:


mattheworiordan:
In future, can I ask that you please include a description of what a PR does in the description field?  See https://ably.atlassian.net/wiki/spaces/ENG/pages/312377505/Git-PR-best-practice

MarkWoulfeAbly:
@SimonWoolf can you take a look at this when you get a moment? Thanks

content/general/events.textile
```diff
@@ -72,9 +72,276 @@ h3(#sources). Sources
...
+Each message envelope in the @items@ array consists of an object with the following attributes:
+
+- name := the event type "@channel.message@"
+- webhookId := an internal unique ID for the configured WebHook
+- source := the source for the WebHook, namely "@channel.message@"
+- timestamp := a timestamp represented as milliseconds since epoch for the published message
+- data := an object containing the attributes defined below in "JSONPath format":http://goessner.net/articles/JsonPath/ @data.*@:
+- data.channelId := name of the channel that the presence event belongs to
+- data.site := an internal site identifier indicating the datacenter from which the message was published
+- data.messages := an @Array@ of messages with fields for each message described below
```
SimonWoolf, line 96: Aren't these lists of fields now duplicated? Looks like Tom moved it down under "examples" and reformatted it a bit, and now you've re-added them at the top so they're there twice?

SimonWoolf: Also it's very nonobvious that is the structure for batched payloads

content/general/events.textile
```diff
@@ -72,9 +72,276 @@ h3(#sources). Sources
...
+                  "data":"bar"
+               }
+            ]
+         }
+      }
+   ]
+}
+```
+
+Please note that if you are planning to receive messages via WebHooks, it is theoretically very easy to exceed the transport rate limits we impose on WebHooks to prevent DoS attacks against your own servers. We recommend you consider "message queues":/general/queues instead as they are far more scalable.
```
SimonWoolf, line 155: Following a rearchitecting in how webhooks work, we're no longer pushing people away from webhooks quite so hard as that -- iirc the limits are the same as for our internal queues

MarkWoulfeAbly: Thanks, removing.

content/general/events.textile
MarkWoulfeAbly: Ahh yes I can see this now looking at the deployment, I missed this as the examples wherein linked partial snippets that I couldn't see by just looking at the textile file itself. Do you think I should remove the parts that aren't about decoding specifically then? A little confused on the best resolution here.

MarkWoulfeAbly:
Following a discussion with @tomczoink, we believe that the decoding examples, and anything else relevant that was missing from the previous versions, should be added within the example partials.

Tom will be commenting on this PR with what the new information should be, and where specifically it should go.

tomczoink:
Decode code snippets, should be added to the end of each relevant partial:


Batched messages:

```[javascript]
webhookMessage.items.forEach((item) => {
  const messages = Ably.Realtime.Message.fromEncodedArray(item.data.messages);
  messages.forEach((message) => {
    console.log(message.toString());
  })
})
```

Presence batched messages:

```[javascript]
webhookMessage.items.forEach((item) => {
  const messages = Ably.Realtime.PresenceMessage.fromEncodedArray(item.data.messages);
  messages.forEach((message) => {
    console.log(message.toString());
  })
})
```

Enveloped message:

```[javascript]
const messages = Ably.Realtime.Message.fromEncodedArray(webhookMessage.messages);
messages.forEach((message) => {
  console.log(message.toString());
})
```

Presence enveloped message:

```[javascript]
const messages = Ably.Realtime.PresenceMessage.fromEncodedArray(webhookMessage.messages);
messages.forEach((message) => {
  console.log(message.toString());
})
```

MarkWoulfeAbly:
Thanks @tomczoink, do we need to include any descriptions from the original? - https://github.com/ably/docs/blob/156ab7a174b66bc588a5973cd99d78b60349c5bf/content/general/versions/v1.0/webhooks.textile

tomczoink:
We should definitely still include the description as it's useful, perhaps with each example?

MarkWoulfeAbly:
@tomczoink to verify decoding message sections are correct before moving onto presence

content/partials/general/events/_batched_events.textile
```diff
@@ -38,6 +38,24 @@ minimize. View batched @message@ payload example
   }
   ```
 
+h5(#batch-example-message-decoding). Decoding batched messages
+
+Messages sent "over the realtime service":/realtime/channels are automatically decoded into "@Messages@":/realtime/types#message object by the Ably client library. With webhooks you need to to do this explicitly, using "@Message.fromEncodedArray@":/realtime/messages#message-from-encoded-array on the @data.messages@ array, or "@Message.fromEncoded@":/realtime/messages#message-from-encoded or an individual member of that array. This will transform them into an array of "@Messages@":/realtime/types#message (or in the case of @fromEncoded@, an individual "@Message@":/realtime/types#message). This has several advantages, e.g.:
```
tomczoink, line 43: `"@Messages@":/realtime/types#message object` => `"@Message@":/realtime/types#message objects` I think?

tomczoink: `or an individual member of that array` => `on an individual member of that array

content/partials/general/events/_enveloped_events.textile
```diff
@@ -30,6 +30,22 @@ minimize. View enveloped @message@ payload example
   }
   ```
 
+h5(#envelope-example-message-decoding). Decoding enveloped messages
+
+Messages sent "over the realtime service":/realtime/channels are automatically decoded into "@Messages@":/realtime/types#message object by the Ably client library. With webhooks you need to to do this explicitly, using "@Message.fromEncodedArray@":/realtime/messages#message-from-encoded-array on the @messages@ array, or "@Message.fromEncoded@":/realtime/messages#message-from-encoded or an individual member of that array. This will transform them into an array of "@Messages@":/realtime/types#message (or in the case of @fromEncoded@, an individual "@Message@":/realtime/types#message). This has several advantages, e.g.:
```
tomczoink, line 35: Same two points as batched messages changes.

MarkWoulfeAbly:
Thanks @tomczoink for your feedback, I have addressed your comments on the batched messages and tried to apply the same to the presence ones I've added. Please give one final review :)

PR #938, opened 2020-08-11
Revert "[WEB-381] Add GA tracking to docs.ably.io"

### LINKED ISSUE #931: GA tracking for docs.ably.io should not reuse the existing www.ably.io GA property
See https://github.com/ably/docs/pull/905

This change introduced GA tracking (well technically it added all GTM tracking), but this site is not part of www.ably.io, and as such, is messing up the reported URLs.

docs.ably.io should have its own GA property to avoid this.

screenshot_2020-08-03_10-56-06_am

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/DOC-38) by [Unito](https://www.unito.io)

MarkWoulfeAbly:
Reverts ably/docs#905

Closes #931 

docs.ably.io should have its own GA property, this will be addressed in a separate PR if necessary.

### DIFF
--- a/layouts/_head.html.erb
+++ b/layouts/_head.html.erb
@@ -1,14 +1,6 @@
 <!DOCTYPE HTML>
 <html lang="en">
   <head>
-    <!-- Google Tag Manager -->
-    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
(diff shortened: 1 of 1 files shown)


mattheworiordan:
I do not understand why you are proposing to create a separate PR to add docs.ably.io tracking instead of just switching out `GTM-TZ37KKW` now with a new GA code? That would require less work (fewer reviews, one small change) and give you visibility into the dosc site usage at the same time.

mattheworiordan (changes requested):
I would prefer to see this fixed (i.e. new tracking code) as opposed to remove tracking completely.

MarkWoulfeAbly:
I had mentioned that possibility to @DominikElmiger but at the time of creating the PR he suggested simply removing it would be faster. Dom, do you have the availability to create a new code to replace the old one instead?

tbedford:
I've created DOC-259 and I will create a new PR to tackle this.

PR #941, opened 2020-08-20
Improved docker errors when Gemfile or Gemfile.lock has changed

kennethkalmer:
Our current entrypoint was a bit lacking, and the recovery advice wasn't correct. If the `Gemfile` or `Gemfile.lock` is out of sync the error now gives instructions on how to run `docker-compose up --build` to trigger a rebuild.

The updates tweak how we build our container so we can be more explicit about the Ruby version, with the upside that we can now compare the `Gemfile.lock` during startup as well.


MarkWoulfeAbly (approved):
@kennethkalmer has agreed to make some small formatting changes and then this is good to land

kennethkalmer:
@MarkWoulfeAbly improvements applied

PR #945, opened 2020-08-27
Redirect fixes

kennethkalmer:
Fixed invalid redirects from 695cffb

Also updated nanoc-redirector, and added more clarifications to the README on the expected values of the `redirect_from` metadata.


README.md
```diff
@@ -161,6 +161,8 @@ redirect_from:
 
 Then, when accessing `/redirect-from-this-path/`, you'll be redirected to the affected template.
 
+**IMPORTANT:** The values used in the `redirect_from` list should be the original URL's, not the name of the file that the URL was generated from.
```
MarkWoulfeAbly, line 164:
```suggestion
**IMPORTANT:** The values used in the `redirect_from` list should be the URL without the file extension. E.g. `general/versions/v0.8/webhooks` not `general/versions/v0.8/webhooks.textile`. 
```

MarkWoulfeAbly (approved):
Made a small edit to the readme that you can accept but other than that LGTM

PR #950, opened 2020-09-02
Channel Enumeration API: add lots of warnings and caveats

SimonWoolf:
context: https://github.com/ably/realtime/issues/3160

as we haven't yet decided on how enumeration will work with decoupling, I settled for adding caveats to allow us to return less results and/or duplicated results in the future, even if we don't currently do so.


paddybyers (approved):
LGTM, thanks

PR #970, opened 2020-10-06
Rewrite and clarify publish() clientId handling

SimonWoolf:
turns out ably-cocoa has had [broken](https://github.com/ably/ably-cocoa/commit/b515697b#diff-79536e232504b8844dec250eb5cb7e43R60) clientId handling for a while. One of the reasons for this may be that the specs for this are a little confusing: they split things up into several different cases in order to apply some extra clientside checks to a couple of the cases, and in doing so might be misinterpreted as saying that those cases should be handled differently, even though they really shouldn't. the actual rule is very simple, it's just 'there's no need to change anything ever, it will all just work'! But that's not obvious from those specs at the moment.

This attempts to rewrite them for clarity, and gets rid of the clientside validation requirement, which adds a lot of complexity and isn't really necessary, we do that perfectly well serverside.

Since this isn't actually a behaviour change, just a clarification, ultimately I figured I was probably ok to reuse the same spec items. But happy to change that to new spec items if @QuintinWillison prefers, since the wording is very different even if there's no actual requirements difference.

(cc @paddybyers @tiholic)


kavalerov:
Definitely like the new wording much better - much more concise and descriptive!

QuintinWillison (changes requested):
I think the new essence for both `RSL1f` and `RSL1g` are so different that they should be new spec items. If nothing else because client library codebases could be referring to the existing points (or sub-points) in some way and those external comments or attributes will end up invalid once this change lands.

Also, in current state, your `RSL1f` has the final sentence bracketed. This feels odd. Was that intentional?

SimonWoolf:
> I think the new essence for both RSL1f and RSL1g are so different that they should be new spec items. 

Sure; I've named the new one RSL1m, so 1f and 1g can be considered removed items.

> Also, in current state, your RSL1f has the final sentence bracketed. This feels odd. Was that intentional?

It was intentional, but I have a (probably bad) habit of overusing parentheticals (or so I have been told); I've removed the parens (and I should probably stop over-nesting them too (though tbh I quite like it (from time to time)))

QuintinWillison (approved):

PR #973, opened 2020-10-07
spec: Make ProtocolMessage private

tcard:
ProtocolMessage, as a structured type, is in the spec, yet it isn't
referred to by any other public part of the interface. It's only
mentioned as a type used by internal operations.

We could just remove it (as it's already defined in the Realtime
Protocol Definition, where it really belongs), but I think there's value
in keeping it here in a higher-level form to keep library codebases
uniform.


QuintinWillison (changes requested):
Thanks, @tcard.

While I agree with the intent that you wish to convey, this doesn't feel like the right way to do it, not least because it's a fairly major structural change.

Also, in the current state of this PR, the addition of `TR5` to specify that `ProtocolMessage` is an internal type is repetition. This would not scale as more types become attributed as internal / private.

My preference would be to take a much lighter weight approach, not introducing any new headings. Something as simple as adding a comment suffix to the type definitions in the Python IDL would suffice - e.g.:

```python
class ProtocolMessage: // Internal
```

And then to add a bullet point to the [Interface Definition](https://docs.ably.io/client-lib-development-guide/features/#idl) pre-amble that explains what the addition of this commentary means.

## Related

A quick search of the spec shows that:

- The `msgSerial` attribute of `PresenceMessage` is defined as "internal"
- The `PresenceMap` is also "internal" - albeit not documented in the IDL

So we already have a level of inconsistency in respect of the "weight" of type definitions in this specification. Hence my preference to just leave this as a lightweight as possible.

tbedford:
@tcard @QuintinWillison - would you be able to take a look at this please? There's a merge conflict - I took a look but wasn't quite sure the best way to resolve. Thanks.

tbedford (approved):

QuintinWillison (approved):

PR #974, opened 2020-10-07
Move RTN14e (transition to SUSPENDED) to RTN15 (state recovery)

tcard:
The transition from DISCONNECTED to SUSPENDED means we've lost the
server-side connection state. But RTN14 describes the retry logic of
first establishing a connection. So it doesn't make sense to transition
to SUSPENDED there, since there's no connection state to lose yet at
that point.

Instead, initial connection attempts keep happening while in
DISCONNECTED state indefinitely.

The transition to SUSPENDED has been thus moved under RTN15, which deals
with connection state recovery.


tcard:
@gernest This is what I mentioned in https://github.com/ably/ably-go/pull/172#discussion_r499546016 . PTAL

SimonWoolf:
:-1: 

> So it doesn't make sense to transition to SUSPENDED there, since there's no connection state to lose yet at that point.

The `suspended` state is not about dropping connection state. Dropping connection state, if there is any, is a one-off action that happens as a side effect of the transition to `suspended`. But that's just a one-off task -- if that was all there was to it, it wouldn't need to be a state in the state machine.

Rather the `suspended` state is about
- no longer queuing messages - per our api we queue publishes clientside for at most two minutes, after which we drop all queued messages and any further publishes are refused
- lowering the frequency of the reconnection attempts. (by default not very much, admittedly - just from 15s to 30s, barely seems any point - but that is modifyable if someone wants to throttle connection attempts after 2 minutes way down)

Both of these should still happen even if the library has never made a successful connection.

tcard:
Right, I didn't take queuing into account.

It's a bit confusing that we're using `connectionStateTtl` also as "queued messages TTL", which seems unrelated. But not worth changing probably.

SimonWoolf:
> It's a bit confusing that we're using connectionStateTtl also as "queued messages TTL", which seems unrelated.

true - there's no fundamental reason they should be tied together, the reasoning was mostly that if you'd be happy for messages to be delivered up to (say) 2 minutes stale, then you probably don't really care whether the delay is due to connectivity issues on the publish side or the subscriber side. so it seemed like a reasonable default to have them both be the same

PR #975, opened 2020-10-07
updated code standards for channel enum tutorial

Srushtika:
Fixes bad variable names and adds let/ const in place of var


kennethkalmer:
@dpiatek @fliptopbox is the browser support for using `let` & `const` good enough? From what I saw on caniuse.com it does look good and this should work without issue.

fliptopbox:
Yup `let` and `const` all good.
Had a look at the code, quickly, and only have this question.
`https://cdn.ably.io/lib/ably-1.js` is this our latest CDN?

Srushtika:
It's at 1.2 at the moment, so it's all good per https://github.com/ably/ably-js#for-browsers @fliptopbox

PR #978, opened 2020-10-09
spec: Fallback behaviour does not apply when port or tlsPort are set

lmars:
This updates the spec to declare that, in the absence of `ClientOptions#fallbackHosts` and `ClientOptions#fallbackHostsUseDefault`, fallback behaviour does not apply when either `ClientOptions#port` or `ClientOptions#tlsPort` is set.

This is inline with the proposed change to ably-ruby in PR https://github.com/ably/ably-ruby/pull/196 which states:

> We need to ensure that custom hosts or custom ports result do not use default environment fallback hosts. We should only use when only the custom environment is set

@vzhikserg @sacOO7 @QuintinWillison I hope this helps to clarify the conversation from https://github.com/ably/ably-java/pull/618.


sacOO7:
Thanks for making a PR. About, custom realtimehost, are we making any changes to spec or fixing code implementation?

lmars:
@sacOO7 

> About, custom realtimehost, are we making any changes to spec

No changes to the spec, no, if `ClientOptions.realtimeHost` is set then it should be used for realtime connections, and I don't think the spec indicates otherwise.

content/client-lib-development-guide/features.textile
```diff
@@ -97,7 +97,11 @@ h3(#restclient). RestClient
 * @(RSC12)@ REST endpoint host is configurable in the Client constructor with the option @restHost@
 * @(RSC13)@ The client library must use the connection and request timeouts specified in the @ClientOptions@, falling back to the defaults described in @ClientOptions@ below
 * @(RSC15)@ Host Fallback
-** @(RSC15b)@ The fallback behavior described below only applies when either @ClientOptions#restHost@ has not been set to an explicit value (see "RSC11a":#RSC11a), the deprecated @ClientOptions#fallbackHostsUseDefault@ option is set to @true@, or an array of @ClientOptions#fallbackHosts@ is provided. If host fallback does not apply, failing HTTP requests that would have "qualified for a retry against a fallback host (see RSC15d)":#RSC15d, will instead result in an error immediately
+** @(RSC15b)@ The fallback behavior described below only applies when either:
+*** @(RSC15b1)@ @ClientOptions#restHost@ has not been set to an explicit value (see "RSC11a":#RSC11a) AND @ClientOptions#port@ is not set AND @ClientOptions#tlsPort@ is not set
```
QuintinWillison, line 101: I don't think the capitalisation of instances of "and" is necessary. From what I can see we don't have a convention, informal or otherwise, of doing that elsewhere in this document.

```suggestion
*** @(RSC15b1)@ @ClientOptions#restHost@ has not been set to an explicit value (see "RSC11a":#RSC11a) and @ClientOptions#port@ is not set and @ClientOptions#tlsPort@ is not set
```

content/client-lib-development-guide/features.textile
```diff
@@ -97,7 +97,11 @@ h3(#restclient). RestClient
 * @(RSC12)@ REST endpoint host is configurable in the Client constructor with the option @restHost@
 * @(RSC13)@ The client library must use the connection and request timeouts specified in the @ClientOptions@, falling back to the defaults described in @ClientOptions@ below
 * @(RSC15)@ Host Fallback
-** @(RSC15b)@ The fallback behavior described below only applies when either @ClientOptions#restHost@ has not been set to an explicit value (see "RSC11a":#RSC11a), the deprecated @ClientOptions#fallbackHostsUseDefault@ option is set to @true@, or an array of @ClientOptions#fallbackHosts@ is provided. If host fallback does not apply, failing HTTP requests that would have "qualified for a retry against a fallback host (see RSC15d)":#RSC15d, will instead result in an error immediately
+** @(RSC15b)@ The fallback behavior described below only applies when either:
```
QuintinWillison, line 100: I know you've inherited these words but I'm not comfortable with "described below" as a turn of phrase, given the context. Would not something like "described by this section, RSC15" be more accurate in terms of making the scope clear?

```suggestion
** @(RSC15b)@ The fallback behavior described by this section, "RSC15":#RSC15, only applies when either:
```

content/client-lib-development-guide/features.textile
```diff
@@ -464,7 +468,10 @@ h3(#realtime-connection). Connection
 ** @(RTN16e)@ If the @recover@ option is missing or no longer valid when connecting to Ably, the client will connect anyway, but emit a @ConnectionStateChange@ with a @reason@, and will additionally set the @Connection#errorReason@ with an @ErrorInfo@ object describing the failure
 ** @(RTN16f)@ The @msgSerial@ component of the @recoveryKey@, unlike the other two components, is not sent to Ably, but rather is used to set the library internal @msgSerial@. (If the recover fails, the counter should be reset to 0 per "RTN15c3":#RTN15c3 )
 * @(RTN17)@ Host Fallback
-** @(RTN17b)@ The fallback behavior described below only applies when either @ClientOptions#realtimeHost@ has not been set to an explicit value (see "RTC1d":#RTC1d), the deprecated @ClientOptions#fallbackHostsUseDefault@ option is set to @true@, or an array of @ClientOptions#fallbackHosts@ is provided.
+** @(RTN17b)@ The fallback behavior described below only applies when either:
```
QuintinWillison, line 471: Matching my comment on `RSC15b`:

```suggestion
** @(RTN17b)@ The fallback behavior described by this section, "RTN17":#RTN17, only applies when either:
```

content/client-lib-development-guide/features.textile
```diff
@@ -464,7 +468,10 @@ h3(#realtime-connection). Connection
...
+*** @(RTN17b1)@ @ClientOptions#realtimeHost@ has not been set to an explicit value (see "RTC1d":#RTC1d) AND @ClientOptions#port@ is not set AND @ClientOptions#tlsPort@ is not set
```
QuintinWillison, line 472: Matching my comment on `RSC15b1`:

```suggestion
*** @(RTN17b1)@ @ClientOptions#realtimeHost@ has not been set to an explicit value (see "RTC1d":#RTC1d) and @ClientOptions#port@ is not set and @ClientOptions#tlsPort@ is not set
```

QuintinWillison (changes requested):
Just a few minor observations. Thanks!

content/client-lib-development-guide/features.textile
lmars: Updated in d3ac5350

content/client-lib-development-guide/features.textile
lmars: Updated in d3ac5350

content/client-lib-development-guide/features.textile
lmars: Clarified in 59e1e737

content/client-lib-development-guide/features.textile
lmars: Clarified in 59e1e737

lmars:
@QuintinWillison comments addresses, PTAL

QuintinWillison (approved):

PR #979, opened 2020-10-09
ClientOptions: Remove deprecated fallbackHostsUseDefault option

SimonWoolf:
Being deprecated in favour of us always creating dns entries for the full set of autogenerated fallbacks


tomczoink (approved):

mattheworiordan:
Out of interest, what happens if a custom hostname is specified?

PR #980, opened 2020-10-12
Presence.history: remove untilAttach param

SimonWoolf:
Remove references to the untilAttach param in the docs for presence history.

Context: https://github.com/ably/realtime/issues/3069#issuecomment-707279282 . Literally no-one uses it, because there's no usecase for it, since you get a full presence sync when you attach. And it adds complexity to the server. So I intend to take advantage of the fact that no-one uses it to remove support for it without breaking anyone's code.

The only snag is that ably-dotnet for some reason includes `untilAttach` as a standalone parameter to the history() call rather than just being one of the other params, so that can't be removed without a breaking change. So for that client lib I've left it in the documentation and marked it as deprecated. We should add new versions of the method that don't take that param and make them the main documented versions.

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -594,7 +594,7 @@ h3(#realtime-channel). Channel
 ** @(RTL10d)@ A test should exist that publishes messages from one client, and upon confirmation of message delivery, a history request should be made on another client to ensure all messages are available
 * @(RTL12)@ An attached channel may receive an additional @ATTACHED@ @ProtocolMessage@ from Ably at any point. (This is typically triggered following a transport being resumed to indicate a partial loss of message continuity on that channel, in which case the @ProtocolMessage@ will have a @resumed@ flag set to false). If and only if the @resumed@ flag is false, this should result in the channel emitting an @UPDATE@ event with a @ChannelStateChange@ object. The @ChannelStateChange@ object should have both @previous@ and @current@ attributes set to @attached@, the @reason@ attribute set to to the @error@ member of the @ATTACHED@ @ProtocolMessage@ (if any), and the @resumed@ attribute set per the @RESUMED@ bitflag of the @ATTACHED@ @ProtocolMessage@. (Note that @UPDATE@ should be the only event emitted: in particular, the library must not emit an @ATTACHED@ event if the channel was already attached, see @RTL2g@).
 * @(RTL15)@ @Channel#properties@ attribute is a @ChannelProperties@ object representing properties of the channel state. @properties@ is a publicly accessible member of the channel, but it is an experimental and unstable API. It has the following attributes:
-** @(RTL15a)@ @attachSerial@ is unset when the channel is instanced, and is updated with the @channelSerial@ from each @ATTACHED@ @ProtocolMessage@ received from Ably with a matching @channel@ attribute. The @attachSerial@ value is used for @untilAttach@ queries, see "RTL10b":#RTL10b and "RTP12b":#RTP12b.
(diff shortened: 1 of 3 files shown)


content/client-lib-development-guide/features.textile
```diff
@@ -682,7 +682,7 @@ h3(#realtime-presence). Presence
 *** @(RTP11c3)@ @connectionId@ filters members by the provided @connectionId@
 * @(RTP12)@ @Presence#history@ function:
 ** @(RTP12a)@ Supports all the same params as REST @Presence#history@
-** @(RTP12b)@ Additionally supports the param @untilAttach@, which if true, will only retrieve messages prior to the moment that the channel was attached or emitted an @UPDATE@ indicating loss of continuity. This bound is specified by passing the querystring param @fromSerial@ with the @Channel#properties.attachSerial@ assigned to the channel in the @ATTACHED@ @ProtocolMessage@ (see "RTL15a":#RTL15a). If the @untilAttach@ param is specified when the channel is not attached, it will result in an error
+** @(RTP12b)@ Does not support the @untilAttach@ param, unlike @Channel#history@
```
QuintinWillison, line 685: For me, leaving this spec point here just adds noise. Would it not be cleaner just to delete it entirely?

QuintinWillison (changes requested):
What's with the `[deprecated]` on the `csharp` code? Is that a known syntax in either C# or our textile-based docs engine?

Also, can we say it's deprecated, despite pending change to be worked on in the .NET library (I created https://github.com/ably/ably-dotnet/issues/449)?

content/client-lib-development-guide/features.textile
SimonWoolf: Fair point - my reasoning was that as part of compliance to spec versions that include this change should include actively removing any now-useless code that does what RTP12b used to require, wasn't sure that just dropping the item would carry the same weight.

But perhaps a better solution to that would be removing the spec item and raising github issues in all the libs to make that transition. I'll do that

SimonWoolf:
> What's with the `[deprecated]` on the csharp code? Is that a known syntax in either C# or our textile-based docs engine?

No, just human-readable annotation - happy to take suggestions of better ways to annotate that

> (I created ably/ably-dotnet#449)

Thanks 🙂

> Also, can we say it's deprecated, despite pending change to be worked on in the .NET library

We can't deprecate that method call until we have a replacement call, sure, but I figured we can still mark that parameter within the method call as deprecated? WDYT?

kavalerov (approved):
This makes sense to me - if something is not being used we definitely should be removing it 👍

content/client-lib-development-guide/features.textile
SimonWoolf: @QuintinWillison this was done

QuintinWillison:
No strong feelings around the "deprecated" annotation style... perhaps would be preferable using `//` comment syntax but happy to leave it like this equally.

QuintinWillison (approved):

mattheworiordan:
Has this been updated in the Ably changelog? https://changelog.ably.com/
I see @SimonWoolf posted something about a change in rewind behaviour.

SimonWoolf:
> Has this been updated in the Ably changelog? https://changelog.ably.com/

I didn't bother on the grounds that the reason we're able to make the change so freely is that per server analytics literally no-one uses it, so no-one'll notice or care that it's gone, so what's the point in advertising that it's gone in the changelog?

mattheworiordan:
> so no-one'll notice or care that it's gone, so what's the point in advertising that it's gone in the changelog?

Well for completeness.  If someone uses an older version of an SDK, which does support it, which is admittedly unlikely, they could later find out it's gone and have no idea why.  

It's not all that important, but it is what the changelog is for.

PR #989, opened 2020-11-16
Adding Flutter docs

tiholic:
First version of flutter docs


markcoxsmith (approved):

content/realtime/channels.textile
```diff
@@ -540,6 +575,13 @@ channel.subscribe("action") { message in
 }
 channel.publish("action", data: "boom!")
 
+bc[flutter]. final channel = realtime.channels.get('chatroom');
+/* Implicit attach when subscribing */
+channel.subscribe(name: 'action').listen((ably.Message message){
+    print('Received: ${message.data}');
```
mattheworiordan, line 581: Inconsistent indentation and use of `{` without a space above.

content/realtime/channels.textile
```diff
@@ -645,6 +691,17 @@ bc[swift]. channel.on(.attached) { stateChange in
   }
 }
 
+bc[flutter]. final stateChangeListener = channel.on(ably.ChannelEvent.attached)
+  .listen((ably.ChannelStateChange state){
+    print('channel ${channel.name} is now attached');
+    if (stateChange.resumed) {
+    System.out.println('Message continuity was preserved');
```
mattheworiordan, line 698: Inconsistent indentation.

content/realtime/channels.textile
```diff
@@ -645,6 +691,17 @@ bc[swift]. channel.on(.attached) { stateChange in
   }
 }
 
+bc[flutter]. final stateChangeListener = channel.on(ably.ChannelEvent.attached)
```
mattheworiordan, line 694: I think this would be more readable as:

```dart
final stateChangeListener = channel
  .on(ably.ChannelEvent.attached)
  .listen((ably.ChannelStateChange state){
    print('channel ${channel.name} is now attached');
    ...
```

Because then all chained calls are indented in the same way.  WDYT?

content/realtime/channels.textile
```diff
@@ -677,6 +734,12 @@ bc[swift]. let listener = channel.on { stateChange in
     print("channel state is \(stateChange.current)")
 }
 
+bc[flutter]. final stateChangeListener = channel.on()
```
mattheworiordan, line 737: As above

content/realtime/messages.textile
```diff
@@ -160,6 +170,11 @@ bc[swift]. channel.subscribe { message in
     print("message data: \(message.data)")
 }
 
+bc[flutter]. final channelMessageSubscription = channel.subscribe().listen((ably.Message message){
+ print('message received for event ${message.name}');
```
mattheworiordan, line 174: Inconsistent indentation.

content/realtime/channels.textile
```diff
@@ -919,6 +997,7 @@ bq(definition#publish-data).
   java:     void publish(String name, Object data, "CompletionListener":#completion-listener listener)
   csharp:   void Publish(string name, object data, Action<bool, ErrorInfo> callback = null, string clientId = null)
   objc,swift: publish(name: String?, data: AnyObject?, callback: (("ARTErrorInfo":/realtime/types#error-info?) -> Void)?)
+  flutter:  Future<void> publish(name: String name, data: Object data, message: "Message":#message message)
```
mattheworiordan, line 1000: Is this correct?  It looks odd with the arg duplicated?  I may be wrong as I am no Dart expert, but wouldn't:

```dart
Future<void> publish(name: String, data: Object, message: "Message":#message)
```

.. be better?

content/realtime/channels.textile
```diff
@@ -777,6 +845,16 @@ bc[swift]. realtime.channels.get("private:chatroom").attach { error in
     }
 }
 
+bc[flutter]. Channel channel = realtime.channels.get("private:chatroom");
```
mattheworiordan, line 848: It is better not to mix use of `"` and `'` in the same code samples, unless there is a compelling reason.  I note above you use `.split('.')` so not a strictly interpolation need I assume.

mattheworiordan (changes requested):
This is awesome.

One observation - why have you gone with `ably.Realtime` instead of `Ably.Realtime`?  The latter is consistent with other libraries (apart from Java) and even idiomatic Dart (as far as I can tell https://dart.dev/samples).  The difference is Java is a namespace `io.ably.*` whereas here it's an import of a class.

Unless there is a compelling reason for lowercase `ably` for a reference to the Class, can we go with Uppercase for consistency please?

Few other questions:

- Why did we opt not to include code samples for channel params & rewind pages?
- There are a lot of trivial whitespace / indentation / line wrap type issues.  Can you have a run through to address across the board?
- There appears to be some missing code samples when I try the rendered version, see https://ably-docs-flutter-docs-wlxvvlu.herokuapp.com/realtime/messages/ for example under Publish, it says no Flutter version exists.  Same for retrieving message history.  Can I ask that you browse the REST & Realtime pages to ensure a code sample exists where other multilang examples exist?
- How are you keeping track of what you have implemented or not for Flutter?  I assume lots of these methods aren't yet available, hence not updated.  This is fine, except a) how do customers know?, b) how will we know what to go back and update?

content/realtime/channels.textile
tiholic: @mattheworiordan do I need to expose the function signature here?

In that case it will have to be, as all the arguments passed here are named (like kwargs in python)

```dart
Future<void> publish({
  name: String,
  data: Object,
  message: "Message":#message
  messages: List<"Message":#message>
});
```

content/realtime/channels.textile
tiholic: Oh, thanks. I missed this.
All dart code samples use `'`

tiholic:
Thanks Matt,

> why have you gone with ably.Realtime instead of Ably.Realtime? 

`ably` is an import prefix and not a class. Dart recommends using snake_casing over camel/PascalCasing
https://dart.dev/guides/language/effective-dart/style#do-name-import-prefixes-using-lowercase_with_underscores

> channel params & rewind pages

I believe you are referring to `ChannelOptions`. Even though we are passing them to native libraries, I've not experimented with that option yet as I'm not fully aware of the cipher -or- how encryption mechanism used.

I came across `rewind` only while updating the docs. Will include docs after we have appropriate test cases for that.

> There appears to be some ... browse the REST & Realtime pages to ensure a code sample exists where other multilang examples exist?

Sure, will do.

> a) how do customers know?

I thought feature matrix would be of a little help.

> b) how will we know what to go back and update

My plan is to update the docs for each release based on the changelog.

content/realtime/channels.textile
tiholic: addressed in e288898cb0f972b74350bd69c2da054346fbb4f0

content/realtime/messages.textile
tiholic: addressed in e288898cb0f972b74350bd69c2da054346fbb4f0

content/realtime/channels.textile
tiholic: addressed in e288898cb0f972b74350bd69c2da054346fbb4f0

content/realtime/channels.textile
tiholic: addressed in e288898cb0f972b74350bd69c2da054346fbb4f0

content/realtime/channels.textile
tiholic: addressed in e288898cb0f972b74350bd69c2da054346fbb4f0

content/realtime/channels.textile
tiholic: addressed in e288898cb0f972b74350bd69c2da054346fbb4f0

kavalerov (approved):
I think there is a general problem with docs (and not just within Flutter context) regarding features which are not supported in certain languages - i.e. Channel Params only have examples in JS and Node.js now. 

I suggest we address those in a separate issue:  https://github.com/ably/docs/issues/990

I think this PR is good. 

The questions about case for import statement and "named parameters"  are potentially outside of scope of this PR (changing them will require changes in the library itself) - I suggest we continue in Slack/Confluence, or raise an issue in https://github.com/ably/ably-flutter

content/realtime/channels.textile
QuintinWillison: I agree with @kavalerov's [comment](https://github.com/ably/docs/pull/989#pullrequestreview-535400038) that this doesn't fall within the scope of this PR. Let's capture this concern as an issue in [ably-flutter](https://github.com/ably/ably-flutter), please, @tiholic - if any API changes are made then we'll have to make a subsequent docs change.

content/realtime/messages.textile
QuintinWillison: @tiholic Can you revisit your fix, please. You're missing a space before the opening curly brace [here](https://github.com/ably/docs/commit/e288898cb0f972b74350bd69c2da054346fbb4f0#diff-ab0455ad7012decae9dc129170e4c90888957d9d6fc2f14ca30fa6ba601b3586R175).

content/realtime/usage.textile
```diff
@@ -155,6 +156,18 @@ blang[objc,swift].
...
 
+blang[flutter].
+  The Ably library is "hosted on Github":https://github.com/ably/ably-flutter and is available as a "Dart package":https://pub.dev/packages/ably-flutter/. Update your @pubspec.yaml@ with appropriate package version:
+
+  bc[yaml]. dependencies:
+    # ...
+    ably-flutter: [version]
+    # ...
+
+  Now, import the package in your dart file
```
QuintinWillison, line 167:
```suggestion
  Now, import the package into your Dart file.
```

I see there's another location in this PR where this will need fixing also.

content/partials/types/_channel_state.textile
```diff
@@ -120,3 +120,18 @@ blang[go].
       StateChanFailed = 32768
     )
   ```
+
+blang[flutter].
+  @ably.ChannelState@ is an enum representing all the "@Realtime Channel@ states":/realtime/channels#channel-states.
+
+  ```[flutter]
+  enum ChannelState {
```
QuintinWillison, line 128: Indentation is perhaps two columns too short, to match the level of the preceding Go example, as well as your Flutter code [here](https://github.com/ably/docs/pull/989/files#diff-a35132c3750f7cf9ed3ee4a13e441713b4a8c3ccd3453ae5de5245b1330ba078R122).

content/realtime/usage.textile
```diff
@@ -155,6 +156,18 @@ blang[objc,swift].
 
   **Note: Since version 1.0.8, the iOS client libraries are thread-safe.** Check "the README":https://github.com/ably/ably-cocoa#thread-safety for details.
 
+blang[flutter].
+  The Ably library is "hosted on Github":https://github.com/ably/ably-flutter and is available as a "Dart package":https://pub.dev/packages/ably-flutter/. Update your @pubspec.yaml@ with appropriate package version:
```
QuintinWillison, line 160: Please avoid using "Dart package" as that is misleading for non-Flutter developers. I would prefer "Flutter plugin" or "plugin" for clarity. e.g.:

```suggestion
  The Ably library is "hosted on Github":https://github.com/ably/ably-flutter and is available as a "Flutter plugin":https://pub.dev/packages/ably-flutter/. Update your @pubspec.yaml@ with appropriate package version:
```

I see there's at least one other instance in this PR where this will need fixing also.

QuintinWillison (changes requested):
I'm not qualified to comment on the textile syntax correctness, but looks sensible to me. Just a few minor observations. Thanks.

content/realtime/channels.textile
tiholic: @kavalerov @QuintinWillison our current implementation in SDK uses named params, but I wasn't sure on what to represent in docs. Should we be showcasing method signature here? Or its usage?

content/realtime/usage.textile
tiholic: addressed 4770624984d162f1ca4416b94db7dd50f643e400

content/realtime/usage.textile
tiholic: Addressed 4770624984d162f1ca4416b94db7dd50f643e400

content/partials/types/_channel_state.textile
tiholic: Addressed 19a969ae337a59f8d8c3c8d8d4bcde6858dbd91c

content/realtime/messages.textile
tiholic: Thanks for pointing this out @QuintinWillison. There were quite a number of usages where I missed that space.
Addressed 19a969ae337a59f8d8c3c8d8d4bcde6858dbd91c

QuintinWillison: Thanks. I'm going to resolve this conversation, on behalf of Matt.

content/realtime/channels.textile
QuintinWillison: @tiholic, I think "usage", matching the surrounding examples from other client libraries.

content/realtime/messages.textile
tiholic: @Quintin
In that case we're good to resolve this conversation, I guess _( unless the syntax isn't clear enough! )_

content/realtime/channels.textile
tiholic: @Quintin
In that case we're good to resolve this conversation, I guess _( unless the syntax isn't clear enough! )_

QuintinWillison (approved):

PR #1016, opened 2021-01-18
[WEB-408] - Support SEO metadata

mattheworiordan:
Content team need to add metadata to the docs content published.  This is a blocker for them so I wanted to unblock them.

Note there is an associated website PR to come.

See https://github.com/ably/website/pull/3524.

PR #1020, opened 2021-01-19
Ensure relative image links for OpenGraph are absolute

mattheworiordan:
Fixes an issue with opengraph images that didn't work correctly with https://github.com/ably/docs/pull/1016.  A separate PR will fix this in full on the website.

screenshot_2021-01-19_07-03-22_pm

As this is a static site builder, we have to hard code a URL unfortunately.  Given the docs site is designed for development use, compromising to use the docs well know host name is OK.

### DIFF
--- a/lib/helpers/link_helper.rb
+++ b/lib/helpers/link_helper.rb
@@ -1,6 +1,6 @@
 module LinkHelper
   CANONICAL_ROOT = 'https://www.ably.io/' unless defined?(CANONICAL_ROOT)
-  DOCS_ROOT = 'https://docs.ably.io/' unless defined?(DOCS_ROOT)
+  DOCS_ROOT = 'https://docs.ably.com/' unless defined?(DOCS_ROOT)
 
   def link(text, target, attributes = {})
     target = "/#{target}" unless target.match(/^(\w+:\/\/|\/)/)
@@ -17,7 +17,7 @@ module LinkHelper
   end
 
   def image_url(image_path)
-    return image_path if image_path.match(/^http/)
+    return image_path if image_path.match(/^http/i)
     "#{DOCS_ROOT}#{image_path.gsub(%r{^/}, '')}"
   end
 end
--- a/lib/helpers/link_helper.rb
+++ b/lib/helpers/link_helper.rb
@@ -1,5 +1,6 @@
 module LinkHelper
   CANONICAL_ROOT = 'https://www.ably.io/' unless defined?(CANONICAL_ROOT)
+  DOCS_ROOT = 'https://docs.ably.io/' unless defined?(DOCS_ROOT)
 
   def link(text, target, attributes = {})
     target = "/#{target}" unless target.match(/^(\w+:\/\/|\/)/)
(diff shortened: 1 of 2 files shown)


lib/helpers/link_helper.rb
```diff
@@ -1,5 +1,6 @@
 module LinkHelper
   CANONICAL_ROOT = 'https://www.ably.io/' unless defined?(CANONICAL_ROOT)
+  DOCS_ROOT = 'https://docs.ably.io/' unless defined?(DOCS_ROOT)
```
fliptopbox, line 3: shouldn't this be `.com`?

lib/helpers/link_helper.rb
```diff
@@ -14,6 +15,11 @@ def canonical_link
       "#{CANONICAL_ROOT}documentation#{clean_item.gsub("/root", "")}".gsub(%r{/$}, '')
     end
   end
+
+  def image_url(image_path)
+    return image_path if image_path.match(/^http/)
```
fliptopbox, line 20: is the regex case insensitive by default ie `HTTP` or `http`?

fliptopbox (changes requested):

lib/helpers/link_helper.rb
mattheworiordan: Mm, I was not aware it was migrated given docs.ably.io still works.  @kennethkalmer @jamienewcomb who migrated docs.ably.io to .com?  It needs redirects in place.  I will change to .com Bruce, thanks.

lib/helpers/link_helper.rb
mattheworiordan: No, good point too, thanks

mattheworiordan: See 
7439acb

lib/helpers/link_helper.rb
mattheworiordan: See 7439acb

fliptopbox (approved):

fliptopbox:
@ramiro-nd `<meta name="twitter:description" content="How to create your first npm package and publish it." />` i checked the twitter meta tags and noticed we are not using `description` or `creator` just checking we are okay with this. @mattheworiordan WDYT?

lib/helpers/link_helper.rb
jamienewcomb: We have a ticket to do this. Its with @kennethkalmer https://ably.atlassian.net/browse/FEA-558

kennethkalmer:
I'm weary to react too quickly again, how do we work with existing images on docs and why would this be any different?

kennethkalmer:
We don't need these changes at all @mattheworiordan, thanks for the PR though. The website already handles assets from docs just fine, it is only missing the host/protocol on the OpenGraph tags. We can zero in on that and get only that fix out the door without interrupting anything on the docs side.

I'll close this off once we've got a fix deployed on the website

kennethkalmer:
Closing this off in favour of ably/website#3530, which has been deployed and the Open Graph images show up correctly.

PR #1022, opened 2021-01-22
Added redirects, updated alt image, and added a new image.

ramiro-nd:
- Added redirects from 4 URLs of the tutorial that Google Analytic shows. (is the format of the URL ok? We don't need to include https://www.ably.io/? 
- Optimized the ALT attribute of the Data flow image
- Added a new image to showcase Syncfusion charts.

Let me know if it's all good

### DIFF
--- a/content/tutorials/realtime-cryptocurrency-app-flutter.textile
+++ b/content/tutorials/realtime-cryptocurrency-app-flutter.textile
@@ -501,10 +501,7 @@ We are all set up and ready! Now we just need the queue to be turned into a list
 
 Here we will use the "**Syncfusion Flutter Charts**":https://www.syncfusion.com/flutter-widgets/flutter-charts package to render the charts. Why Syncfusion? Their Flutter Charts package is a beautifully-crafted charting widget to visualize data. It contains a gallery of 30+ charts and graphs that can be fully customized with options to include animations and render huge amounts of data in seconds. You can "try it yourself for free":https://www.syncfusion.com/account/login.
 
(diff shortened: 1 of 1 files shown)


tbedford:
The image looks kind of large. I will have a play and see if I can fix it - if not today then tomorrow first thing.

image

tbedford:
@ramiro-nd - also the image is adjacent to the code that follows it. We really need a line of text before the following code block to break that up. If you can suggest a brief introductory text (a sentence) then I will add that before the code block. That is, we just need to say what the code block does so there is some context. Thanks.

ramiro-nd:
> @ramiro-nd - also the image is adjacent to the code that follows it. We really need a line of text before the following code block to break that up. If you can suggest a brief introductory text (a sentence) then I will add that before the code block. That is, we just need to say what the code block does so there is some context. Thanks.

@tbedford that's a good point, maybe the image should go after the code then so it shows the result?

tbedford:
> > @ramiro-nd - also the image is adjacent to the code that follows it. We really need a line of text before the following code block to break that up. If you can suggest a brief introductory text (a sentence) then I will add that before the code block. That is, we just need to say what the code block does so there is some context. Thanks.
> 
> @tbedford that's a good point, maybe the image should go after the code then so it shows the result?

Yes, I can try that, in which case I'll need to add some text on what the diagram is before we have the diagram.

tbedford (approved):
LGTM. Reviewed locally (as it's a fork) using the latest website release.

PR #1023, opened 2021-01-22
DOC-238: Add incoming webhooks topic

tbedford:
## Description

Documentation topic for incoming webhooks.

DOC-238

## Review

* [Page to review](https://ably-docs-doc-238-add-i-v0orbd.herokuapp.com/general/incoming-webhooks/)


tbedford:
@ginotoro - interested in getting your feedback. Anything else we need to add?

ginotoro:
@kennethkalmer also nice to get a review from some team members.

ginotoro:
> @ginotoro - interested in getting your feedback. Anything else we need to add?

@tbedford the content is great. I see a rendering issue with the curl command. Don't think it's the content causing it.

tbedford:
> > @ginotoro - interested in getting your feedback. Anything else we need to add?
> 
> @tbedford the content is great. I see a rendering issue with the curl command. Don't think it's the content causing it.

@ginotoro - this is because unfortunately what you see in the review app is not what appears on the website. I've attached a screenshot, which hopefully shows that on the website you won't get the text overflow problem you see in the review app (or if you are running `nanoc` locally). 

image

ginotoro (approved):
Looks good to me

content/general/incoming-webhooks.textile
```diff
@@ -0,0 +1,104 @@
...
+    - Configuring your incoming webhook#configure
+    - Test your incoming webhook#test
+    - Receiving the incoming webhook#receive
+    - Specifying optional headers#headers
+    - Summary#summary
+    - See also#see-also
+redirect_from:
+---
+
+Incoming webhooks provide a way to integrate external web services with Ably. The incoming webhooks are routed to Ably channels as "messages":/realtime/messages.
```
paddybyers, line 20: I think this paragraph is misleading, in that it implies it is necessary to configure a webhook to make this possible. In fact, these incoming webhooks are just a convenience feature - it's perfectly possible to configure a webhook with an existing key - and I think that should be clear.

content/general/incoming-webhooks.textile
```diff
@@ -0,0 +1,104 @@
...
+channel.subscribe(function(message) {
+    console.log("webhook: " + JSON.stringify(message.data));
+});
+```
+
+h2(#headers). Specifying optional headers
+
+The request body of incoming webhooks is treated as a message to be published. You can optionally set headers and parameters in your external service, if the service supports this, as described in the "REST API":/rest-api#unenveloped documentation.
+
+The following example shows how the message name could be set via a request header:
```
paddybyers, line 78: `name` should be marked up as an identifier `@name@`

content/general/incoming-webhooks.textile
```diff
@@ -0,0 +1,104 @@
...
+
+Your code to receive the named message is:
+
+```[nodejs]
+channel.subscribe('webhook-message', function(message) {
+    console.log("webhook: " + JSON.stringify(message.data));
+});
+```
+
+h2(#summary). Summary
```
paddybyers, line 94: I think this is missing some more general information about how these requests will be handled. For example, what happens if it succeeds, and what happens if it fails?

I think it's also worth saying generally that incoming webhooks are just REST publishes, and so they will reflect the behaviour of the REST publish API. (I realise there's a link to that documentation but I think it should make that connection explicitly.) Therefore a successful publish will respond with a 201 statusCode; failure will result in a 4xx or 5xx code and an `ErrorInfo` response body.

Rate-limiting will be applied as with REST publishes.

Idempotent publishing is possible by including a unique message identifier in an `X-Ably-MessageId` header.

content/general/incoming-webhooks.textile
```diff
@@ -0,0 +1,104 @@
...
+```[sh]
+curl -X POST 'https://rest.ably.io/channels/webhook-test/messages?key=key:secret&enveloped=false' \
+     -H 'content-type: application/json' --data '{"some":"json"}' \
+     -H 'X-Ably-Name: webhook-message'
+```
+
+Your code to receive the named message is:
+
+```[nodejs]
+channel.subscribe('webhook-message', function(message) {
```
paddybyers, line 89: this should use arrow syntax

content/general/incoming-webhooks.textile
```diff
@@ -0,0 +1,104 @@
...
+You can receive the incoming webhook as you would a normal Ably message on a channel. For example, if the incoming webhook data is in JSON format, you can display the data using the following code:
+
+```[nodejs]
+var ably = new require("ably").Realtime('{{API_KEY_NAME}}:{{API_KEY_SECRET}}');
+
+/* Obtain the channel that has been configured to receive webhooks */
+var channel = ably.channels.get('webhook-test');
+
+/* Incoming webhook messages can be displayed using this function */
+channel.subscribe(function(message) {
```
paddybyers, line 69: This should be arrow syntax

paddybyers (commented):
A few comments/suggestions

ginotoro (approved):

tbedford:
@ginotoro - it looks like Integrations tab and Incoming webhooks UI is now live in the Dashboard, so unless you have any other issues I will merge this. Thanks.

PR #1024, opened 2021-01-27
Add maintainers file

niksilver:
Add maintainers file

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,0 +1,1 @@
+This repository is owned by the Ably Devrel team.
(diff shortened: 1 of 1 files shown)


tbedford:
Is this the case? I thought it might be owned by product @kavalerov 

Thought I would check before merging...

kavalerov:
@tbedford agreed, and I think it is even more precise than that - it is Docs part of the product team

kavalerov:
@tbedford agreed, and I think it is even more precise than that - it is Docs part of the product team

tbedford (approved):

PR #1034, opened 2021-02-10
Feature spec: add Ably-Agent header

owenpearson:
Will help us to make more informed decisions about whether to drop support for older versions of a platform.


content/client-lib-development-guide/features.textile
```diff
@@ -80,6 +80,7 @@ h3(#restclient). RestClient
...
 ** @(RSC7c)@ If the @addRequestIds@ client option is enabled, every REST request to Ably should include in a @request_id@ query string parameter a random string obtained by url-safe base64-encoding a sequence of at least 9 bytes obtained from a source of randomness. This request ID must remain the same if a request is retried to a fallback host per @RSC15@. Any log messages associated with the request should include the request ID. If the request fails, the request ID must be included in the @ErrorInfo@ returned to the user.
+** @(RSC7d)@ Where possible, the header @X-Ably-Platform-Version: [lib][.optional variant]?-[version]@ should be included in all REST requests to the Ably endpoint where @[lib]@ is the name of the library such as @js@ for @ably-js@, @[.optional variant]@ is an optional library variant, such as @laravel@ for the @php@ library, which is always delimited with a period such as @php.laravel@, and where @[version]@ is a string representing the version of the library's runtime environment. For example, in NodeJS this would be the output of @process.version@.
```
lmars, line 83: I wonder if we should have a more flexible format that supports a list of versions? For example, could adopt the User-Agent style `<product>/<product-version>` format from https://tools.ietf.org/html/rfc7231#section-5.5.3:

- `X-Ably-Lib: ably-js/1.2.3 Node.js/14.15.5`
- `X-Ably-Lib: ably-go/1.2.1 Go/1.15.8`

QuintinWillison: I believe @paddybyers had some concerns about making this too complicated and/or trying to be too generic - I think grounded in the real world experience of `User-Agent` misuse / masquerade by "browsers".

However, given it'll be our client library code generating these strings then I would be inclined to go with @lmars' suggestion... which would allow us to point to a standard rather than inventing our own.

SimonWoolf: AFAICT the proposed `X-Ably-Platform-Version` has two parts:
1. The lib/variant part
 
This is redundant to x-ably-lib, RSC7b

2. The platform/runtime part

Surely this is exactly what the `User-Agent` header is for, per the relevant RFC?

Instead of "adopting the user-agent style" why not just start sending a `user-agent` header for non-browsers? There's nothing ably-specific about sending the platform version.

An advantage of doing it this way is that it means non-browsers present the same header set as browsers. Like, there's not much point copying the browser user-agent into the second part of `x-ably-lib`, it's redundant, we'll get the user-agent header too.

And it's true that user-agent is a mess, but, well, it's not like there's a secret 'navigator.actualPlatformVersion' that we can use to populate some new thing; the user-agent, as imperfect as it, is what there is. (Well, there's chrome's new `Sec-CH-UA` user-agent-but-starting-fresh thing, but that's its own header (that the server needs to opt into with Accept to be sent), so it would still be redundant to copy that to another ably-specific header)

paddybyers: I think we should try to get the best available information, and include it in a header that permits identification of the multiple layers at play as per Lewis' example. In the flutter client, for example, we would want to include the Flutter SDK version; the native (ie cocoa/android) plugin version; the OS version. Should we send that, in the cases that there's no platform-provided `User-Agent`, as `X-Ably-User-Agent`? And then we can take the value of `X-Ably-User-Agent` if it's there, or `User-Agent` if it's not, and process those consistently?

owenpearson: I definitely agree with the idea of multiple layers of identification, but is anyone able to comment on how easy it would be in this case to get statistics about a single 'layer'. At the moment in Grafana I can, for example, filter results by 'js' clients and quickly get a breakdown of the versions of ably-js which are being used in web browsers. I don't know anything about how that data gets from InfluxDB in to Grafana but would it be possible to add some kind of middleware to parse `User-Agent` style strings from requests into useful data so that I could, for example, see a breakdown of Flutter SDK versions without caring about what OS versions they're associated with.

lmars: > I don't know anything about how that data gets from InfluxDB in to Grafana

If you squint hard enough, you'll see it's handled by this code in the realtime router:

https://github.com/ably/go-services/blob/main/router/request_handler.go#L546-L596

So it:

- reads the lib version from either the `X-Ably-Lib` header or the `lib` query param
- splits it into two parts at the first hyphen (fails if there is no hyphen present)
- records HTTP and WS metrics tagged with `agent=<parts[0]>` and `version=<parts[1]>`

> would it be possible to add some kind of middleware to parse User-Agent style strings from requests into useful data so that I could, for example, see a breakdown of Flutter SDK versions without caring about what OS versions they're associated with.

Absolutely yes, whatever we decide here, we can extract in the router and emit metrics appropriately.

lmars: Re-reading the comments so far, it seems there's agreement to go with the user-agent style formatting, and to the question of whether it should be `X-Ably-Lib` or `User-Agent` in both browsers and non-browsers, I'm pretty indifferent really, although I do think we should standardise on one or the other being the source of information about the various versions we want to run analytics queries on.

If I were to be pushed for a decision, I'd probably stick with `X-Ably-Lib` and the `lib` query param, but if we went for `User-Agent` and the `ua` query param instead, I wouldn't lose any sleep 🙂

paddybyers: OK, so it sounds like there's consensus on there being a single string value, that contains multiple components, and the concatenation of those components follows the `ua` convention.

If we send it as a query param, this can be `lib`.

I'm willing to go with a decision that it's consistent across all libraries, and not either `ua` or `x-ably-lib`.

Then it's just a choice about whether it's `ua` or `x-ably-lib`. The pros and cons we see are:

a) `x-ably-lib` we can always control what it contains, and there isn't risk of conflict with other usages of `ua`;
b) using `x-ably-lib` we will duplicate information vs `ua`, which will make every request that bit larger;
c) on some platforms we don't have access to all the platform information we would like to include in `x-ably-lib`, so that info will be omitted.

If we go with `x-ably-lib`, can't we just have something in the router that appends some or all of the `ua` details to that header, so then in our logs (and anything downstream of the router) it's all complete and consistent?

And do we really prefer `x-ably-lib` to `x-ably-user-agent` ?

QuintinWillison: While I fully support adopting the same standard that's used for user agent strings (and we can reference them as our inspiration), I would avoid using `-User-Agent` / `ua` or similar in the definition of our protocol, especially on reading [section 5.5.3 of RFC 7231](https://tools.ietf.org/html/rfc7231#section-5.5.3) as it could be confusing.

My preference would be `X-Ably-Lib`.

niksilver: I'm fine with this conclusion, including `X-Ably-Lib`.

There's only one bit I now have a view on:

> If we go with x-ably-lib, can't we just have something in the router that appends some or all of the ua details to that header, so then in our logs (and anything downstream of the router) it's all complete and consistent?

My preference is that if a header is created in one place then its content is preserved, and remains untouched. Otherwise it's potentially problematic debugging the contents - you end up having to look in two places... and you need to know to look in two places. If you forget (or don't realise) you need to look in two places then it'll be really confusing. For me that's too likely to cause problems when troubleshooting.

However, if different components want to add their own headers, that seems fine to me.

owenpearson: Straw poll for what the header should be called:
React with 🎉   for `Ably-Client` and the `client` query string param (bear in mind this may cause minor confusion alongside the `clientId` param)
React with 🚀   for `Ably-Agent` and the `agent` query string param.

niksilver (commented):
One comment and one question+comment...

A comment, regarding "Please note this header has now been superseded...". If RCS7b is superseded by RCS7b then I'd like to be told that at the start of the paragraph, rather than at the end after I've slogged through some quite difficult text.  So for me, ideally, it would begin:

> (RSC7b) (Please note this clause and the associated header has now been superseded by RCS7d.) The header X-Ably-Lib: [lib][.optional variant]?-[version] should be included...

A question+comment. If RSC7b is no longer relevant, and is superseded by RCS7d, then what happens to the recommendation in RSC7b1? Presumably we also need a correspondng recommendation for RSC7d, which I imagine would be RSC7d1.

owenpearson:
Thanks @niksilver, those are valid points and I've addressed them both in de54665.

niksilver:
Okay, now I'm going to be really picky...

* "...this clause and the associated header has now been superseded..." should be "...this clause and the associated header have now been superseded...". ("have" not "had".) And I realise you just copied my error, so apologies for not getting that right first time.
* RSC7d1 makes it look as though you can choose any name you like for the query param. And although I don't think anyone will really do that I do think we should make it clear. So the text "...should be sent as a query param such as..." should be "...should be sent as an @agent@ query param such as...". (But check my markdown syntax, and do check that it reads well as a whole.)

content/client-lib-development-guide/features.textile
```diff
@@ -77,9 +77,11 @@ h3(#restclient). RestClient
...
+** @(RSC7d)@ Where possible, the header @Ably-Agent@ should be included in all REST requests to the Ably endpoint and contain a series of products and associated versions in the format described by "RFC 7321 Section 5.5.3":https://tools.ietf.org/html/rfc7231#section-5.5.3 (ie a list of <product>/<product-version> joined by spaces). The products described in this header should include the SDK itself and any relevant runtime platforms, for example ably-js version 1.2.8 running on NodeJS 14.16.0 should send "ably-js/1.2.8 NodeJS/14.16.0".
+*** @(RSC7d1)@ When it is not possible to send the @Ably-Agent@ header, such as for @JSONP@ requests, the library version should be sent as a query param such as @agent=ably-js/1.2.8 NodeJS/14.16.9@ (in practice this query param should also be URL escaped).
```
paddybyers, line 84: I think we can restructure this to be (fill in blanks):

```
@(RSC7d)@ An @Agent@ library identifier is used to identify the library type and version, together with relevant information describing any underlying library layers and the client platform.

@(RSC7d1)@ In general the @Agent@ library identifier is composed of a series of @key/value@ entries joined by spaces as defined in "....". This must include at least the @<library name>/<library version>@ entry followed optionally, in dependency order, the @name/version@ entry for underlying library layer or platform layers. An example would be @ably-flutter/1.2.0 ably-java 1.2.1 android/24@. The documentation for each library must define the content and meaning of the @Agent@ identifiers advertised by that library.

@(RSC7d2)@ Where possible, the @Agent@ library identifier should be included in all REST requests to the Ably endpoint via an @Ably-Agent@ HTTP header.

@(RSC7d3)@ Where the library is unable to specify a header when making a request (such as for @JSONP@ requests), the @Agent@ library identifier should be sent via an @agent@ query param whose value is the URI-encoded @Agent@ library identifier.
```

paddybyers (commented):
I've suggested restructuring the requirement to separate the definition of the content of the identifier from the ways of sending the identifier.

Shouldn't there be something added that says what happens for realtime connection requests?

owenpearson:
> Shouldn't there be something added that says what happens for realtime connection requests?

@paddybyers by this do you mean WebSocket handshake requests? If so, are you aware of anywhere that this is specified for the existing `X-Ably-Lib` header? And would it be an adequate solution to change your suggestion from "Where possible, the @Agent@ library identifier should be included in all REST requests to the Ably endpoint" to "Where possible, the @Agent@ library identifier should be included in _all HTTP requests made by the library_"?

paddybyers:
> by this do you mean WebSocket handshake requests?

Yes

> If so, are you aware of anywhere that this is specified for the existing `X-Ably-Lib` header?

No, the websocket API doesn't allow you to specify headers, so libraries use the `lib` param. I've checked the spec and I can't see where it says to include `lib` in ws connection URL, but that's definitely a requirement.

> And would it be an adequate solution to change your suggestion from "Where possible, the @agent@ library identifier should be included in all REST requests to the Ably endpoint" to "Where possible, the @agent@ library identifier should be included in _all HTTP requests made by the library_"?

There is a structural problem with the spec in that there's repetition between the REST and realtime sections. Based on the present structure, there should be a new analogous requirement in the realtime bit. However, I don't really think we should be proliferating this duplication, and rationalising that should be a separate PR from this one. So yes, I'd support extending this requirement to also apply to websocket connection requests. Maybe it should be "...included in all HTTP and Websocket requests made by the library".

I think there should be an issue raised to restructure the spec to put common definitions and requirements into a separate section.

owenpearson:
@paddybyers Thanks for the feedback, I've added a commit to change to pretty much exactly what you suggested. I've also raised [an issue](https://github.com/ably/docs/issues/1069) for your point about common definitions/requirements.

niksilver (approved):
Looks good to me.

paddybyers (approved):
LGTM, thanks

content/client-lib-development-guide/features.textile
```diff
@@ -77,9 +77,13 @@ h3(#restclient). RestClient
...
+** @(RSC7d)@ An @Agent@ library identifier is used to identify the library type and version, as well as relevant information describing any underlying layers and the client platform.
+*** @(RSC7d1)@ In general, the @Agent@ library identifier is composed of a series of @key/value@ entries joined by spaces as described in described by "RFC 7321 Section 5.5.3":https://tools.ietf.org/html/rfc7231#section-5.5.3. This must include at least the @<library name>/<library version>@ entry followed optionally, in dependency order, the @name/version@ entry for underlying library layer or platform layers. An example would be @ably-flutter/1.2.0 ably-java 1.2.1 android/24@. The documentation for each library must define the content and meaning of the @Agent@ identifiers advertised by that library.
```
SimonWoolf, line 84: > ably-flutter/1.2.0 ably-java 1.2.1 android/24

Missing a forwardslash in ably-java?

SimonWoolf (approved):

content/client-lib-development-guide/features.textile
owenpearson: good spot, thanks. fixed in ab565f6.

Issue #1041, opened 2021-02-18
Proposal for spec tweak wrt receiving an ATTACHED while detaching


SimonWoolf:
Unlike receiving a DETACHED while attaching, which is specified by RTL13, the spec does not define the behaviour if the client library receives an ATTACHED while detaching.

All we have is the spec for detach (`(RTL5d) Otherwise a DETACH ProtocolMessage is sent to the server, the state transitions to DETACHING and the channel becomes DETACHED when the confirmation DETACHED ProtocolMessage is received`), and the normal spec for what happens when the client receives an `ATTACHED`, which is in the section for `channel#attach()` and is not clear on whether it applies only in that context or at any time.

Paddy's answer when this sort of question has come up in the past has been ["the server sends the definitive channel state"](https://github.com/ably/ably-js/pull/324#discussion_r85799402); the client adopts whatever channel state the server tells it it's in. Ably-js is written accordingly: when the client receives an ATTACHED while detaching, it obediently goes into the `attached` state.

The problem with this is that the channel state is overloaded; it represents both a reflection of the server state, as well as a reflection of the intentions of the user. Hence the existence of pending states, and the `suspended` state (for 'I'm detached and not currently attaching but the user wants me to eventually be attached'), and the automatic re-attach when you're attached and there's a server-initiated detach.

This last one is the one that's causing trouble here. Have the following sequence of events:
- user calls attach(), client sends ATTACH, client state `attaching`
- user calls detach(), client sends DETACH, client state `detaching`
- client receives ATTACHED from server, client state `attached`
- client receives DETACHED from server, client sends an ATTACH per RTL13, client state `attaching`
- client receives ATTACHED from server, client state `attached`.

...So the channel ends up attached, despite that the last thing the user did was signal their intention to detach the channel.

Additional side-problem: the detach() callback is called when the client receives the first ATTACHED, it doesn't know to wait for the next channel state message.

(additional side problem on top of the side problem: not populating the callback error when this happens)

We know the solution to this - we [decided it back in 2017](https://github.com/ably/ideas/issues/160#issuecomment-282755252).  It's to add ids to client-initiated ATTACH/DETACH protocol messages and reflect them back in the corresponding ATTACHED/DETACHED, so that the client library can properly pipeline requests. Then it would know that when it gets the first ATTACHED that the DETACH is still pending, and keep the channel state as `detaching`. And it would know not to call the callback of the `detach` until it gets a detached with the right id (or something that overrides, like a fatal error).

[...]

PR #1044, opened 2021-02-23
Spec: add RTL5i (attached received while in the detaching state)

SimonWoolf:
As proposed by https://github.com/ably/docs/issues/1041, see that issue for more details and justification.

Already approved by Paddy in principle, this just implements it.


SimonWoolf:
@QuintinWillison poke

tbedford:
I've not looked into the specs much yet, but is this error "RTL5i has 1 duplicate(s)" something that needs to be looked into I wonder?

SimonWoolf:
> I've not looked into the specs much yet, but is this error "RTL5i has 1 duplicate(s)" something that needs to be looked into I wonder?

Thanks, missed that, fixed

paddybyers:
@QuintinWillison this spec change was implemented in ably-js but not the other libs. How do you intend to track this? Should issues be raised against the other libs?

QuintinWillison:
Thanks, @paddybyers ... sorry for the glacial delay, but issues now created in the realtime-capable SDK repositories ☝️.

PR #1049, opened 2021-03-10
Architectural patterns to implement popular use-cases with Ably (wip)

Srushtika:
This PR updates existing and adds new architectural patterns to make it easier for clients to start using Ably effectively for popular realtime use-cases.


kavalerov:
As this is a public-facing repo, and "goat" is an internal process - should we change the title of the PR to be more friendly for external observers and contributors? I.e. "Add patters pages to make it easier for clients to start using Ably effectively"

PR #1062, opened 2021-03-23
Define behaviour when device secret is lost

cruickshankpg:
I've tried to tighten up the language around the device ID and secret to make it clear they are a pair.


content/client-lib-development-guide/features.textile
```diff
@@ -1000,7 +1000,7 @@ h3(#activation-state-machine). Activation State Machine
...
 ***** @(RSH3a2a4)@ Transitions to @WaitingForRegistrationSync@.
-**** @(RSH3a2b)@ If the local device does not have @id@ and @deviceSecret@, both are generated locally. The @id@ must be a "ulid":https://github.com/ulid/spec or similar globally-unique identifier. The @deviceSecret@ must be created using secure random data with sufficient entropy to generate a digest of at least 32 bytes (eg using sha256) and encoding that digest with base64. The local @DeviceDetails@ is updated with the resulting @deviceId@ and @deviceSecret@.
+**** @(RSH3a2b)@ If the local device does not have @id@ or @deviceSecret@, both are generated locally. The @id@ must be a "ulid":https://github.com/ulid/spec or similar globally-unique identifier. The @deviceSecret@ must be created using secure random data with sufficient entropy to generate a digest of at least 32 bytes (eg using sha256) and encoding that digest with base64. The local @DeviceDetails@ is updated with the resulting @deviceId@ and @deviceSecret@. If either the @id@ or the @deviceSecret@ is lost then a new pair must be created.
```
paddybyers, line 1003: I think there needs to be an explicit statement in RSH8a, or in RSH8 as a separate requirement, stating that if during initialisation of the activation state machine the retrieval of existing `LocalDevice` details fails, then the machine should transition to `NotActivated` (which will result in new local device details being created on the next activation event).

paddybyers (changes requested):
See comment

content/client-lib-development-guide/features.textile
```diff
@@ -1110,6 +1110,7 @@ h3(#local-device). LocalDevice
...
 ** @(RSH8h)@ If an attempt to obtain the push transport details for local device (eg an FCM registration token) fails, a @GettingPushDeviceDetailsFailed@ event containing the indicated error is sent to "the state machine":#RSH3.
 ** @(RSH8i)@ Each time the library is instanced, if the LocalDevice has push device details (eg an APNS deviceToken), and if the platform supports it, it must verify the validity of those details (eg by requesting a token from the platform and comparing that with the already-known token). If as a result there are updated details, then an update to the Ably server is triggered by sending a @GotPushDeviceDetails@ event to "the state machine":#RSH3.
+** @(RSH8j)@ If during library initialisation the @LocalDevice@ @id@ or @deviceSecret@ attributes are not able to be loaded then the ActivationStateMachine machine should transition to the @NotActivated@ state. New @LocalDevice@ @id@ and @deviceSecret@ attributes should be generated on the next activation event.
```
paddybyers, line 1113: .... are not able to be loaded then **those LocalDevice details must be discarded and** the ...

paddybyers (approved):
ok if you make the suggested change, thanks

PR #1064, opened 2021-03-26
Asset Tracking SDK overview

tbedford:
## Description

Initial asset tracking overview docs.

Pages to review:

* [Overview](https://ably-docs-pr-1064.herokuapp.com/asset-tracking/)
* [Using the example apps](https://ably-docs-pr-1064.herokuapp.com/asset-tracking/example-apps/)
* [Using the SDKs](https://ably-docs-pr-1064.herokuapp.com/asset-tracking/using-the-sdks/)


content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+In a typical asset tracking scenario the key participants are:
+
+- asset (trackable) := Item that the customer wants to track such as driver, vehicle, container, and so on. This also known as a @Trackable@ in the SDKs.
+- driver := Person driving the asset physically, or entity controlling the asset programatically.
+- customer := Entity that uses the Ably Asset Tracking SDKs to build their product to solve the asset tracking problem for their users.
+- end user := Person that uses the product built by the customer and uses the asset tracking capabilities.
+
+Each asset being tracked must have a @trackableId@, and this must be globally unique within an Ably asset tracking account.
+
+The Ably Asset Tracking SDKs also allow for various battery optimization scenarios, so frequency of updates can be modified based on various parameters such as distance to destination, subscriber status, map zoom level, and existing battery levels.
```
kavalerov, line 49: "subscriber preferences" instead of "subscriber status"

kavalerov: Should it be just "battery levels" instead of "existing battery levels"?

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+
+h2(#battery-optimization). Battery optimization scenarios
+
+One of the biggest challenges in asset tracking is combining high resolution of updates with low battery consumption on the asset tracking device. For example, in a delivery tracking use case you may want your users to have a smooth and realtime visualization of where their delivery is on the map, but at the same time sending realtime locations from the driver's device frequently will rapidly drain its battery.
```
kavalerov, line 53: "realtime locations from the driver's device frequently" - I am not sure having both "realtime" and "frequently" in one sentence makes sense

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+
+There are two key factors that impact asset tracking device battery usage:
+
+1. GPS sensor - obtaining location of the asset.
+2. Networking - sending location updates over network.
+
+It is hard to reduce the battery consumption of the GPS sensor in some situations, such as when the device is being used for navigation at the same time as tracking an asset. However, it is possible to adjust the networking aspect of resolution.
```
kavalerov, line 60: > However, it is possible to adjust the networking aspect of resolution.

Probably this was copied from the doc that I wrote, but I don't think "networking aspect of resolution" makes sense. I think it should just say "networking aspect", or "networking frequency".

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+
+The following scenarios are examples of how update resolution can be changed depending on the state of the asset and subscribers:
+
+* Updates can be sent to the subscriber more often when asset is closer to destination.
+* Updates don't need to happen as often when the user of the subscribing SDK is not observing the location of the asset, for example, the user of delivery app is not checking the app for the driver's location.
```
kavalerov, line 65: "don't need to be sent" instead of "don't need to happen"

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+* Different subscribers may have different preferences regarding the resolution of location updates.
+* When the driver's battery is running low, location updates can happen with lower frequency.
+
+Using the approaches similar to these, an optimal resolution of updates can be identified, which provides a good user experience when it is needed, such as when the driver is close to the destination, and the user is observing the location of the driver in the app. This approach also reduces battery usage when possible, for example when the driver is far from the destination and the user is not observing location updates.
+
+While developers can create their own code to implement battery optimization logic, Ably also provides a set of default constraints and policies, which are suitable for common use cases.
```
kavalerov, line 71: Should we emphasise that Ably provides an extensible interface for building custom logic? Right now it sounds like the emphasis is on the default implementations, but I would argue that both are important: default implementation covers majority of basic use-cases, but SDKs are also very flexible and extensible because of the way we've built them, this extensibility is a feature.

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+- @desiredInterval@ := Desired time between updates, in milliseconds. Lowering this value increases the temporal resolution. Location updates whose timestamp differs from the last captured update timestamp by less that this value are to be filtered out. Used to govern the frequency of updates requested from the underlying location provider, as well as the frequency of messages broadcast to subscribers.
+- @minimumDisplacement@ := Minimum positional granularity required, in metres. Lowering this value increases the spatial resolution. Location updates whose position differs from the last known position by a distance smaller than this value are to be filtered out. Used to configure the underlying location provider, as well as to filter the broadcast of updates to subscribers.
+
+h2(#resolution-set). Resolution set
+
+A resolution set, is a set of resolutions that must be provided by the app in order to inform the resolution policy. While developers can define their own resolution sets, Ably also provides a default resolution set that can be used for common use cases.
```
kavalerov, line 85: Ably provides a default resolution set implementation not just to be used with common use-cases, but specifically for use with default resolution policy.

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+    batteryLevelThreshold = 10.0f,
+    lowBatteryMultiplier = 2.0f
+)
+```
+
+h2(#resolution-policy). Resolution policy
+
+Resolution policy defines the strategy by which the various @ResolutionRequests@ and preferences are translated by @Publisher@ instances into a target @Resolution@.
+
+As different use cases have different requirements for battery optimization, Ably Asset Tracking SDKs provide a flexible @ResolutionPolicy@ interface, that allows SDK users to provide their own implementation of resolution optimization logic. Ably also provides a "default resolution policy":#default-resolution-policy, which is suitable for common use cases.
```
kavalerov, line 133: > "default resolution policy"

Should we just write "default resolution policy implementation" (without the quotemarks)?

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,232 @@
...
+    - See also#see-also
+---
+
+This topic explains how to get started using the Ably Asset Tracking SDKs.
+
+h2(#supported-platforms). Supported platforms
+
+There are two Asset Tracking SDKs, one for publishing and one for subscribing. The following platforms are supported:
+
+* Android (Java and Kotlin) - publisher and subscriber SDK
```
kavalerov, line 30: The terminology that we use at the moment is "publishing" and "subscribing" SDKs, not "publisher" and "subscriber" (@QuintinWillison correct me if I am wrong)

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,232 @@
...
+* JavaScript - any suitable environment of your choice 
+
+You also need to have suitable credentials for the various SDK components:
+
+* @ABLY_API_KEY@ - Your Ably API key
+* @MAPBOX_DOWNLOADS_TOKEN@ - Mapbox credentials
+* @MAPBOX_ACCESS_TOKEN@ - Mapbox credentials
+* @GOOGLE_MAPS_API_KEY@ - Google Maps credentials
+
+You can set these values in your @~/.gradle/gradle.properties@ file.
```
kavalerov, line 57: This is specific to Android, this line is not applicable to iOS and Javascript

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,232 @@
...
+try {
+    publisher.track(
+        Trackable(
+            trackingId, // provide a tracking identifier for the asset
+            constraints = exampleConstraints // provide a set of Resolution Constraints
+        )
+    )
+    // TODO handle asset tracking started successfully
+} catch (exception: Exception) {
+    // TODO handle asset tracking could not be started
```
kavalerov, line 146: Should we clarify wording of these comments? I.e. Handle failure to start tracking a trackable

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,232 @@
...
+
+h3(#publisher-start-tracking). Start tracking
+
+You can start tracking an asset by calling the @track@ method of the publisher. You need to supply the tracking identifier of the asset to be tracked, and any "resolution constraints":/asset-tracking#resolution-constraints that need to be applied.
+
+The following code example demonstrates how to start tracking an asset:
+
+```[android]
+// Start tracking an asset
+try {
```
kavalerov, line 137: Does this try-catch statement provide any value as part of this code snippet? I am wondering if we should just keep the actual SDK call

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,232 @@
...
+    .start() // start listening for updates
+```
+
+h3(#subscriber-subscribe). Subscribe to updates
+
+You can subscribe to updates from the publisher, specifying a function that is called when each update is received. This is shown in the following example:
+
+```[android]
+// Listen for location updates
+locations
```
kavalerov, line 199: We still need to specify what this `locations` method is called on.

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,232 @@
...
+    .launchIn(scope) // coroutines scope on which the locations are received
+```
+
+h3(#subscriber-subscribe-asset-state-changes). Subscribe to asset state changes
+
+You can subscribe to asset state changes from the publisher, specifying a function that is called when each state change is received. This is shown in the following example:
+
+```[android]
+// Listen for asset state changes
+trackableStates
```
kavalerov, line 210: Same here as above - code snippet right now is incomplete as doesn't show what this is called on.

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,232 @@
...
+    .onEach { } // provide a function to be called when the asset changes its state
+    .launchIn(scope) // coroutines scope on which the statuses are received
+```
+
+h3(#subscriber-subscribe-asset-state-changes). Request a different resolution
+
+The subscriber can always request a different resolution preference by calling the @resolutionPreference@ method, passing in the required @Resolution@. This is shown in the following example:
```
kavalerov, line 217: I think it is important to explain somewhere why we have this feature. The use-case is simple - subscriber may require different resolutions of updates depending on state - i.e. wether map is displayed to the user, what zoom level is set on the map, etc. This feature allows subscriber to express different preferences depending on all these factors. 

Maybe this can be described on the index page?

kavalerov (changes requested):
I left some comments through the PR. In general I think this is a good structure, and we can iron things out and shape it into something we can merge soon :)

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+For the @DefaultResolutionSet@ the following resolutions are expected to be included:
+
+- @farWithoutSubscriber@ := The resolution to select if above the "@proximityThreshold@":#default-resolution-constraints), with no subscribers.
+- @farWithSubscriber@ := The resolution to select if above the "@proximityThreshold@":#default-resolution-constraints), with one or more subscribers.
+- @nearWithoutSubscriber@ := The resolution to select if below the "@proximityThreshold@":#default-resolution-constraints), with no subscribers.
+- @nearWithSubscriber@ := The resolution to select if below the "@proximityThreshold@":#default-resolution-constraints), with one or more subscribers.
+
+h2(#resolution-constraints). Resolution constraints
+
+Resolution constraints specify the thresholds and corresponding logical mappings for tracking @Resolution@s that are required by the default @ResolutionPolicy@, which can be adopted by @Publisher@ instances. While developers can define their own resolution constraints, Ably also provides default resolution constraints that can be used for common use cases.
```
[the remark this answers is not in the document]
KacperKluka: > While developers can define their own resolution constraints

That's not true at the moment, we have a sealed class `ResolutionConstraints` (it's like an abstract class with a fixed number of implementations) that has only one implementation `DefaultResolutionConstraints`. This doesn't allow the users to make their own versions of `ResolutionConstraints` :thinking: @kavalerov is that how it's supposed to work?

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+
+h3(#default-resolution-constraints). Default resolution constraints
+
+The Ably-provided default resolution constraints consist of:
+
+- @DefaultResolutionSet@ := The set of resolutions which must be defined in order to populate the @DefaultResolutionConstraints@.
+- @DefaultProximity@ := This is a distance to destination at which resolution of updates is increased to improve user experience.
```
KacperKluka, line 107: Proximity can be spatial, temporal or both. If the word `distance` means both distance in meters and the time required to arrive at some destination then it's fine (I'm not a native speaker so it isn't that clear to me :wink: )

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+- @batteryLevelThreshold@ := This is a level of battery on the device that runs Publishing SDK that is considered low. When the battery level falls below this, the resolution of updates is reduced. Default value: 10%. In the range 0.0f (no battery) to 100.0f (full battery).
```
KacperKluka, line 108: > Default value: 10%

We're not providing any default value for any of the `DefaultResolutionConstraints` fields. A user has to specify all of its fields in order to create it.

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+
+Resolution policy defines the strategy by which the various @ResolutionRequests@ and preferences are translated by @Publisher@ instances into a target @Resolution@.
+
+As different use cases have different requirements for battery optimization, Ably Asset Tracking SDKs provide a flexible @ResolutionPolicy@ interface, that allows SDK users to provide their own implementation of resolution optimization logic. Ably also provides a "default resolution policy":#default-resolution-policy, which is suitable for common use cases.
+
+h3(#default-resolution-policy). Default resolution policy
+
+Ably Asset Tracking SDKs come with a @DefaultResolutionPolicy@ implementation provided out of the box, which is suitable for common use cases.
+
+Each @Trackable@ added to the @Publisher@ has a destination and set of @ResolutionConstraints@. Ably provides a set of "default resolution constraints":#default-resolution-constraints, @DefaultResolutionContraints@.
```
KacperKluka, line 139: Both `destination` and `resolutionConstraints` are optional for `Trackable`, however, having them improves the resolution policy.

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+
+When the SDK is running, on each change of the device or trackable status, the @ResolutionPolicy@ recalculates which resolution should be applied. Example changes include: 
+
+* Subscribers added or removed from the list of subscribers.
+* Battery or proximity threshold hit.
+* Subscriber closes app.
+
+The following logic is used:
+
+1. When comparing resolutions, the more precise resolution is always selected.
```
KacperKluka, line 149: To be more precise we are comparing each field of the resolution (accuracy, desiredInterval, minimumDisplacement) and we're keeping the most precise value for each of those fields. So the final resolution could be a hybrid of all the provided resolutions. 
For example, it can have the accuracy of resolution A, the desiredInterval of resolution B and the minimumDisplacement of resolution C :wink:

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,232 @@
...
+The required methods are:
+
+- connection := Called to provide Ably connection information, such as API keys, and any other configuration parameters as needed. 
+- map := Called to provide Mapbox configuration, such as API keys, any other configuration parameters as needed. 
+- resolutionPolicy := Sets the policy factory to be used to define the target resolution for publishers created from this builder.
+
+The optional methods are:
+
+- profile := Called to set the means of transport being used for the initial state of publishers created from this builder. If not set then the default value is @RoutingProfile.DRIVING@.
+- context := Called to provide the Android runtime context (on Android only).
```
KacperKluka, line 89: This is a required method on Android and it's called `androidContext` :wink:

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,232 @@
...
+
+During initialization of the publisher various methods can be called to configure the @Publisher@.
+
+The required methods are:
+
+- connection := Called to provide Ably connection information, such as API keys, and any other configuration parameters as needed. 
+- map := Called to provide Mapbox configuration, such as API keys, any other configuration parameters as needed. 
+- resolutionPolicy := Sets the policy factory to be used to define the target resolution for publishers created from this builder.
+
+The optional methods are:
```
KacperKluka, line 86: One optional method is missing - `locationSource`

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,232 @@
...
+    .map(MapConfiguration(MAPBOX_ACCESS_TOKEN)) // provide Mapbox configuration with credentials
+    .androidContext(this) // provide Android runtime context
+    .resolutionPolicy(DefaultResolutionPolicyFactory(defaultResolution, this)) // provide either the default resolution policy factory or your custom implementation
+    .profile(RoutingProfile.DRIVING) // provide mode of transportation for better location enhancements
+    .start()
+```
+
+h3(#publisher-start-tracking). Start tracking
+
+You can start tracking an asset by calling the @track@ method of the publisher. You need to supply the tracking identifier of the asset to be tracked, and any "resolution constraints":/asset-tracking#resolution-constraints that need to be applied.
```
KacperKluka, line 131: The resolution constraints are optional, as a non-native speaker that's not clear from this sentence to me :wink:

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,232 @@
...
+// Prepare the default resolution for the Resolution Policy
+val defaultResolution = Resolution(Accuracy.BALANCED, desiredInterval = 1000L, minimumDisplacement = 1.0)
+```
+
+Once these are created you can then initialize the publisher with the constraints and default resolution, and start the publisher:
+
+```[android]
+// Initialise and Start the Publisher
+val publisher = Publisher.publishers() // get the Publisher builder in default state
+    .connection(ConnectionConfiguration(ABLY_API_KEY, CLIENT_ID)) // provide Ably configuration with credentials
```
KacperKluka, line 121: After landing of the [Token Auth PR]( https://github.com/ably/ably-asset-tracking-android/pull/222 ) this line will be obsolete

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,232 @@
...
+The optional methods are:
+
+- resolution := Request a specific resolution of updates to be requested from the remote publisher.
+
+The following code example demonstrates initializing and starting the subscriber:
+
+```[android]
+// Initialise and Start the Subscriber
+val subscriber = Subscriber.subscribers() // Get an AssetSubscriber
+    .connection(ConnectionConfiguration(ABLY_API_KEY, CLIENT_ID)) // provide Ably configuration with credentials
```
KacperKluka, line 185: After landing of the [Token Auth PR]( https://github.com/ably/ably-asset-tracking-android/pull/222 ) this line will be obsolete

KacperKluka (changes requested):
I've found some things to improve but overall the docs look good :+1: 

One more thing, for longer lines in the code examples, the text is sticking out of the code box :wink: 
Screenshot from 2021-04-19 09-41-46

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+  - java
+  - swift
+jump_to:
+  Help with:
+    - Key participants#key-participants
+    - Resolution policy#resolution-policy
+    - See also#see-also
+---
+
+The Ably asset tracking solution provides a way to track multiple assets in realtime. An asset is any object that moves geographically, and whose position needs to be monitored. The asset tracking solution is powered by the Ably realtime network, the Ably Asset Tracking SDKs, and the Mapbox Navigation SDK with location enhancement.
```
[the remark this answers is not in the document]
tbedford: > In the first sentence - probably makes sense to clarify "location of multiple assets".

Not sure where I say that! Also, asset is defined in the next sentence.

> should we use this naming here as well, instead of "Ably asset tracking solution"?

Yes we should! Fixed.

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+
+The Ably asset tracking solution provides two SDKs:
+
+* Publisher SDK - for embedding in apps on the asset to be tracked.
+* Subscriber SDK - for embedding in apps that want to observe the asset being tracked.
+
+As Ably is used as the underlying transport, you have direct access to your data and can use Ably integrations for a wide range of applications, in addition to direct realtime subscriptions. Examples include:
```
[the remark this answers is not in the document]
tbedford: Good spot! Fixed.

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+As Ably is used as the underlying transport, you have direct access to your data and can use Ably integrations for a wide range of applications, in addition to direct realtime subscriptions. Examples include: 
+
+* Passing data to a third-party service.
+* Persistence of data to a database for later retrieval.
+
+As the asset being tracked moves, it publishes its position through the app to an Ably channel. Any subscribers to the channel can then receive positional information, which can be displayed on a detailed map.
+
+The following diagram provides an overview of an asset tracking use case:
+
+<a href="/images/screenshots/asset-tracking/overview.png" target="_blank">
```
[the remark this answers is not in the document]
tbedford: Already done last week.

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+  <img src="/images/screenshots/asset-tracking/overview.png" style="width: 100%" alt="Overview">
+</a>
+
+In a typical asset tracking scenario the key participants are:
+
+- asset (trackable) := Item that the customer wants to track such as driver, vehicle, container, and so on. This also known as a @Trackable@ in the SDKs.
+- driver := Person driving the asset physically, or entity controlling the asset programatically.
```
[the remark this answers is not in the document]
tbedford: Yes it's vague. Changed to "Person driving the asset physically. Could also be a machine moving the asset under software direction, such as in the case of a delivery drone."

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+- customer := Entity that uses the Ably Asset Tracking SDKs to build their product to solve the asset tracking problem for their users.
+- end user := Person that uses the product built by the customer and uses the asset tracking capabilities.
+
+Each asset being tracked must have a @trackableId@, and this must be globally unique within an Ably asset tracking account.
```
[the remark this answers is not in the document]
tbedford: Yes, it should be discussed in SDK section. I also thought it was worth mentioning here too, as it's such an important concept.

content/asset-tracking/index.textile
tbedford: > Should it be just "battery levels" instead of "existing battery levels"?

I've changed to "current battery levels". The idea being if the battery level is low, resolution can be modified.

content/asset-tracking/index.textile
tbedford: Yes :) I removed realtime.

content/asset-tracking/index.textile
tbedford: Yes, I also wasn't too keen on it. I've changed to "adjust the frequency of updates".

content/asset-tracking/index.textile
tbedford: I've rewritten all those bullet points.

content/asset-tracking/index.textile
tbedford: Yes, I've reworded.

content/asset-tracking/example-apps.textile
```diff
@@ -0,0 +1,57 @@
...
+  <img src="/images/screenshots/asset-tracking/android/sdk-build-tools.png" style="width: 100%" alt="Installed SDK build tools">
+</a>
+
+3. Clone the Ably Asset Tracking SDK Android "GitHub repo":https://github.com/ably/ably-asset-tracking-android.
+4. Open the cloned directory in Android Studio.
+5. From the drop-down menu in Android Studio, select the app to run: @publishing-example-app@ and a device to run it on, then click Run.
+
+<a href="/images/screenshots/asset-tracking/android/select-app.png" target="_blank">
+  <img src="/images/screenshots/asset-tracking/android/select-app.png" style="width: 100%" alt="Select the publishing app">
+</a>
```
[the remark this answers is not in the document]
tbedford: Oops. Changed screenshot.

content/asset-tracking/index.textile
tbedford: Fixed.

content/asset-tracking/index.textile
tbedford: > Should we just write "default resolution policy implementation"

I couldn't find that text. However, there is a Textile link for default resolution policy, which uses quotes for link syntax.

content/asset-tracking/using-the-sdks.textile
tbedford: Clarified.

content/asset-tracking/using-the-sdks.textile
tbedford: Can do. Removed try-catch to simplify code, and also in the example further down.

content/asset-tracking/using-the-sdks.textile
tbedford: Removed try-catch

content/asset-tracking/using-the-sdks.textile
tbedford: Thanks for clarifying. Also changed in overview.

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,157 @@
...
+h3(#default-resolution-set). Default resolution set
+
+The @DefaultResolutionSet@ is the set of resolutions which must be defined in order to satisfy the "default resolution constraints":#default-resolution-constraints. These are required in order to inform the "default resolution policy":#default-resolution-policy.
+
+For the @DefaultResolutionSet@ the following resolutions are expected to be included:
+
+- @farWithoutSubscriber@ := The resolution to select if above the "@proximityThreshold@":#default-resolution-constraints), with no subscribers.
+- @farWithSubscriber@ := The resolution to select if above the "@proximityThreshold@":#default-resolution-constraints), with one or more subscribers.
+- @nearWithoutSubscriber@ := The resolution to select if below the "@proximityThreshold@":#default-resolution-constraints), with no subscribers.
+- @nearWithSubscriber@ := The resolution to select if below the "@proximityThreshold@":#default-resolution-constraints), with one or more subscribers.
```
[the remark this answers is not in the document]
tbedford: Good spot, thanks. Fixed.

content/asset-tracking/index.textile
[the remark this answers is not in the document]
tbedford: Waiting for input from @kavalerov

content/asset-tracking/index.textile
tbedford: You are quite correct - distance implies distance in metres, that is spatial only. I've hopefully clarified it can be spatial, temporal, or both.

content/asset-tracking/using-the-sdks.textile
tbedford: Fixed!

content/asset-tracking/index.textile
tbedford: Removed.

tbedford:
On the code lines overrunning the code box, this only happens for the staging site, and not on the actual documentation website. Tooling issue that we are working to resolve.

content/asset-tracking/example-apps.textile
```diff
@@ -0,0 +1,57 @@
...
+
+The app builds and the emulator is displayed. It takes a while to boot up.
+
+6. In the emulator, use the Extended Tools to create a route.
```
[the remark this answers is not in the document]
tbedford: > Shouldn't we add here that there are many ways of providing the device's location? Using the tools for mocking GPS is only one of them, the other would be using an Ably channel or a history file from S3.

So, to be clear, it's possible for the examples apps to use these other data sources?

> I'd also add that they need to use the Extended Tools to create a route in order to simulate the GPS readings for the emulator, or is it obvious and not worth mentioning?

I did say in step 6: "In the emulator, use the Extended Tools to create a route." Is there something else I should add here do you think?

content/asset-tracking/index.textile
tbedford: Fixed.

content/asset-tracking/index.textile
tbedford: Useful info added. Fixed.

content/asset-tracking/example-apps.textile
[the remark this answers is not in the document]
KacperKluka: > So, to be clear, it's possible for the examples apps to use these other data sources?

Yes, when you click on the settings button on the main screen in the Publisher example app you will see the settings screen. On that screen, you can select the "Location source" to be either Phone (GPS from the device) or Ably (an Ably channel that sends location updates) or S3 (a location history file downloaded from AWS S3 that replays the location data).

> I did say in step 6: "In the emulator, use the Extended Tools to create a route." Is there something else I should add here do you think?

Maybe we could append something like this "[...] to create a route that will simulate the GPS movement of the publisher" ? :thinking: However, I'm not convinced if that's better than it was so I'll leave it up to you to decide :wink:

content/asset-tracking/using-the-sdks.textile
tbedford: Fixed. `locationSource` added to optional methods.

content/asset-tracking/using-the-sdks.textile
tbedford: Quite right. Fixed.

content/asset-tracking/using-the-sdks.textile
tbedford: @KacperKluka - cool, if possible can you let me know what this will be replaced with and I will update accordingly. Thanks.

content/asset-tracking/using-the-sdks.textile
tbedford: @KacperKluka - cool, if possible can you let me know what this will be replaced with and I will update accordingly. Thanks.

content/asset-tracking/using-the-sdks.textile
tbedford: Yes, I don't understand this bit - could do with some advice if possible.

content/asset-tracking/using-the-sdks.textile
tbedford: Yes, I don't understand this bit - could do with some advice if possible. @KacperKluka

content/asset-tracking/using-the-sdks.textile
KacperKluka: I think that it should be `subscriber.locations` :wink:

content/asset-tracking/using-the-sdks.textile
KacperKluka: Like above, it should be `subscriber.trackableStates` :wink:

content/asset-tracking/using-the-sdks.textile
tbedford: Fixed.

content/asset-tracking/example-apps.textile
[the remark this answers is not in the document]
tbedford: I've reworked this a fair bit, and added a couple of new screen shots to hopefully make this clearer.

content/asset-tracking/using-the-sdks.textile
tbedford: Fixed.

content/asset-tracking/using-the-sdks.textile
tbedford: Fixed.

content/asset-tracking/using-the-sdks.textile
tbedford: I've added a new section in the overview. 

> "depending on all these factors."

Do we have a complete list of these factors anywhere?

content/asset-tracking/index.textile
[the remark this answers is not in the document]
tbedford: I've removed the untrue statement. Leaving this comment unresolved for now.

content/asset-tracking/using-the-sdks.textile
tbedford: After digging around in the PR I've replaced the above code with what I think is going to be the new code:

```
// Initialise and Start the Subscriber
val subscriber = Subscriber.subscribers() // Get an AssetSubscriber
    .connection(
                ConnectionConfiguration(
                    Authentication.basic(
                        CLIENT_ID,
                        ABLY_API_KEY,
                    )
                )
            ) // provide Ably configuration with credentials
    .resolution( // request a specific resolution to be considered by the publisher
        Resolution(Accuracy.MAXIMUM, desiredInterval = 1000L, minimumDisplacement = 1.0)
    )
    .trackingId(trackingId) // provide the tracking identifier for the asset that needs to be tracked
    .start() // start listening for updates
```

content/asset-tracking/index.textile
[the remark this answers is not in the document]
tbedford: Fixed. Clarified wording.

content/asset-tracking/example-apps.textile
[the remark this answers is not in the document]
tbedford: Updated before I saw your comments @kavalerov - however, I don't have a real Android device so I just tested with the emulator.

content/asset-tracking/using-the-sdks.textile
KacperKluka: Right, we should replace 
``` kotlin
.connection(ConnectionConfiguration(ABLY_API_KEY, CLIENT_ID))
```
with
``` kotlin
.connection(ConnectionConfiguration(Authentication.basic(ABLY_API_KEY, CLIENT_ID)))
```

tbedford: Interesting, in the code I looked at last night API_KEY and CLIENT_ID were the other way around. I assume your comment is correct @KacperKluka ?

KacperKluka: Good catch, sorry for not paying enough attention to the order of the params, I've focused on adding the `Authentication.basic` and overlooked that the params are now passed in a different order :sweat_smile: So the correct version is
``` kotlin
.connection(ConnectionConfiguration(Authentication.basic(CLIENT_ID, ABLY_API_KEY)))
```

content/asset-tracking/using-the-sdks.textile
```diff
@@ -86,7 +86,8 @@ The required methods are:
 The optional methods are:
 
 - profile := Called to set the means of transport being used for the initial state of publishers created from this builder. If not set then the default value is @RoutingProfile.DRIVING@.
-- context := Called to provide the Android runtime context (on Android only).
+- locationSource := Sets the location source to be used instead of the GPS. The location source provides location updates for the @Publisher@.
+- androidContext := Called to provide the Android runtime context (on Android only).
```
KacperKluka, line 90: The name is correct now however this thing is required for Android but here it's under "The optional methods" wink

content/asset-tracking/using-the-sdks.textile
tbedford: Thanks @KacperKluka !

content/asset-tracking/index.textile
```diff
@@ -80,32 +81,36 @@ Resolution governs how often to sample locations, at what level of positional ac
...
+- @farWithSubscriber@ := The resolution to select if above the "@proximityThreshold@":#default-resolution-constraints, with one or more subscribers.
+- @nearWithoutSubscriber@ := The resolution to select if below the "@proximityThreshold@":#default-resolution-constraints, with no subscribers.
+- @nearWithSubscriber@ := The resolution to select if below the "@proximityThreshold@":#default-resolution-constraints, with one or more subscribers.
 
 h2(#resolution-constraints). Resolution constraints
 
-Resolution constraints specify the thresholds and corresponding logical mappings for tracking @Resolution@s that are required by the default @ResolutionPolicy@, which can be adopted by @Publisher@ instances. While developers can define their own resolution constraints, Ably also provides default resolution constraints that can be used for common use cases.
+Resolution constraints specify the thresholds and corresponding logical mappings for tracking @Resolution@ objects that are required by the default @ResolutionPolicy@, which can be adopted by @Publisher@ instances.
```
KacperKluka, line 105: To be very precise the `ResolutionConstraints` are always optional and they're not required in the default `ResolutionPolicy` implementation :wink:

KacperKluka (changes requested):
Looks very good! I've only found a few minor things :wink:

content/asset-tracking/index.textile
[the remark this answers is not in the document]
KacperKluka: @kavalerov I don't think that there was any specific reason for sealing this class :thinking: We could turn it into an interface but then we'd need to decide how we will handle a scenario in which a user sends his custom resolution constraints to our default `ResolutionPolicy`. We can either:
1. Throw an error when a user provides any other resolution constraints than `DefaultResolutionConstraints` or
2. Ignore resolution constraints from a trackable if it isn't the `DefaultResolutionConstraints`

I think that we should go with option 1 because it will immediately inform the developer that he isn't correctly using our default `ResolutionPolicy`. If we'll go with option 2 then he won't get any feedback that he's doing something wrong and may not even notice it and get frustrated that his resolution constraints aren't working :wink: @QuintinWillison WDYT?

content/asset-tracking/using-the-sdks.textile
tbedford: Fixed. `androidContext` moved to required.

content/asset-tracking/index.textile
tbedford: Yes, I worded that very confusingly, sorry. Does the following make more sense: "Resolution constraints specify the thresholds and corresponding logical mappings for tracking @Resolution@ objects. These @Resolution@ objects are required by the default @ResolutionPolicy@, which may be adopted by @Publisher@ instances."

KacperKluka: As I read it now I'm having more doubts if what's written here is true :thinking: And I've noticed that we have the same description written in the kdocs in our code :thinking: 

**This is how I understand this sentence** (please correct me if I'm wrong):
We have a thing called "resolution constraints". It specifies some thresholds and logical mappings. Those thresholds and mappings are then used for tracking some "resolution objects". The "resolution objects" are required by the default "Resolution Policy".

**A detailed description of how it works in the code**:
We have a thing called "resolution constraints". It specifies:
-  the proximity threshold
-  a set of resolutions that are used depending on some factors (i.e. threshold)
- the battery level threshold together with a low battery multiplier (applied for desired interval if battery is below the threshold)

This "resolution constraints" is an optional part of a `Trackable` object. The `Trackable` object represents a thing that the `Publisher` is tracking. If we create a `Trackable` object with those "resolution constraints" then they will be used by `ResolutionPolicy` in order to resolve the best possible `Resolution` for that `Trackable`. However, if the "resolution constraints" are not provided, the `ResolutionPolicy` still works and resolves a valid `Resolution` but its calculations are simpler.

**How it's described in our code docs:**

``` kotlin
/**
 * Specifies the thresholds and corresponding logical mappings for tracking [Resolution]s that are required by the
 * default [ResolutionPolicy], which can adopted by [Publisher] instances using the [Builder][Publisher.Builder]'s
 * [resolutionPolicy][Publisher.Builder.resolutionPolicy] method, providing it with an instance of the
 * [DefaultResolutionPolicyFactory] class.
 */
```

@QuintinWillison given my above description of how it works in the code, are our docs accurate or should we change them? :thinking:

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,234 @@
...
+
+The SDKs can be found in the following GitHub repositories:
+
+* "Android":https://github.com/ably/ably-asset-tracking-android
+* "iOS":https://github.com/ably/ably-asset-tracking-cocoa
+* "JavaScript":https://github.com/ably/ably-asset-tracking-js
+
+h2(#prerequisites). Prerequisites
+
+You need to have a suitable developer environment installed, for example:
```
owenpearson, line 44: I'd go with "development environment" here.

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,234 @@
...
+```[android]
+publisher.remove(trackable)
+```
+
+h2(#subscribing-sdk). Using the Subscribing SDK
+
+Common operations you will need to carry out on the subscriber include:
+
+* Initialization of the subscriber.
+* Listen for updates sent from from the publisher.
```
owenpearson, line 161: Maybe put "location updates" here for clarity? If I didn't know about asset tracking before reading this I'd probably assume that "listening for asset status updates" is already encompassed by "listening for updates".

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,234 @@
...
+
+```[android]
+publisher.remove(trackable)
+```
+
+h2(#subscribing-sdk). Using the Subscribing SDK
+
+Common operations you will need to carry out on the subscriber include:
+
+* Initialization of the subscriber.
```
owenpearson, line 160: The other items in this list are in infinitive form so it might be worth changing this to "Initialize the subscriber", just for the sake of consistency.

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,165 @@
...
+
+The following scenarios are examples of how update resolution can be changed depending on the state of the asset and subscribers:
+
+* Updates can be sent to the subscriber more often when asset is closer to the destination.
+* Updates don't need to occur as frequently when the user of the subscribing SDK is not observing the location of the asset. The frequency of updates can be reduced, for example, if the user of delivery app is not checking the app for the driver's location.
+* Different subscribers may have different preferences regarding the resolution of location updates.
+* When the driver's battery is running low, location updates can be transmitted at lower frequency.
+
+Using the approaches similar to these, an optimal resolution of updates can be identified, which provides a good user experience when it is needed, such as when the driver is close to the destination, and the user is observing the location of the driver in the app. This approach also reduces battery usage when possible, for example when the driver is far from the destination and the user is not observing location updates.
```
owenpearson, line 73: I'm not too sure what "Using _the_ approaches similar to these" is supposed to mean in this context. I'm wondering if this would be better written as either:

- "Using these approaches"
- "Using approaches similar to these"

content/asset-tracking/using-the-sdks.textile
```diff
@@ -0,0 +1,234 @@
...
+
+This topic explains how to get started using the Ably Asset Tracking SDKs.
+
+h2(#supported-platforms). Supported platforms
+
+There are two Asset Tracking SDKs, one for publishing and one for subscribing. The following platforms are supported:
+
+* Android (Java and Kotlin) - publisher and subscriber SDK
+* iOS (Objective-C and Swift) - publisher and subscriber SDK
+* Web (JavaScript) - subscriber SDK
```
owenpearson, line 32: I wonder if it's worth flaunting the fact that the JavaScript SDK has first class TypeScript support? @kavalerov WDYT?

owenpearson (changes requested):
Mainly just nitpicky language stuff. The technical side of things looks good to me 👍 (although I'm sure others in this thread will have more context).

content/asset-tracking/using-the-sdks.textile
owenpearson: Actually, I've just read through it again and along with the preceding "Common operations you will need to carry out on the subscriber include:" I think this line does sound better as "Initialization". Perhaps the other items on the list should instead be changed to "_Listening_ for updates sent from the publisher" etc. I'm probably just being overly pedantic at this point though 🤷

content/asset-tracking/index.textile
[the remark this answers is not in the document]
kavalerov: Ok, on this one - I think we first need to make sure that the docs represent what we have in the code right now. (and @tbedford did this already). 

Regarding fixing this - I suggest we take this into a new issue in the ably-asset-tracking-android, and continue conversation there.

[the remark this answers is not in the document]
kavalerov: Created https://github.com/ably/ably-asset-tracking-android/issues/299 , resolving this now.

content/asset-tracking/using-the-sdks.textile
tbedford: Fixed.

content/asset-tracking/using-the-sdks.textile
tbedford: Good spot.

content/asset-tracking/using-the-sdks.textile
tbedford: I've reworked slightly, yes there was something off.

content/asset-tracking/using-the-sdks.textile
tbedford: Seems worth mentioning. Added note.

owenpearson (approved):
LGTM, thanks

content/asset-tracking/index.textile
kavalerov: I actually think the wording is wrong now. 

`ResolutionConstraints` are defined on per `Trackable` basis, and the are used to calculate appropriate `Resolution` by the `ResolutionPolicy`.

paddybyers:
I'm going to leave some high-level comments here, but will do a review in detail of the overview doc.

Use of Android Studio: both in the example apps and the SDKs documentation, we say you have to have Android Studio. That's not true - we should support building and running apps using Gradle, with the Android SDK installed. The documentation needs to make this explicit, and the Android Studio descriptions should state that they are illustrative, but equivalent steps will work if performed directly from the command line.

There are rendering artifacts in the SDK docs where, for example, text overflows the bounding box. I assume those issues will be reviewed and dealt with separately.

content/asset-tracking/index.textile
kavalerov: @KacperKluka do you think what I wrote above makes sense?

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,165 @@
...
+    - Resolution policy#resolution-policy
+    - See also#see-also
+---
+
+The Ably Asset Tracking solution provides a way to track the location of multiple assets in realtime. An asset is any object that moves geographically, and whose position needs to be monitored. The Ably Asset Tracking solution is powered by the Ably realtime network, the Ably Asset Tracking SDKs, and the Mapbox Navigation SDK with location enhancement.
+
+The Ably Asset Tracking solution provides two SDKs:
+
+* Publishing SDK - for embedding in apps on the asset to be tracked.
+* Subscribing SDK - for embedding in apps that want to observe the asset being tracked.
```
paddybyers, line 28: add "... via a realtime subscription"

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,165 @@
...
+
+As Ably is used as the underlying transport, you have direct access to your data and can use "Ably Integrations":https://ably.com/integrations for a wide range of applications, in addition to direct realtime subscriptions. Examples include: 
+
+* Passing data to a third-party service.
```
paddybyers, line 32: We need to capture the idea that it includes any realtime processing that's not via a realtime connection. It doesn't have to be a third-party.

Maybe "passing data to another service for realtime processing or tracking"

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,165 @@
...
+The following diagram provides an overview of an asset tracking use case:
+
+<a href="/images/screenshots/asset-tracking/overview.png" target="_blank">
+  <img src="/images/screenshots/asset-tracking/overview.png" style="width: 100%" alt="Overview">
+</a>
+
+In a typical asset tracking scenario the key participants are:
+
+- asset (trackable) := Item that the customer wants to track such as driver, vehicle, container, and so on. This also known as a @Trackable@ in the SDKs.
+- driver := Person driving the asset physically. Could also be a machine moving the asset under software direction, such as in the case of a delivery drone.
```
paddybyers, line 47: In what way is this a "participant"? There's nothing in the SDKs that refer to this. I know that in the text that follows there are multiple references to "driver", but I'm not sure it qualifies for definition here. If it is defined here, I would prefer that it's not explicitly a "person".  We could define a more generic "carrier" term, maybe, and define that as the entity that is physically transporting the asset.

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,165 @@
...
+- customer := Entity that uses the Ably Asset Tracking SDKs to build their product to solve the asset tracking problem for their users.
+- end user := Person that uses the product built by the customer and uses the asset tracking capabilities.
```
paddybyers, line 49: I think this should be an "entity", not a "person".

content/asset-tracking/index.textile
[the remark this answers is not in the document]
paddybyers: If we're doing this, then should there be a comment that the `trackableId` corresponds directly to the channel used to disseminate updates?

content/asset-tracking/index.textile
KacperKluka: Yes, it's how it works in the code and how I understand it too :wink:

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,165 @@
...
+
+Each asset being tracked must have a @trackableId@, and this must be globally unique within an Ably asset tracking account.
+
+The Ably Asset Tracking SDKs also allow for various battery optimization scenarios, so frequency of updates can be modified based on various parameters such as distance to destination, subscriber preference, map zoom level, and current battery levels.
```
paddybyers, line 53: I suggest we remove this paragraph. I think it needs clarification (see below) but I think it would just end up repeating what's said immediately in the next section.

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,165 @@
...
+
+h2(#battery-optimization). Battery optimization scenarios
+
+One of the biggest challenges in asset tracking is combining high resolution of updates with low battery consumption on the asset tracking device. For example, in a delivery tracking use case you may want your users to have a smooth and realtime visualization of where their delivery is on the map, but at the same time sending locations from the driver's device frequently will rapidly drain its battery.
```
paddybyers, line 57: I think this focusses too much on battery optimisation as the reason for controlling resolution. It's not just that - it might be bandwidth costs, for example.

The heading needs to change I think to be about "dynamic resolution" or "dynamic update frequency".

I suggest something like:

The Asset Tracking SDKs allow the frequency of publication of location updates to be controlled based on a range of parameters, both by the publisher and the subscriber. Having this control, to vary the frequency of updates, and sensor(s) used to obtain location information, is important to be able to control the resource usage of the application. In many mobile applications, for example, conservation of the battery life of the courier's device is paramount; high frequency and high resolution location updates that are broadcast unconditionally are not compatible with the battery constraints for mobile delivery riders, for example. The Asset Tracking SDKs aim to provide a sufficient level of control to be able to optimise the multiple parameters that influence bandwidth and energy usage.

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,165 @@
...
+
+There are two key factors that impact asset tracking device battery usage:
```
paddybyers, line 59: "... that impact asset tracking resource usage:"

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,165 @@
...
+
+1. GPS sensor - obtaining location of the asset.
```
paddybyers, line 61: "The sensor used to obtain the location of the asset, such as GPS"

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,165 @@
...
+2. Networking - sending location updates over network.
```
paddybyers, line 62: "The frequency with which location updates are sent over the network"

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,165 @@
...
+
+It is hard to reduce the battery consumption of the GPS sensor in some situations, such as when the device is being used for navigation at the same time as tracking an asset. However, it is possible to adjust the frequency of updates.
```
paddybyers, line 64: The actual resource usage, in practice, also depends on certain factors that can be beyond the control of the Asset Tracking application; for example, a delivery rider might be using an app to navigate (which will then be causing the GPS to be used continuously). This means that not all aspects of the resource usage are always controllable by the Asset Tracking application. However, even in this situation it is important that the application is able to control the frequency of updates being sent over the network.

paddybyers (changes requested):
These are my comments on the overview doc so far.

content/asset-tracking/index.textile
kavalerov: @tbedford we probably should change the wording to express what I wrote 2 comments above.

content/asset-tracking/index.textile
kavalerov: I agree with the change of wording, but I also think that battery usage was highlighted as one of the primary value drivers for customers. We probably need to address it in some way in our docs eventually, as people will come to them with question in mind "How do I optimise battery usage by using Ably Asset Tracking?" as this is one of the things we will advertise. Therefore, I think docs should give an answer to this, that is easy to find. 

Should we potentially add Battery Optimisations as a separate topic (page) in the Asset Tracking section? Or should it be something like a Tutorial? 

@tbedford / @paddybyers  what do you think?

paddybyers: When this doc is expanded there's going to have to be more extensive discussion relating to resolution policies, and we should have a whole section on how resolution policy should be tuned to optimise battery life, and what parameters to pass to the default policy to optimise battery life. I just don't think it should be the high-level statement of what dynamic resolution is for.

content/asset-tracking/index.textile
tbedford: @kavalerov I think I did already! Can you double check please.

KacperKluka: FYI I've created an [issue](https://github.com/ably/ably-asset-tracking-android/issues/305) in the android repo to fix code docs :wink:

kavalerov (approved):
LGTM, thanks!

KacperKluka (approved):
LGTM :+1:

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,163 @@
...
+- @minimumDisplacement@ := Minimum positional granularity required, in metres. Lowering this value increases the spatial resolution. Location updates whose position differs from the last known position by a distance smaller than this value are to be filtered out. Used to configure the underlying location provider, as well as to filter the broadcast of updates to subscribers.
+
+h3(#request-different-resolution). Request a different Resolution 
+
+The subscriber may require different resolutions of updates depending on state. For example, whether the map is currently displayed to the user or not, what zoom level the map is set to, and so on. This feature allows the subscriber to specify different preferences depending on any factors of interest to the client.
+
+h2(#resolution-set). Resolution set
+
+A resolution set, is a set of resolutions that must be provided by the app in order to inform the resolution policy. While developers can define their own resolution sets, Ably also provides a default resolution set that can be used with the default resolution policy.
```
paddybyers, line 91: "Ably also provides" -> "the library also provides"

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,163 @@
...
+- @nearWithoutSubscriber@ := The resolution to select if below the "@proximityThreshold@":#default-resolution-constraints, with no subscribers.
+- @nearWithSubscriber@ := The resolution to select if below the "@proximityThreshold@":#default-resolution-constraints, with one or more subscribers.
+
+h2(#resolution-constraints). Resolution constraints
+
+Resolution constraint objects, @ResolutionConstraints@, are defined on a per-@Trackable@ basis, and are then used to calculate an appropriate @Resolution@ by the @ResolutionPolicy@.
+
+h3(#default-resolution-constraints). Default resolution constraints
+
+Ably provides default resolution constraints consisting of:
```
paddybyers, line 110: Same comment as above - I would have said "The library provides...". I don't feel strongly though

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,163 @@
...
+
+- @DefaultResolutionSet@ := The set of resolutions which must be defined in order to populate the @DefaultResolutionConstraints@.
+- @DefaultProximity@ := This is the proximity to the destination at which resolution of updates is increased to improve user experience. Proximity can be spatial, temporal, or both. 
+- @batteryLevelThreshold@ := This is a level of battery on the device that runs Publishing SDK that is considered low. When the battery level falls below this, the resolution of updates is reduced.
```
paddybyers, line 114: "This is a level of battery on the device that runs Publishing SDK that is considered low" -> "This is a battery level for the publisher's device that is considered to be low"

content/asset-tracking/index.textile
```diff
@@ -0,0 +1,163 @@
...
+
+Resolution policy defines the strategy by which the various @ResolutionRequests@ and preferences are translated by @Publisher@ instances into a target @Resolution@.
+
+As different use cases have different requirements for battery optimization, Ably Asset Tracking SDKs provide a flexible @ResolutionPolicy@ interface, that allows SDK users to provide their own implementation of resolution optimization logic. Ably also provides a "default resolution policy":#default-resolution-policy, which is suitable for common use cases.
+
+h3(#default-resolution-policy). Default resolution policy
+
+Ably Asset Tracking SDKs come with a @DefaultResolutionPolicy@ implementation provided out of the box, which is suitable for common use cases.
+
+Each @Trackable@ added to the @Publisher@ has a destination and set of @ResolutionConstraints@. Both @destination@ and @resolutionConstraints@ are optional for @Trackable@, but having them improves the resolution policy. Ably provides an optional set of "default resolution constraints":#default-resolution-constraints, @DefaultResolutionContraints@.
```
paddybyers: "improves the resolution policy" -> "enables the resolution policy to reflect those constraints" maybe

content/asset-tracking/index.textile
tbedford: I went with "the SDK also provides...".

content/asset-tracking/index.textile
tbedford: Much better, thanks.

paddybyers (approved):
LGTM, thanks

PR #1073, opened 2021-04-19
CSharp API usage correction

tomkirbygreen:
Small CSharp API usage correction.  In the latest version of the SDK when
establishing a callback for 'Connection' events the first parameter should
be a 'ConnetionEvent' and not a 'ConnectionState'.

I have only changed this for the latest version of the docs.  I need to do
more research and see if this applies to older versions of the SDK.

### DIFF
--- a/content/realtime/connection.textile
+++ b/content/realtime/connection.textile
@@ -69,7 +69,7 @@ ably.connection.on('connected', new ConnectionStateListener() {
 });
 
 bc[csharp]. AblyRealtime ably = new AblyRealtime("{{API_KEY}}");
-ably.Connection.On(ConnectionState.Connected, args => {
+ably.Connection.On(ConnectionEvent.Connected, args => {
   Console.WriteLine("Connected, that was easy");
 });
 
--- a/content/root/quick-start-guide.textile
+++ b/content/root/quick-start-guide.textile
@@ -186,7 +186,7 @@ ably.connection.on(.connected) {
 }
 
 bc[csharp]. var ably = new AblyRealtime("{{API_KEY}}");
-ably.Connection.On(ConnectionState.Connected, args =>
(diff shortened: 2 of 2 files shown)


content/realtime/connection.textile
```diff
@@ -69,7 +69,7 @@ ably.connection.on('connected', new ConnectionStateListener() {
 });
 
 bc[csharp]. AblyRealtime ably = new AblyRealtime("{{API_KEY}}");
-ably.Connection.On(ConnectionState.Connected, args => {
+ably.Connection.On(ConnectionEvent.Connected, args => {
```
tbedford, line 72: I'm not a .NET expert, but looking at the repo [readme](https://github.com/ably/ably-dotnet#connection), the examples there seem to use `ConnectionState` and not `ConnectionEvent`. If so, the ably-dotnet readme also needs updating.  

Also, this code seems to be in `ConnectionStateListener()`. Should it be `ConnectioEventListener`? 

Just noticed, corresponding [reference guide](https://ably.com/documentation/realtime/connection#connection-state) looks incorrect:  It seems to have both `ConnectionState` and `ConnectionEvent`, but the documentation for `ConnectionEvent` just seems to duplicate the previous section for `ConnectionState`.

image

@tomkirbygreen - would you be able to look into these items for .NET and confirm the situation please?

tbedford (changes requested):

tbedford:
@tomkirbygreen I also just noticed this is on a fork rather than a clone. Can you create a PR on a clone, if not it's a permission problem I need to chase up.

tbedford (approved):
I tested this out as part of this ticket https://ably.atlassian.net/browse/DOC-420 - I will merge this into main. We have a follow up PR on branch fea-1041-update-quickstart that also updates all the Quickstart guides. I will also create a ticket to review .NET types reference guide.

PR #1077, opened 2021-04-27
Make ably-agent versions optional, introduce semver requirement

owenpearson:
As discussed in [this internal slack thread](https://ably-real-time.slack.com/archives/CURL4U2FP/p1619001193081400).


content/client-lib-development-guide/features.textile
```diff
@@ -81,7 +81,7 @@ h3(#restclient). RestClient
...
+*** @(RSC7d1)@ In general, the @Agent@ library identifier is composed of a series of @key[/value]@ entries joined by spaces where @key@ is the name of a produce and @value@ is an optional version of the associated product which must be a valid "semantic version":https://semver.org/. This must include at least the @<library name>/<library version>@ entry followed optionally, in dependency order, the @name[/version]@ entry for underlying library layer or platform layers. An example would be @ably-flutter/1.2.0 ably-java/1.2.1 android/24@. The documentation for each library must define the content and meaning of the @Agent@ identifiers advertised by that library.
```
lmars, line 84:
```diff
- produce
+ product
```

lmars: Also, why MUST it be a semantic version, rather than that being a recommendation?

lmars: Also, why does this spec item start with "In general"? Is there a non-general equivalent?

owenpearson: @lmars I fixed the spelling mistake, thanks. I also removed the "in general", I agree that it had no purpose there.

As for the SemVer requirement we've discussed this internally quite a bit. The general consensus has been that (1) It's a lot easier to relax this requirement later on than to start with a more liberal requirement and then make it more restrictive later on, and (2) If all of the versions are SemVer then this enforces consistency which may be useful later on if we need to parse this data - if we know everything will be SemVer then we can leverage the ton of already-existing open source SemVer parser out there rather than having to work with the versions on a case-by-case basis. I would be open to changing it to "should" but are there any particular reasons you have in mind for doing so?

lmars: ah ok thanks, I formally remove my semver question 🙂

paddybyers (approved):
LGTM, thanks

QuintinWillison (approved):

PR #1079, opened 2021-04-27
FEA-961: Add sample connection and request events to metachannels page

m-hulbert:
Adds the sample connection and request event metachannels into the existing [metachannels](https://ably.com/documentation/realtime/metachannels) page.

View [here](http://ably-docs-pr-1079.herokuapp.com/realtime/metachannels/) for testing.


tbedford:
@marklewin - I won't get to this until next week, so in the mean time would you be able to do a quick review please? Thanks.

marklewin (changes requested):
Looks good! A few minor comments and suggestions.

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
+    console.log('connection closed: ', msg.data);
+})
+
+channel.subscribe('connection.failed', (msg) => {
+    console.log('connection failed: ', msg.data);
+})
+
+channel.subscribe('connection.refused', (msg) => {
+    console.log('connection refused: ', msg.data);
+})
```
SimonWoolf, line 155: Showing the client subscribing to every message name there is one by one is a little bit of an antipattern -- if you want to subscribe to everything you can just do that. If you want to show both, maybe:
```js
// Subscribe to all event types
channel.subscribe((msg) => {
    console.log('Event type: ' + msg.name, msg.data);
});

// Subscribe to one particular event type
channel.subscribe('connection.opened', (msg) => {
    console.log('connection opened: ', msg.data);
});
```

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
+```
+
+h3(#sample). Sample rate
+
+The rate of sampling dictates the number of events published to the metachannels. For example, a rate of 0.1% for connections would publish every 1000th connection event. The sample rate can be configured independently for connections and API requests within the same app.
```
SimonWoolf, line 160: > would publish every 1000th connection event

it's not every 1000th event exactly, it's just events at random with a 0.1% probability. "...would publish an average of one message in 1000"

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
+
+- connection.opened := The connection was successfully made by the client to Ably.
+- connection.closed := The connection was explicitly closed by the client.
+- connection.refused := The connection was rejected for an expected reason, such as invalid credentials or a malformed request.
+- connection.failed :=  The connection was rejected for an unexpected reason, such as a network failure.
+
+Events contain the following metadata in the @data@ field:
+
+- name := The event type, for example @connection.opened@.<br>__Type: @String@__
+- time := The time the connection was attempted, as milliseconds since the epoch.<br>__Type: @Integer@__
```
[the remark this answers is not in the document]
SimonWoolf: I don't think there's a `time` field at all? (there's a message.timestamp as usual, but I don't think there's a separate `time` in the data field)

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
+- host := The host the connection was made to.<br>__Type: @String@__
+- clientId := The ID of the client that attempted the connection.<br>__Type: @String@__
+- requestId := The unique ID of the connection.<br>__Type: @String@__
+- region := The region the connection was made from.<br>__Type: @String@__
+- headers := The headers sent with the connection.<br>__Type: @JSON Object@__
+- query := The details of the connection query.<br>__Type: @JSON Object@__
```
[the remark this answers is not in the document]
SimonWoolf: this could be pretty explicit, eg "The parsed query string of the connect request, redact to remove auth params, containing connection information sent by the client library such as the library version and any custom transport parameters"

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
+- channels := A list of channels included in the request. This is only relevant where channels are supported as part of the connection request, such as with "SSE":/sse <br>__Type: @Array@__
```
SimonWoolf, line 183: missing connectionId

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
+Sampled connection events for an app are published to the channel @[meta]clientEvents:connections@. Connection events can be used to compile statistics relating to client population.
+
+The connection event types are:
+
+- connection.opened := The connection was successfully made by the client to Ably.
+- connection.closed := The connection was explicitly closed by the client.
+- connection.refused := The connection was rejected for an expected reason, such as invalid credentials or a malformed request.
+- connection.failed :=  The connection was rejected for an unexpected reason, such as a network failure.
+
+Events contain the following metadata in the @data@ field:
```
SimonWoolf, line 173: Not all events will contain all of these. "contain some subset of the following metadata in the `data` field", maybe?

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
+
+- name := The event type, for example @connection.opened@.<br>__Type: @String@__
+- time := The time the connection was attempted, as milliseconds since the epoch.<br>__Type: @Integer@__
+- host := The host the connection was made to.<br>__Type: @String@__
+- clientId := The ID of the client that attempted the connection.<br>__Type: @String@__
+- requestId := The unique ID of the connection.<br>__Type: @String@__
```
SimonWoolf, line 179: "The unique ID of the connection" is a better description of the connectionId than the requestId. Maybe "An identifier that Ably can use to correlate a connection event with our internal logs, if necessary"

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
+The connection event types are:
+
+- connection.opened := The connection was successfully made by the client to Ably.
+- connection.closed := The connection was explicitly closed by the client.
+- connection.refused := The connection was rejected for an expected reason, such as invalid credentials or a malformed request.
+- connection.failed :=  The connection was rejected for an unexpected reason, such as a network failure.
+
+Events contain the following metadata in the @data@ field:
+
+- name := The event type, for example @connection.opened@.<br>__Type: @String@__
```
SimonWoolf, line 175: `name` isn't a field in the `data` field, it's a top-level field of the message

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
+
+The request event types are:
+
+- request.succeeded := The request was successful.
+- request.refused := The request was rejected for an expected reason, such as a malformed request or insufficient privileges.
+- request.failed  := The request was rejected for an unexpected reason, such as a network failure.
+
+Events contain the following metadata in the @data@ field:
+
+- name := The event type, for example @request.succeeded@.<br>__Type: @String@__
```
SimonWoolf, line 222: same comments apply to most of these as above

content/realtime/metachannels.textile
SimonWoolf: also missing error (only relevant in a few places)

SimonWoolf: also missing duration

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
+- time := The time the request was made, as milliseconds since the epoch.<br>__Type: @Integer@__
+- host := The host the request was made to.<br>__Type: @String@__
+- clientId := The ID of the client that made the request.<br>__Type: @String@__
+- requestId := The unique ID of the request.<br>__Type: @String@__
+- region := The region the request was made in.<br>__Type: @String@__
+- headers := The headers sent with the request.<br>__Type: @JSON Object@__
+- query := The details of the request query.<br>__Type: @JSON Object@__
+- channels := A list of channels the request was made against.<br>__Type: @Array@__
+- error := The details of any error received whilst making the request.<br>__Type: @JSON Object@__
```
SimonWoolf, line 231: maybe say 'encountered' rather than 'received'

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
+
```
SimonWoolf, line 232: missing `path`

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
 
-If you want webhooks for channel lifecycle events, creating a "channel lifecycle webhook":/general/events#sources is an easy alternative to creating a message rule for the @[meta]channel.lifecycle@ metachannel; such webhooks will have simpler POST bodies as the lifecycle events won't be wrapped in "@Message@":/realtime/types#message objects.
+If you want webhooks for metachannel events, creating a "webhook":/general/events#sources is an easy alternative to creating a message rule for the metachannel; such webhooks will have simpler POST bodies as the lifecycle events won't be wrapped in "@Message@":/realtime/types#message objects.
```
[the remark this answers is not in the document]
SimonWoolf: The sentence doesn't make sense now, but just dropping "If you want webhooks for metachannel events" wouldn't help much -- "Creating a webhook is an easy alternative to creating a message rule for the metachannel" still doesn't make much sense, a webhook is a kind of message rule.

The point of this paragraph is that a "channel lifecycle webhook" is a _special kind of webhook_ -- one with the rule source set to "channel.lifecycle". So if you want to get a webhook (or whatever) for every channel lifecycle event, creating a channel lifecycle webhook rule is an alternative to creating a message rule for every message on the channel lifecycle metachannel.

(And in future it will be a much better alternative, because that won't be subject to the rate-limiting you have by funneling everything through the metachannel)

SimonWoolf (changes requested):

SimonWoolf:
@m-hulbert if you message me an appId of an app you own on slack I can enable sampled events for it so you can play around with it and directly see what fields there are in the version that made it into prod

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
 
 * "Inband channel occupancy events":/realtime/inband-occupancy - for use cases where the entity that needs to determine occupancy is attached to the channel. 
 * Occupancy rules - for use cases where the entity needing to determine channel occupancy is not connected to the channel. A channel occupancy reactor rule allows you to obtain occupancy updates from a set of channels in a centralized way, that doesn't involve attaching to all of them. Please "get in touch":mailto:enterprise-sales@ably.io for further information.
 
-h3(#reactor). Reactor rules
+h2(#sampled-events). Sampled connection and request events
+
+Sampled connection and request events can be used to observe the usage of Ably services. Events are published to metachannels that contain metadata for a sample of connections and API requests. These events can be used to compile statistics on app usage, enabling you to perform arbitrary data processing and aggregation related to client population and client activity.
+
+Note that this is an Enterprise-only feature and must be enabled for each app you would like to sample. Please "contact us":https://ably.com/contact to enable this functionality.
```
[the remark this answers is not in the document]
m-hulbert: I've checked this and it seems best to leave it as "/contact" for now, but I agree it isn't very consistent so I've logged a JIRA to tidy up the different options we have for getting in touch.

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
 
 * "Inband channel occupancy events":/realtime/inband-occupancy - for use cases where the entity that needs to determine occupancy is attached to the channel. 
 * Occupancy rules - for use cases where the entity needing to determine channel occupancy is not connected to the channel. A channel occupancy reactor rule allows you to obtain occupancy updates from a set of channels in a centralized way, that doesn't involve attaching to all of them. Please "get in touch":mailto:enterprise-sales@ably.io for further information.
 
-h3(#reactor). Reactor rules
+h2(#sampled-events). Sampled connection and request events
+
+Sampled connection and request events can be used to observe the usage of Ably services. Events are published to metachannels that contain metadata for a sample of connections and API requests. These events can be used to compile statistics on app usage, enabling you to perform arbitrary data processing and aggregation related to client population and client activity.
```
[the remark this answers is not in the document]
m-hulbert: Went back and forth on this whilst writing it - so have changed to "monitor" 👍

content/realtime/metachannels.textile
```diff
@@ -115,17 +116,150 @@ metaChannel.subscribe('channel.region.inactive', (msg) => {
...
+h3(#connections). Connections
+
+Sampled connection events for an app are published to the channel @[meta]clientEvents:connections@. Connection events can be used to compile statistics relating to client population.
+
+The connection event types are:
+
+- connection.opened := The connection was successfully made by the client to Ably.
+- connection.closed := The connection was explicitly closed by the client.
+- connection.refused := The connection was rejected for an expected reason, such as invalid credentials or a malformed request.
+- connection.failed :=  The connection was rejected for an unexpected reason, such as a network failure.
```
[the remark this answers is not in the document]
m-hulbert: Fixed, thanks.

content/realtime/metachannels.textile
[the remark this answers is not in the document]
m-hulbert: Removed `time`.

content/realtime/metachannels.textile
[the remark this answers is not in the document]
m-hulbert: Updated based on this conversation.

content/realtime/metachannels.textile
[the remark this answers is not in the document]
m-hulbert: I've re-written this as a separate heading to make it clearer. Thanks for the explanation, Simon.

content/realtime/metachannels.textile
m-hulbert: Very good point. Applied the same logic to the `channel.lifecycle` example too.

content/realtime/metachannels.textile
m-hulbert: Ah, didn't know that. Fixed, thanks.

content/realtime/metachannels.textile
m-hulbert: Added and removed as necessary after having being able to try these out, thanks!

content/realtime/metachannels.textile
m-hulbert: Included subset, thanks.

content/realtime/metachannels.textile
m-hulbert: Thanks for the explanation - updated accordingly.

content/realtime/metachannels.textile
m-hulbert: Good point.

content/realtime/metachannels.textile
m-hulbert: Updated as per above.

marklewin (approved):
LGTM - can't comment much on the content, but it reads well!

content/realtime/metachannels.textile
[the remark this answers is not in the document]
paddybyers: `/contact` is correct. `channel.occupancy` lifecycle events in their present form are shortly going to be removed in any case.

content/realtime/metachannels.textile
```diff
@@ -79,53 +84,171 @@ For further information on capabilities, please refer to the following resources
...
 
-You can associate "Reactor rules":https://www.ably.com/integrations with metachannels in the same way as for regular channels; configure these using the "Message":/realtime/messages source and use a @channelFilter@ of, for example, @^\[meta\]log@. Normally the @channelFilter@ regex will only match normal, non-meta channels; however, a regex of the form @^\[qualifier\]pattern@ is special-cased to match channels with that qualifier. The @pattern@ is a regex as normal.
+h3(#lifecycle-webhooks). Channel lifecycle outbound webhooks
 
-If you want webhooks for channel lifecycle events, creating a "channel lifecycle webhook":/general/events#sources is an easy alternative to creating a message rule for the @[meta]channel.lifecycle@ metachannel; such webhooks will have simpler POST bodies as the lifecycle events won't be wrapped in "@Message@":/realtime/types#message objects.
+When creating an outbound webhook Reactor rule it is possible to select @Channel Lifecycle@ as the "source":general/events#sources. This will set all channel lifecycle events as triggers for the webhook, rather than having to create individual message rules for each event type.
```
SimonWoolf, line 254: > rather than having to create individual message rules for each event type.

That's not a thing you would need to do either way, a message rule gets every message on the channel. The reason to use lifecycle webhooks is the one in the previous version -- simpler and more readable post bodies -- and also that we have the freedom to move lifecycle events to a new mechanism, which we're planning on doing soon.

How about:

`If you're considering creating a message rule on the @[meta]channel.lifecycle@ metachannel, we suggest that you instead create a rule with @channel.lifecycle@ as the "source":general/events#sources. Events from this will be much simpler to parse as the lifecycle events won't be wrapped in "@Message@":/realtime/types#message objects. This is currently only selectable in batched Webhook rules; however we are shortly planning to extend this to all rule types, "get in touch":https://ably.com/contact if you'd be interested in this.`

SimonWoolf (approved):
👍 with one comment

content/realtime/metachannels.textile
m-hulbert: Thanks Simon - reworded this to point that out and removed the individual rule sentence.

PR #1089, opened 2021-05-11
Spec: remove optional wording from idle timeout detection

SimonWoolf:
The trouble with the 'This requirement is not mandatory' wording is it gives the false impression that the item isn't important. It's
also not very clear what a client library developer is supposed to base their decision to implement or not on. Perhaps best to remove the choice and just make it mandatory


content/client-lib-development-guide/features.textile
```diff
@@ -481,7 +481,7 @@ h3(#realtime-connection). Connection
...
+** @(RTN23a)@ If a transport does not receive any indication of activity on a transport for a period greater than the sum of the @maxIdleInterval@ (which will be sent in the @connectionDetails@ of the most recent @CONNECTED@ message received on that transport) and the @realtimeRequestTimeout@, that transport should be disconnected. Any message (or non-message indicator, see @RTN23b@) received counts as an indication of activity and should reset the timer, not merely heartbeat messages. However, it must be received (that is, sent from the server to the client); client-sent data does not count.
```
tbedford, line 484: Is there a use case in which it would not be possible to implement this for a client? I'm not sure what the underlying constraints are, but could there be a transport for which it's not possible to implement this?

tbedford (approved):
Minor note added.

Minor style note, it's usually more readable if lengthy bracketed text is avoided in most cases, otherwise looks good.

PR #1090, opened 2021-05-12
RTN11c: Explicit connect should keep trying to recover connection state

tcard:
"Aborts the retry process" suggests that we should just abandon the
whole connection state recovery logic. In implementation terms, this
may be translated to breaking an actual loop with the associated
connection state TTL ([this happened recently in Go](https://github.com/ably/ably-go/pull/297)).

This doesn't make much sense, and it's really not useful. What should
happen instead is that the reconnection loop should skip the wait, and
attempt the next try right away.

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -414,7 +414,7 @@ h3(#realtime-connection). Connection
 * @(RTN11)@ @Connection#connect@ function:
 ** @(RTN11a)@ Explicitly connects to the Ably service if not already connected
 ** @(RTN11b)@ If the state is @CLOSING@, the client should make a new connection with a new transport instance and remove all references to the old one. In particular, it should make sure that, when the @CLOSED@ @ProtocolMessage@ arrives for the old connection, it doesn't affect the new one.
-** @(RTN11c)@ If the state is @DISCONNECTED@ or @SUSPENDED@, aborts the retry process described in "RTN14d":#RTN14d and "RTN14e":#RTN14e and immediately tries to reconnect.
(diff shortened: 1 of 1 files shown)


QuintinWillison:
I would appreciate it if @SimonWoolf could have a look at this one. This spec point was introduced in #112 and my concern is that if the nature of the spec point changes too much then how do we track which libraries are adhering to which version of the spec point. We have some internal guidance which we're building in Confluence but I don't think this particular type of change and when it's applicable has been discussed and decided anywhere canonical yet, as far as I'm aware, but I know Simon has an aptitude for remembering and/or finding these things... also way more historical context than me!

AndyNicks (approved):

owenpearson (approved):
LGTM, thanks

SimonWoolf (approved):
👍

>  I don't think this particular type of change and when it's applicable has been discussed and decided anywhere canonical yet

FWIW this seems to me like a clarification rather than a change of nature.

The old place that had guidance on different kinds of changes is https://github.com/ably/docs#branch-and-tag-scheme-for-features-spec in the readme.

The main mechanism for keeping track of this was nominally that the diff from one spec version to another, which client lib devs can use as the starting point for seeing what needs to be changed, is the old release tag to the new one - so that incorporates all changes since that tag, whether made to 1.2 after the 1.2 release, or 1.3-only made in the 1.3 integration branch. Each dev at the time then decides whether any particular change means something needs changing in their lib or not.

At least that was the idea at some point. The extent to which that was actually done in practice is unclear. And that picture is muddied further by The Spreadsheet, which might present its own view of which specs had substantive changes between versions.

Ultimately this has been a bit of a mess before, so up to Quintin to decide what he thinks is the right mechanism to use going forwards for how we should do spec changes.

tcard:
> my concern is that if the nature of the spec point changes too much then how do we track which libraries are adhering to which version of the spec point

I wouldn't worry too much in this case, because the only consequence of doing it the "old" way is that we might believe we're DISCONNECTED for longer than we should. This is already possible if the server evicts the connection state for other reasons.

PR #1096, opened 2021-05-21
Expand Ably-Agent header spec points

QuintinWillison:
Specifically around the product component.

In response to various comments under https://github.com/ably/ably-cocoa/pull/1119.


QuintinWillison:
I'm not sure why my [latest commit](https://github.com/ably/docs/pull/1096/commits/898ca82833be9fdb43b595336e6684f2bb31ad93) hasn't resulted in a new deployment, but either way I think this is ready for reviewers.

content/client-lib-development-guide/features.textile
```diff
@@ -84,6 +84,8 @@ h3(#restclient). RestClient
...
 *** @(RSC7d2)@ Where possible, the @Agent@ library identifier should be included in all HTTP and WebSocket requests to the Ably endpoint via an @Ably-Agent@ HTTP header.
 *** @(RSC7d3)@ Where the library is unable to specify a header when making a request (such as for @JSONP@ requests), the @Agent@ library identifier should be sent via an @agent@ query param whose value is the URI-encoded @Agent@ library identifier.
+*** @(RSC7d4)@ Product names must be conformed by the client to contain only the primary set of characters allowed in a @token@ as defined by "RFC 7230, Section 3.2.6":https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6 (that is, only: @DIGIT@, @ALPHA@ and the set @!#$%&'*+-.^_`|~@).
+*** @(RSC7d5)@ Clients should conform product names using a common function, before they are packed into the agent string. This common function should always succeed, dropping any characters not in the supported set as defined by "RSC7d4":#RSC7d4. This function should be unit tested.
```
owenpearson, line 88: Is the intention here to actually drop disallowed characters completely rather than just replace them or URL encode them? I think we should be more explicit about what developers should do if they want to add a value to the headers that contains disallowed characters, even if it's more of a recommendation rather than a rule.

owenpearson (changes requested):

content/client-lib-development-guide/features.textile
QuintinWillison: Very good point. Have you got a suggestion of swaps? Perhaps anything unsupported gets replaced with a dash `-`?

ben-xD: I found that section of the RFC very hard to read/ understand, and referring to it suggests we should read it or is somehow significant, but actually the only thing we're using here is a set of characters (defined as `tchar` in the RFC). Can we remove reference to the RFC completely, that will reduce cognitive burden? Can we avoid using the terms `ALPHA`, `DIGIT` which were defined in rfc7230, but not here. We also haven't defined `product name`. Should we instead say `key` or `value`, because that is a previously used term (in `RSC7d1`)?

1. I tried to condense it down those 2 new lines:
> Only alphanumeric characters or characters in `!#$%&'*+-.^_`|~` should be included in a key or value. Any other characters should be deleted.

2. Can we have examples? (that would help me know if ` `(space) needs to be deleted too.

mattheworiordan: Given we have common error codes (see https://github.com/ably/ably-common/blob/main/protocol/errors.json), should we not insist that we have common product/platform/lib names defined in the open too?  This will have the following benefits:

* Collisions won't just happen by mistake.  Expecting ably-X developer to sync with ably-* developers, is impractical.  If however they are instructed to add to the list in `ably-common`, that creates sync.
* The router team then have a bounded list of products/platforms etc. to allow, which means we won't inadvertently  create a cardinality issue, or ignore valid product/platforms.
* There can be debate on naming by the wider team, when appropriate, using PRs
* The server / router could issue warnings in a `Ably-Agent-Error` header indicating something is wrong.  For developers building SDKs, they can test this out live and get immediate feedback, as opposed to hoping someone downstream later identifies an issue and tells them.
WDYT?

ben-xD: ## I support seeing more examples given I've only seen 1, but:
- Having a list will immediately require more work now because someone has to enumerate this list, and this would be a long list
- It will also add future bookkeeping/ logic (Also, new variations will mean more changes to this list, and subsequent changes needed by router team?). 
  - This adds friction to _growing_ this list. I assume we want this list to grow easily, so we can garner insights from the agent/ platforms that are in use
-`Ably-Agent-Error` response header/ errors will come too late (after implementation/ the SDK is already released). Because the strings are generated dynamically, these errors can happen long after a library is released, or when a new device is released, or OS is updated.

## But can we pause for a second:
This Ably Agent header is getting more and more complicated. Will Ably SDK's really be affected by all the data we're collecting here? The User Agent already gives us the platform, JS engine, etc **already**. If there are specific extra information we need like the Flutter version, Ably SDK version, let's add those specifically if there is a good use case. But do we need to know the e.g. Flutter version is, or are we over-engineering? (Yes there might be a chance that Flutter methods channels get a bug and we need to know the version, but if that's worrying you, let's add a Flutter version header, most platforms don't need this, making it generic is costly.). Don't make every client SDK implement redundant information which already comes in the user agent.

To avoid scope creep and spiralling complexity, I think the only reasonable thing left to do is go back to the drawing board: with motivations of this work (usage analytics?, detecting bugs/ issues?), with many examples, and also encoding/ decoding into/from header strings to avoid missing things. Preferably, we start with motivation and perhaps choose a better option. **Maybe an `AblySDK library name + version` string is enough for now?** I noticed many companies just stick extra info in the User Agent comment. 

I don't think many of us are aligned on this Ably Agent stuff.

Have a nice long weekend 😅

lmars: @ben-xD you make some good points, and I'll respond to them here, but it seems we should move the various conversations were having to one place (i.e. this PR, [this Slack thread](https://ably-real-time.slack.com/archives/CURL4U2FP/p1621854952027600), [this router PR](https://github.com/ably/go-services/pull/559), [this Confluence page](https://ably.atlassian.net/wiki/spaces/SDK/pages/1522401281/Ably-Agent)), perhaps a proposal document owned by the SDK team that we can all contribute to? @QuintinWillison WDYT?

> Having a list will immediately require more work now because someone has to enumerate this list, and this would be a long list

But it seems an explicit list is needed, otherwise we just get [a whole load of junk being collected](https://github.com/ably/go-services/pull/559#issuecomment-849954399), so I don't think we can get away from doing the work to construct this list, but it should be as simple as reading through the current list of tags and picking just those that are obviously SDK versions and not junk?

> It will also add future bookkeeping/ logic (Also, new variations will mean more changes to this list, and subsequent changes needed by router team?)

Again, it is true that the list requires maintenance, but I don't believe the alternative of allowing anything set in the `lib` or `agent` query param to become a metric is acceptable. The proposal of doing this in `ably-common` means there is minimal maintenance on the router side, we just update the submodule reference as and when needed.

> if that's worrying you, let's add a Flutter version header, most platforms don't need this, making it generic is costly

But then aren't we just re-hashing the same problem as needing to maintain a list of headers and their expected format that the router needs to parse and emit as logs / metrics?

> with motivations of this work (usage analytics?, detecting bugs/ issues?), with many examples, and also encoding/ decoding into/from header strings to avoid missing things

As per my opening comment, I definitely support this 👍

lmars: cc @kavalerov (I would tag Andrey too but don't know his GitHub)

mattheworiordan: > but it seems we should move the various conversations were having to one place (i.e. this PR, this Slack thread, this router PR, this Confluence page), perhaps a proposal document owned by the SDK team that we can all contribute to? @QuintinWillison WDYT?

Please can we do this cc @kavalerov @AndyNicks?  

Note in relation to this discussion:
* See https://ably-real-time.slack.com/archives/CURL4U2FP/p1621854952027600. I also asked for a reference doc  which didn't exist at the time. Given our move to written-first, we need to do this please.
* @lmars request for above in Slack https://ably-real-time.slack.com/archives/CURL4U2FP/p1622539113068500

mattheworiordan: > Because the strings are generated dynamically, these errors can happen long after a library is released, or when a new device is released, or OS is updated.

We explicitly want the SDKs to use meaningful agent strings that we can use for analysis.  If there is a risk we'll end up with unexpected variants, then it will make it more difficult for us to use this data.  From my perspective, this is exactly one of the reasons we should have control over this.  

> If there are specific extra information we need like the Flutter version, Ably SDK version, let's add those specifically if there is a good use case.

Please can write up a document detailing the reasons for capturing this information.  If you recall, the change to the agent in the spec came about in response to specific shortcomings in the current agent set up, which now seems to have been forgotten.  Please can a document be written up now to reconfirm for everyone (now and future us) why we capture agents.  We should additionally link to that from the SDK spec.

> To avoid scope creep and spiralling complexity, I think the only reasonable thing left to do is go back to the drawing board: with motivations of this work (usage analytics?, detecting bugs/ issues?), with many examples, and also encoding/ decoding into/from header strings to avoid missing things. Preferably, we start with motivation and perhaps choose a better option. Maybe an AblySDK library name + version string is enough for now? I noticed many companies just stick extra info in the User Agent comment.

That is what we had before https://github.com/ably/docs/pull/1034 :)  I was not involved in the motivation behind the proposed agent change in https://github.com/ably/docs/pull/1034, what I was aware of was that Flutter was incorrectly reporting itself as macOS or Android, and I raised that as a concern during an update from the SDK team as we had no way of knowing if the Flutter SDK was being used or not.  I was not involved in subsequent conversations.  

I think it's clear we need a document that clearly describes the motivations for the change and the expected outcomes from the agent capture.  @kavalerov @AndyNicks I believe we should focus on that and get consensus?

QuintinWillison:
I've converted to Draft because we may end up rejecting some or all of the content of this pull request. See also this internal comment: https://github.com/ably/go-services/pull/559#issuecomment-855960121

PR #1108, opened 2021-06-09
DOC-299: Correct and update webhook batched event payload examples

m-hulbert:
This PR corrects errors in the `source` for the batched webhook examples. There are also some minor changes to readability in the examples and associated text.

The updated page can be viewed [here](http://ably-docs-pr-1108.herokuapp.com/general/events/#examples) and compared to the live version [here](https://ably.com/documentation/general/events#examples).

### DIFF
--- a/content/partials/general/events/_batched_events.textile
+++ b/content/partials/general/events/_batched_events.textile
@@ -1,8 +1,8 @@
 Each batched message will have the following fields:
 
-- name := the event type, aka "@presence.message@", "@channel.message@", "@channel.closed@", etc
+- name := the event type, for example @presence.message@, @channel.message@ or @channel.closed@
 - webhookId := an internal unique ID for the configured webhook
-- source := the source for the webhook, namely "@channel.message@", "@channel.presence@", "@channel.lifecycle@"
+- source := the source for the webhook, which will be @channel.message@, @channel.presence@ or @channel.lifecycle@
 - timestamp := a timestamp represented as milliseconds since the epoch for the presence event
 - data := an object containing the data of the event defined below in "JSONPath format":https://goessner.net/articles/JsonPath/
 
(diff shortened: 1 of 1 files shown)


content/partials/general/events/_batched_events.textile
```diff
@@ -1,8 +1,8 @@
 Each batched message will have the following fields:
 
-- name := the event type, aka "@presence.message@", "@channel.message@", "@channel.closed@", etc
+- name := the event type, for example @presence.message@, @channel.message@ or @channel.closed@
 - webhookId := an internal unique ID for the configured webhook
-- source := the source for the webhook, namely "@channel.message@", "@channel.presence@", "@channel.lifecycle@"
+- source := the source for the webhook, which will be @channel.message@, @channel.presence@ or @channel.lifecycle@
 - timestamp := a timestamp represented as milliseconds since the epoch for the presence event
```
marklewin, line 6: Looks good. I just wondered if it would be better to use "Unix timestamp" here instead of "milliseconds since the epoch"?

m-hulbert: I definitely agree we need to move away from "milliseconds since the epoch". My thinking is that we should address these in bulk though, as we have about ~150 occurrences of it in the docs. I'll raise a JIRA to replace them and leave this for now so we're consistently "wrong", if you agree?

marklewin: Sounds like a good plan to me @m-hulbert !

marklewin (approved):
LGTM

content/partials/general/events/_batched_events.textile
m-hulbert: Great! Created that and assigned it to me.

SimonWoolf: (fwiw I'm used to "Unix timestamp" with no further qualifications meaning seconds since the epoch, not milliseconds. "Unix timestamp in milliseconds" might work)

m-hulbert: Thanks @SimonWoolf - that's a good point and I'll make a note of it in the issue I raised.

PR #1109, opened 2021-06-11
Spec: Slight change in semantics to RTP1 (HAS_PRESENCE)

SimonWoolf:
Clarify that HAS_PRESENCE doesn't mean that there is definitely ≥1 present member, it just means that the client lib should expect a SYNC (and so that presence ops that wait_for_sync need to wait for that, etc.).

Context: third bullet point of https://github.com/ably/realtime/pull/3326

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -617,7 +617,7 @@ h3(#realtime-channel). Channel
 
 h3(#realtime-presence). Presence
 
-* @(RTP1)@ When a channel @ATTACHED@ @ProtocolMessage@ is received, the @ProtocolMessage@ may contain a @HAS_PRESENCE@ bit flag indicating that there are currently members present on the channel, see "TR3":#TR3 . If the flag is 1, the server will shortly perform a @SYNC@ operation as described in "RTP18":#RTP18 . If that flag is 0 or there is no @flags@ field, the presence map should be considered in sync immediately with no members present on the channel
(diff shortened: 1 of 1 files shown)


SimonWoolf:
@QuintinWillison Could you let me know if there's anything more you'd like me to do to move this forward?

QuintinWillison:
Sorry, @SimonWoolf, I do keep revisiting this one and then immediately hit by hesitancy to approve based on the challenge that we're modifying a spec point. We don't currently have any system in place to coherently and predictably manage/emit client library codebase changes arising from mutation to spec points.

It's this specific issue which I've also not got around to addressing in our [internal documentation on the subject](https://ably.atlassian.net/wiki/spaces/SDK/pages/683409414/Client+Library+Features+Spec). I need to put some more thought into this, so please bear with me.

FYI, @AndyNicks

QuintinWillison:
@paddybyers' comment on [this internal document](https://ably.atlassian.net/wiki/spaces/SDK/pages/683409414/Client+Library+Features+Spec?focusedCommentId=1753448492#comment-1753448492) in relation to whether we should allow mutation of spec points at all probably also applies here:

> The only caveat I would suggest is where the spec point change is a rewording or clarification that doesn’t change the intended behaviour.

As, if my understanding is correct, this spec point change relates to clarifying realtime behaviour... and, as such, it does not change requirements of client library implementations.

@SimonWoolf could you confirm my understanding, please?

SimonWoolf:
> this spec point change relates to clarifying realtime behaviour... and, as such, it does not change requirements of client library implementations.

Pretty much. It's a bit of an edge case because there was an actual change to _realtime_ behaviour, not just a clarification. But there was no change to the _client lib_ behaviour requirements, which remain 'if you get has_presence, expect a sync'. Does that then count as a clarification for spec purposes? I figured yes on the grounds that the change isn't something that a client lib can implement - it gets the new realtime behaviour whether it wants to or not. But if you disagree happy to change the pr accordingly.

QuintinWillison (approved):

PR #1119, opened 2021-06-17
Replace existing tutorial step 1 partial

marklewin:
[DOC-345](https://ably.atlassian.net/browse/DOC-345)

Update step 1 tutorial partial to reflect changed UI, choose either to create an app or use an existing one, and clarify info about Ably apps and API keys.

Ensure that tutorials using the existing partial now use the new one.


tbedford (approved):
LGTM. Grepped for missing replacements (number was correct). Viewed enumeration tutorial as a use case for partial. Looked good.

PR #1121, opened 2021-06-18
Use 'instantiate x' rather than 'instance x'

marklewin:
[DOC-206](https://ably.atlassian.net/browse/DOC-206)

Replace:

- 'instance the ' -> 'instantiate the'
- 'instance with(out)' -> 'instantiate with(out)'
- 'instanced' -> 'instantiated'


tbedford (changes requested):
Diff files should never be modified manually. They are generated by a script when creating new versions of the docs.

tbedford (approved):
Second review: just checked that diff files had been excluded. LGTM.

PR #1122, opened 2021-06-21
Replace ULID with UUID

ben-xD:
ULID is referenced in the feature spec.

ULID is uncommon, and introduces a 3rd party and uncommon dependency. It was also used only once in ably-cocoa, when UUID could have done the job. It has been removed by https://github.com/ably/ably-cocoa/pull/1128, and removing ULID from the feature spec would avoid having other libraries use the uncommon ULID instead of the often built-in UUID.


SimonWoolf:
I'm not sure what the reasoning was for the ULID requirement so not sure what the consequences would be for removing it. @tcard was the one so introduced it so is the best person to evaluate this, so I'm untagging myself and tagging him

tcard:
No need for ULID in particular, at least originally. The requirement was for some universal unique ID to be generated on clients, and I had used ULID in the first (Java/Android) push implementation for no special reason.

(Of course, I don't know if, since then, in application of Hyrum's Law now some part of Ably depends on those IDs being ULIDs in particular, but I don't think that's likely.)

tcard (approved):
This change is fine, but given we're doing it at all, I'd specify what we actually _require_ from the IDs: guaranteed global uniqueness and randomness (can't guess it for a particular device), and not too long (how long?).

Why does the spec "prefer" UUID? I think just mentioning it as a suggestion is good enough and raises fewer questions.

Also, there are 5 UUID versions. Since we're suggesting UUID at all, we should be specific about the version that complies with the requirements (4).

ben-xD:
Thanks @tcard , good point. I have amended it in b9e2b39

> If the local device does not have @id@ or @deviceSecret@, both are generated locally. The @id@ must be globally-unique identifier (e.g. UUID, GUID).

QuintinWillison (approved):
Thanks, @ben-xD. 😁 

As I've mentioned elsewhere (mainly internally), I do have concerns about the mutation of spec points and how that change is then relayed out to the client libraries that had already implemented that spec point - however, that concern need not block this PR landing.

The status quo is that this practice has been going on for a while, so until we create some tighter guidelines around it then this is what happens. If you are aware of any client library repositories that are using ULID at the moment then perhaps you could add issues to them specifically to get it replaced.

tcard:
> I do have concerns about the mutation of spec points

> If you are aware of any client library repositories that are using ULID at the moment then perhaps you could add issues to them specifically to get it replaced.

This change is backwards-compatible. Using ULID still complies with the spec.

PR #1135, opened 2021-07-14
Clarify our approach to semantic versioning

marklewin:

## Description

Make it clear that, although we tend to follow the semantic versioning scheme, there are differences.

* [DOC-314](https://ably.atlassian.net/browse/DOC-314).

## Review

Please check the intro para on this page:

* [Ably Versioning](https://ably-docs-pr-1135.herokuapp.com/client-lib-development-guide/versioning/)

### DIFF
--- a/content/client-lib-development-guide/versioning.textile
+++ b/content/client-lib-development-guide/versioning.textile
@@ -4,7 +4,7 @@ section: client-lib-development-guide
 index: 19
 ---
 
-Ably aims to follow the "semantic versioning scheme":https://semver.org/ for our "client library SDKs":https://ably.com/download, our "REST API endpoints":/rest-api, and all other public endpoints that support versioning such as "Server-Sent Events":/sse.
+Ably's general approach to versioning is to use the "semantic versioning scheme":https://semver.org/, with a few differences that are explained below. This applies to our "client library SDKs":https://ably.com/download, our "REST API endpoints":/rest-api, and all other public endpoints that support versioning such as "Server-Sent Events":/sse.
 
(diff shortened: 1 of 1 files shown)


content/client-lib-development-guide/versioning.textile
```diff
@@ -4,7 +4,7 @@ section: client-lib-development-guide
 index: 19
 ---
 
-Ably aims to follow the "semantic versioning scheme":https://semver.org/ for our "client library SDKs":https://ably.com/download, our "REST API endpoints":/rest-api, and all other public endpoints that support versioning such as "Server-Sent Events":/sse.
+Ably's general approach to versioning is to use the "semantic versioning scheme":https://semver.org/, with a few differences that are explained below. This applies to our "client library SDKs":https://ably.com/download, our "REST API endpoints":/rest-api, and all other public endpoints that support versioning such as "Server-Sent Events":/sse.
```
tbedford, line 7: > few differences that are explained below

Instead of "explained below" could we perhaps say something along the lines of "in the main section"? To make it clear we are not talking about in the next sentence...

> This applies to our "client

Could we perhaps say something like: Ably's versioning approach applies to... 

I only say that as for some reason I read it as: the differences applying to...

tbedford (changes requested):
Looks good. I've made a couple of minor suggestions.

content/client-lib-development-guide/versioning.textile
marklewin: All good points, thanks

tbedford (approved):

PR #1139, opened 2021-07-15
Tidy pub/sub tutorial live demo instructions

marklewin:

## Description

The live demo refers to a QR code which is not visible, so I'm suggesting that we just remove it.

Because it's not clear from the formatting that there is an editable text box and a button, I've adjusted the layout of and captions for each of those.

* [DOC-398](https://ably.atlassian.net/browse/DOC-398).

## Review

Please check that reference to QR code is removed and that it is clear how to run the live demo.

* [Page to review](https://doc-398-remove-qr-code.herokuapp.com/tutorials/publish-subscribe#live-demo)

### DIFF
--- a/content/tutorials/publish-subscribe.textile
+++ b/content/tutorials/publish-subscribe.textile
@@ -741,11 +741,7 @@ We're done, it's that simple. We have now shown you how to subscribe to messages
 
 h2(#live-demo). Live demo
 
-<div id="qr-code" style="float: right; margin: 0 0 15px 15px; width: 200px; display: none; text-align: center">
-  <div style="margin-top: 5px">Scan this on your mobile</div>
-</div>
-
-<a href="#" id="new-browser" target="_blank">Open this demo in a new browser window</a> to see publish & subscribe in action.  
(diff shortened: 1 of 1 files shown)


content/tutorials/publish-subscribe.textile
```diff
@@ -745,16 +745,18 @@ h2(#live-demo). Live demo
   <div style="margin-top: 5px">Scan this on your mobile</div>
 </div>
 
-<a href="#" id="new-browser" target="_blank">Open this demo in a new browser window</a> to see publish & subscribe in action.  Alternatively, scan the QR code to see the published message delivered in real time.
+<a href="#" id="new-browser" target="_blank">Open this demo in a new browser window</a> to see publish & subscribe in action.
```
tbedford, line 748: > to see publish & subscribe

Can we use 'and' instead of '&' in this case?

tbedford (approved):
Apart from the minor quibble, LGTM!

content/tutorials/publish-subscribe.textile
marklewin: Good shout! Done.

PR #1144, opened 2021-07-21
Refine Ably-Agent

QuintinWillison:
It's become obvious as we've started work on implementing `Ably-Agent` across client library codebases that our initial spec needed some evolution to match requirements as they've changed. This pull request addresses that need, which should hopefully make requirements clearer for client library developers.

Specifically:

- https://github.com/ably/docs/commit/f788a54b11a5c22310ea2d582f3c99e1eba18f71: Product versions do not need to be [SemVer](https://semver.org/) - and, in fact, this requirement was too restrictive for some implementations - e.g. as explicitly discovered by @tomkirbygreen for [ably-dotnet](https://github.com/ably/ably-dotnet) OS versions, and clearly present [for ably-java](https://github.com/ably/ably-java/pull/671) where we're using simpler sources of information, such as [Android's](https://developer.android.com/reference/android/os/Build.VERSION#SDK_INT) `Build.VERSION.SDK_INT` (a plain integer).
- https://github.com/ably/docs/commit/67ebf583bd182ec45ac08756e249200faa5d999a (new spec point): Semantic versioning is still a good aspiration for products emitted as part of agent strings, but this can only be best-effort as we must not misrepresent other forms (e.g. [Microsoft's](https://docs.microsoft.com/en-us/dotnet/standard/assembly/versioning#assembly-version-number), which as stated in my previous point, doesn't conform to SemVer).
- https://github.com/ably/docs/commit/7af33b11e8ccab8e9a17d9abb9968fa2e21ba2f1 (modified spec point), https://github.com/ably/docs/commit/825d52b3302b66c799576ac805fd4b7489e36f18 (new spec point): Our canonical reference location for agent products is [agents.json](https://github.com/ably/ably-common/blob/main/protocol/agents.json).

The following work, (mostly) currently in progress, needs to conform according to the changes proposed under this pull request:

- https://github.com/ably/ably-js/pull/740
- https://github.com/ably/ably-java/pull/671
- https://github.com/ably/ably-cocoa/pull/1119
- https://github.com/ably/ably-python/pull/195
- https://github.com/ably/ably-go/pull/360
- https://github.com/ably/ably-ruby/pull/248

Other work to be done, which will be impacted by the changes proposed under this pull request:

- https://github.com/ably/ably-flutter/issues/100
- https://github.com/ably/ably-php/issues/104
- https://github.com/ably/ably-dotnet/issues/498

Adding @lmars and @cruickshankpg to this review in order to check that there's no implications for the server-side implementation of this relaxation in requirements.


QuintinWillison:
Now that this has been deployed for this pull request, it may assist reviewers for me to point you [here](https://ably-docs-pr-1144.herokuapp.com/client-lib-development-guide/features/#RSC7d) to view the rendered output for this portion of the spec. 😄

owenpearson (approved):

AndyNicks (approved):
LGTM

tomkirbygreen (approved):
Totally applaud the pragmatic relaxation around `SemVer`.  Personally I think each client SDK should gather as much 'version / environment' detail as they decently can and send such back to the server with zero additional client side post-processing. Server side we should archive the raw telemetry and then we can, should we wish, post process this to normalise it into a `SemVer`-like format for public presentation on the web-site or to normalise it to feed into the whole business intelligence function. 

Keeping the raw, unprocessed telemetry can be incredibly useful when it comes to informing future Product decisions or in 11th hour debugging issues where you really care what service pack the client is running or what DLLs (or equivalent) the client has loaded into their address space.

With respect to the latter, I think client SDKs should have a 'off-by-default'  mode that returns additional telemetry (happy to help define the scope of this). Then in the event of a serious customer issue, First Line Support can ask the customer if they're happy to enable this and attach the detailed log to the issue before passing it onto the `SDK Team`. In my experience having a capability like this 'ready to go' goes a long way to reassure customers when they encounter pain that may (or may not) be down to us. Personally I think this is particularly important as we scale towards out 1-billion-connected-clients-a-month goal.

SimonWoolf (approved):

ben-xD (approved):

content/client-lib-development-guide/features.textile
```diff
@@ -81,9 +81,11 @@ h3(#restclient). RestClient
...
-*** @(RSC7d1)@ The @Agent@ library identifier is composed of a series of @key[/value]@ entries joined by spaces where @key@ is the name of a product and @value@ is an optional version of the associated product which must be a valid "semantic version":https://semver.org/. This must include at least the @<library name>/<library version>@ entry followed optionally, in dependency order, the @name[/version]@ entry for underlying library layer or platform layers. An example would be @ably-flutter/1.2.0 ably-java/1.2.1 android/24@. The documentation for each library must define the content and meaning of the @Agent@ identifiers advertised by that library.
+*** @(RSC7d1)@ The @Agent@ library identifier is composed of a series of @key[/value]@ entries joined by spaces where @key@ is the name of a product and @value@ is an optional version of the associated product. This must include at least the @<library name>/<library version>@ entry followed optionally, in dependency order, the @name[/version]@ entry for underlying library layer or platform layers. An example would be @ably-flutter/1.2.0 ably-java/1.2.1 android/24@.
```
lmars, line 84: What does "in dependency order" mean, and why is that ordering desirable?

content/client-lib-development-guide/features.textile
```diff
@@ -81,9 +81,11 @@ h3(#restclient). RestClient
...
 *** @(RSC7d2)@ Where possible, the @Agent@ library identifier should be included in all HTTP and WebSocket requests to the Ably endpoint via an @Ably-Agent@ HTTP header.
 *** @(RSC7d3)@ Where the library is unable to specify a header when making a request (such as for @JSONP@ requests), the @Agent@ library identifier should be sent via an @agent@ query param whose value is the URI-encoded @Agent@ library identifier.
+*** @(RSC7d4)@ Where possible, product versions should be represented as a valid "semantic version":https://semver.org/. This is best-effort as we recognise that strict SemVer is not always available, or otherwise possible to derive from information available at runtime, especially in the case of products that represent operating system versions.
```
lmars, line 87: This will require an equivalent relaxation of the server-side parsing of the Ably-Agent header, which currently expects a semantic version.

We could completely remove the parsing of the version and allow any string of non-whitespace characters, but this was previously problematic since absolute junk was appearing in our internal metrics.

What do you think about adding a `versionFormat` field to each agent entry in agents.json, and mapping those formats to regular expressions which are used for server-side parsing to avoid the junk? For example:

```
"agents": [
    {
        "identifier": "ably-java",
        "versioned": true,
        "emitters": [
            "ably-java"
        ],
        "type": "sdk",
        "versionFormat": "semver"
    },
    {
        "identifier": "android",
        "versioned": true,
        "emitters": [
            "ably-java",
            "ably-dotnet"
        ],
        "type": "os",
        "versionFormat": "integer"
    }
],
"versionFormats": {
    "semver": "<semver regex>",
    "integer": "\d+"
},
```

content/client-lib-development-guide/features.textile
QuintinWillison: Well spotted. I think that was slipped in original for a requirement that has now gone away. I'll remove this.

content/client-lib-development-guide/features.textile
QuintinWillison: The problem we're facing at the moment is that I think we'll struggle to define precise formats for every version we'll receive from some products (e.g. OS versions, custom Java runtimes, etc..) - as in, perhaps by definition, some level of "junk" will arrive.

Now that we've pinned down the product components, I think the problem will be much more manageable. My preference would be to remove the parsing of versions server-side for the moment, for our first iteration of this feature.

If it really becomes an issue down the line then perhaps we can look to add more filtering at the server-side. At least that will be managed in the one place, without need for complex client-side logic for every single client library.

content/client-lib-development-guide/features.textile
QuintinWillison: Removed: https://github.com/ably/docs/pull/1144/commits/45ba3b95759af30cbda06468ae774dee06f46acd.

sacOO7:
> Now that this has been deployed for this pull request, it may assist reviewers for me to point you [here](https://ably-docs-pr-1144.herokuapp.com/client-lib-development-guide/features/#RSC7d) to view the rendered output for this portion of the spec.

I clicked on the link, it says `No such app`

content/client-lib-development-guide/features.textile
lmars: Ok, I'm happy to just remove the parsing and then re-consider if we see any issues re-surface 👍

lmars (approved):

cruickshankpg (approved):

content/client-lib-development-guide/features.textile
```diff
@@ -81,9 +81,11 @@ h3(#restclient). RestClient
...
```
paddybyers, line 87: I don't think "where possible, use semver" is now what we really intend. We're saying "use the version identifier advertised by the component in question". There presumably also should be some rules about how we escape those things so as to not interfere with the parsing of the full agent string, and to ensure that the full string does not contain characters that are not allowed in a header string.

QuintinWillison: This comes down to interpretation of this spec point so if I've made it unclear then I am happy to try to evolve it.

What you capture in your comment here was intended to implicit in what I wrote in the spec point. I am, indeed, saying that the version identifier used by the component in question (a.k.a. product) should be used. But what I am also saying is that if we have any control over the versioning system in use then we should elect to use SemVer (i.e. for our own libraries as well as wrappers we create - upper layers like Flutter plugin).

QuintinWillison: @paddybyers and I have further discussed this [internally](https://ably-real-time.slack.com/archives/C030C5YLY/p1626940588283900?thread_ts=1626839568.280300&cid=C030C5YLY) and it has been agreed that we're both, fundamentally saying the same thing.

As such, and given the number of approvers on this PR already, we'll land this pull request as it stands. This area of the spec can always be subsequently iterated if required, but for the time being it's fit for purpose.

PR #1160, opened 2021-08-03
Make RTL5K also apply to detached state

owenpearson:
As suggested by Simon in [this comment](https://github.com/ably/ably-js/pull/784#pullrequestreview-721164875).

> Imagine: client lib is in detached after an explicit detach() while in suspended after a failed attach, and then the server finally catches up and sends an ATTACHED; we want to enforce the user's stated preference and send a DETACH rather than just acquiesce and end up in the attached state after the user called detach()

As far as I can tell there's no existing special behaviour defined by the spec when ATTACHED is received in the `detached` state so this doesn't conflict with any existing spec points but it's worth having a second pair of eyes to confirm this?

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -544,7 +544,7 @@ h3(#realtime-channel). Channel
 ** @(RTL5h)@ If the connection state is @CONNECTING@ or @DISCONNECTED@, do the detach operation once the connection state is @CONNECTED@
 ** @(RTL5d)@ Otherwise a @DETACH@ ProtocolMessage is sent to the server, the state transitions to @DETACHING@ and the channel becomes @DETACHED@ when the confirmation @DETACHED@ ProtocolMessage is received
 ** @(RTL5f)@ Once a @DETACH@ @ProtocolMessage@ is sent, if a @DETACHED@ @ProtocolMessage@ is not received within the "default realtime request timeout":#defaults, the detach request should be treated as though it has failed and the channel will return to its previous state
-** @(RTL5k)@ If the channel receives an @ATTACHED@ message while in the @DETACHING@ or @DETACHED@ state, it should send a new @DETACH@ message and remain in the @DETACHING@ state
+** @(RTL5k)@ If the channel receives an @ATTACHED@ message while in the @DETACHING@ or @DETACHED@ state, it should send a new @DETACH@ message and remain in (or transition to) the @DETACHING@ state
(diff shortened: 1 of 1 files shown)


content/client-lib-development-guide/features.textile
```diff
@@ -544,7 +544,7 @@ h3(#realtime-channel). Channel
 ** @(RTL5h)@ If the connection state is @CONNECTING@ or @DISCONNECTED@, do the detach operation once the connection state is @CONNECTED@
 ** @(RTL5d)@ Otherwise a @DETACH@ ProtocolMessage is sent to the server, the state transitions to @DETACHING@ and the channel becomes @DETACHED@ when the confirmation @DETACHED@ ProtocolMessage is received
 ** @(RTL5f)@ Once a @DETACH@ @ProtocolMessage@ is sent, if a @DETACHED@ @ProtocolMessage@ is not received within the "default realtime request timeout":#defaults, the detach request should be treated as though it has failed and the channel will return to its previous state
-** @(RTL5k)@ If the channel receives an @ATTACHED@ message while in the @DETACHING@ state, it should send a new @DETACH@ message and remain in the @DETACHING@ state
+** @(RTL5k)@ If the channel receives an @ATTACHED@ message while in the @DETACHING@ or @DETACHED@ state, it should send a new @DETACH@ message and remain in the @DETACHING@ state
```
SimonWoolf, line 547: > `remain in the @DETACHING@ state`

should probably now be 'remain (or transition) to'

SimonWoolf (approved):

content/client-lib-development-guide/features.textile
owenpearson: whoops, 31e2c7f.

AndyNicks (approved):

QuintinWillison (changes requested):
I believe this is a change in client library behaviour, so should probably be a new spec point. See [this internal comment](https://ably.atlassian.net/wiki/spaces/SDK/pages/683409414/Client+Library+Features+Spec?focusedCommentId=1753448492#comment-1753448492).

content/client-lib-development-guide/features.textile
QuintinWillison: Resolving this conversation on @SimonWoolf's behalf so that we can merge this pull request cleanly, given it seems clear that @owenpearson has included the suggested change.

QuintinWillison:
Given [my comment](https://github.com/ably/docs/pull/1160#pullrequestreview-758430769) was not responded to, I'll take it that my point was disregarded. Therefore I'm going to land this pull request as it's been open for a long time.

PR #1161, opened 2021-08-04
DOC-369: Update Ruby connection example to generate demo API key in Quickstart

m-hulbert:

## Description

This PR fixes the Ruby connection example code so that it generates a demo API key when published. See [JIRA](https://ably.atlassian.net/browse/DOC-369) for further information.

## Review

Run the website locally against this PR and check that the Ruby code block generates a demo API key.

### DIFF
--- a/content/root/quick-start-guide.textile
+++ b/content/root/quick-start-guide.textile
@@ -231,9 +231,9 @@ ably.connection.on(ConnectionState.connected, new ConnectionStateListener()
 
 bc[ruby]. EventMachine.run do
   ably = Ably::Realtime.new(key: '{{API_KEY}}')
-end
-ably.connection.on(:connected) do
-  puts "Connected to Ably!"
+  ably.connection.on(:connected) do
+    puts "Connected to Ably!"
+  end
 end
 
 bc[swift]. let ably = ARTRealtime(key: "{{API_KEY}}")
--- a/content/root/quick-start-guide.textile
+++ b/content/root/quick-start-guide.textile
@@ -230,7 +230,7 @@ ably.connection.on(ConnectionState.connected, new ConnectionStateListener()
 });
 
 bc[ruby]. EventMachine.run do
-  ably = Ably::Realtime.new(api_key)
+  ably = Ably::Realtime.new(key: '{{API_KEY}}')
 end
 ably.connection.on(:connected) do
   puts "Connected to Ably!"


content/root/quick-start-guide.textile
```diff
@@ -230,7 +230,7 @@ ably.connection.on(ConnectionState.connected, new ConnectionStateListener()
 });
 
 bc[ruby]. EventMachine.run do
-  ably = Ably::Realtime.new(api_key)
+  ably = Ably::Realtime.new(key: '{{API_KEY}}')
```
marklewin, line 233: I'm in no way a ruby expert, but I don't think that `API_KEY` should be surrounded by `{{ }}`. To me, that makes it look like some [sort of handlebars-style templating](https://github.com/SmartBear/ruby-handlebars) is going on. If you look at a Ruby version of [one of our tutorials](https://ably.com/tutorials/publish-subscribe#step-2), you'll see the following form, which I think would be preferable, but it might be worth asking one of the Rubyists on the web team what they think!

```ruby
require 'ably'
EventMachine.run do
  ably = Ably::Realtime.new(key: 'API_KEY')
end
```

marklewin (changes requested):

content/root/quick-start-guide.textile
m-hulbert: Sorry, I should have written better instructions for testing this one (I'll update the description).

If you look at the [live quickstart](https://ably.com/documentation/quick-start-guide) in Ruby, the code block is:

```ruby
EventMachine.run do
  ably = Ably::Realtime.new(api_key)
end
ably.connection.on(:connected) do
  puts "That was simple, you're now connected to Ably in realtime"
end
```

`{{API_KEY}}` is a variable that injects a demo API key into the code block, but this only gets replaced in the website view (so won't appear correctly in the demo app for this PR). If you run the website locally against this PR then it will be replaced with a valid demo API key.

lmars: Hey guys, I just saw this in passing, but this Ruby snippet is incorrect, all the code should be inside the `EventMachine.run` block, like this:

```ruby
EventMachine.run do
  ably = Ably::Realtime.new(key: '{{API_KEY}}')

  ably.connection.on(:connected) do
    puts "Connected to Ably!"
  end
end
```

There is a reference to this requirement [here](https://github.com/ably/ably-ruby#introduction), although it could perhaps be clearer.

m-hulbert: Thanks for spotting that @lmars - I've fixed that as well!

marklewin (approved):
LGTM - thanks both for the clarifications!

PR #1192, opened 2021-09-07
Add missing param to `GET rest.ably.io/push/channelSubscriptions` and update PushChannel docs with this param.

ben-xD:
Add documentation on `PushChannel#listSubscriptions`.
- The documentation was copy and pasted from an irrelevant method, and was completely wrong.
- The `concatParams` parameter was missing from the Rest API docs.
- This method actually requires `Push Admin` capability. This seems like a bug in realtime, but it does, so I have documented it so users understand this behaviour.


content/partials/types/_push_channel.textile
```diff
@@ -41,9 +41,9 @@ h6(#list-subscriptions).
 
 bq(definition).
   java,android: "PaginatedResult":#paginated-result<"PushChannelSubscription":#push-channel-subscription> listSubscriptions(String deviceId, String clientId, String deviceClientId, String channel)
-  objc,swift: listSubscriptions(deviceId: String?, clientId: String?, deviceClientId: String?, channel: String?, callback: ("ARTPaginatedResult":#paginated-result<"PushChannelSubscription":#push-channel-subscription>?, ARTErrorInfo?) -> Void)
+  objc,swift: listSubscriptions(deviceId: String?, clientId: String?, deviceClientId: String?, `channel: String?, callback: ("ARTPaginatedResult":#paginated-result<"PushChannelSubscription":#push-channel-subscription>?, ARTErrorInfo?) -> Void)
 
-"Unsubscribe all devices associated with your device's clientId":/general/push/activate-subscribe#subscribing-client-id from the channel's push notifications.
+Lists push subscriptions on a specified channel name (@channel@) for @clientId@, @deviceId@ or both if @concatFilters@ is set to @true@. This method requires clients to have the "Push Admin capability":/general/push/admin. For more information, take a look at the @GET rest.ably.io/push/channelSubscriptions@ "Rest API":/rest-api.
```
marklewin, line 46: Maybe this would be a little clearer:

"Lists push subscriptions on a specified channel (@channel@). You can filter the list of subscriptions by specifying either a @deviceId@ or @clientId@. To include both @deviceId@ and @clientId@ in your filter, set @concatFilters@ to @true@."

content/rest-api/index.textile
```diff
@@ -1004,6 +1004,7 @@ h5. Parameters
 - channel := Filter to restrict to subscriptions associated with that @channel@.
 - clientId := Filter to restrict to subscriptions associated with that @clientId@. Cannot be used with @deviceId@.
```
marklewin, line 1005: Might be clearer if we say "Do not specify both @deviceId@ and @clientId@, unless @concatFilters@ is set to true."

marklewin (changes requested):
Some suggestions to improve clarity

content/partials/types/_push_channel.textile
ben-xD: I took that suggestion and did something similar here: 40063a5d016a91252616fc754d4f1fef997d001c

content/rest-api/index.textile
ben-xD: I have done something slightly different, but mostly takes care of this concern:  d152b7835730411d29d9c5f9f5ba7345e6084c05

The main annoyance here is the API itself though. If `clientId` and `deviceId` are set, it should return subscriptions for both, not throw an error. Then `concatFilters` can be removed. More of my concerns about this API is https://ably-real-time.slack.com/archives/CURL4U2FP/p1630998668011900

content/partials/types/_push_channel.textile
```diff
@@ -41,9 +41,9 @@ h6(#list-subscriptions).
...
-  objc,swift: listSubscriptions(deviceId: String?, clientId: String?, deviceClientId: String?, channel: String?, callback: ("ARTPaginatedResult":#paginated-result<"PushChannelSubscription":#push-channel-subscription>?, ARTErrorInfo?) -> Void)
+  objc,swift: listSubscriptions(deviceId: String?, clientId: String?, deviceClientId: String?, `channel: String?, callback: ("ARTPaginatedResult":#paginated-result<"PushChannelSubscription":#push-channel-subscription>?, ARTErrorInfo?) -> Void)
 
-"Unsubscribe all devices associated with your device's clientId":/general/push/activate-subscribe#subscribing-client-id from the channel's push notifications.
+Lists push subscriptions on a channel specified by it's channel name (@channel@). These subscriptions can be either client (@clientId@) subscriptions, device (@deviceId@) subscriptions, or if @concatFilters@ is set to @true@, both. This method requires clients to have the "Push Admin capability":/general/push/admin. For more information, take a look at the @GET rest.ably.io/push/channelSubscriptions@ "Rest API":/rest-api.
```
marklewin, line 46:
```suggestion
Lists push subscriptions on the specified channel (@channel@). These subscriptions are either client (@clientId@) subscriptions, device (@deviceId@) subscriptions, or if @concatFilters@ is set to @true@, both. This method requires clients to have the "Push Admin capability":/general/push/admin. For more information, see the @GET rest.ably.io/push/channelSubscriptions@ "Rest API":/rest-api.
```

marklewin: (it's -> its)

marklewin (changes requested):

content/partials/types/_push_channel.textile
ben-xD: I did some of those things on purpose. 😅 e.g.  
- `a channel specified by it's channel name (@channel@)`. The argument is the channel name, the entity is a channel. 
- `These subscriptions can be either`. There is no such thing as a subscription that is simultaneously both clientId and deviceId. _**They can be either**_. 🙂. Your suggestion makes it seem like a subscription can be both. 
- Ill change `take a look at` to `see`, but I'm not sure if it makes a difference here (happy to be convinced though 🤓).

ben-xD: I've elaborated the original text for point 1 above: ea5363f58a472b753ed2a510b003a2af00ee02c6
I've fixed the 3rd and 4th point above: b5003df7b537c6f11de77159b41c7427a1d03395

content/rest-api/index.textile
```diff
@@ -1002,8 +1002,9 @@ bc[sh]. curl https://rest.ably.io/push/channelSubscriptions \
 h5. Parameters
 
 - channel := Filter to restrict to subscriptions associated with that @channel@.
-- clientId := Filter to restrict to subscriptions associated with that @clientId@. Cannot be used with @deviceId@.
-- deviceId := Filter to restrict to subscriptions for that @deviceId@. Cannot be used with @clientId@.
+- clientId := Filter to restrict to subscriptions associated with that @clientId@. Cannot be used with @deviceId@, unless @concatFilters@ is set to @true@.
+- deviceId := Filter to restrict to subscriptions for that @deviceId@. Cannot be used with @clientId@, unless @concatFilters@ is set to @true@.
+- concatFilters := _false_ Find all device registrations which match either @clientId@ or @deviceId@. Set this to @true@ and set both @clientId@ and @deviceId@.
```
QuintinWillison, line 1007: Perhaps it's me misunderstanding the function being described, but I think this sentence needs fixing like this:

```suggestion
- concatFilters := _false_ Find all device registrations which match either @clientId@ or @deviceId@. Set this to @true@ in order to set both @clientId@ and @deviceId@.
```

QuintinWillison (changes requested):

marklewin (approved):
LGTM

PR #1238, opened 2021-10-21
DOC-384: Update attach vs subscribe best practice section

m-hulbert:

## Description

This PR updates the attach vs subscribe section of the best practice guide to be more explicit.

## Review

View the [best practice guide](https://ably-docs-pr-1236.herokuapp.com/best-practice-guide#attach-vs-subscribe) to review this PR.

### DIFF
--- a/content/root/best-practice-guide.textile
+++ b/content/root/best-practice-guide.textile
@@ -83,9 +83,13 @@ h2(#for-experienced-developers). For experienced developers
 
 In order to make the best use of Ably, you need to be aware of the following information:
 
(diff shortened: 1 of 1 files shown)


paddybyers:
@SimonWoolf  fyi
I think it is worth saying explicitly (here, or maybe elsewhere) that if you do `subscribe()`, then `unsubscribe()`, the effect of that is that you remain attached (and explain why)

SimonWoolf:
IMO we should be stating things like this - which are ultimately how attach/subscribe work -- in the api docs for attach, subscribe, detach, and unsubscribe. Not just in a best-practices doc, which most people reading the api docs won't see, or probably even know exists. (_I'd_ forgotten it exists!)

PR #1250, opened 2021-11-04
Token revocation docs

tbedford:

## Description

This PR adds a Token Revocation section to the Auth & Security topic, and also a small section in the REST API Specification topic.

* [Jira epic ticket](https://ably.atlassian.net/browse/DOC-517) - DOC-538, DOC-537, DOC-540

## Review

* [Main topic page](https://ably-docs-pr-1250.herokuapp.com/core-features/authentication/#token-revocation)
* [REST API Spec page](http://ably-docs-pr-1250.herokuapp.com/rest-api/#revoke-tokens)

Note: For content review must be tested under web view for actual rendering.

### DIFF
--- a/content/core-features/authentication.textile
+++ b/content/core-features/authentication.textile
@@ -449,7 +449,7 @@ There can be a maximum of 100 target specifiers in a single request.
 
 The token revocation API is rate-limited, so there is a maximum global aggregate rate of revocation requests per app. The rate is configurable by Ably at the application level. As part of the process of enabling revocation on an app Ably will ask you what the maximum rate of token revocations on that app needs to be, and then provision that capacity. The decision is not permanent, it can be changed by Ably on request at any time.
 
-Note that revocation is not a permanent prohibition of any client, connection, @clientId@ or channel. It invalidates token credentials issued before a certain point in time. Ultimately, the app that issues tokens is the arbiter of whether or not a client is able to continue to interact with the Ably service.
+Note that revocation does not permanently forbid a client, connection, @clientId@ or channel. It invalidates token credentials issued before a certain point in time. Ultimately, the app that issues tokens is responsible for whether or not a client is able to continue to interact with the Ably service.
 
 Do not expect to be able to enable token revocation on an app, or the @revocableTokens@ attribute of a key, and then revoke a token that's been used since before you made that change. If you want to be able to revoke a token, revocation should have been enabled at the time that the token was issued and first used.
 
--- a/content/core-features/authentication.textile
+++ b/content/core-features/authentication.textile
@@ -392,7 +392,7 @@ h3. Revocable tokens
 
 Tokens are revocable if they are created from an API key that has a @revocableTokens@ attribute set to @true@. 
 
-Tokens created from a key with the @revocableTokens@ attribute set to @true@ are restricted to a TTL of 1 hour; an attempt to create, or present, a token that is revocable, but has an TTL of longer than 1 hr, is rejected with an error.
+Tokens created from a key with the @revocableTokens@ attribute set to @true@ are restricted to a TTL of 1 hour; an attempt to create, or present, a token that is revocable, but has a TTL of longer than 1 hr, is rejected with an error.
 
 A JWT may include the following additional claim:
 
--- a/content/core-features/authentication.textile
+++ b/content/core-features/authentication.textile
@@ -386,7 +386,7 @@ Token revocation has to be enabled for you by Ably. There is a performance cost 
 
 h3(#development-status). Development status
 
-The revocable tokens feature is currently in Beta.
+The token revocation feature is currently in Beta.
 
 h3. Revocable tokens
 
@@ -399,6 +399,8 @@ A JWT may include the following additional claim:
 - @x-ably-revocation-key@ := a string used to identify which token(s) to revoke in the revocation request.
 
 Designating a revocation key for a token, or a group of tokens, enables the revocation process to be used at any level of granularity, depending on the needs of the application.
+
+**Note:** Token revocation can only revoke tokens that were issued by an API key that had the @revocableTokens@ attribute set before the token was issued.
 
 h3. Token revocation API
 
@@ -431,17 +433,17 @@ In that request, the request body has the following form:
 
 ```[json]
 {
-    targets: [/* ... array of target specifier strings ... */],
-    issuedBefore: number (timestamp), /* (optional) a Unix timestamp where tokens issued before this time are invalidated */
-    allowReauthMargin: bool /* (optional) if true, allows a 30 seconds margin for clients to obtain a new auth token online, default false */
+    targets: [ "clientId:client1@example.com", "clientId:client2@example.com" ],
+    issuedBefore: 1636556083000,
+    allowReauthMargin: true
 }
 ```
 
 The request body has the following properties:
 
 - targets := An array of "target specifier":#target-specifiers strings.
-- issuedBefore := If @issuedBefore@ is not specified, it is set to the current time. A request with an @issuedBefore@ in the future or more than an hour in the past will be rejected.
-- allowReauthMargin := The @allowReauthMargin@ bool permits a token renewal cycle to take place without needing established connections to be dropped, by postponing enforcement to 30 seconds in the future, and sending any existing connections a hint to obtain (and upgrade the connection to use) a new token. It defaults to @false@, meaning that the effect is near-immediate.
+- issuedBefore := Optional number. A Unix timestamp in milliseconds where tokens issued before this time are invalidated. If @issuedBefore@ is not specified, it is set to the current time. A request with an @issuedBefore@ in the future or more than an hour in the past will be rejected.
+- allowReauthMargin := Optional boolean. The @allowReauthMargin@ boolean permits a token renewal cycle to take place without needing established connections to be dropped, by postponing enforcement to 30 seconds in the future, and sending any existing connections a hint to obtain (and upgrade the connection to use) a new token. It defaults to @false@, meaning that the effect is near-immediate.
 
 When invoking the revocation API, a client must prove possession of the key used to issue the tokens to be revoked. You can't use a key to revoke a token issued from a different key. This can be done by using basic authentication, using the API key itself.
 
@@ -450,8 +452,6 @@ There can be a maximum of 100 target specifiers in a single request.
 The token revocation API is rate-limited, so there is a maximum global aggregate rate of revocation requests per app. The rate is configurable by Ably at the application level. As part of the process of enabling revocation on an app Ably will ask you what the maximum rate of token revocations on that app needs to be, and then provision that capacity. The decision is not permanent, it can be changed by Ably on request at any time.
 
 Note that revocation does not permanently forbid a client, connection, @clientId@ or channel. It invalidates token credentials issued before a certain point in time. Ultimately, the app that issues tokens is responsible for whether or not a client is able to continue to interact with the Ably service.
-
-Do not expect to be able to enable token revocation on an app, or the @revocableTokens@ attribute of a key, and then revoke a token that's been used since before you made that change. If you want to be able to revoke a token, revocation should have been enabled at the time that the token was issued and first used.
 
 h4. Examples
 
--- a/content/core-features/authentication.textile
+++ b/content/core-features/authentication.textile
@@ -384,13 +384,15 @@ Without token revocation, in order to restrict potential rogue access, you would
 
 Token revocation has to be enabled for you by Ably. There is a performance cost to enabling this feature. "Contact Ably":https://ably.com/contact if you'd like to enable this feature for a specific application.
 
+h3(#development-status). Development status
+
+The revocable tokens feature is currently in Beta.
+
 h3. Revocable tokens
 
-Tokens are revocable if they are created from an API key that has a @revocable@ attribute set to @true@. 
+Tokens are revocable if they are created from an API key that has a @revocableTokens@ attribute set to @true@. 
 
-Tokens created from a key with the @revocable@ attribute set to @true@ are restricted to a TTL of 1 hour; an attempt to create, or present, a token that is revocable, but has an TTL of longer than 1 hr, is rejected with an error.
-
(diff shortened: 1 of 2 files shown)


content/core-features/authentication.textile
```diff
@@ -373,6 +374,125 @@ The following JavaScript example demonstrates how to issue an "Ably Token":#ably
...
+
+The main use case of token revocation is as a method to combat malicious clients. In cases where the client behaves safely, you don't need token revocation - your code can instruct the connection to Ably to close (or detach from a channel it shouldn't have access to). But if you want to prevent any chance of a user obtaining the token before it is disconnected, and using it in their own custom client to gain access to Ably, then token revocation can be used.
+
+Without token revocation, in order to restrict potential rogue access, you would make tokens with a short TTL (say 10 minutes) and then simply not issue a new token to a suspect client. Unfortunately, reducing the TTL would increase load on your authentication server, so token revocation can provide a more scalable solution to the issue of rogue clients.
+
+Token revocation has to be enabled for you by Ably. There is a performance cost to enabling this feature. "Contact Ably":https://ably.com/contact if you'd like to enable this feature for a specific application.
+
+h3. Revocable tokens
+
+Tokens are revocable if they are created from an API key that has a @revocable@ attribute set to @true@.
```
SimonWoolf, line 389: The attribute is `revocableTokens` rather than `revocable` (keys themselves are always revocable)

(not that 99% of people will care about the actual raw attribute name, they'll just be ticking a box in the UI, to be decided what that box will be labelled)

content/core-features/authentication.textile
```diff
@@ -373,6 +374,125 @@ The following JavaScript example demonstrates how to issue an "Ably Token":#ably
...
+
+The main use case of token revocation is as a method to combat malicious clients. In cases where the client behaves safely, you don't need token revocation - your code can instruct the connection to Ably to close (or detach from a channel it shouldn't have access to). But if you want to prevent any chance of a user obtaining the token before it is disconnected, and using it in their own custom client to gain access to Ably, then token revocation can be used.
+
+Without token revocation, in order to restrict potential rogue access, you would make tokens with a short TTL (say 10 minutes) and then simply not issue a new token to a suspect client. Unfortunately, reducing the TTL would increase load on your authentication server, so token revocation can provide a more scalable solution to the issue of rogue clients.
+
+Token revocation has to be enabled for you by Ably. There is a performance cost to enabling this feature. "Contact Ably":https://ably.com/contact if you'd like to enable this feature for a specific application.
```
SimonWoolf, line 385: Should we say explicitly that it's in early access / beta and there's a chance that API may be changed before general availability? cc @ben-pickering

content/core-features/authentication.textile
```diff
@@ -373,6 +374,125 @@ The following JavaScript example demonstrates how to issue an "Ably Token":#ably
...
+
+h3. Revocable tokens
+
+Tokens are revocable if they are created from an API key that has a @revocable@ attribute set to @true@. 
+
+Tokens created from a key with the @revocable@ attribute set to @true@ are restricted to a TTL of 1 hour; an attempt to create, or present, a token that is revocable, but has an TTL of longer than 1 hr, is rejected with an error.
+
+This functionality applies to client-signed JWTs and Ably-issued tokens, the latter being tokens obtained using the "@requestToken()@ API":/realtime/authentication#request-token).
```
SimonWoolf, line 393: This paragraph seems unnecessary -- all tokens are either JWTs or ably-issued tokens, so it's not excluding anything. (the way it's worded might make someone think that it doesn't apply if you're using [the auth flow where](https://ably.com/documentation/core-features/authentication#token-request-process) you supply a tokenrequest to a client and it turns that into a token? but that works fine, because the client just turns it into a token).

content/core-features/authentication.textile
```diff
@@ -373,6 +374,125 @@ The following JavaScript example demonstrates how to issue an "Ably Token":#ably
...
+}
+```
+
+The request body has the following properties:
+
+- targets := An array of "target specifier":#target-specifiers strings.
+- issuedBefore := If @issuedBefore@ is not specified, it is set to the current time. A request with an @issuedBefore@ in the future or more than an hour in the past will be rejected.
+- allowReauthMargin := The @allowReauthMargin@ bool permits a token renewal cycle to take place without needing established connections to be dropped, by postponing enforcement to 30 seconds in the future, and sending any existing connections a hint to obtain (and upgrade the connection to use) a new token. It defaults to @false@, meaning that the effect is near-immediate.
+
+When invoking the revocation API, a client must prove possession of the API key in question. This can be done by using basic authentication, using the API key itself.
```
SimonWoolf, line 444: Probably worth being explicit here that "the API key in question" means the key from which the tokens you're trying to revoke were issued. You can't use a key to revoke a token issued from a different key.

content/core-features/authentication.textile
```diff
@@ -373,6 +374,125 @@ The following JavaScript example demonstrates how to issue an "Ably Token":#ably
...
+
+There can be a maximum of 100 target specifiers in a single request.
+
+The token revocation API is rate-limited, so there is a maximum global aggregate rate of revocation requests per app. The rate is configurable by Ably at the application level. As part of the process of enabling revocation on an app Ably will ask you what the maximum rate of token revocations on that app needs to be, and then provision that capacity. The decision is not permanent, it can be changed by Ably on request at any time.
+
+Note that revocation is not a permanent prohibition of any client, connection, @clientId@ or channel. It invalidates token credentials issued before a certain point in time. Ultimately, the app that issues tokens is the arbiter of whether or not a client is able to continue to interact with the Ably service.
+
+It may be that if you enable token revocation, and attempt to revoke a token already created earlier, then that token cannot be guaranteed to be revoked if it has already been cached.
```
SimonWoolf, line 452: This could maybe be a bit more explicit? Maybe "Do not expect to be able to enable token revocation on an app, or the `revocableTokens` attribute of a key, and then revoke a token that's been in-use since before you made that change. If you want to be able to revoke a token, revocation should have been enabled at the time that the token is issued and first used."

(the fact that it _might_ work anyway is probably irrelevant - we can't guarantee it so for security purposes the customer should assume it won't work)

content/core-features/authentication.textile
```diff
@@ -373,6 +374,125 @@ The following JavaScript example demonstrates how to issue an "Ably Token":#ably
...
+
+```[json]
+{
+    "targets": [ "revocationKey:client1@example.com" ]
+}
+```
+
+h3. Revocation errors
+
+System error responses arising from the use of an invalidated token will include an Ably @ErrorInfo@ with a @code@ of 40146 and an informative error message. This includes REST API error responses, and Ably connection @disconnected@ state changes. Such a token error is non-fatal, which means that client libraries can react to that error by triggering a token renewal where the given library options allow (that is, if an @authCallback@ or @authURL@ is specified).
```
SimonWoolf, line 490: 40141, not 40146 - https://github.com/ably/ably-common/blob/main/protocol/errors.json#L55

content/core-features/authentication.textile
```diff
@@ -373,6 +374,125 @@ The following JavaScript example demonstrates how to issue an "Ably Token":#ably
...
+
+There can be a maximum of 100 target specifiers in a single request.
+
+The token revocation API is rate-limited, so there is a maximum global aggregate rate of revocation requests per app. The rate is configurable by Ably at the application level. As part of the process of enabling revocation on an app Ably will ask you what the maximum rate of token revocations on that app needs to be, and then provision that capacity. The decision is not permanent, it can be changed by Ably on request at any time.
+
+Note that revocation is not a permanent prohibition of any client, connection, @clientId@ or channel. It invalidates token credentials issued before a certain point in time. Ultimately, the app that issues tokens is the arbiter of whether or not a client is able to continue to interact with the Ably service.
+
+It may be that if you enable token revocation, and attempt to revoke a token already created earlier, then that token cannot be guaranteed to be revoked if it has already been cached.
+
+h4. Curl examples
```
SimonWoolf, line 454: While I see the reasons the `rest-api` docs to use curl to show what's actually going on, seems like the main docs should show how to do this with client libraries, ie `rest.request`? eg as https://ably.com/documentation/rest/batch#batch-publish does. After all we do [recommend that people use a client library rather than the raw rest api whenever possible](https://faqs.ably.com/should-i-use-the-rest-api-directly)

SimonWoolf (commented):
few comments, mostly looks good so far 👍

content/core-features/authentication.textile
tbedford: Changed.

content/core-features/authentication.textile
tbedford: Good point. I've added a section stating current development status is Beta.

content/core-features/authentication.textile
tbedford: Removed.

content/core-features/authentication.textile
tbedford: Improved the wording.

content/core-features/authentication.textile
tbedford: Reworded to be clearer.

content/core-features/authentication.textile
tbedford: Changed to 40141

content/core-features/authentication.textile
tbedford: I've changed the examples to use the REST client library SDK.

content/core-features/authentication.textile
```diff
@@ -373,6 +374,127 @@ The following JavaScript example demonstrates how to issue an "Ably Token":#ably
...
+
+```
+POST /keys/<keyName>/revokeTokens
+```
+
+Where the @keyName@ is @appId.keyId@.
+
+In that request, the request body has the following form:
+
+```[json]
```
m-hulbert, line 432: I'm not sure the comments add much to this code sample given the properties are defined below. I wonder if example values would be better?

content/core-features/authentication.textile
```diff
@@ -373,6 +374,127 @@ The following JavaScript example demonstrates how to issue an "Ably Token":#ably
...
+
+There can be a maximum of 100 target specifiers in a single request.
+
+The token revocation API is rate-limited, so there is a maximum global aggregate rate of revocation requests per app. The rate is configurable by Ably at the application level. As part of the process of enabling revocation on an app Ably will ask you what the maximum rate of token revocations on that app needs to be, and then provision that capacity. The decision is not permanent, it can be changed by Ably on request at any time.
+
+Note that revocation is not a permanent prohibition of any client, connection, @clientId@ or channel. It invalidates token credentials issued before a certain point in time. Ultimately, the app that issues tokens is the arbiter of whether or not a client is able to continue to interact with the Ably service.
+
+Do not expect to be able to enable token revocation on an app, or the @revocableTokens@ attribute of a key, and then revoke a token that's been used since before you made that change. If you want to be able to revoke a token, revocation should have been enabled at the time that the token was issued and first used.
```
m-hulbert, line 454: I wonder if this paragraph should be under the "revocable tokens" header and a one-liner? E.g:

"Token revocation can only revoke tokens that were issued by an API key that had the @revocableTokens@ attribute set before the token was issued."

m-hulbert (changes requested):
Looks good! Just a couple of suggestions.

content/core-features/authentication.textile
```diff
@@ -373,6 +374,127 @@ The following JavaScript example demonstrates how to issue an "Ably Token":#ably
...
+
+When invoking the revocation API, a client must prove possession of the key used to issue the tokens to be revoked. You can't use a key to revoke a token issued from a different key. This can be done by using basic authentication, using the API key itself.
+
+There can be a maximum of 100 target specifiers in a single request.
+
+The token revocation API is rate-limited, so there is a maximum global aggregate rate of revocation requests per app. The rate is configurable by Ably at the application level. As part of the process of enabling revocation on an app Ably will ask you what the maximum rate of token revocations on that app needs to be, and then provision that capacity. The decision is not permanent, it can be changed by Ably on request at any time.
+
+Note that revocation is not a permanent prohibition of any client, connection, @clientId@ or channel. It invalidates token credentials issued before a certain point in time. Ultimately, the app that issues tokens is the arbiter of whether or not a client is able to continue to interact with the Ably service.
```
m-hulbert, line 452: Sorry missed one suggestion:

```suggestion
Note that revocation does not a permanently forbid a client, connection, @clientId@ or channel. It invalidates token credentials issued before a certain point in time. Ultimately, the app that issues tokens is responsible for whether or not a client is able to continue to interact with the Ably service.
```

content/core-features/authentication.textile
```diff
@@ -373,6 +374,127 @@ The following JavaScript example demonstrates how to issue an "Ably Token":#ably
...
+
+The main use case of token revocation is as a method to combat malicious clients. In cases where the client behaves safely, you don't need token revocation - your code can instruct the connection to Ably to close (or detach from a channel it shouldn't have access to). But if you want to prevent any chance of a user obtaining the token before it is disconnected, and using it in their own custom client to gain access to Ably, then token revocation can be used.
+
+Without token revocation, in order to restrict potential rogue access, you would make tokens with a short TTL (say 10 minutes) and then simply not issue a new token to a suspect client. Unfortunately, reducing the TTL would increase load on your authentication server, so token revocation can provide a more scalable solution to the issue of rogue clients.
+
+Token revocation has to be enabled for you by Ably. There is a performance cost to enabling this feature. "Contact Ably":https://ably.com/contact if you'd like to enable this feature for a specific application.
+
+h3(#development-status). Development status
+
+The revocable tokens feature is currently in Beta.
```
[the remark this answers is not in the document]
tbedford: Yes! :)

content/core-features/authentication.textile
tbedford: Removed and expanded properties below where necessary. Replaced by actual values.

content/core-features/authentication.textile
tbedford: Modified and moved.

m-hulbert (approved):

SimonWoolf (approved):

PR #1262, opened 2021-11-23
DOC-326: Add a section on detaching to channels docs

m-hulbert:

## Description

This PR adds a section to the realtime channel documentation on detaching from channels. See [JIRA](https://ably.atlassian.net/browse/DOC-326) for further information. 

## Review

View the [channels](https://ably-docs-pr-1262.herokuapp.com/realtime/channels) documentation to review the PR.

### DIFF
--- a/content/realtime/channels.textile
+++ b/content/realtime/channels.textile
@@ -24,6 +24,7 @@ jump_to:
     - Publishing to multiple channels#multi-publish
     - Transient Publishing#transient-publish
     - Channel states#channel-states
+    - Detaching from a channel#channel-detach
     - Handling failures#handling-failures
     - Channel namespaces#channel-namespaces
     - Presence#presence-api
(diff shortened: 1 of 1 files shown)


content/realtime/channels.textile
```diff
@@ -813,6 +814,55 @@ Previously registered listeners can be removed individually or all together.
     stateChangeListener.cancel();
 ```
 
+
+h3(#channel-detach). Detaching from a channel
+
+A client can detach from a channel so that they no longer receive any messages published to that channel. Detaching is different to unsubscribing from a channel as "@unsubscribe()@":#unsubscribe is a client-side operation, meaning Ably is unaware of whether or not a client has unsubscribed from the channel. Messages will continue to be streamed to the client until "@detach()@":#detach is called.
```
marklewin, line 820: Suggest referring to a client as "it" to make it clear that we're talking about a bit of code rather than a company that pays for our services?

```suggestion
A client can detach from a channel so that it no longer receives any messages published to that channel. Detaching is different to unsubscribing from a channel as "@unsubscribe()@":#unsubscribe is a client-side operation, meaning Ably is unaware of whether or not a client has unsubscribed from the channel. Messages will continue to be streamed to the client until "@detach()@":#detach is called.
```

marklewin: Perhaps we can simplify this sentence, for example:

"A client can detach from a channel so that it no longer receives any messages published to that channel. Detaching is different to unsubscribing from a channel because "@unsubscribe()@":#unsubscribe is a client-side operation. The Ably platform does not know that a client has unsubscribed and will continue to stream messages to that client until "@detach()@":#detach is called."

content/realtime/channels.textile
```diff
@@ -813,6 +814,55 @@ Previously registered listeners can be removed individually or all together.
...
+
+bc[flutter]. channel.detach();
+final stateChangeListener = channel
+  .on(ably.ChannelEvent.detached)
+  .listen((ably.ChannelStateChange state) {
+    print('detached from the channel ${channel.name}');
+  }
+);
+
+A channel will automatically close when there are no more realtime clients attached to it and approximately one minute has passed since the last client detached and since the last message was published to the channel.
```
marklewin, line 864: Is this correct? I'm thinking that, for a message to be published, a client must be attached to the channel to do the publishing. Therefore, the last client leaving will always be later than the last message being published? 🤔

marklewin (changes requested):

content/realtime/channels.textile
m-hulbert: Good point on the "it" and "client" point. I've taken your suggestion for the rewrite as it's far less wordy!

content/realtime/channels.textile
m-hulbert: This is true as far as I'm aware, yes. You can always publish by REST so this can occur after you no longer have any realtime attachments. 

Found the [FAQ](https://faqs.ably.com/how-does-ably-count-peak-channels) I read this in for reference.

paddybyers: Correct, REST publishes can take place without attachment. (And also equivalently, realtime publishes via a transient attach - technically they are attaches under the hood, but clients don't see that - they just call `publish()`).

m-hulbert: Thanks for the confirmation, Paddy.

marklewin (approved):
LGTM!

PR #1271, opened 2021-12-14
DOC-558: Add sections for heartbeatInterval and dormantAfter

m-hulbert:

## Description

This PR adds documentation for two `transportParams`: `heartbeatInterval` and `dormantAfter`. It also updates the reference for `clientOptions` to link to the descriptions added for these two parameters.

## Review

View the following pages to review this PR:

* [Connections]()
* [Presence]()
* [clientOptions]()


tbedford (approved):

content/realtime/connection.textile
```diff
@@ -393,6 +394,27 @@ h4. Connection recovery constraints
...
 
+h3(#heartbeats). Heartbeats
+
+Heartbeats enable Ably to identify clients that abruptly disconnect from the service, such as where an internet connection drops out or a client changes networks. 
+
+Ably sends a heartbeat to connected clients every 15 seconds. The lack of two consecutive heartbeats will indicate to Ably that a client has disconnected and the "connection state":#connection-states will be updated to @disconnected@. Note that this is only in the case of an abrupt disconnect. Closing a connection, such as by calling "@close()@":#close, is instant.
```
SimonWoolf, line 401: This is mixing two different things: the client detecting the connection has dropped, and Ably detecting it. The claim that Ably discovering that a client has disconnected results in that client's connection state being updated is confused: if we're here then we're not talking to the client, so we're not able to communicate anything to it.

```suggestion
Ably sends a heartbeat to connected clients every 15 seconds. If the client goes more than 25 seconds (the heartbeat interval plus a 10s margin of error to allow for network delays) without seeing any server activity from Ably, it assumes that something has gone wrong with the connection and the "connection state":#connection-states will become @disconnected@.

(Ably also uses this mechanism to detect dropped client connections from our end, though the details of this vary by transport).

To be clear, this mechanism is not usually used: it's not needed for the case of a connection deliberately closing (by calling "@close()@":#close ) or being disconnected (by the server or othewise). It is only for when something disrupts communication but does not properly terminate the TCP connection.
```

content/realtime/presence.textile
```diff
@@ -505,6 +506,27 @@ Any time a channel is re-attached and the presence set is re-synced, e.g. after
...
+
+The Ably client library will attempt to "reconnect":/realtime/connection#connection-state-recovery after a disconnect. If the connection is reestablished before 15 seconds have passed, a @leave@ event will not be sent. If the connection is reestablished after 15 seconds, a @leave@ event will be sent and the presence set will need to be rejoined.
+
+Note that the 15 second delay from being removed from the presence set is only for abrupt or unplanned disconnects. If a client calls "@leave()@":#leave or "@close()@":/realtime/connection#close they immediately send a @leave@ event.
+
+The time taken before a @leave@ event is sent in the case of an abrupt disconnect can be reduced to a minimum of 1 second by setting a value for @dormantAfter@, in milliseconds. This property is set within the @transportParams@ property of the "@clientOptions@":/realtime/usage#client-options object.
```
SimonWoolf, line 517: We should probably make a decision on what we actually want to call this parameter before publishing our docs on it. It's been pointed out that "dormant" isn't a concept the customer actually knows about. cf https://github.com/ably/realtime/pull/3008#issuecomment-631565738

cc @ben-pickering

SimonWoolf (approved):
Couple comments, generally looks fine, though I'm a little uneasy about the amount of attention we're lavishing on these small parameters -- there's dozens of these sort of barely-documented params, if we're planning to document them all in this level of detail then a little worried our docs are going to get even more impenetrable

content/realtime/connection.textile
m-hulbert: Ah thanks for clarifying that. I've updated the wording based on your suggestion.

content/realtime/presence.textile
m-hulbert: Updated `dormantAfter` to `remainPresentFor` as per https://github.com/ably/realtime/pull/3869

PR #1276, opened 2021-12-30
[DEX-26] Modernize jsbin sample

artismarti:
## Description

This work changes the following:
- Replaces `var` with `const` in code sample that users see when they finish the registration process.
- Adds comments about API key and channel name.
- Replaces the old Ably logo.

### Jira ticket:
* [DEX-26](https://ably.atlassian.net/browse/DEX-26)

## Review
Existing Version:
image

Changed Version:
image

### DIFF
--- a/content/code/website-examples/home-page.code
+++ b/content/code/website-examples/home-page.code
@@ -1,6 +1,13 @@
 [--- Javascript ---]
-var ably = new Ably.Realtime('{{API_KEY}}');
-var channel = ably.channels.get('{{RANDOM_CHANNEL_NAME}}');
+/**
+ * This is a DEMO API KEY. Your own API key is within your app in the dashboard: 'https://ably.com/accounts/' .
+ */
+const ably = new Ably.Realtime('{{API_KEY}}');
+
+/**
+ * This is a random channel name, update it to your channel name if testing it on an Ably App.
+ */
+const channel = ably.channels.get('{{RANDOM_CHANNEL_NAME}}');
 
 $('input#publish').on('click', function() {
   channel.publish('greeting', 'Hello from the browser');
@@ -18,7 +25,7 @@ function show(status) {
 [--- HTML ---]
 <script type="text/javascript" src="//cdn.ably.io/lib/ably.min-1.js"></script>
 <script src="//jsbin-files.ably.io/js/jquery-1.8.3.min.js"></script>
-<h1>Ably: Simple API Example</h1>
+<h1>Simple API Example</h1>
 
 <p>Ably provides a simple API that keeps developers happy. <a href="https://ably.com/documentation/">View the API documentation.</a></p>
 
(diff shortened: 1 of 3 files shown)


content/code/website-examples/home-page.code
```diff
@@ -1,6 +1,13 @@
 [--- Javascript ---]
-var ably = new Ably.Realtime('{{API_KEY}}');
-var channel = ably.channels.get('{{RANDOM_CHANNEL_NAME}}');
+/**
+ * This is a DEMO API KEY. Your own API key is within your app in the dashboard: 'https://ably.com/accounts/' .
```
m-hulbert, line 3:
```suggestion
 *  Note: this example uses a demo API key. Your own API keys are available in your account dashboard: 'https://ably.com/accounts/'.
```

content/code/website-examples/home-page.code
```diff
@@ -1,6 +1,13 @@
...
+ */
+const ably = new Ably.Realtime('{{API_KEY}}');
+
+/**
+ * This is a random channel name, update it to your channel name if testing it on an Ably App.
```
m-hulbert, line 8:
```suggestion
 * Note: this example uses a random channel name. You can use your own channel name when running this code sample against an Ably app.
```

data/jsbins.yaml
```diff
@@ -53,6 +51,8 @@ jsbin_hash:
   EZuljFh/oQj0cOkrUZpU7XoBjQM=: alohep
   aCl/jytvZrNnguSEukgdTXNIja4=: epovij
   MpDgpT2wvuGs712waej3+FufE7A=: oxewul
+  EjhhGNY3f3dw55pyTen7hGEhUow=: ohidar
```
m-hulbert, line 54: I'm not sure why this JSBin was updated but it looks identical to its previous version.

m-hulbert (changes requested):
The example config file shouldn't be deleted as it's referenced in the README. 

I added a couple of suggested tweaks on the code comments, but feel free to ignore these if you disagree. There's also a JSBin that was updated to a new version, but it looks identical and only appears to be used in the formatting guide as an example.

data/jsbins.yaml
artismarti: I accidentally changed the wrong code sample file & reverted the changes. Could that be it?

Do I need to change this?

artismarti:
> The example config file shouldn't be deleted as it's referenced in the README.
> 
> I added a couple of suggested tweaks on the code comments, but feel free to ignore these if you disagree. There's also a JSBin that was updated to a new version, but it looks identical and only appears to be used in the formatting guide as an example.

Thanks @m-hulbert 

I've made the requested changes! :)

data/jsbins.yaml
m-hulbert: Yes, that will what triggered it. It should be fine though - just wanted to check incase it was something you'd meant to change.

m-hulbert (approved):

PR #1296, opened 2022-02-11
docs: add missing time unit

ferdipret:

## Description

While looking through "how ably works", I've noticed in the presence section, that we do not specify a time unit for how long presence events are stored on disk. I assume this should be `hours`.

## Review

See diff


m-hulbert (approved):
Thanks @ferdipret !

PR #1297, opened 2022-02-11
Remove QR code functionality

dpiatek:

## Description

- this code used the Google URL shortner, a service that was closed down in 2019 https://developers.google.com/apps-script/advanced/url-shortener
- the error the call to this service made broke the whole QR code functionality, so this was not working since 2019
- because this is not necessary for the page, and should be done in a lot easier to debug fashion, we decided to remove this feature and potentially bring it back when tutorials are re-platformed
- this will prevent an obnoxious error in Sentry https://sentry.io/organizations/ably-uq/issues/2976824031/?project=6156014&query=is%3Aunresolved
- and reduce the page size by around 50kb (gapi + qr code lib) https://www.npmjs.com/package/kjua

## Review

When merged, this error should no longer appear on the changed tutorial pages:
Screenshot 2022-02-11 at 15 59 38

### DIFF
--- a/content/tutorials/encryption.textile
+++ b/content/tutorials/encryption.textile
@@ -557,7 +557,6 @@ h2(#next-steps). Next steps
 4. "Get in touch if you need help":https://ably.com/contact
 
 <script src="//cdn.ably.io/lib/ably.min-1.js" crossorigin="anonymous"></script>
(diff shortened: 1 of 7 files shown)


kennethkalmer (approved):
✂️ 🎉

PR #1309, opened 2022-02-28
Remove extraneous enclosing scopes from mqtt examples

owenpearson:
## Description

Removes enclosing scopes (`{}`) from the MQTT code snippets. These were unnecessary, and the examples look better without them.

image

## Review

Code samples on this page:

* [MQTT](https://ably-docs-mqtt-remove-s-r7kvzw.herokuapp.com/mqtt/)

### DIFF
--- a/content/mqtt/index.textile
+++ b/content/mqtt/index.textile
@@ -33,12 +33,12 @@ To use the Ably MQTT protocol adapter, you'll need to ensure you correctly confi
 For example, in the NodeJS "MQTT package":https://www.npmjs.com/package/mqtt, you'd need to specify the following:
 
 bc[nodejs]. var options = {
-    keepalive: 30,
-    username: '{{API_KEY_NAME}}', /* API key's name */
-    password: '{{API_KEY_SECRET}}', /* API key's secret */
-    port: 8883
-  };
-  var client = mqtt.connect('mqtts:mqtt.ably.io', options);
+  keepalive: 30,
+  username: '{{API_KEY_NAME}}', /* API key's name */
+  password: '{{API_KEY_SECRET}}', /* API key's secret */
+  port: 8883
+};
+var client = mqtt.connect('mqtts:mqtt.ably.io', options);
 
 
 This will connect using TLS through MQTT to Ably.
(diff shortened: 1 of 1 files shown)


tbedford (changes requested):
Syntax is not quite right. I think the `p.` is unnecessary. Needs some more testing.

image

We also need to test under web view.

Sapples (changes requested):
Not super familiar with MQTT but it looks like the code examples now have extraneous "p." at the end of the code snippets. Let me know if I'm wrong and they are needed though!

tbedford:
@Sapples - correct! If you get time today, would you be able to push up a fix? It should be just a case of removing the outer `{}` and testing. Thanks!

owenpearson:
Huh, that's odd. I didn't build the docs and check for myself so apologies for the error. I found [these textile docs](https://textile-lang.com/doc/block-code) which say:
> For long blocks of code with blank lines in between, use the extended block directive `bc.`. and end the block with `p.` or with another block tag

I'll convert this PR back to a draft state for now. I'm guessing just removing the `p.`s will suffice but I'll check it for myself and provide screenshots before reopening. @Sapples @tbedford do either of you have any immediate ideas as to why the `p.`s aren't working as I expected?

tbedford:
>  @Sapples @tbedford do either of you have any immediate ideas as to why the p.s aren't working as I expected?

I've not had time to test this out as yet. I do prefer using the GitHub syntax for codeblocks that we also support, which has an explicit closure.

m-hulbert:
I'm not sure if we have a slightly different implementation to account for the different languages, but we just list a blank line to end a block in our own [formatting guide](https://github.com/ably/docs/blob/main/content/client-lib-development-guide/documentation-formatting-guide.textile#code-blocks).

Sapples:
@tbedford - Removed the trailing .p from the code blocks.

tbedford (changes requested):
Code needs realigning (VSCode can do this for us):

image

tbedford (changes requested):
@Sapples - ah, I think I see why they used enclosing brackets before - blank lines were ending the code block. I've pushed up a fix using GitHub-flavoured codeblocks, and spaced the code out so that functions are separated from other code by blank lines. I'll review again when the review app deploys.

tbedford (approved):
Tested via review app. Outer brackets removed, and code spaced out nicely.

tbedford:
@owenpearson - thanks Owen, if you're happy [with this](https://ably-docs-mqtt-remove-s-r7kvzw.herokuapp.com/mqtt/) I think we can merge.

owenpearson:
@tbedford that looks good to me, thanks a lot for your help 👍

Sapples (approved):
LGTM now!

PR #1312, opened 2022-03-02
DOC-552: Correct API key format

m-hulbert:
## Description

This PR corrects the API key format from using a`:` as the first separator to a `.`.


Sapples (approved):
LGTM

tbedford (approved):
LGTM! However, we should not be linking to FAQs for things as basic as creating and managing keys. 

@Sapples - could you create a DOC ticket to create a suitable topic in the docs, and make a note in the ticket to circle back and change links to FAQs for this topic to the new topic in the docs please? Thanks!

PR #1313, opened 2022-03-07
Show how to specify auth options

tbedford:
## Description

Show how to specify auth options.

* [DOC-560](https://ably.atlassian.net/browse/DOC-560).

## Review

Sections / pages to view:

* [Realtime - Token auth](https://ably-docs-doc-560-auth--7otqeh.herokuapp.com/realtime/authentication/#token-authentication)
* [REST - Token auth](https://ably-docs-doc-560-auth--7otqeh.herokuapp.com/rest/authentication/#token-authentication)


tbedford:
@owenpearson - your feedback on the code examples would be greatly appreciate - thanks.

content/realtime/authentication.textile
```diff
@@ -77,42 +77,110 @@ Here is a straightforward example of using Basic Authentication to connect to Ab
...
+const ably = new Ably.Realtime({ authUrl: "/auth", authMethod: "POST", authParams: { p1: param1, b: param2}, authHeaders: {h1: header1, h2: header2} });
+```
+
+h3(#auth-callback). Token refresh using @authCallback@
+
+For non-browser clients, an authentication callback is the preferred method of implementing token refresh, although browser clients can also use @authCallback@ if the use case requires it. This relies on your application communicating securely with your own servers to obtain a token.
+
+The following code sample shows an example of using an @authCallback@ in a client:
+
+```[jsall](code-editor:realtime/auth-token-callback)
```
owenpearson, line 140: This code block tag says 'jsall' but uses the NodeJS standard library so I'm wondering if it should say 'jsnode' or something similar?

content/realtime/authentication.textile
```diff
@@ -77,42 +77,110 @@ Here is a straightforward example of using Basic Authentication to connect to Ab
...
+const https = require("https");
+require("dotenv").config();
```
owenpearson, line 142: There are a lot of ways to supply environment variables to a NodeJS script. Generally most code samples I've seen will use environment variables and expect the reader to choose their own method to supply them, so I'd opt for just removing this line.

content/realtime/authentication.textile
```diff
@@ -77,42 +77,110 @@ Here is a straightforward example of using Basic Authentication to connect to Ab
...
+
+const options = {
+  hostname: process.env.HOSTNAME,
+  port: 443, // SSL
```
owenpearson, line 146: SSL is deprecated and has been superseded by TLS.

content/realtime/authentication.textile
```diff
@@ -77,42 +77,110 @@ Here is a straightforward example of using Basic Authentication to connect to Ab
...
+  path: "/auth",
+  method: "POST",
+  headers: {
+    "User-Agent": "Realtime Client v0.2",
```
owenpearson, line 150: It's not clear to me what this user agent is supposed to do or where the v0.2 comes from. If there's a valid reason to keep this then I'd say we need a comment to explain.

content/realtime/authentication.textile
```diff
@@ -77,42 +77,110 @@ Here is a straightforward example of using Basic Authentication to connect to Ab
...
+          callback(null, JSON.parse(tokenRequest));
+        });
+      } else {
+        console.error("Response code was not 200.");
+        callback(res.statusCode, null);
+      }
+    });
+
+    req.on("error", (error) => {
+      console.error(error);
```
owenpearson, line 170: There ought to be a `callback(error)` below this line.

content/realtime/authentication.textile
```diff
@@ -77,42 +77,110 @@ Here is a straightforward example of using Basic Authentication to connect to Ab
...
+    "User-Agent": "Realtime Client v0.2",
+    username: process.env.USERNAME,
+    password: process.env.PASSWORD,
+  },
+};
+
+const ably = new require("ably").Realtime({
+  authCallback: (tokenParams, callback) => {
+    const req = https.request(options, (res) => {
+      if (res.statusCode == 200) {
```
owenpearson, line 159: Any `2xx` response code indicates success so we might want to change this condition to `res.statusCode >= 200 && res.statusCode < 300`.

Also it's generally wise to use strict equals (`===`) where possible in JS. In this case `res.statusCode` is always a number so we should definitely change it.

content/realtime/authentication.textile
```diff
@@ -77,42 +77,110 @@ Here is a straightforward example of using Basic Authentication to connect to Ab
...
+  path: "/auth",
+  method: "POST",
+  headers: {
+    "User-Agent": "Realtime Client v0.2",
+    username: process.env.USERNAME,
+    password: process.env.PASSWORD,
+  },
+};
+
+const ably = new require("ably").Realtime({
```
owenpearson, line 156: We should put `const Ably = require('ably');` at the top rather than importing it here.

owenpearson: I'd also consider using a different variable name from `ably` here. I personally use `client` or `realtime`. Based on my quick scan through the existing docs it seems `realtime` is the most commonly used.

content/realtime/authentication.textile
```diff
@@ -77,42 +77,110 @@ Here is a straightforward example of using Basic Authentication to connect to Ab
...
+  authCallback: (tokenParams, callback) => {
+    const req = https.request(options, (res) => {
+      if (res.statusCode == 200) {
+        res.on("data", (tokenRequest) => {
```
owenpearson, line 160: This usage of the NodeJS HTTP client is a little bit off; the "data" event can be emitted more than once if the response is sufficiently large (I'm not sure exactly what the conditions are for this to happen, I'd assume it depends on the server implementation). I think generally an Ably token request should be small enough to come through as a single chunk so maybe it's fine but it might raise a few eyebrows. The more 'correct' way of doing this is to initialise a string and append each chunk to the string whenever "data" is emitted. Then you can do the JSON parsing once "end" is emitted:
```
let tokenRequest = '';

res.on("data", (chunk) => {
  tokenRequest += chunk;
});

res.on("end", () => {
  callback(null, JSON.parse(tokenRequest));
});
```

Worth noting as well that JSON.parse can throw an error if the string supplied to it is not valid JSON. In a real world application I would definitely advise adding a try/catch block, although I realise we want these examples to be simple so I'm on the fence about adding one here.

content/realtime/authentication.textile
owenpearson: If you change the success condition then remember to update the `console.error` on line 164.

owenpearson (changes requested):
I've made a few comments on the realtime code snippet. I think these should all apply to the rest snippet too.

content/realtime/authentication.textile
```diff
@@ -77,42 +77,110 @@ Here is a straightforward example of using Basic Authentication to connect to Ab
...
 
-Below is a rather contrived yet straightforward example that instances a Realtime library using "Token Authentication":/core-features/authentication#token-authentication with a means to reissue tokens as required. Typically, in a browser environment, the <span lang="default">"@authUrl@":#auth-options</span><span lang="csharp">"@AuthUrl@":#auth-options</span> provided would be a relative URL to a local endpoint that is used to issue tokens to trusted clients. Client requests can, for example, be trusted based on session cookies. For non-browser clients, an authentication callback is preferred thus relying on your application to communicate securely with your own servers to obtain a token.
+h3(#auth-url). Token refresh using @authUrl@
 
-```[jsall](code-editor:realtime/auth-token-callback)
-  var realtime = new Ably.Realtime({ authUrl: 'http://my.website/auth' });
+Below is an example that instances a Realtime library using "Token Authentication":/core-features/authentication#token-authentication with a means to reissue tokens as required.
```
lmars, line 86: We use the term "instances" to mean "to instance something" in a few places at Ably, but it never reads well to me, I can't recall ever seeing other technical documentation that uses the term in this way, it's typically either "instantiates" or "creates an instance of". Just thought it worth saying here.

content/realtime/authentication.textile
```diff
@@ -77,42 +77,110 @@ Here is a straightforward example of using Basic Authentication to connect to Ab
...
+h3(#auth-url). Token refresh using @authUrl@
 
-```[jsall](code-editor:realtime/auth-token-callback)
-  var realtime = new Ably.Realtime({ authUrl: 'http://my.website/auth' });
+Below is an example that instances a Realtime library using "Token Authentication":/core-features/authentication#token-authentication with a means to reissue tokens as required. 
+
+Typically, in a browser environment, the <span lang="default">"@authUrl@":#auth-options</span><span lang="csharp">"@AuthUrl@":#auth-options</span> provided would be a relative URL to a local endpoint that is used to issue tokens to trusted clients. Client requests can, for example, be trusted based on session cookies.
```
lmars, line 88: I think "local endpoint" here is incorrect (local to what?), I think it should be "endpoint under the same origin"

content/realtime/authentication.textile
```diff
@@ -77,42 +77,110 @@ Here is a straightforward example of using Basic Authentication to connect to Ab
...
+
+const options = {
+  hostname: process.env.HOSTNAME,
+  port: 443, // SSL
+  path: "/auth",
+  method: "POST",
+  headers: {
+    "User-Agent": "Realtime Client v0.2",
+    username: process.env.USERNAME,
+    password: process.env.PASSWORD,
```
lmars, line 152: I don't think putting these in `headers` will do the right thing, is we want to demonstrate using basic auth, I think it should be an `auth` key on the options.

content/realtime/authentication.textile
owenpearson: Yeah I did consider mentioning fetch earlier. The fetch API is obviously a lot simpler and I would assume the vast majority of people implementing auth with Ably won't want to use a low-level HTTP client (like the one in Node's standard lib). That being said, this particular snippet is supposedly aimed at Node users and fetch still isn't available in any LTS Node version so we'd at the very least need to recommend using a polyfill (such as node-fetch) for users who aren't using Node 17.5+. Definitely worth considering but I feel like it's gonna be noisy either way 🤷

lmars: Whilst `fetch` isn't in LTS yet, I assume it'll be in the foreseeable future, then this code using `https` will seem outdated then?

tbedford:
@owenpearson - thanks for the feedback! I think I will completely rethink the callback example, bearing in mind what you've said. Thanks again.

tbedford:
This ticket was mostly about how to specify auth options, so I've taken my (bad) callback code out, and replaced it by a stub. I think we should do a tutorial on auth callback instead, and use fetch rather than https. Thanks!

content/realtime/authentication.textile
owenpearson: Yeah that's a good point. I've thought about it a bit more now and I'd be in favour of using fetch 👍

content/realtime/authentication.textile
tbedford: Replaced code with stub.

content/realtime/authentication.textile
tbedford: Replaced code with stub.

content/realtime/authentication.textile
tbedford: This was mainly added as Glitch requires it - but not necessary for this example. Replaced with stub.

content/realtime/authentication.textile
tbedford: Good point - will bear this in mind for the tutorial. Replaced code with stub.

content/realtime/authentication.textile
tbedford: Yes, changed examples to use `realtime` or `rest`.

content/realtime/authentication.textile
tbedford: Duly noted, fetch seems to be the way to go.

content/realtime/authentication.textile
tbedford: Quite right - changed to "creates an instance of".

content/realtime/authentication.textile
tbedford: Reworded to be correct, "endpoint under the same origin" is what I actually meant!

content/realtime/authentication.textile
tbedford: Replaced code with stub.

content/rest/authentication.textile
```diff
@@ -91,63 +91,127 @@ Here is a straightforward example of using Basic Authentication when instancing
...
+h3(#auth-url). Token refresh using @authUrl@
 
-Below is a rather contrived yet straightforward example that instances a Realtime library using "Token Authentication":/core-features/authentication#token-authentication with a means to reissue tokens as required. Typically, in a browser environment, the <span lang="default">"@authUrl@":#auth-options</span><span lang="csharp">"@AuthUrl@":#auth-options</span> provided would be a relative URL to a local endpoint that is used to issue tokens to trusted clients. Client requests can, for example, be trusted based on session cookies. For non-browser clients, an authentication callback is preferred thus relying on your application to communicate securely with your own servers to obtain a token.
+Below is an example that instances a Realtime library using "Token Authentication":/core-features/authentication#token-authentication with a means to reissue tokens as required.
```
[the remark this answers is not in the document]
tbedford: Good spot. Fixed.

owenpearson (approved):
LGTM 👍

lmars (approved):

Sapples (changes requested):
@tbedford LGTM. However, I did slightly read past and noticed a typo under [Selecting Auth](https://ably-docs-doc-560-auth--7otqeh.herokuapp.com/rest/authentication/#selecting-auth):

image

Sapples (approved):
LGTM.

PR #1317, opened 2022-03-15
Update rest or realtime links

Sapples:

## Description

* [Bugs](https://ably.atlassian.net/browse/DOC-608)
* [DOC-497](https://ably.atlassian.net/browse/DOC-497).

## Review

Check the new link for REST or Realtime goes to Docs and not FAQs.

* [Quickstart guide](https://ably-docs-doc-497-updat-tq1yhr.herokuapp.com/quick-start-guide/)

### DIFF
--- a/content/root/best-practice-guide.textile
+++ b/content/root/best-practice-guide.textile
@@ -58,7 +58,7 @@ The main factors are summarized here:
 
 * If you want to maintain a persistent connection with Ably's Realtime platform in order to subscribe to messages sent on a channel, or be notified whenever the presence set changes, you need to use the "realtime library":/realtime.
 * If you just want to publish data to the platform and don't need to subscribe or manage channel attachment, you can the "REST library":/rest.
-* If you want to publish at very high rates or with the lowest possible realtime latencies, "Ably's realtime library would be best as it can publish at higher rates and lower latencies than is possible with the REST library due to message pipelining":https://ably.com/documentation/best-practice-guide#rest-or-realtime.
+* If you want to publish at very high rates or with the lowest possible realtime latencies, Ably's realtime library would be best as it can publish at higher rates and lower latencies than is possible with the REST library due to message pipelining.
 * If you want to get one-time info about the presence set of a channel, you can use the "REST library":/rest.
 * If you want to set up realtime communication with IoT devices on platforms for which there isn't an Ably Client Library SDK available, you can use the "MQTT protocol adapter":/mqtt.
 * If you wish to migrate from PubNub or Pusher to Ably, you can use the relevant "protocol adapter":https://ably.com/protocols to "switch to Ably":https://faqs.ably.com/can-you-help-ease-the-pain-of-migration-from-another-data-stream-network, and then gradually move over to using the Ably Client Library SDKs.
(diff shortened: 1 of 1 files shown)


content/root/best-practice-guide.textile
```diff
@@ -58,7 +58,7 @@ The main factors are summarized here:
 
 * If you want to maintain a persistent connection with Ably's Realtime platform in order to subscribe to messages sent on a channel, or be notified whenever the presence set changes, you need to use the "realtime library":/realtime.
 * If you just want to publish data to the platform and don't need to subscribe or manage channel attachment, you can the "REST library":/rest.
-* If you want to publish at very high rates or with the lowest possible realtime latencies, "Ably's realtime library would be best as it can publish at higher rates and lower latencies than is possible with the REST library due to message pipelining":https://faqs.ably.com/should-i-use-the-rest-or-realtime-library.
+* If you want to publish at very high rates or with the lowest possible realtime latencies, "Ably's realtime library would be best as it can publish at higher rates and lower latencies than is possible with the REST library due to message pipelining":https://ably.com/documentation/best-practice-guide#rest-or-realtime.
```
m-hulbert, line 61: I don't think this needs to be a link as it's linking to itself.

m-hulbert (commented):
Just a small link removal. I also had a quick scan of the article being linked to and I think it might be worth at least moving over the part on the use case of using REST to issue tokens as I don't think that's mentioned in the BPG.

content/root/best-practice-guide.textile
Sapples: Agreed, I've just removed the link totally now.

content/root/best-practice-guide.textile
```diff
@@ -58,7 +58,7 @@ The main factors are summarized here:
 
 * If you want to maintain a persistent connection with Ably's Realtime platform in order to subscribe to messages sent on a channel, or be notified whenever the presence set changes, you need to use the "realtime library":/realtime.
 * If you just want to publish data to the platform and don't need to subscribe or manage channel attachment, you can the "REST library":/rest.
-* If you want to publish at very high rates or with the lowest possible realtime latencies, "Ably's realtime library would be best as it can publish at higher rates and lower latencies than is possible with the REST library due to message pipelining":https://faqs.ably.com/should-i-use-the-rest-or-realtime-library.
+* If you want to publish at very high rates or with the lowest possible realtime latencies, Ably's realtime library would be best as it can publish at higher rates and lower latencies than is possible with the REST library due to message pipelining.
```
tbedford, line 61: This seems to remove the link, but I think we need to link to where we have this info in the main docs. For example:

* https://ably.com/documentation/realtime#realtime-vs-rest
* https://ably.com/documentation/rest#rest-vs-realtime
* https://ably.com/documentation/best-practice-guide#rest-or-realtime

Generally, we want to change links to KB/FAQ articles to link to the same info in the docs. Where that info is not currently in the main docs, we need to create new tickets to capture that.

As an aside, "pipelining" needs to be clarified, but that's probably beyond the scope of this PR and covered by [DOC-410](https://ably.atlassian.net/browse/DOC-410).

tbedford (changes requested):

content/root/best-practice-guide.textile
Sapples: I've updated it now to point to the [rest landing page](https://ably.com/documentation/realtime) as the other links just go to the relevant landing page and Mark made a good point that linking from Rest vs Realtime in one page to Rest vs Realtime in another felt a bit odd.

tbedford (approved):

m-hulbert (approved):

PR #1349, opened 2022-04-06
Edx 89 definition list display

MymmiJ:
## Description

* [Jira ticket](https://ably.atlassian.net/browse/EDX-89).

## Review

Before:
image

After: 
Screenshot 2022-04-06 at 12 31 51


DuncanEdwards (approved):
The fix works but I'm not sure that 2nd screenshot is the correct one. I'd have expected something more like:
image

PR #1367, opened 2022-04-12
Edx 45 menu items

MymmiJ:

## Description

Replacing all menu items with styled menu items that match our style guide.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-45).

## Review

Check that menu items render correctly per https://www.figma.com/file/4KdtIEGY2BfZPq5tT1u6D3/Migration-Guide?node-id=21%3A346

Verify that updated scrolling behaviour makes sense (in code or by checking): intended behaviour is that:
* Gatsby never takes over scrolling when moving between variations of a particular page (e.g. different language variations)
* Gatsby never takes over scrolling when moving to or from a specific section link (e.g. #channel-states)
* Gatsby scrolls user to top when viewing a completely new page.

Other changes:
* `    background: linear-gradient(0deg, white 0%, rgba(255, 255, 255, 0) 100%);` - this ensures that the gradient 'scroll more' hint works on older Safari versions, cf. https://stackoverflow.com/questions/38391457/linear-gradient-to-transparent-bug-in-latest-safari
* TSConfig has been fixed so it won't throw false positive errors, by giving it an output directory & include/exclude paths

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -48,6 +48,7 @@ const shouldUpdateScroll = ({ prevRouterProps, routerProps: { location } }) => {
 
   const plainCurr = curr.replace(versionRegex, '').replace(languageRegex, '').replace(hashRegex, '');
   const plainPrev = prev.replace(versionRegex, '').replace(languageRegex, '').replace(hashRegex, '');
+  console.log(plainCurr, plainPrev);
 
   return (
     plainCurr !== plainPrev && // If the current location is a variant of the previous location, do not update scroll
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -12,7 +12,7 @@ import {
 
 import sprites from '@ably/ui/core/sprites.svg';
 
-const PAGE_OFFSET = 94;
+const PAGE_OFFSET = 128;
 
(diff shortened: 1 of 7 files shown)


gatsby-browser.js
```diff
@@ -46,8 +51,9 @@ const shouldUpdateScroll = ({ prevRouterProps, routerProps: { location } }) => {
 
   return (
```
DuncanEdwards, line 52: Don't need brackets

DuncanEdwards (commented):
I feel like there is a lot of gap above the menus on the side?
image
Is that intentional?

gatsby-browser.js
MymmiJ: This is a weird one, I noticed your linting isn't showing the same things as mine while pairing; this will be highlighted by prettier for me if I change it, are you using something else?

DuncanEdwards: I probably need to install the prettier extension

MymmiJ:
Not sure why Input/index.tsx appears in this; I rebased from new-toolchain and they've appeared in the diff now even though they're present in both; can't seem to get rid of them.

PR #1385, opened 2022-04-14
[EDX-102] correct spacing on code text element

MymmiJ:

## Description

* [Jira ticket](https://ably.atlassian.net/browse/EDX-102).

## Review

There's quite a lot going on here for the small size of change, here's a summary:
* The user agent stylesheet was overriding some CSS because we were in quirks mode, this is because various <table> elements were trying to render text especially '/n/t'. This has been fixed: (table, tbody, tr).
* There's some reset styles in Ably UI that do quite a lot of base styling that we have to ignore in `global.css`
* The base font-family and font-weight have been wrong, that had to be corrected in the global CSS reset/normalize styles to pull through the changes we need to Code (Museo Sans/Menlo => NEXT Book/Arial)
* This messed with a few subsequent styles, mostly font-weight being weird/wrong (dd, td, dt, li) and Sidebar menus not appearing correctly.
* Once this is all fixed, the line-height can be successfully set and display as expected in the SyntaxHighlighter styles for 'hljs' classes & the code text element now displays correctly.


src/components/Sidebar/RightSidebar/index.js
```diff
@@ -8,7 +8,7 @@ const mapMenuItemToSidebarItem = ({ name, id, level }) => ({
   level,
 });
 
-const RightSidebar = ({ menuData, className }) => {
+const RightSidebar = ({ menuData, languages, className }) => {
```
DuncanEdwards, line 11: Convert to TS (if you get stuck I can help!)

src/components/Sidebar/index.js
```diff
@@ -8,12 +8,13 @@ import { EXPAND_MENU } from './consts';
 const Sidebar = ({
```
DuncanEdwards, line 8: Convert to TS (if you get stuck I can help!)

DuncanEdwards (changes requested):
I thought I'd stop at making you move two to TS! We should do stuff like the table and list components in bulk at some later point

src/components/Sidebar/index.js
MymmiJ: Have had a go at TS definitions for this and RightSidebar now!

src/components/Sidebar/sidebar-props.ts
```diff
@@ -0,0 +1,11 @@
+import { EXPAND_MENU } from './expand-menu-enum';
```
DuncanEdwards, line 1: As it isn't used by more than one component, this doesn't need to be in its own file, can just be placed inline in Sidebar.

src/components/Sidebar/RightSidebar/right-sidebar-props.ts
```diff
@@ -0,0 +1,6 @@
+import { MenuData } from './menu-data';
```
DuncanEdwards, line 1: Only used by RightSidebar, so can just be inline in the `RightSidebar` component

DuncanEdwards (changes requested):

src/components/Sidebar/sidebar-props.ts
MymmiJ: This is also used by SidebarLinkItem, SidebarLinkMenu and left-sidebar.js as an enum now, replacing the frozen object 'enum' used before, should I still move it?

MymmiJ: Ah I understand you mean the whole file now after reading the other comment, got it!

DuncanEdwards (approved):

PR #1400, opened 2022-04-21
[DOC-742] Add kafka group to tutorials

tbedford:
## Description

* [DOC-742](https://ably.atlassian.net/browse/DOC-742).

Tutorial did not display on Tutorials homepage, although it does display if the URL is invoked directly. The reason is the tutorial is in the group `kafka`, but that group does not exist. This PR does two things:

1. Adds the kafka group in `tutorials.json`.
2. Also adds the tutorial to the Integrations tab (as well as the channels tab). 

The effect of this is the tutorial will be displayed as follows:

image

## Review

I've added multiple reviewers as I want to make sure this approach is acceptable. Thanks.

Currently, this can only be reviewed by running the webview locally.

### DIFF
--- a/config/tutorials/tutorials.json
+++ b/config/tutorials/tutorials.json
@@ -43,6 +43,12 @@
       "name": "Using Ably Client Library SDKs",
       "desc": "A selection of tutorials demonstrating the use of Realtime and/or REST libraries within devices such as browsers, desktops, mobiles or servers."
     },
+    "kafka": {
+      "id": "kafka",
+      "tab": "Channels, Push, Reactor & API Hub",
+      "name": "With Using Ably with Kafka",
+      "desc": "Tutorials on using Kafka with Ably."
+    },
     "sse": {
       "id": "sse",
       "tab": "Channels, Push, Reactor & API Hub",
--- a/config/tutorials/tutorials.json
+++ b/config/tutorials/tutorials.json
@@ -27,7 +27,7 @@
     {
       "id": "reactor-integrations",
       "icon": "lightning",
-      "name": "Reactor Integrations",
+      "name": "Integrations",
       "desc": "Fun tutorials on how to make use of various libraries and JS frameworks with Ably."
     },
     {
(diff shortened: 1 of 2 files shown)


Srushtika (commented):
I'm not sure if this is the right approach. 

We already know that the categorisation of the existing cards between SDKs, MQTT and SSE needs to be revisited at some point but if we look at the basis for that categorisation, it was how a client can access the Ably service. With that basis in mind, this tutorial, if anything, should fall into a category called 'Using REST API' as we don't have one for it (not suggesting we do that) given that the tutorial uses the raw rest API to retrieve historical messages. 

Having a section called 'Using Kafka Connector' calls for a need to have separate sections for each of the inbound and outbound streaming integrations we have tutorials for. (for example, WebHooks with options, AWS SQS, etc).

I'd suggest that as a temporary measure, we update the copy of the 'Using Client SDKs' section to include the REST API and add it there. We can take it up again when we get a chance to look at the tutorials site as a whole.

config/tutorials/tutorials.json
```diff
@@ -43,6 +43,12 @@
       "name": "Using Ably Client Library SDKs",
       "desc": "A selection of tutorials demonstrating the use of Realtime and/or REST libraries within devices such as browsers, desktops, mobiles or servers."
     },
+    "kafka": {
+      "id": "kafka",
+      "tab": "Channels, Push, Reactor & API Hub",
+      "name": "With Using Ably with Kafka",
```
m-hulbert, line 49: I think the first `With` is unnecessary.

tbedford: Well spotted, left over from experimentation! Thanks

Srushtika (commented):
The categorisation looks good to me, thanks. 

I do think we can add all tags from the existing list applicable to the tutorial, so it comes up in the tag-based search. For example, integrations, language-based tags, etc. The tag system and the search experience as a whole need to be looked at as well at a later stage but for now we can optimise for what we have I think :) 

Screenshot 2022-04-21 at 14 47 26

tbedford:
> The categorisation looks good to me, thanks.
> 
> I do think we can add all tags from the existing list applicable to the tutorial, so it comes up in the tag-based search. For example, integrations, language-based tags, etc. The tag system and the search experience as a whole need to be looked at as well at a later stage but for now we can optimise for what we have I think :)
> 
> Screenshot 2022-04-21 at 14 47 26

Good idea. It's in the kafka tag, but I'll add a few more.

config/tutorials/tutorials.json
```diff
@@ -41,25 +41,19 @@
...
       "id": "sdk",
       "name": "Using Ably Client Library SDKs",
-      "desc": "A selection of tutorials demonstrating the use of Realtime and/or REST libraries within devices such as browsers, desktops, mobiles or servers."
-    },
-    "kafka": {
-      "id": "kafka",
-      "tab": "Channels, Push, Reactor & API Hub",
-      "name": "With Using Ably with Kafka",
-      "desc": "Tutorials on using Kafka with Ably."
+      "desc": "A selection of tutorials demonstrating the use of the Ably Client library SDKs, or the Ably REST API, within devices such as browsers, desktops, mobiles or servers."
```
marklewin, line 44: I think I'd be inclined to say "on devices such as" rather than "within devices"?

tbedford: Agreed. There's a few things I'd reword on that page! I've fixed a couple while I was there.

marklewin (approved):
LGTM 👍

PR #1401, opened 2022-04-21
Remove RSC20 and TO3m from client lib feature spec

### LINKED ISSUE #1381: Remove `RSC20` and `TO3m` (`logExceptionReportingUrl`)
We have decided that SDKs have no need to report errors to an external service. Our Sentry service is also being retired, so there would be nowhere for these calls to go anyway.

- Removed from `ably-js` by: https://github.com/ably/ably-js/pull/908 (February 2022)
- Removed from `ably-cocoa` by: https://github.com/ably/ably-cocoa/pull/1073 (October 2020)

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/DOC-734) by [Unito](https://www.unito.io)

owenpearson:
Resolves #1381.

An alternative approach could be to label this section as deprecated but I decided that since the functionality has already been removed from all SDKs it's unlikely to provide any value keeping it.


lmars (approved):

tbedford (approved):

QuintinWillison (changes requested):
Yay 🎉  - though we had agreed previously that we wouldn't delete the clauses entirely, as there's a risk someone might add a new clause using that old number, hence the need to replace clause textual contents with "This clause has been deleted." per [our guidelines](https://github.com/ably/engineering/blob/main/sdk/features-specification.md#guidelines).

owenpearson:
@QuintinWillison Good point, I've changed it now to just add "(deprecated)" to each bit instead of removing. This convention is already used elsewhere in the spec (see TS12d for example).

QuintinWillison (approved):

PR #1415, opened 2022-05-04
Added channel lifecycle status spec point

Rosalita:

## Description

Add support to get channel status [Channel status / REST / Docs | Ably Realtime](https://ably.com/docs/rest/channel-status#metadata-rest).

* [Jira epic ticket](https://ably.atlassian.net/browse/SDK-1790) 
* [Jira ticket](https://ably.atlassian.net/browse/SDK-1796).

## Review

* [Page to review](https://docs.ably.io/client-lib-development-guide/features/)


content/client-lib-development-guide/features.textile
```diff
@@ -300,6 +300,9 @@ h3(#rest-channel). Channel
...
 ** @(RSL6b)@ If, for example, incompatible encryption details are provided or invalid Base64 is detected in the message payload, an error message will be sent to the logger, but the message will still be delivered with last successful decoding and the @encoding@ field. For example, if a message had a decoding of "utf-8/cipher+aes-128-cbc/base64", and the payload was successfully Base64 decoded but the payload could not be decrypted because the @CipherParam@ details were not configured, the message would be delivered with a binary payload and an @encoding@ with the value "utf-8/cipher+aes-128-cbc". Additional steps need to be taken if decoding failed on "vcdiff" encoding; see "RTL18":#RTL18
 * @(RSL7)@ @Channel#setOptions@ takes a @ChannelOptions@ object and sets or updates the stored channel options, then indicates success
+* @(RSL8)@ @Channel#status@ function:
+** @(RSL8a)@ Returns a @ChannelDetails@ object
+** @(RSL8b)@ A side-effect of this request, is that it will cause the channel in question to become activated
```
QuintinWillison, line 305: Would this benefit from the suffix "if it isn't already"?

content/client-lib-development-guide/features.textile
```diff
@@ -300,6 +300,9 @@ h3(#rest-channel). Channel
...
 ** @(RSL6b)@ If, for example, incompatible encryption details are provided or invalid Base64 is detected in the message payload, an error message will be sent to the logger, but the message will still be delivered with last successful decoding and the @encoding@ field. For example, if a message had a decoding of "utf-8/cipher+aes-128-cbc/base64", and the payload was successfully Base64 decoded but the payload could not be decrypted because the @CipherParam@ details were not configured, the message would be delivered with a binary payload and an @encoding@ with the value "utf-8/cipher+aes-128-cbc". Additional steps need to be taken if decoding failed on "vcdiff" encoding; see "RTL18":#RTL18
 * @(RSL7)@ @Channel#setOptions@ takes a @ChannelOptions@ object and sets or updates the stored channel options, then indicates success
+* @(RSL8)@ @Channel#status@ function:
+** @(RSL8a)@ Returns a @ChannelDetails@ object
```
QuintinWillison, line 304: It might be helpful to reference your new `CHD1` spec. point from here.

QuintinWillison (changes requested):
A few thoughts, plus the convention in this document is to also add new types to [the 'API IDL'](https://docs.ably.com/client-lib-development-guide/features/#idl).

content/client-lib-development-guide/features.textile
```diff
@@ -300,6 +300,9 @@ h3(#rest-channel). Channel
...
 ** @(RSL6b)@ If, for example, incompatible encryption details are provided or invalid Base64 is detected in the message payload, an error message will be sent to the logger, but the message will still be delivered with last successful decoding and the @encoding@ field. For example, if a message had a decoding of "utf-8/cipher+aes-128-cbc/base64", and the payload was successfully Base64 decoded but the payload could not be decrypted because the @CipherParam@ details were not configured, the message would be delivered with a binary payload and an @encoding@ with the value "utf-8/cipher+aes-128-cbc". Additional steps need to be taken if decoding failed on "vcdiff" encoding; see "RTL18":#RTL18
 * @(RSL7)@ @Channel#setOptions@ takes a @ChannelOptions@ object and sets or updates the stored channel options, then indicates success
+* @(RSL8)@ @Channel#status@ function:
```
owenpearson, line 303: I think the spec should include details about what request the library sends (eg, GET vs POST, which URI to use). `RSC16` (Rest#time) is an example of this, although it's actually kind of incorrect since it specifies a static URI but the request URI will change depending on environment (see RSC11b). Maybe there's a better example in the spec somewhere but it would take some time to check.

content/client-lib-development-guide/features.textile
owenpearson, line 305: I think this should be removed since as far as I can tell it's not really relevant to client library implementation. I would argue that, as a rule, for any given spec point we should be able to ask "Does ably-[whatever language] implement this spec point?" and I don't think that applies here.

content/client-lib-development-guide/features.textile
```diff
@@ -1336,6 +1339,10 @@ h4. ChannelProperties
 * @(CP2)@ The attributes of @ChannelProperties@ consist of:
 ** @(CP2a)@ @attachSerial@ string - contains the last @channelSerial@ received in an @ATTACHED@ @ProtocolMesage@ for the channel, see "RTL15a":#RTL15a
 
+h4. ChannelDetails
+
+* @(CHD1)@ A @ChannelDetails@ is a type that represents information for a channel including channelID, name, status and occupancy
```
owenpearson, line 1344: We need to add `Channel#status` and the `ChannelDetails` type to the IDL section at the bottom of this document.

owenpearson (changes requested):

content/client-lib-development-guide/features.textile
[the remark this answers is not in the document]
QuintinWillison: Did you have any thoughts around indicating optionality for these members? - i.e. are any of them allowed to be `null`/`nil`.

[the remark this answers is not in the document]
Rosalita: They are returned from the request, so I don't think any of them are optional. 

In my mind, it's values that are passed in (not values that come out) that can be optional. 

In the case of the `channel.status` endpoint, the only thing that needs to be known to make the request is the channelID (which appears to be the same as the channel name). So if the request is made by calling a method on the REST channel object, no parameters need to be passed, therefore no optional parameters.

Using the existing `REST.Request` method the data that comes back looks like
```
[ map[
    channelId:test 
    name:test 
    status:map[
      isActive:true 
      occupancy:map[
        metrics:map[
          connections:1 
          presenceConnections:1 
          presenceMembers:1 
          presenceSubscribers:1 
          publishers:1 
          subscribers:1
        ]
      ]
    ]
  ]
]
```

content/client-lib-development-guide/features.textile
```diff
@@ -1336,6 +1339,32 @@ h4. ChannelProperties
 * @(CP2)@ The attributes of @ChannelProperties@ consist of:
 ** @(CP2a)@ @attachSerial@ string - contains the last @channelSerial@ received in an @ATTACHED@ @ProtocolMesage@ for the channel, see "RTL15a":#RTL15a
 
+h4. ChannelDetails
+
+* @(CHD1)@ @ChannelDetails@ is a type that represents information for a channel including channelID, name, status and occupancy
```
lmars, line 1344: The API returns a `channelId` field (note the lower case `d`), I think we should be consistent and refer to it as `channelId` here in the document too.

content/client-lib-development-guide/features.textile
```diff
@@ -1336,6 +1339,32 @@ h4. ChannelProperties
 * @(CP2)@ The attributes of @ChannelProperties@ consist of:
 ** @(CP2a)@ @attachSerial@ string - contains the last @channelSerial@ received in an @ATTACHED@ @ProtocolMesage@ for the channel, see "RTL15a":#RTL15a
 
+h4. ChannelDetails
+
+* @(CHD1)@ @ChannelDetails@ is a type that represents information for a channel including channelID, name, status and occupancy 
+* @(CHD2)@ The attributes of @ChannelDetails@ consist of:
+** @(CHD2a)@ @channelID@ string - the identifier of the channel
+** @(CHD2b)@ @channelName@ string - the name of the channel
```
lmars, line 1347:
```suggestion
** @(CHD2b)@ @name@ string - the name of the channel
```

content/client-lib-development-guide/features.textile
```diff
@@ -1336,6 +1339,32 @@ h4. ChannelProperties
...
+** @(CHD2c)@ @status@ @ChannelStatus@ - status an object containing channel status information
```
lmars, line 1348:
```suggestion
** @(CHD2c)@ @status@ @ChannelStatus@ - the status of the channel
```

content/client-lib-development-guide/features.textile
```diff
@@ -1336,6 +1339,32 @@ h4. ChannelProperties
...
+** @(CHD2a)@ @channelID@ string - the identifier of the channel
+** @(CHD2b)@ @channelName@ string - the name of the channel
+** @(CHD2c)@ @status@ @ChannelStatus@ - status an object containing channel status information 
+
+h4. ChannelStatus
+
+* @(CHS1)@ @ChannelStatus@ is a type that contains status and occupancy for a channel 
+* @(CHS2)@ The attributes of @ChannelStatus@ consist of:
+** @(CHS2a)@ @isActive@ boolean - represents if the channel is active
+** @(CHS2b)@ @occupancy@ @ChannelMetrics@ - occupancy is an object containing the metrics for the channel
```
lmars, line 1355: `occupancy` is actually an object with a `metrics` field, and the value of `metrics` is a `ChannelMetrics` object:

```json
{
        "name": "example",
        "channelId": "example",
        "status": {
                "isActive": true,
                "occupancy": {
                        "metrics": {
                                "connections": 1,
                                "publishers": 1,
                                "subscribers": 1,
                                "presenceConnections": 1,
                                "presenceMembers": 0,
                                "presenceSubscribers": 1
                        }
                }
        }
}
```

So I think we need an extra object in between.

lmars: Also, it concerns me that we are basically duplicating what is on the following docs page:

https://ably.com/docs/realtime/channel-metadata#types

the source of which is already contained within this same GitHub repository:

https://github.com/ably/docs/blob/main/content/partials/types/_channel_details.textile

content/client-lib-development-guide/features.textile
Rosalita: Good spot, I've pushed this change.

content/client-lib-development-guide/features.textile
Rosalita: Thanks for the feedback. I've removed `RSL8b`.

content/client-lib-development-guide/features.textile
Rosalita: I've expanded this description and also the description of RSL8a as the GET http request could potentially return an error.

content/client-lib-development-guide/features.textile
Rosalita: Not sure how to action this, would be great if you could make the change you would like using the suggestion tool.

content/client-lib-development-guide/features.textile
Rosalita: This has been added

content/client-lib-development-guide/features.textile
owenpearson: Looks good, although I would suggest changing it to something along the lines of:
> makes a http get request to @\<restHost\>/channels/\<channelId\>@ where @\<restHost\>@ represents the current rest host as described by "@RSC11@":#RSC11.

Also, although it's fairly obvious that it's passed in as an argument to `status`, I think we should explicitly state where \<channelId\> comes from just to avoid any ambiguity.

content/client-lib-development-guide/features.textile
```diff
@@ -300,6 +300,8 @@ h3(#rest-channel). Channel
...
 ** @(RSL6b)@ If, for example, incompatible encryption details are provided or invalid Base64 is detected in the message payload, an error message will be sent to the logger, but the message will still be delivered with last successful decoding and the @encoding@ field. For example, if a message had a decoding of "utf-8/cipher+aes-128-cbc/base64", and the payload was successfully Base64 decoded but the payload could not be decrypted because the @CipherParam@ details were not configured, the message would be delivered with a binary payload and an @encoding@ with the value "utf-8/cipher+aes-128-cbc". Additional steps need to be taken if decoding failed on "vcdiff" encoding; see "RTL18":#RTL18
 * @(RSL7)@ @Channel#setOptions@ takes a @ChannelOptions@ object and sets or updates the stored channel options, then indicates success
+* @(RSL8)@ @Channel#status@ function: makes a http get request to @rest.ably.io/channels/<channelId>@
+** @(RSL8a)@ Returns a @ChannelDetails@ object or an @ErrorInfo@ object with details of the error that has occurred
```
owenpearson, line 304: I would expect that in most languages the `ErrorInfo` would be thrown rather than returned. I'd suggest either:
1. Change this to "Returns a @ChannelDetails@ object or results in an @ErrorInfo@ object with details of the error that has occurred"
2. Remove this spec point and consider it already covered by `TI2`

content/client-lib-development-guide/features.textile
```diff
@@ -1336,6 +1338,38 @@ h4. ChannelProperties
...
+h4. ChannelStatus
+
+* @(CHS1)@ @ChannelStatus@ is a type that contains status and occupancy for a channel 
+* @(CHS2)@ The attributes of @ChannelStatus@ consist of:
+** @(CHS2a)@ @isActive@ boolean - represents if the channel is active
+** @(CHS2b)@ @occupancy@ @ChannelOccupancy@ - occupancy is an object containing the metrics for the channel
+
+h4. ChannelOccupancy
+
+* @(CHO1)@ is a type that contain channel metrics
```
owenpearson, line 1358: This line and the one below are missing a capital letter at the start.

content/client-lib-development-guide/features.textile
```diff
@@ -1701,6 +1736,26 @@ class ChannelOptions:
   params?: Dict<String, String> // TB2c
   modes?: [ChannelMode] // TB2d
 
+class ChannelDetails:
+  channelId: String //CHD2a
```
owenpearson, line 1740: We should add a space in between the comment symbol and the comment (ie `//CHD2a` -> `// CHD2a`) for consistency with the rest of the comments in this section.

owenpearson (changes requested):

content/client-lib-development-guide/features.textile
Rosalita: I've just pushed the first suggestion in this comment.

Regarding your second suggestion. I wasn't expecting channelId to be passed into the channel.Status function at all. This is because Status is a method on a channel object that will already know the its own name, which is the channelID. 

If I have misunderstood something here, please feel free to use the suggestion tool to show me what you mean.

content/client-lib-development-guide/features.textile
owenpearson: Oh wait, I can see that this line kind of follows on from the `h4` above... Maybe change it to:
> * @(CHO1)@ A @ChannelOccupancy@ is a type that contain channel metrics

(Inspired by [`TG1`](https://github.com/ably/docs/blob/main/content/client-lib-development-guide/features.textile#paginatedresult)).

The line below should definitely start with a capital though.

content/client-lib-development-guide/features.textile
Rosalita: So this is was written from my point of view as a Go developer working in a language where it's idiomatic to communicate errors via an explicit, separate return values. In my Go world errors are not thrown, they are always returned and can be nil. 

I think this certainly proves it's exceptionally hard to write a specification that makes sense in multiple languages. So I've gone with suggestion 2 and removed the information about `ErrorInfo`.

content/client-lib-development-guide/features.textile
Rosalita: Done

content/client-lib-development-guide/features.textile
Rosalita: Done

owenpearson: There are still some more places where this needs to be changed

Rosalita: Is there a way to auto format this document?

content/client-lib-development-guide/features.textile
```diff
@@ -1701,6 +1736,26 @@ class ChannelOptions:
...
+
+class ChannelStatus:
+  isActive: Boolean // CHS2a
+  occupancy: ChannelOccupancy // CHS2b
+
+class ChannelOccupancy:
+  metrics: ChannelMetrics // CHO2a
+
+class ChannelMetrics:
+  connections: Int //CHM2a
```
owenpearson, line 1752:
```suggestion
  connections: Int // CHM2a
```

owenpearson (approved):
LGTM, nice work 🙂

Gemfile.lock
```diff
@@ -96,6 +96,8 @@ GEM
       nanoc (~> 4.9)
     nenv (0.3.0)
     netrc (0.11.0)
+    nokogiri (1.13.4-arm64-darwin)
```
mohyour, line 99: It appears changes to the Gemfile shouldn't be committed

lmars: Given someone decided in the past to commit Gemfile.lock rather than adding it to .gitignore, I think we can assume that if it changes, we should commit the change.

lmars (approved):

Gemfile.lock
Rosalita: I'm not going to create an issue to add `Gemfile.lock` to `.gitignore` because it's used as a snapshot of all the required gems and their dependencies. By committing `Gemfile.lock` the bundler doesn't have to recalculate all the gem dependencies each time the project is run. This file would have updated itself when I ran this project locally.

content/client-lib-development-guide/features.textile
QuintinWillison: Sorry, I wasn't very clear. I think this should work:

```suggestion
** @(RSL8a)@ Returns a @ChannelDetails@ object - see "@CHD1@":#CHD1
```

That should create a clickable, code-formatted hyperlink... which, as far as I've seen, is the 'house style' for this document in respect of internal cross references.

QuintinWillison (approved):

PR #1421, opened 2022-05-09
Batch operations spec points

Peter-Maguire:

## Description

Adds spec points for the Batch Operations functions.

* [Jira epic ticket](https://ably.atlassian.net/browse/SDK-1792).
* [Jira ticket](https://ably.atlassian.net/browse/SDK-1810).

## Review

* [Page to review](https://docs.ably.io/client-lib-development-guide/features/)

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -343,9 +343,9 @@ h3(#rest-compatibility). Forwards compatibility
 h3(#batch-operations). Batch Operations
 * @(BO1)@ The batch operations functions must use the REST endpoints in Batch Mode, sending a single request containing all specified data
 * @(BO2)@ Batch operations must be able to be performed for the following:
-** @(BO2a)@ Publishing messages against one or more channels with one or more messages
-*** @(B02a1)@ Functions should be provided to pass either an array or a single object for both Message and Channel
-** @(BO2b)@ Retrieving the presence data for one or more channels
+** @(BO2a)@ @BatchOperations::publish@ publishes messages against one or more channels with one or more messages
+*** @(B02a1)@ Functions should be provided to pass either an array or a single object for both Message and Channel ID
+** @(BO2b)@ @BatchOperations::getPresence@ retrieves the presence data for one or more channels
 * @(BO3)@ When all passed arrays contain a single object, the underlying request is functionally identical to its non-batch equivalent, but the returned result should be a @BatchResponse@ object.
 
 h2(#realtime). Realtime client library features
@@ -1357,7 +1357,7 @@ h4. ChannelDetails
 
 h4. ChannelStatus
 
-* @(CHS1)@ @ChannelStatus@ is a type that contains status and occupancy for a channel 
+* @(CHS1)@ @ChannelStatus@ is a type that contains status and occupancy for a channel
 * @(CHS2)@ The attributes of @ChannelStatus@ consist of:
 ** @(CHS2a)@ @isActive@ boolean - represents if the channel is active
 ** @(CHS2b)@ @occupancy@ @ChannelOccupancy@ - occupancy is an object containing the metrics for the channel
@@ -1370,13 +1370,13 @@ h4. ChannelOccupancy
 
 h4. ChannelMetrics
 
-* @(CHM1)@ @ChannelMetrics@ is a type that contains the count of publishers and subscribers, connections and presenceConnections, presenceMembers and presenceSubscribers 
+* @(CHM1)@ @ChannelMetrics@ is a type that contains the count of publishers and subscribers, connections and presenceConnections, presenceMembers and presenceSubscribers
 * @(CHM2)@ The attributes of @ChannelMetrics@ consist of:
 ** @(CHM2a)@ @connections@ integer - the total number of connections to the channel
 ** @(CHM2b)@ @presenceConnections@ integer - the total number of presence connections to the channel
 ** @(CHM2c)@ @presenceMembers@ integer - the total number of presence members for the channel
 ** @(CHM2d)@ @presenceSubscribers@ integer - the total number of presence subscribers for the channel
-** @(CHM2e)@ @publishers@ integer - the total number of publishers to the channel 
+** @(CHM2e)@ @publishers@ integer - the total number of publishers to the channel
 ** @(CHM2f)@ @subscribers@ integer - the total number of subscribers to the channel
 
 h4. BatchResult
@@ -1722,7 +1722,6 @@ class RealtimeChannel:
     end: Time api-default now(), // RTL10a
     direction: .Backwards | .Forwards api-default .Backwards, // RTL10a
     limit: int api-default 100, // RTL10a
-    limit: int api-default 100, // RTL10a
     untilAttach: Bool default false // RTL10b
   ) => io PaginatedResult<Message> // RSL2a
   publish(Message) => io // RTL6i
@@ -1736,10 +1735,10 @@ class RealtimeChannel:
   setOptions(options: ChannelOptions) => io // RTL16
 
 class BatchOperations:
-  publish([Channel], [Message]) => BatchResult<BatchPublishResponse> // BO2a
-  publish(Channel, [Message]) => BatchResult<BatchPublishResponse> // BO2a
-  publish([Channel], Message) => BatchResult<BatchPublishResponse> // BO2a
-  getPresence([Channel]) => BatchResult<BatchPresenceResponse> // BO2b
+  publish([String], [Message]) => BatchResult<BatchPublishResponse> // BO2a
+  publish(String, [Message]) => BatchResult<BatchPublishResponse> // BO2a
+  publish([String], Message) => BatchResult<BatchPublishResponse> // BO2a
+  getPresence([String]) => BatchResult<BatchPresenceResponse> // BO2b
 
 class BatchResult<T>:
   error: ErrorInfo? // BPA2b
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -1336,6 +1336,33 @@ h4. ChannelProperties
 * @(CP2)@ The attributes of @ChannelProperties@ consist of:
 ** @(CP2a)@ @attachSerial@ string - contains the last @channelSerial@ received in an @ATTACHED@ @ProtocolMesage@ for the channel, see "RTL15a":#RTL15a
 
+h4. BatchResult
+* @(BPA1)@ Contains the results from the batch operation
+* @(BPA2)@ @BatchResult@ has the following attributes:
+** @(BPA2a)@ @responses@ is an array of batch response objects. undefined if the request failed completely (e.g. for an invalid key)
+** @(BPA2b)@ @error@ is an @ErrorInfo@ object which is populated if one or more batch publish requests failed
+
+h4. BatchPublishResponse
+* @(BPB1)@ Contains information for each batch publish request within a @BatchResult@
+* @(BPB2)@ @BatchPublishResponse@ has the following attributes:
+** (@BPB2a)@ @channel@ is the channel name which this publish request was directed to
+** (@BPB2b)@ @messageId@ contains the resultant message ID, if the request succeeds and is undefined if @error@ is present
+** (@BPB2c)@ @error@ contains an @ErrorInfo@ object if this publish request failed, and is undefined if it succeeded
+
+h4. BatchPresenceResponse
+* @(BPD1)@ Contains information for each batch presence request within a @BatchResult@
+* @(BPD2)@ @BatchPresenceResponse@ contains the following attributes:
+** @(BPD2a)@ @channel@ is the channel name which this presence request
+** @(BPD2b)@ @presence@ is an array of presence data for the @channel@
+
+h4. BatchPresence
+* @(BPE1)@ Is a partial @PresenceMessage@ object containing @clientId@ and @action@, or @error@ if the presence failed
+* @(PBE2)@ @BatchPresence@ contains the following attributes:
+** @(PBE2a)@ @clientId@ - identical to #TP3c
+** @(PBE2a)@ @action@ - identical to #TP3b - undefined if @error@ is present
+** @(PBE2a)@ @error@ - an @ErrorInfo@ object representing the failure reason for this channel - undefined if @action@ is present
+
+
 h3(#options). Option types
 
 h4. ClientOptions
@@ -1664,27 +1691,23 @@ class RealtimeChannel:
   setOptions(options: ChannelOptions) => io // RTL16
 
 class BatchOperations:
-  publish([Channel], [Message]) => BatchPublishResult
-  publish(Channel, [Message]) => BatchPublishResult
-  publish([Channel], Message) => BatchPublishResult
-  getPresence([Channel]) => BatchPresenceResult
+  publish([Channel], [Message]) => BatchResult<BatchPublishResponse>
+  publish(Channel, [Message]) => BatchResult<BatchPublishResponse>
+  publish([Channel], Message) => BatchResult<BatchPublishResponse>
+  getPresence([Channel]) => BatchResult<BatchPresenceResponse>
 
-class BatchPublishResult:
-  error: ErrorInfo?
-  responses: []BatchPublishResponse?
+class BatchResult<T>:
+  error: ErrorInfo? // BPA2b
+  responses: []T? // BPA2a
 
 class BatchPublishResponse:
-  channel: String
-  messageId: String?
-  error: ErrorInfo?
+  channel: String // BPB2a
+  messageId: String? // BPB2b
+  error: ErrorInfo? // BPB2c
 
-class BatchPresenceResult:
-  error: ErrorInfo?
-  responses: []BatchPresenceResponse?
-
-class BatchPresenceResponse
-  channel: String
-  presence: []BatchPresence
+class BatchPresenceResponse:
+  channel: String // BPD2a
+  presence: []BatchPresence // PBD2b
 
 class BatchPresence:
   clientId: string
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -1402,8 +1402,8 @@ h4. BatchPresence
 * @(BPE1)@ Is a partial @PresenceMessage@ object containing @clientId@ and @action@, or @error@ if the presence failed
 * @(PBE2)@ @BatchPresence@ contains the following attributes:
 ** @(PBE2a)@ @clientId@ - identical to #TP3c
-** @(PBE2a)@ @action@ - identical to #TP3b - undefined if @error@ is present
-** @(PBE2a)@ @error@ - an @ErrorInfo@ object representing the failure reason for this channel - undefined if @action@ is present
+** @(PBE2b)@ @action@ - identical to #TP3b - undefined if @error@ is present
+** @(PBE2c)@ @error@ - an @ErrorInfo@ object representing the failure reason for this channel - undefined if @action@ is present
 
 
 h3(#options). Option types
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -344,7 +344,7 @@ h3(#batch-operations). Batch Operations
 * @(BO1)@ The batch operations functions must use the REST endpoints in Batch Mode, sending a single request containing all specified data
 * @(BO2)@ Batch operations must be able to be performed for the following:
 ** @(BO2a)@ @BatchOperations::publish@ publishes messages against one or more channels with one or more messages
-*** @(B02a1)@ Functions should be provided to pass either an array or a single object for both Message and Channel ID
+*** @(B02a1)@ Functions should be provided to pass either an array or a single object for both Message and Channel ID, or an array of @BatchSpec@ objects
 ** @(BO2b)@ @BatchOperations::getPresence@ retrieves the presence data for one or more channels
 * @(BO3)@ When all passed arrays contain a single object, the underlying request is functionally identical to its non-batch equivalent, but the returned result should be a @BatchResponse@ object.
 
@@ -1738,6 +1738,7 @@ class BatchOperations:
   publish([String], [Message]) => BatchResult<BatchPublishResponse> // BO2a
   publish(String, [Message]) => BatchResult<BatchPublishResponse> // BO2a
   publish([String], Message) => BatchResult<BatchPublishResponse> // BO2a
+  publish([BatchSpec]) => BatchResult<BatchPublishResponse> // BO2a
   getPresence([String]) => BatchResult<BatchPresenceResponse> // BO2b
 
 class BatchResult<T>:
@@ -1764,6 +1765,10 @@ class PushChannel:
   unsubscribeDevice() => io // RSH7c
   unsubscribeClient() => io // RSH7d
   listSubscriptions() => io PaginatedResult<PushChannelSubscription> // RSH7e
+
+class BatchSpec:
+  channels: [String]
+  messages: [Message]
 
 enum ChannelState:
   INITIALIZED
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -1516,6 +1516,7 @@ class Realtime:
   auth: Auth // RTC4
   push: Push
   device() => io LocalDevice
+  batch: BatchOperations
   channels: Channels<RealtimeChannel> // RTC3, RTS1
   clientId: String? // proxy for RSA7
   connection: Connection // RTC2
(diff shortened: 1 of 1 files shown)


content/client-lib-development-guide/features.textile
```diff
@@ -339,6 +340,14 @@ h3(#rest-encryption). Encryption
 h3(#rest-compatibility). Forwards compatibility
 * @(RSF1)@ The library must apply the "robustness principle":https://en.wikipedia.org/wiki/Robustness_principle in its processing of requests and responses with the Ably system. In particular, deserialization of Messages and related types, and associated enums, must be tolerant to unrecognised attributes or enum values. Such unrecognised values must be ignored.
 
+h3(#batch-operations). Batch Operations
+* @(BO1)@ The batch operations functions must use the REST endpoints in Batch Mode, sending a single request containing all specified data
+* @(BO2)@ Batch operations must be able to be performed for the following:
```
owenpearson, line 345: I know you've got method names in the IDL but I think generally in the spec we mention the name of the method in the spec too, see RSE1 and RSP3 for examples.

content/client-lib-development-guide/features.textile
```diff
@@ -339,6 +340,14 @@ h3(#rest-encryption). Encryption
...
+** @(BO2a)@ Publishing messages against one or more channels with one or more messages
+*** @(B02a1)@ Functions should be provided to pass either an array or a single object for both Message and Channel
+** @(BO2b)@ Retrieving the presence data for one or more channels
+* @(BO3)@ When all passed arrays contain a single object, the underlying request is functionally identical to it's non-batch equivalent, but the returned result should be a @BatchResponse@ object.
```
owenpearson, line 349:
```suggestion
* @(BO3)@ When all passed arrays contain a single object, the underlying request is functionally identical to its non-batch equivalent, but the returned result should be a @BatchResponse@ object.
```

owenpearson (changes requested):
Looks good, just a few comments. Also, I notice that the REST API supports multiple BatchSpec objects in one HTTP request (as in specifying different messages for different channels in the same batch). Did you consider supporting this in SDKs? Even if we don't support it initially it might be good to design this such that we can support it in the future without a breaking change.

content/client-lib-development-guide/features.textile
Peter-Maguire: [Done](https://github.com/ably/docs/pull/1421/commits/2cdba4a3cc09efeb703bacc864b0e1f6489e068a)

content/client-lib-development-guide/features.textile
```diff
@@ -1697,6 +1735,30 @@ class RealtimeChannel:
   unsubscribe(String, (Message) ->) // RTL8a
   setOptions(options: ChannelOptions) => io // RTL16
 
+class BatchOperations:
+  publish([Channel], [Message]) => BatchResult<BatchPublishResponse> // BO2a
```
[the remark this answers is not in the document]
Peter-Maguire: Good shout, [done](https://github.com/ably/docs/pull/1421/commits/2cdba4a3cc09efeb703bacc864b0e1f6489e068a)

content/client-lib-development-guide/features.textile
```diff
@@ -1685,6 +1722,7 @@ class RealtimeChannel:
     end: Time api-default now(), // RTL10a
     direction: .Backwards | .Forwards api-default .Backwards, // RTL10a
     limit: int api-default 100, // RTL10a
+    limit: int api-default 100, // RTL10a
```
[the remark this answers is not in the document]
Peter-Maguire: Good spot, [done](https://github.com/ably/docs/pull/1421/commits/2cdba4a3cc09efeb703bacc864b0e1f6489e068a)

Peter-Maguire:
> Looks good, just a few comments. Also, I notice that the REST API supports multiple BatchSpec objects in one HTTP request (as in specifying different messages for different channels in the same batch). Did you consider supporting this in SDKs? Even if we don't support it initially it might be good to design this such that we can support it in the future without a breaking change.

@owenpearson this is a good question, I definitely agree that it should be implemented here, although I can't quite figure out how the rest way to do it would be. I've [pushed a commit](https://github.com/ably/docs/pull/1421/commits/859971218fb5ea2cf83a84586812ea050125e6ed) for how I think it should work, could you take a look and see if it's how you were thinking too?

content/client-lib-development-guide/features.textile
```diff
@@ -1361,15 +1370,42 @@ h4. ChannelOccupancy
...
 ** @(CHM2c)@ @presenceMembers@ integer - the total number of presence members for the channel
 ** @(CHM2d)@ @presenceSubscribers@ integer - the total number of presence subscribers for the channel
-** @(CHM2e)@ @publishers@ integer - the total number of publishers to the channel 
+** @(CHM2e)@ @publishers@ integer - the total number of publishers to the channel
 ** @(CHM2f)@ @subscribers@ integer - the total number of subscribers to the channel
 
+h4. BatchResult
+* @(BPA1)@ Contains the results from the batch operation
+* @(BPA2)@ @BatchResult@ has the following attributes:
+** @(BPA2a)@ @responses@ is an array of batch response objects. undefined if the request failed completely (e.g. for an invalid key)
```
Rosalita, line 1385: This sentence doesn't start with a capital letter so maybe it's been cut off? `undefined` also isn't a type in some programming languages (like Go) maybe this description can be more programming language neutral?

Peter-Maguire: 'responses' is lowercase because it references the field in the IDL of the same name. I can see your point on the use of undefined, although I was using it in the literally sense of "this value is undefined" rather than referencing a language specific concept. I can see how this could be confusing, though, so I'm open to changing it to 'null' (or 'not set' to be fully language agnostic) if you think that would be clearer.

owenpearson: `null` is used a lot throughout the spec so I'd vote we go with that.

owenpearson:
@Peter-Maguire Nice, that is pretty much the API I was imagining, although now that I see it side-by-side with the pre-existing API I almost wonder if it's worth keeping the `(string[], Message[])` API, since it would be almost identical to using the `(BatchSpec[])` API, for example:
```ts
// Using the (string[], Message[]) API
client.batch.publish(['my_cool_channel', 'my_other_cool_channel'], { foo: 'bar' });

// Using the (BatchSpec[]) API
client.batch.publish([{
  channels: ['my_cool_channel', 'my_other_cool_channel'],
  messages: [{ foo: 'bar' }]
}]);
```

I haven't thought about it too much but my current thinking is maybe we just support two `batch.publish` signatures: one which takes a single `BatchSpec`, and another which takes a `BatchSpec[]`. WDYT?

content/client-lib-development-guide/features.textile
```diff
@@ -339,6 +340,14 @@ h3(#rest-encryption). Encryption
 h3(#rest-compatibility). Forwards compatibility
 * @(RSF1)@ The library must apply the "robustness principle":https://en.wikipedia.org/wiki/Robustness_principle in its processing of requests and responses with the Ably system. In particular, deserialization of Messages and related types, and associated enums, must be tolerant to unrecognised attributes or enum values. Such unrecognised values must be ignored.
 
+h3(#batch-operations). Batch Operations
```
Rosalita, line 343: We should maybe add something to say that batch operations "best effort", and not transactional (all or nothing) in nature.

Like If a user was to publish a batch of 2 messages. If the first publish failed, an attempt would still be made to publish the second message.  The reason to add this info would be that in some contexts, batch operations (like database writes) the batch can be transactional where either everything is written or nothing is written.

Peter-Maguire:
@owenpearson my initial thinking with including the various inputs for single channels/messages and arrays of them was for convenience. Perhaps having a combination of each plus the BatchSpec methods would be too many different variations, though. Especially for languages that don't support function overloading. I will remove the shorthand functions and replace them with BatchSpec and []BatchSpec, I think that will be easier  to implement more consistently.

content/client-lib-development-guide/features.textile
Peter-Maguire: Would this be a consideration from an SDK perspective? I would imagine that the handling for the returned value is the same as the transactional nature is only relevant to the server-side implementation. If there is a difference in client-side behavior that you can foresee I will put a spec point in.

content/client-lib-development-guide/features.textile
Peter-Maguire: [Done](https://github.com/ably/docs/pull/1421/commits/1e1ded87b7b3c1cc7219e46f0aac1828736be373)

content/client-lib-development-guide/features.textile
```diff
@@ -339,6 +340,14 @@ h3(#rest-encryption). Encryption
...
+* @(BO1)@ The batch operations functions must use the REST endpoints in Batch Mode, sending a single request containing all specified data
+* @(BO2)@ Batch operations must be able to be performed for the following:
+** @(BO2a)@ @BatchOperations::publish@ publishes messages against one or more channels with one or more messages
+*** @(B02a1)@ Functions should be provided to pass either an array or a single object for both Message and Channel ID, or an array of @BatchSpec@ objects
```
Rosalita: I would avoid 'array or a single object' as this language feels prescriptive and not programming language neutral (this could be called a slice in Go.)

Also is it worth mentioning that the BatchSpec object behaves like a matrix? In my mind I imagine it working like a matrix where all messages are applied to all channelIDs like...

```
[ 
    {
        channels: []channel{"1", "2"},
        messages: []messages{"1", "2"}
    },

    {
        channels: []channels{"3"},
        messages: []messages{"1"}
    }
]
```
This example batch spec would apply message "1" & "2" to both channels "1" & "2" and only apply message "1" to channel "3"

content/client-lib-development-guide/features.textile
```diff
@@ -339,6 +340,14 @@ h3(#rest-encryption). Encryption
...
+** @(BO2b)@ @BatchOperations::getPresence@ retrieves the presence data for one or more channels
+* @(BO3)@ When all passed arrays contain a single object, the underlying request is functionally identical to its non-batch equivalent, but the returned result should be a @BatchResponse@ object.
```
Rosalita, line 349:
```suggestion
* @(BO3)@ When a batch operation only contains one batch, the underlying request is functionally identical to its non-batch equivalent, but the returned result should be a @BatchResponse@ object.
```
An attempt to simplify this wording and use language that's a bit more programming language neutral.

content/client-lib-development-guide/features.textile
```diff
@@ -1361,15 +1370,42 @@ h4. ChannelOccupancy
...
 ** @(CHM2b)@ @presenceConnections@ integer - the total number of presence connections to the channel
 ** @(CHM2c)@ @presenceMembers@ integer - the total number of presence members for the channel
 ** @(CHM2d)@ @presenceSubscribers@ integer - the total number of presence subscribers for the channel
-** @(CHM2e)@ @publishers@ integer - the total number of publishers to the channel 
+** @(CHM2e)@ @publishers@ integer - the total number of publishers to the channel
 ** @(CHM2f)@ @subscribers@ integer - the total number of subscribers to the channel
 
+h4. BatchResult
+* @(BPA1)@ Contains the results from the batch operation
+* @(BPA2)@ @BatchResult@ has the following attributes:
```
Rosalita, line 1384: Idiomatic Go would return a pointer to a result (nil if no result) and and error. 
Like ` result, err := REST.batch.publish()` 
So I'm not sure if we should be prescribing that BatchResult has an error nested inside it. Maybe we can leave this open to being implemented idiomatically by each language.

content/client-lib-development-guide/features.textile
```diff
@@ -1361,15 +1370,42 @@ h4. ChannelOccupancy
...
+** @(BPA2a)@ @responses@ is an array of batch response objects. undefined if the request failed completely (e.g. for an invalid key)
+** @(BPA2b)@ @error@ is an @ErrorInfo@ object which is populated if one or more batch publish requests failed
+
+h4. BatchPublishResponse
+* @(BPB1)@ Contains information for each batch publish request within a @BatchResult@
+* @(BPB2)@ @BatchPublishResponse@ has the following attributes:
+** (@BPB2a)@ @channel@ is the channel name which this publish request was directed to
+** (@BPB2b)@ @messageId@ contains the resultant message ID, if the request succeeds and is undefined if @error@ is present
+** (@BPB2c)@ @error@ contains an @ErrorInfo@ object if this publish request failed, and is undefined if it succeeded
```
Rosalita: Same comment again as line 1384
Idiomatic Go would return a pointer to a result (nil if no result) and and error.
Like result, err := REST.batch.publish()
So I'm not sure if we should be prescribing that BatchPublishResult has an error nested inside it. Maybe we can leave this open to being implemented idiomatically by each language.

content/client-lib-development-guide/features.textile
```diff
@@ -1697,13 +1734,42 @@ class RealtimeChannel:
...
   setOptions(options: ChannelOptions) => io // RTL16
 
+class BatchOperations:
+  publish([String], [Message]) => BatchResult<BatchPublishResponse> // BO2a
+  publish(String, [Message]) => BatchResult<BatchPublishResponse> // BO2a
+  publish([String], Message) => BatchResult<BatchPublishResponse> // BO2a
+  publish([BatchSpec]) => BatchResult<BatchPublishResponse> // BO2a
+  getPresence([String]) => BatchResult<BatchPresenceResponse> // BO2b
+
+class BatchResult<T>:
```
Rosalita, line 1742: This looks like a generic where batch result is an array of type T. 
I'm not sure how a programming language which doesn't have generics (like php) would implement this.

Generics are still very, very new in Go and the release notes advising using caution when deploying code containing generics to production (https://go.dev/doc/go1.18) 

Will languages like PHP (and in the immediate short term Go) be able to make their own implementation choices here?

content/client-lib-development-guide/features.textile
Rosalita: Thinking about it now, I'm not sure how a publish operation could be transactional as the SDK only knows if a publish has failed when it tries to publish, by which time the message has gone and can't be rolled back.  I guess it's inferred that all the batch operations for publish are "best effort" as they are sending data.

content/client-lib-development-guide/features.textile
Peter-Maguire: > I would avoid 'array or a single object' as this language feels prescriptive and not programming language neutral (this could be called a slice in Go.)  
  
Ideally I think there would be a terminology guide for things like this. The point I wanted to get across with this wording vs something like 'one or more' is that, language permitting, it should be possible to pass either a single object or an array, as opposed to just an array with one or more objects.

content/client-lib-development-guide/features.textile
Peter-Maguire: Yes, generics are used elsewhere in the IDL and I believe the correct implementation in a language without generics here would be to make `BatchPublishResult` and `BatchPresenceResult`, but that would be up to the implementer.

content/client-lib-development-guide/features.textile
Peter-Maguire: The issue here is that when a partial response is received, an error is supplied. Having an error response and a success response is impossible or discouraged in a lot of languages, perhaps either the partial error code should be ignored or full errors should be handled separately in a language-specific way?

content/client-lib-development-guide/features.textile
Rosalita: I was possibly thinking about this from a strongly typed language point of view. But the more I think about it, single object or multiple objects should be possible to get into a slice in Go with the `..` varadic operator.

Rosalita (approved):
Thank you for addressing all my comments, I have no more comments to make 👍

owenpearson (approved):
LGTM 👍

mattheworiordan:
This PR is great, apologies for feedback after the fact, but I was not aware of this PR until today.  It's so great to see the spec evolving 🥳 

I do have on question though. Why does the batch method not exist on the Realtime SDK?  I believe this creates inconsistency in what we've done historically where all Rest functionality is available in the Realtime SDK by default.  As you can see from the screenshot below, by not doing this, we're deviating from what we've done historically.  Was this intentional or just an omission?

MO screenshot 2022-05-13 at 15 23 40

content/client-lib-development-guide/features.textile
```diff
@@ -1697,13 +1735,40 @@ class RealtimeChannel:
...
+  error: ErrorInfo? // BPB2c
+
+class BatchPresenceResponse:
+  channel: String // BPD2a
+  presence: []BatchPresence // PBD2b
+
+class BatchPresence:
+  clientId: string
+  action: string?
+  error: ErrorInfo?
```
lmars, line 1759: I think this error belongs on `BatchPresenceResponse`

content/client-lib-development-guide/features.textile
```diff
@@ -1697,13 +1735,40 @@ class RealtimeChannel:
...
+
 class PushChannel:
   subscribeDevice() => io // RSH7a
   subscribeClient() => io // RSH7b
   unsubscribeDevice() => io // RSH7c
   unsubscribeClient() => io // RSH7d
   listSubscriptions() => io PaginatedResult<PushChannelSubscription> // RSH7e
 
+class BatchSpec:
```
lmars, line 1768: I know the documentation calls this `BatchSpec` too, but it seems specific to publishing a batch of messages, so perhaps it should be called something like `BatchPublishRequest`?

content/client-lib-development-guide/features.textile
```diff
@@ -1697,13 +1735,40 @@ class RealtimeChannel:
...
+  responses: []T? // BPA2a
+
+class BatchPublishResponse:
+  channel: String // BPB2a
+  messageId: String? // BPB2b
+  error: ErrorInfo? // BPB2c
+
+class BatchPresenceResponse:
+  channel: String // BPD2a
+  presence: []BatchPresence // PBD2b
```
lmars, line 1754: I think this should be `[]PresenceMessage`, not sure an extra type is needed.

content/client-lib-development-guide/features.textile
```diff
@@ -1697,13 +1735,40 @@ class RealtimeChannel:
   unsubscribe(String, (Message) ->) // RTL8a
   setOptions(options: ChannelOptions) => io // RTL16
 
+class BatchOperations:
+  publish([BatchSpec]) => BatchResult<BatchPublishResponse> // BO2a
```
lmars, line 1739: Thinking about partial success, I think I'd like to get back something which helps me understand which of the things I tried to publish failed.

Whilst I understand why the REST API is designed as it is, I think there's an opportunity to wrap it in a client-side abstraction which might be more useful.

For example, what if it looked like this:

```
batch := client.newBatchPublish()
req1 := batch.publish("channel1", "msg1")
req2 := batch.publish("channel1", "msg2")
req3 := batch.publish("channel2", "msg3")
results, err := batch.send()
```

and `results` somehow helps me correlate any errors with one or more of `req1`, `req2`, or `req3`.

lmars (commented):
@Peter-Maguire @owenpearson apologies for being late with my review too! Happy to move the conversation elsewhere, but I thought for now this was the most appropriate place.

content/client-lib-development-guide/features.textile
```diff
@@ -1525,6 +1561,7 @@ class Rest:
   constructor(ClientOptions) // RSC1
   auth: Auth // RSC5
   push: Push
+  batch: BatchOperations // BO1
```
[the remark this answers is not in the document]
Peter-Maguire: I agree, this didn't feel like a 'perfect' solution to me either. I was stuck for a better way to do it though, especially for batch publishing. I could add batch presence to RestPresence easily, but batch publishing to multiple channels didn't feel like it would work in the same way as publishing happens on a Channel object

content/client-lib-development-guide/features.textile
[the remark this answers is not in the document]
Peter-Maguire: The outer response is already genericised as BatchResult<BatchPublishResponse>. I'll update this line to reflect that.

content/client-lib-development-guide/features.textile
```diff
@@ -1370,6 +1379,33 @@ h4. ChannelMetrics
 ** @(CHM2e)@ @publishers@ integer - the total number of publishers to the channel 
 ** @(CHM2f)@ @subscribers@ integer - the total number of subscribers to the channel
 
+h4. BatchResult
+* @(BPA1)@ Contains the results from the batch operation
+* @(BPA2)@ @BatchResult@ has the following attributes:
+** @(BPA2a)@ @responses@ is an array of batch response objects. undefined if the request failed completely (e.g. for an invalid key)
```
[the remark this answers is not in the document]
Peter-Maguire: There was a discussion around this [here](https://github.com/ably/docs/pull/1421#discussion_r870121878) where it was changed to `null`.

content/client-lib-development-guide/features.textile
Peter-Maguire: The [batch publish documentation](https://ably.com/docs/rest/batch#publish-requests) reads as if the BatchSpec object is the one passed into the body of the request either as a single object or within an array as it's being used here. In any case I would argue that the spec should be naming things as consistently as possible with the REST documentation.

content/client-lib-development-guide/features.textile
Peter-Maguire: PresenceMessage has far more fields on it than what appears in the response for a batch presence request. If those fields are all optional or can be retrieved from context then I would be happy to use that type instead.

content/client-lib-development-guide/features.textile
lmars: > In any case I would argue that the spec should be naming things as consistently as possible with the REST documentation.

I agree with that, so I guess what I'm saying is, how about we rename it to `BatchPublishRequest` in both places? This term isn't something that appears in the API response itself, so we should be free to change it, and now would be the time to change it if we think it could be named better before it starts to appear in SDK code.

content/client-lib-development-guide/features.textile
lmars: > PresenceMessage has far more fields on it than what appears in the response for a batch presence request

Can you give an example? I'm not sure why retrieving presence for multiple channels in one request should return less fields than retrieving for a single channel, if that is the case it sounds more like a problem with the API that we should fix rather than something we should be encoding into the type system.

Peter-Maguire: According to the spec, PresenceMessage has `connectionId`, `id` and `timestamp` along with `action` and `clientId` which are the only two fields that the REST API returns for batch presence.  
[Docs link for PresenceMessage](https://ably.com/docs/api/rest-api#presence) | [Docs link for BatchPresence](https://ably.com/docs/rest/batch#presence-success)

lmars: > PresenceMessage has connectionId, id and timestamp along with action and clientId which are the only two fields that the REST API returns for batch presence.

The docs may seem to indicate that, but it's not accurate, you get the same fields from both endpoints:

```
$ curl -u "${ABLY_API_KEY}" "https://rest.ably.io/channels/test1/presence"
[
        {
                "id": "LnUWZJf1G-:0:0",
                "clientId": "lmars",
                "connectionId": "LnUWZJf1G-",
                "timestamp": 1653686295156,
                "data": "",
                "action": 1
        }
]

$ curl -u "${ABLY_API_KEY}" "https://rest.ably.io/presence?channel=test1"
[
        {
                "channel": "test1",
                "presence": [
                        {
                                "id": "LnUWZJf1G-:0:0",
                                "clientId": "lmars",
                                "connectionId": "LnUWZJf1G-",
                                "timestamp": 1653686295156,
                                "data": "",
                                "action": 1
                        }
                ]
        }
]
```

content/client-lib-development-guide/features.textile
```diff
@@ -339,6 +340,14 @@ h3(#rest-encryption). Encryption
 h3(#rest-compatibility). Forwards compatibility
 * @(RSF1)@ The library must apply the "robustness principle":https://en.wikipedia.org/wiki/Robustness_principle in its processing of requests and responses with the Ably system. In particular, deserialization of Messages and related types, and associated enums, must be tolerant to unrecognised attributes or enum values. Such unrecognised values must be ignored.
 
+h3(#batch-operations). Batch Operations
+* @(BO1)@ The batch operations functions must use the REST endpoints in Batch Mode, sending a single request containing all specified data
```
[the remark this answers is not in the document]
owenpearson: Removed this sentence and added some clarity about the request in [015aa44](https://github.com/ably/docs/pull/1450/commits/015aa4425b540a281f1c49a4c23f81c3d3007e5c)

content/client-lib-development-guide/features.textile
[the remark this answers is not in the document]
owenpearson: The only alternative I can think of is to have `RestChannels#batchPublish` and `RestPresence#batchGet`, @paddybyers WDYT?

content/client-lib-development-guide/features.textile
owenpearson: I've updated to use `PresenceMessage` in [384b4fe](https://github.com/ably/docs/pull/1450/commits/384b4fe48a169b3f7b40b839ee25611035a7965b). I've also created https://github.com/ably/docs/issues/1451 to address this on the docs side.

content/client-lib-development-guide/features.textile
owenpearson: Good spot, I've moved it over on the new PR in [a33bfbe](https://github.com/ably/docs/pull/1450/commits/a33bfbe04c7b196e1af80824752e352c83294f74) and [384b4fe](https://github.com/ably/docs/pull/1450/commits/384b4fe48a169b3f7b40b839ee25611035a7965b)

content/client-lib-development-guide/features.textile
owenpearson: I've renamed it to `BatchPublishRequest` both in the spec and in the docs in [a2eee7d](https://github.com/ably/docs/pull/1450/commits/a2eee7de39d8b63bf0a706919e75f90dfdfd9381)

PR #1426, opened 2022-05-16
Token revocation spec 

ikbalkaya:

## Description

Client library spec for token revocation

* [Revoke Tokens (JS, Ruby, PHP, Python, Java android, Cocoa)](https://ably.atlassian.net/browse/SDK-1809)

## Review

This feature spec is likely to change a bit after some reviews. I would like to highlight the areas that may need to be addressed better. 
* Exception messages in client side errors (invalid request etc) 
* Exception messages from REST: I saw there are some fixed error codes on other spec points but wasn't sure what the error codes would be for this spec.
* Did I use spec numbering correctly? I just used  arbitrary numbering that was similar to others and unique

* [Page to review](https://docs.ably.io/client-lib-development-guide/features)


content/client-lib-development-guide/features.textile
```diff
@@ -1242,6 +1245,20 @@ h4. TokenRequest
 * @(TE4)@ @ttl@ attribute represents time to live (expiry) of this token in milliseconds
 * @(TE6)@ @TokenRequest::fromJson@ static factory method that accepts either a @JsonObject@ or a string (which should be parsed as a JSON string), and returns a new @TokenRequest@. Statically typed languages (that expect the @authCallback@ to result in a typed @TokenRequest@ object, rather than, say, a hashmap) must implement this; others may at their discretion.
 
+h4. TokenRevocationRequest
+
+* @(TF1)@ @TokenRevocationRequest@ is a type containing the token revocation request details sent to the "REST revokeTokens endpoint":/rest-api/#revokeTokens
+* @(TF2)@ @TokenRevocationRequest#keys@ attribute is an array of strings containing the keys to revoke. For request to be valid, it must contain at least one key. A test must exist to ensure there is at least one key in the request before it is sent.
```
paddybyers, line 1251: I think we should generally avoid adding statements that say "A test must exist ....".

The REST endpoint is `POST /keys/appId.keyId/revokeTokens` which revokes tokens associated with a specific key, whereas this request permits multiple keys. This introduces complexity because there is the possibility that the caller has rights to revoke tokens for one key but not others. I suggest we restrict this API to mirror the REST API, and only support token revocation for a single key.

content/client-lib-development-guide/features.textile
```diff
@@ -1242,6 +1245,20 @@ h4. TokenRequest
...
+* @(TF3)@ @TokenRevocationRequest#targets@ is an array of @TokenRevocationTarget@s to revoke. For request to be valid, it must contain at least one target. A test must exist to ensure there is at least one target in the request before it is sent.
+* @(TF4)@ @TokenRevocationRequest#issuedBefore@ is an optional timestamp (in milliseconds since the epoch) to revoke tokens issued before this time.
+* @(TF5)@ @TokenRevocationRequest#allowReauthMargin@ is an optional boolean that permits a token renewal cycle to take place without needing established connections to be dropped.Default for this value is false if not provided.
```
paddybyers, line 1254: Missing space after period

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,9 @@ h3(#rest-auth). Auth
 ** @(RSA16b)@ If the library is provided with a token without the corresponding @TokenDetails@, then this holds a @TokenDetails@ instance in which only the @token@ attribute is populated with that token string
 ** @(RSA16c)@ Is set with the current token (if applicable) on instantiation and each time it is replaced, whether the result of an explicit @Auth#authorize@ operation, or a library-initiated renewal resulting from expiry or a token error response
 ** @(RSA16d)@ Is empty if there is no current token, including after a previous token has been determined to be invalid or expired, or if the library is using basic auth
+* @(RSA17)@ @Auth#revokeToken@ function:
```
paddybyers, line 235: Why not `revokeTokens` to reflect (i) what it does, and (ii) the REST endpoint name ?

content/client-lib-development-guide/features.textile
```diff
@@ -1348,7 +1365,7 @@ h4. ChannelDetails
 
 h4. ChannelStatus
 
-* @(CHS1)@ @ChannelStatus@ is a type that contains status and occupancy for a channel 
+* @(CHS1)@ @ChannelStatus@ is a type that contains status and occupancy for a channel
```
paddybyers, line 1368: These changes are whitespace removal that presumably has been done automatically by the editor and they're not intended to be part of this change. They should usually be in a separate commit.

paddybyers (changes requested):
Comments so far

content/client-lib-development-guide/features.textile
ikbalkaya: Thank you. I think reading spec and following and example I mistook the . with , and somehow  thought we support multiple keys. I made some changes here 1c220a6fc382fa4744baec3a9e5796e16013b431

content/client-lib-development-guide/features.textile
ikbalkaya: Changed method name ea9c6dce7fa0f04a08bf3494a73236c9813fd2f1 and added explanation and REST endpoint name 3819be5d3eeca7380d45921323ae953837fc0bcc

content/client-lib-development-guide/features.textile
ikbalkaya: 0bbf2aaa4ba3df5052019e95ca0d784b7d6b42e0

content/client-lib-development-guide/features.textile
ikbalkaya: 0ab5af22c5284c4e17e6324413a2207af54a393a

content/client-lib-development-guide/features.textile
```diff
@@ -1242,6 +1258,31 @@ h4. TokenRequest
...
+* @(TX1)@ @TokenRevocationResponse@ is a type representing a target for "REST revokeTokens endpoint":/rest-api/#revokeTokens.
+* @(TX2)@ @TokenRevocationResponse#revokedTokens@ is a collection of @RevokedToken@s representing the result of the token revocation request.
+
+h4. RevokedToken
+* @(TY1)@ @RevokedToken@ is an object representing a single token that is revoked. 
+* @(TY2)@ @RevokedToken#target@ is a @TokenRevocationTarget@ representing the target of revoked token
+* @(TY3)@ @RevokedToken#issuedBefore@ is a timestamp (in milliseconds since the epoch) representing the time the the revoked token issued before.
+* @(TY4)@ @RevokedToken#appliesAt@ is a timestamp (in milliseconds since the epoch) representing the time the revocation takes effect for this token.
+
+
```
owenpearson, line 1285:
```suggestion

```

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,22 @@ h3(#rest-auth). Auth
...
+* @(RSA17)@ @Auth#revokeTokens@ function: This function is intended to be used to revoke tokens using provided @TokenRevocationRequest@ details. It internally creates request for  "REST revokeTokens endpoint":/rest-api/#revokeTokens and handles the response.
+** @(RSA17a)@ Method signature is @revokeTokens(TokenRevocationRequest)@.
+** @(RSA17b)@ If the @TokenRevocationRequest#key@ attribute is not provided, then the function must throw an error with a @"Key is not provided"@ message.
+** @(RSA17c)@ If the @TokenRevocationRequest#targets@ attribute is null or invalid (as per @TF3@ ), then the function must throw an error with a @"targets are invalid"@ message.
+** @(RSA17d)@ If the @TokenRevocationRequest#key@ is provided and @TokenRevocationRequest#targets@ valid, then the function must create the request and execute it.
+** @(RSA17e)@ Function returns a @TokenRevocationResponse@ object that contains the result of the request if request is successful.
+** @(RSA17f)@ Function throws an appropriate error/exception  if the request fails either for the reasons of invalid request, or as  a result of request failing.
```
mattheworiordan, line 241: Tiny point but you have have an additional space before `if the request`

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,22 @@ h3(#rest-auth). Auth
 ** @(RSA16b)@ If the library is provided with a token without the corresponding @TokenDetails@, then this holds a @TokenDetails@ instance in which only the @token@ attribute is populated with that token string
 ** @(RSA16c)@ Is set with the current token (if applicable) on instantiation and each time it is replaced, whether the result of an explicit @Auth#authorize@ operation, or a library-initiated renewal resulting from expiry or a token error response
 ** @(RSA16d)@ Is empty if there is no current token, including after a previous token has been determined to be invalid or expired, or if the library is using basic auth
+* @(RSA17)@ @Auth#revokeTokens@ function: This function is intended to be used to revoke tokens using provided @TokenRevocationRequest@ details. It internally creates request for  "REST revokeTokens endpoint":/rest-api/#revokeTokens and handles the response.
```
mattheworiordan, line 235: Am I doing something wrong, I cannot see the REST API documentation for this, in either https://ably.com/docs/rest-api#revoke-tokens or https://docs.ably.io/rest-api/#revoke-tokens.  Where is that documentation?  Should it not exist in the REST API documentation?

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,22 @@ h3(#rest-auth). Auth
...
+** @(RSA17a)@ Method signature is @revokeTokens(TokenRevocationRequest)@.
+** @(RSA17b)@ If the @TokenRevocationRequest#key@ attribute is not provided, then the function must throw an error with a @"Key is not provided"@ message.
+** @(RSA17c)@ If the @TokenRevocationRequest#targets@ attribute is null or invalid (as per @TF3@ ), then the function must throw an error with a @"targets are invalid"@ message.
```
mattheworiordan, line 238: Extra space after `@TF3@`

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,22 @@ h3(#rest-auth). Auth
...
+** @(RSA17d)@ If the @TokenRevocationRequest#key@ is provided and @TokenRevocationRequest#targets@ valid, then the function must create the request and execute it.
+** @(RSA17e)@ Function returns a @TokenRevocationResponse@ object that contains the result of the request if request is successful.
+** @(RSA17f)@ Function throws an appropriate error/exception  if the request fails either for the reasons of invalid request, or as  a result of request failing.
+** @(RSA17g)@ Tests in this section must exist in client library.
+*** @(RSA17g1)@ Given a null @TokenRevocationRequest#key@ attribute, verify that the function throws an error with a @"Key is not provided"@ message.
+*** @(RSA17g2)@ Given a null @TokenRevocationRequest#targets@ attribute, verify that the function throws an error with a @"targets are invalid"@ message.
```
mattheworiordan, line 244: Could you not simplify this to `Given a null or empty ...` and remove the next spec item?

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,22 @@ h3(#rest-auth). Auth
...
+*** @(RSA17g3)@ Given an empty @TokenRevocationRequest#targets@ attribute, verify that the function throws an error with a @"targets are invalid"@ message.
+*** @(RSA17g4)@ Given a non-null @TokenRevocationRequest#key@ attribute and a valid @TokenRevocationRequest#targets@ attribute, verify that the function creates the request and executes it.
+*** @(RSA17g5)@ Given a non-null but invalid @TokenRevocationRequest#key@ attribute and a valid @TokenRevocationRequest#targets@ attribute, verify that client library throws an API error.
+*** @(RSA17g6)@ Given a valid @TokenRevocationRequest#key@ attribute and a valid @TokenRevocationRequest#targets@ attribute, verify that the function executes the request and returns a @TokenRevocationResponse@ with targets matching with the original targets in request.
+*** @(RSA17g7)@ Given valid required attributes in @TokenRevocationRequest@ and a timestamp in the future for @TokenRevocationRequest#issuedBefore@ attribute, verify that the request is rejected.
```
mattheworiordan, line 249: This is not clear, does this rely on the local clock only or rely on the server clock only?  I would advocate for simplicity and rely on the server clock as no clock sync is needed then.

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,22 @@ h3(#rest-auth). Auth
...
+*** @(RSA17g4)@ Given a non-null @TokenRevocationRequest#key@ attribute and a valid @TokenRevocationRequest#targets@ attribute, verify that the function creates the request and executes it.
+*** @(RSA17g5)@ Given a non-null but invalid @TokenRevocationRequest#key@ attribute and a valid @TokenRevocationRequest#targets@ attribute, verify that client library throws an API error.
+*** @(RSA17g6)@ Given a valid @TokenRevocationRequest#key@ attribute and a valid @TokenRevocationRequest#targets@ attribute, verify that the function executes the request and returns a @TokenRevocationResponse@ with targets matching with the original targets in request.
+*** @(RSA17g7)@ Given valid required attributes in @TokenRevocationRequest@ and a timestamp in the future for @TokenRevocationRequest#issuedBefore@ attribute, verify that the request is rejected. 
+*** @(RSA17g8)@ Given valid required attributes in @TokenRevocationRequest@ and a timestamp in the past > 1 hour for @TokenRevocationRequest#issuedBefore@ attribute, verify that the request is rejected.
```
mattheworiordan, line 250: As above.

content/client-lib-development-guide/features.textile
```diff
@@ -1242,6 +1258,31 @@ h4. TokenRequest
...
 
+h4. TokenRevocationRequest
+
+* @(TF1)@ @TokenRevocationRequest@ is a type containing the token revocation request details sent to the "REST revokeTokens endpoint":/rest-api/#revokeTokens
+* @(TF2)@ @TokenRevocationRequest#key@ attribute is a string representing key to revoke (appId.Key).
+* @(TF3)@ @TokenRevocationRequest#targets@ is an array of @TokenRevocationTarget@s to revoke. For request to be valid, it must contain at least one target. A test must exist to ensure there is at least one target in the request before it is sent.
+* @(TF4)@ @TokenRevocationRequest#issuedBefore@ is an optional timestamp (in milliseconds since the epoch) to revoke tokens issued before this time.
+* @(TF5)@ @TokenRevocationRequest#allowReauthMargin@ is an optional boolean that permits a token renewal cycle to take place without needing established connections to be dropped. Default for this value is false if not provided.
```
mattheworiordan, line 1267: @paddybyers @ikbalkaya what is the thinking behind defaulting to disconnecting clients? Wouldn't it be more seamless by default to not disconnect clients, and instead let a customer escalate the urgency?

content/client-lib-development-guide/features.textile
```diff
@@ -1242,6 +1258,31 @@ h4. TokenRequest
...
+
+h4. TokenRevocationTarget
+
+* @(TL1)@ @TokenRevocationTarget@ is a type representing a target for "REST revokeTokens endpoint":/rest-api/#revokeTokens.
+* @(TL2)@ @TokenRevocationTarget#key@ is a string containing the target key. Target key can be "clientId","revocationKey" or "channel". Where possible, clients should predefine those strings (enums for example) to prevent customers passing an invalid value.
```
mattheworiordan, line 1272: Should we be `enum`ing this?  This would mean if we want to support new targets, we have to roll out updates to all clients.  Would it not be better in fact to keep this as strings and instead provide advice in the inline docs for developers to reduce overhead in maintaining this?  I think overall we're erring towards client specs that don't need to change when server-side functionality expands.

content/client-lib-development-guide/features.textile
```diff
@@ -1242,6 +1258,31 @@ h4. TokenRequest
...
+* @(TL3)@ @TokenRevocationTarget#value is a string representing the value of specified for target key. This value cannot be empty or null.
```
mattheworiordan, line 1273: I don't understand this statement "is a string representing the value of specified for target key". Can you reword please?

content/client-lib-development-guide/features.textile
```diff
@@ -1242,6 +1258,31 @@ h4. TokenRequest
...
+
+h4. TokenRevocationTarget
+
+* @(TL1)@ @TokenRevocationTarget@ is a type representing a target for "REST revokeTokens endpoint":/rest-api/#revokeTokens.
+* @(TL2)@ @TokenRevocationTarget#key@ is a string containing the target key. Target key can be "clientId","revocationKey" or "channel". Where possible, clients should predefine those strings (enums for example) to prevent customers passing an invalid value.
+* @(TL3)@ @TokenRevocationTarget#value is a string representing the value of specified for target key. This value cannot be empty or null. 
+
+h4. TokenRevocationResponse
+* @(TX1)@ @TokenRevocationResponse@ is a type representing a target for "REST revokeTokens endpoint":/rest-api/#revokeTokens.
+* @(TX2)@ @TokenRevocationResponse#revokedTokens@ is a collection of @RevokedToken@s representing the result of the token revocation request.
```
mattheworiordan, line 1277: @paddybyers OOI, what happens if there are thousands+ of revoked tokens?

content/client-lib-development-guide/features.textile
mattheworiordan: In addition, if we later changed the response for `RevokedToken` below, and we had a new target type, we'd break the SDK as it would end up with a null value / error trying to create the `RevokedToken` object.

mattheworiordan (commented):
I hope my comments are useful.  Shout if you want to discuss.

content/client-lib-development-guide/features.textile
ikbalkaya: b66744b0566cb12d25d401781b169d06b8195d4b

content/client-lib-development-guide/features.textile
ikbalkaya: 1ceadf2c30128fefecfd387c89da7dd94e3724b6

content/client-lib-development-guide/features.textile
ikbalkaya: My intention was to separate the null and empty situations however it looks like the spec does not separate them. I am keen on adding a new spec item for throwing a different error when target is null. WDYT?

content/client-lib-development-guide/features.textile
ikbalkaya: I think enum'ing will add some clarity and safety but as you mentioned it will also make it incompatible with any new changes made to API. I wasn't particularly sure what's the guidance around how we interpret such API objects and suggested enuming based on supported attributes here
https://ably.com/docs/core-features/authentication#target-specifiers
Now you pointed out I will update this so it's more permissive

content/client-lib-development-guide/features.textile
ikbalkaya: 46bef24067f78d6812f1a98d93abe4126ccdf33f

content/client-lib-development-guide/features.textile
ikbalkaya: 1be9698fd4c603d5c303ec97f4b9ba1390f9612c

content/client-lib-development-guide/features.textile
ikbalkaya: @mattheworiordan I am not very sure about this 
There is a longer description here 
https://ably.com/docs/core-features/authentication#token-revocation

> Optional boolean. The allowReauthMargin boolean permits a token renewal cycle to take place without needing established connections to be dropped, by postponing enforcement to 30 seconds in the future, and sending any existing connections a hint to obtain (and upgrade the connection to use) a new token. It defaults to false, meaning that the effect is near-immediate.

I added a cut version of description from here

content/client-lib-development-guide/features.textile
ikbalkaya: Removed the wording on enuming key 6868b0b3971edbc99fd4262749e702caafbc77f6

content/client-lib-development-guide/features.textile
ikbalkaya: Added a spec item and split two cases 1ae5367e5358f2e7845785dfed9e0cf9a81e71f9

mattheworiordan: Well if you feel strongly it's needed.  IMHO, it creates "more spec", which I believe we are all not keen on. So whatever we can do to keep the spec as terse as possible, and have single spec items that cover a wider array of problems without being overly explicit, the better.

content/client-lib-development-guide/features.textile
mattheworiordan: No, what I am saying is that the client does not need to enforce this, this creates extra code in the SDK, a dependency on time sync, and what gain do we get from this?  The client will get an error code anyway.  We should be explicit

content/client-lib-development-guide/features.textile
ikbalkaya: My initial motivation was to be as clear as possible to clients about any input they provide. But I get your point. I have made change bac65329d67ecbd6896a1e0b34dc2ccb0285b55b with also clarifying the message so at least we are more descriptive in message side

content/client-lib-development-guide/features.textile
ikbalkaya: Because `issuedBefore` is a timestamp that can (optionally) be provided by client, I thought it would be a good way to verify client SDK handles passing this correctly to the REST API. It wasn't to verify  the API behaviour.

ikbalkaya: Removed test cases relating to `issuedBefore` 4207e8ec4d0d7483e64df76ec287741ba43ec719

content/client-lib-development-guide/features.textile
ikbalkaya: 4207e8ec4d0d7483e64df76ec287741ba43ec719

content/client-lib-development-guide/features.textile
```diff
@@ -1242,6 +1255,30 @@ h4. TokenRequest
 * @(TE4)@ @ttl@ attribute represents time to live (expiry) of this token in milliseconds
 * @(TE6)@ @TokenRequest::fromJson@ static factory method that accepts either a @JsonObject@ or a string (which should be parsed as a JSON string), and returns a new @TokenRequest@. Statically typed languages (that expect the @authCallback@ to result in a typed @TokenRequest@ object, rather than, say, a hashmap) must implement this; others may at their discretion.
 
+h4. TokenRevocationRequest
+
+* @(TF1)@ @TokenRevocationRequest@ is a type containing the token revocation request details sent to the "REST revokeTokens endpoint":/rest-api/#revokeTokens
+* @(TF2)@ @TokenRevocationRequest#key@ attribute is a string representing key to revoke (appId.Key).
```
SimonWoolf, line 1261: Why require the key to be passed in here? This is a method on an instantiated client lib instance, which will already have some auth params; if it has access to an api key it would presumably already have been instantiated with it?

(In the unusual event that someone needs to revoke things from multiple keys, they can always instantiate a new rest client lib for each one)

content/client-lib-development-guide/features.textile
```diff
@@ -1242,6 +1255,30 @@ h4. TokenRequest
...
+* @(TF3)@ @TokenRevocationRequest#targets@ is an array of @TokenRevocationTarget@s to revoke. For request to be valid, it must contain at least one target. A test must exist to ensure there is at least one target in the request before it is sent.
+* @(TF4)@ @TokenRevocationRequest#issuedBefore@ is an optional timestamp (in milliseconds since the epoch) to revoke tokens issued before this time.
+* @(TF5)@ @TokenRevocationRequest#allowReauthMargin@ is an optional boolean that permits a token renewal cycle to take place without needing established connections to be dropped. Default for this value is false if not provided.
+
+h4. TokenRevocationTarget
+
+* @(TL1)@ @TokenRevocationTarget@ is a type representing a target for "REST revokeTokens endpoint":/rest-api/#revokeTokens.
+* @(TL2)@ @TokenRevocationTarget#key@ is a string that is the target key.This value cannot be null.
```
SimonWoolf, line 1269: The key again, in the target specifier? Why?

content/client-lib-development-guide/features.textile
```diff
@@ -1242,6 +1255,30 @@ h4. TokenRequest
...
+* @(TL3)@ @TokenRevocationTarget#value is a string that is the target value. This value cannot be empty or null.
```
SimonWoolf, line 1270: Why bother having a struct that just contains a string? Either have a rich target with an enumerated types like `clientId` and `revocationKey`, or just make `targets` an array of strings, but this seems like disadvantages of both worlds

content/client-lib-development-guide/features.textile
SimonWoolf: The endpoint doesn't return a list of revoked tokens. It returns a list of revoked _targets_. See https://ably.com/docs/api/rest-api#revoke-tokens. Problem here is just the naming of `RevokedToken`

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,19 @@ h3(#rest-auth). Auth
...
+** @(RSA17g)@ Tests in this section must exist in client library.
+*** @(RSA17g1)@ Given a null @TokenRevocationRequest#key@ attribute, verify that the function throws an error with a @"Key is not provided"@ message.
+*** @(RSA17g2)@ Given a null or empty @TokenRevocationRequest#targets@ attribute, verify that the function throws an error with a @"targets cannot be null or empty"@ message.
+*** @(RSA17g3)@ Given a non-null @TokenRevocationRequest#key@ attribute and a valid @TokenRevocationRequest#targets@ attribute, verify that the function creates the request and executes it.
+*** @(RSA17g4)@ Given a non-null but invalid @TokenRevocationRequest#key@ attribute and a valid @TokenRevocationRequest#targets@ attribute, verify that client library throws an API error.
+*** @(RSA17g5)@ Given a valid @TokenRevocationRequest#key@ attribute and a valid @TokenRevocationRequest#targets@ attribute, verify that the function executes the request and returns a @TokenRevocationResponse@ with targets matching with the original targets in request.
```
SimonWoolf, line 247: Don't think client libs can have this test now, they're just creating simulation apps which won't have revocation enabled

(we might be able to have them specify in the appspec that we want revocable keys etc, but not sure that'll work as-is right now)

content/client-lib-development-guide/features.textile
ikbalkaya: This was renamed to `RevokedTarget` 2bcd1cdcc69c3cd5427268e291ca32a488f15bd7

content/client-lib-development-guide/features.textile
ikbalkaya: @SimonWoolf The wording wasn't very descriptive and it wasn't necessary after all. I got appId.Key from here https://ably.com/docs/api/rest-api#revoke-tokens 
I suppose it is just the key associated with the current Rest client. \ I removed all key and references here 029fe8ab29162690d758649f60a587a4b4e62bc7

content/client-lib-development-guide/features.textile
ikbalkaya: This is the key as in target key like for example key for "clientId:client1@example.com"  is "clientId"

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,19 @@ h3(#rest-auth). Auth
 ** @(RSA16b)@ If the library is provided with a token without the corresponding @TokenDetails@, then this holds a @TokenDetails@ instance in which only the @token@ attribute is populated with that token string
 ** @(RSA16c)@ Is set with the current token (if applicable) on instantiation and each time it is replaced, whether the result of an explicit @Auth#authorize@ operation, or a library-initiated renewal resulting from expiry or a token error response
 ** @(RSA16d)@ Is empty if there is no current token, including after a previous token has been determined to be invalid or expired, or if the library is using basic auth
+* @(RSA17)@ @Auth#revokeTokens@ function: This function is intended to be used to revoke tokens using provided @TokenRevocationRequest@ details. It internally creates request for  "REST revokeTokens endpoint":/rest-api/#revokeTokens and handles the response.
+** @(RSA17a)@ Method signature is @revokeTokens(TokenRevocationRequest)@.
```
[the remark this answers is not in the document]
ikbalkaya: I thought that while two of parameters are optional, it could be a waste of parameter space to move these as parameters of the function. Also change on parameters in the future would cause the function interface to change. 
However I think this is a subjective matter and an approach will suit better in a language than other. I see the benefit of using function params, that the passed object will be un-encapsulated and more reachable.  I have used my OOP instincts when writing this. Can I get some opinions about this please @QuintinWillison ?

content/client-lib-development-guide/features.textile
ikbalkaya: Removed all tests here 6dafadef8882fc493a11ee695cb4c627ca8b1de5

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,19 @@ h3(#rest-auth). Auth
...
+** @(RSA17b)@ If the @TokenRevocationRequest#key@ attribute is not provided, then the function must throw an error with a @"Key is not provided"@ message.
+** @(RSA17c)@ If the @TokenRevocationRequest#targets@ attribute is null or empty (as per @TF3@), then the function must throw an error with a @"targets cannot be null or empty"@ message.
+** @(RSA17d)@ If the @TokenRevocationRequest#key@ is provided and @TokenRevocationRequest#targets@ valid, then the function must create the request and execute it.
+** @(RSA17e)@ Function returns a @TokenRevocationResponse@ object that contains the result of the request if request is successful.
+** @(RSA17f)@ Function throws an appropriate error/exception if the request fails either for the reasons of invalid request, or as  a result of request failing.
+** @(RSA17g)@ Tests in this section must exist in client library.
```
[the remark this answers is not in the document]
ikbalkaya: I agree with this too and I think the spec itself should be enough for writing comprehensive tests. I was just blindly following the old way of doing things 🤦  Removed all test points 6dafadef8882fc493a11ee695cb4c627ca8b1de5

content/client-lib-development-guide/features.textile
ikbalkaya: I added some explanation on usage c4067a078a685bef9837339a736503c9cc5f7ff1. This is hopefully more clear now

content/client-lib-development-guide/features.textile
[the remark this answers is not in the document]
mattheworiordan: I agree with @SimonWoolf, this makes it harder for the user, they have to create an object, propagage the object, then pass that object to the method, instead of just calling a method.  It's unnecessary indirection.

content/client-lib-development-guide/features.textile
mattheworiordan: > they're just creating simulation apps which won't have revocation enabled

Why not?  This seems insane that we are creating simulation apps that don't have a feature we need for end-to-end tests in our SDKs?  @SimonWoolf can this not be trivially supported?

In addition, what is the plan for token revocation more broadly?  Is our plan not to make this feature self-servicable?  Why are we holding off offering this to customers?

mattheworiordan:
I see this feature is currently in beta (https://ably.com/docs/core-features/authentication#development-status) and we state "Token revocation has to be enabled for you by Ably. There is a performance cost to enabling this feature. [Contact Ably](https://ably.com/contact) if you’d like to enable this feature for a specific application."

I understand we may do that if we want to gate access to a feature during testing, but what is our thinking behind keeping this behind a contact form, and not making it a feature all customers have access to?  I appreciate this is not an @ikbalkaya question, but it's very relevant to why we're adding this feature to SDKs, so I think needs answering by the product team.

content/client-lib-development-guide/features.textile
SimonWoolf: > Why not?

No reason not to, was just an oversight on my part -- done now https://github.com/ably/realtime/pull/4143

> In addition, what is the plan for token revocation more broadly? Is our plan not to make this feature self-servicable? Why are we holding off offering this to customers?

My understanding was it's in beta until we have a few customers using it, who can give feedback to us about how it works in practice, whether the API needs any changes, etc., at which point we'll make any final API tweaks, then enable it (defaulting to 1 revocation shard) for everyone.

I can't find the conversation now, but IIRC I was chatting to @ben-pickering about the status of this a few months ago and he said that there had been slower than expected progress getting customers to try the beta, so this process was taking a bit of time.

content/client-lib-development-guide/features.textile
ikbalkaya: https://github.com/ably/docs/issues/1446

content/client-lib-development-guide/features.textile
paddybyers: > defaulting to 1 revocation shard
Would we consider just making the shard count a function of the number of connections in the package? (I was going to suggest the number of `tokenRequests` or `apiRequests`, but with JWT neither of those correlate with actual token usage. We would set the shards as a simple multiple of the peak connections, and only ever increase it from that point forward

SimonWoolf: > Would we consider just making the shard count a function of the number of connections in the package?

Aren't we about to move to a pricing model where we're not using packages/quotas for self-service any more? So we wouldn't have that information.

I was assuming that for self-service we were just going to have it be a fixed limit (1 shard = 50 revocations per second), and for business/enterprise who can request a higher limit, we'd just divide whatever limit they've requested/negotiated by 50 and set that as the number of shards

paddybyers: ok, makes sense, thanks

content/client-lib-development-guide/features.textile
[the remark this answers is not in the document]
ikbalkaya: I changed function signature and removed the request object here. 3cc1dab873bfbf3ddf9f3209235fdc0720bafda5

content/client-lib-development-guide/features.textile
ikbalkaya: @mattheworiordan Sorry I didn't know that was supposed to turn into a link. I made it so it goes to the full path 3127a70c431000bbfd6e052768f048011949fa91

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,17 @@ h3(#rest-auth). Auth
...
+** @(RSA17b)@ Method signature is @revokeTokens(Array<TokenRevocationTarget>. issuedBefore, allowReauthMargin)@.
+*** @(RSA17b1)@ @Array<TokenRevocationTarget>@ is an array of @TokenRevocationTarget@s to revoke. This array of token revocation targets must be used to construct the individual @TokenRevocationTarget@s in the form of "@TokenRevocationTarget#key@:@TokenRevocationTarget#value@" for the "targets" array in the request body.
+*** @(RSA17b2)@ @issuedBefore@ is an optional timestamp (in milliseconds since the epoch) to revoke tokens issued before this time.
+*** @(RSA17b3)@ @allowReauthMargin@ is an optional boolean that permits a token renewal cycle to take place without needing established connections to be dropped. Default for this value is false if not provided.
+** @(RSA17c)@ key to build the request must be retrieved from stored @Auth#AuthOptions#key@
+** @(RSA17d)@ If the @targets@ parameter is null or empty, then the function must throw an error with a @"targets cannot be null or empty"@ message.
```
paddybyers, line 242: "throw" is too language-specific. Elsewhere more portable language is used, eg "respond with an error" or "indicate an error"

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,17 @@ h3(#rest-auth). Auth
...
+** @(RSA17e)@ If the @targets@ parameter is valid, then the function must create the request and execute it.
+** @(RSA17f)@ Function returns a @TokenRevocationResponse@ object that contains the result of the request if request is successful.
```
paddybyers, line 244: "if request is successful" -> "if the request is successful"

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,17 @@ h3(#rest-auth). Auth
...
+** @(RSA17g)@ Function throws an appropriate error/exception if the request fails either for the reasons of invalid request, or as  a result of request failing.
```
paddybyers, line 245: "throws" again

content/client-lib-development-guide/features.textile
paddybyers: I'm really not keen on overloading `key` in this way. I think it's confusing. Can we not find a different word? Isn't it an array of specifiers `<type>:<target>`, so we could use `type` instead of `key` ?

paddybyers (changes requested):
Mostly editorial

content/client-lib-development-guide/features.textile
ikbalkaya: I agree `type` makes more sense in this context. The change is here ff5c856d4b60859cc07d42845c5ba31faad17f82

content/client-lib-development-guide/features.textile
```diff
@@ -232,6 +232,17 @@ h3(#rest-auth). Auth
...
+* @(RSA17)@ @Auth#revokeTokens@ function
+** @(RSA17a)@ Revokes tokens based on provided function arguments. It internally creates request for  "REST revokeTokens endpoint":https://ably.com/docs/api/rest-api#revoke-tokens and handles the response.
+** @(RSA17b)@ Method signature is @revokeTokens(Array<TokenRevocationTarget>. issuedBefore, allowReauthMargin)@.
+*** @(RSA17b1)@ @Array<TokenRevocationTarget>@ is an array of @TokenRevocationTarget@s to revoke. This array of token revocation targets must be used to construct the individual @TokenRevocationTarget@s in the form of "@TokenRevocationTarget#key@:@TokenRevocationTarget#value@" for the "targets" array in the request body.
+*** @(RSA17b2)@ @issuedBefore@ is an optional timestamp (in milliseconds since the epoch) to revoke tokens issued before this time.
+*** @(RSA17b3)@ @allowReauthMargin@ is an optional boolean that permits a token renewal cycle to take place without needing established connections to be dropped. Default for this value is false if not provided.
```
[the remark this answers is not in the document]
ikbalkaya: f9420343b5b5f7ea07c0254459b6c98886d7de19

content/client-lib-development-guide/features.textile
ikbalkaya: f9420343b5b5f7ea07c0254459b6c98886d7de19

content/client-lib-development-guide/features.textile
ikbalkaya: 2924615143d316accbb5647a5de7c49a4fb5868b

content/client-lib-development-guide/features.textile
ikbalkaya: 2924615143d316accbb5647a5de7c49a4fb5868b

ikbalkaya (commented):
I addressed @paddybyers 's editorial comments

QuintinWillison:
From @ikbalkaya [via Slack (internal)](https://ably-real-time.slack.com/archives/C01LR5XFHFX/p1663320356849839?thread_ts=1662389340.590889&cid=C01LR5XFHFX):

> This work was paused as spec for batch operations would change. It's current state is based on current REST api

So I am going to move this pull request back to the 'Draft' state and label it as `on hold`. It shall remain open in this repository (`ably/docs`) for visibility only, until the features spec has moved over to `ably/specification` (see https://github.com/ably/specification/issues/1). At that point, it will be re-formed in that repository, most likely targeting a v2 integration branch there.

QuintinWillison:
This changes proposed in this pull request has been [moved](https://github.com/ably/docs/pull/1426#ref-issue-1393772361) to the new home of the features spec, in [the `ably/specification` repository](https://github.com/ably/specification).

content/client-lib-development-guide/features.textile
QuintinWillison: @paddybyers If this conversation needs to continue, then that will need to happen against the new pull request that we will open against https://github.com/ably/specification/issues/10. Please resolve this conversation _here_ once you have re-raised your concern _there_, or if you are already happy that this matter is resolved. Thanks.

content/client-lib-development-guide/features.textile
QuintinWillison: @paddybyers If this conversation needs to continue, then that will need to happen against the new pull request that we will open against https://github.com/ably/specification/issues/10. Please resolve this conversation _here_ once you have re-raised your concern _there_, or if you are already happy that this matter is resolved. Thanks.

PR #1427, opened 2022-05-16
Make all TokenDetail fields apart from token optional

Morganamilo:

## Description

RSA16b says:

If the library is provided with a token without the corresponding
TokenDetails, then this holds a TokenDetails instance in which only
the token attribute is populated with that token string

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -1665,10 +1665,10 @@ class Auth:
 
 class TokenDetails:
   +fromJson(String | JsonObject) -> TokenDetails// TD7
-  capability: String // TD5
+  capability: String? // TD5
   clientId: String? // TD6
-  expires: Time // TD3
-  issued: Time // TD4
+  expires: Time? // TD3
+  issued: Time? // TD4
   token: String // TD2
 
 class TokenRequest:


Morganamilo:
Similarly, `Auth#authorize` and `Auth#requestToken` both return `TokenDetails` yet AuthCallback and AuthUrl permit returning just a token leading me to assume they must also return a TokenDetails with just token set.

lmars:
@Morganamilo having chatted with Paddy, I think if we are to make a change here, it should probably be to stop wrapping token strings in a `TokenDetails` object, have a return type that better communicates the fact that a token may either by a string literal or a `TokenDetails` object, and continue to expect these `TokenDetails` fields to be non-null.

In the Rust SDK, we have the [`Token` enum](https://github.com/ably/ably-rust/blob/main/src/auth.rs#L621-L628) which is returned form an auth callback, and that can either be a token request, token details, or token literal (inline with spec point [RSA8d](https://docs.ably.io/client-lib-development-guide/features/#RSA8d)), so perhaps that (or something similar) needs to make its way into the IDL?

QuintinWillison:
This work will be revisited once the features spec has been moved to its new home under https://github.com/ably/specification/issues/1.

PR #1444, opened 2022-05-30
Message Interactions

Peter-Maguire:

## Description

Adds message interaction object into message extras, and a method for subscribing to channel messages with a filter.

* [Jira epic ticket](https://ably.atlassian.net/browse/DEP-74).
* [Jira ticket](https://ably.atlassian.net/browse/DEP-85).
* [Ably JS Implementation](https://github.com/ably/ably-js/pull/1003).

## Review

The context behind this PR can be obtained from [this document](https://docs.google.com/document/d/1gptZ5meelzeWhXLB5EvDsUMRb1Af7EYR1XPYRUbKO-0)

* [Page to review](https://docs.ably.io/client-lib-development-guide/features)


content/client-lib-development-guide/features.textile
```diff
@@ -599,6 +599,16 @@ h3(#realtime-channel). Channel
...
 ** @(RTL7e)@ If a message cannot be decoded or decrypted successfully, it should be delivered to the listener with the @encoding@ attribute set indicating the residual encoding state, and an error should be logged
 ** @(RTL7f)@ A test should exist ensuring published messages are not echoed back to the subscriber when @echoMessages@ is set to false in the @Realtime@ library constructor
+** @(RTL7g)@ Subscribe with a dictionary argument subscribes a listener which filters which messages to accept based on the dictionary values
+*** @(RTL7g1)@ An empty dictionary functions identically to "RTL7a":#RTL7a
+*** @(RTL7g2)@ A dictionary with multiple key/value pairs should require each entry to be satisfied in order to accept the Message
+*** @(RTL7g3)@ Keys currently must be one of @ref@ or @type@ or a negated version of these (@-ref@ and @-type@)
+**** @(RTL7g3a)@ @ref@ and @type@ refer to @ref.timeserial@ and @ref.type@ in the Message body respectively
```
cruickshankpg, line 606: Had a chat with Paddy this morning, we're going to put the ref and timeserial fields in extras rather than the main message body.

This is because it allows SDKs that we haven't added official support in to still see and use these features.

content/client-lib-development-guide/features.textile
```diff
@@ -599,6 +599,16 @@ h3(#realtime-channel). Channel
...
+**** @(RTL7g3b)@ Consideration should be made to allow for easy expansion of the allowed fields here in the future 
+*** @(RTL7g4)@ Keys which are negated (starting with a @-@) should be treated as a NOT and not accept any Messages which match that value
+*** @(RTL7g5)@ A value should be matched exactly, unless it is @*@ which should match any value that is not null
```
cruickshankpg, line 609: I don't think this is needed, if you leave the field blank you get any value and I've not seen evidence for needing to support more complex regexes.

If we think that if more complex filters are needed then we should us a functional argument instead e.g. `filter(msg): bool`. The benefit I see in using exact strings is that you can put the listeners in a hashmap and have more efficient lookups on a per message basis. However if/when the dictionary approach is extended to support new fields it will be a bit of a nightmare of hierarchies.

If we assume that there will be <5 handlers per message in view in the chat and <20 messages in view at a time then a new inbound message will call < 400 filter functions. It's not amazing but it is definitely the most generally useful.

So I think we should go with a function rather than dict. WDYT?

cruickshankpg (commented):
Good start

content/client-lib-development-guide/features.textile
Peter-Maguire: >  if you leave the field blank you get any value and I've not seen evidence for needing to support more complex regexes.  
The asterisk's purpose is mostly for being able to handle only messages with no ref, e.g `"-ref": "*"`. I felt that was more extensible than the `includeRefs` we drafted. A function would be much more flexible and definitely easier to implement so I'm definitely happy to go that route instead. Would we still have the dictionary version for simple filtering or would be only use the function?

content/client-lib-development-guide/features.textile
```diff
@@ -599,6 +599,7 @@ h3(#realtime-channel). Channel
 ** @(RTL7d)@ Messages delivered are automatically decoded based on the @encoding@ attribute; see REST @Channel@ encoding features. Tests should exist to publish and subscribe to encoded messages using the "AES 128":https://github.com/ably/ably-common/blob/main/test-resources/crypto-data-128.json and "AES 256":https://github.com/ably/ably-common/blob/main/test-resources/crypto-data-256.json fixture test data
 ** @(RTL7e)@ If a message cannot be decoded or decrypted successfully, it should be delivered to the listener with the @encoding@ attribute set indicating the residual encoding state, and an error should be logged
 ** @(RTL7f)@ A test should exist ensuring published messages are not echoed back to the subscriber when @echoMessages@ is set to false in the @Realtime@ library constructor
+** @(RTL7g)@ Subscribe with a function argument subscribes a listener which only fires when the function returns @true@
```
lmars, line 602: I do think we may have talked our way out of this being particularly convenient.

My suggestion in [Proposal: Message References (May 2022)](https://docs.google.com/document/d/1gptZ5meelzeWhXLB5EvDsUMRb1Af7EYR1XPYRUbKO-0/edit#heading=h.i8z9qmamfguq) was to support specific functions for things like "subscribe to reactions to this specific message":

```
channel
  .subscriber()
  .refId("v1b25XrTDg:0")
  .refType("com.ably.reaction.v1")
  .forEach(msg => { … })
```

whereas with this function as described here, that'd be something like:

```
channel.subscribe(
  msg => msg.ref && msg.ref.id === "v1b25XrTDg:0" && msg.ref.type === "com.ably.reaction.v1",
  msg => { ... },
)
```

which makes me wonder if that's really much more convenient than:

```
channel.subscribe(msg => {
  if(!msg.ref || msg.ref.id !== "v1b25XrTDg:0" || msg.ref.type !== "com.ably.reaction.v1") {
    return;
  )
  ...
)
```

Or, do we imagine providing convenience functions which return filters? If so, could they be composed?

Peter-Maguire: My main issue with the builder style is that it's completely different from how any other part of the SDKs work (as far as I'm aware) and would feel out of place to me, unless we were looking to move subscriptions over to this style in the future and deprecate `subscribe()`. 

Perhaps a builder for filter functions similar to your original concept, except it returns a `(Message)=>bool` that can be inserted as a filter would be the best of both worlds? e.g `channel.subscribe(filter.withRef("v1b25XrTDg:0").withType("com.ably.reaction.v1"), (msg)=>{})`  
  
Aside from that, the lack of usefulness of a filter function makes me wonder what benefit we actually get from modifying the SDK here at all. Perhaps all the logic should be within a higher level ably-chat SDK?

content/client-lib-development-guide/features.textile
```diff
@@ -1730,11 +1731,14 @@ class RealtimeChannel:
   publish(name: String?, data: Data?) => io // RTL6i
   subscribe((Message) ->) => io // RTL7a
   subscribe(String, (Message) ->) => io // RTL7b
+  subscribe((Message) -> Bool, (Message) ->) // RTL7g
```
owenpearson, line 1734: This actually wouldn't work in ably-js because `RealtimeChannelCallbacks.subscribe` is async and optionally takes a callback which resolves when the channel is attached (see https://github.com/ably/ably-js/blob/main/ably.d.ts#L715) so this signature is already being used for something else.

Also would we also want a `subscribe(String, (Message) -> Bool, (Message) ->)` signature?

content/client-lib-development-guide/features.textile
```diff
@@ -1730,11 +1731,14 @@ class RealtimeChannel:
...
   unsubscribe() // RTL8a, RTE5
   unsubscribe((Message) ->) // RTL8a
   unsubscribe(String, (Message) ->) // RTL8a
   setOptions(options: ChannelOptions) => io // RTL16
 
+
+
 class BatchOperations:
```
owenpearson, line 1742:
```suggestion
class BatchOperations:
```

content/client-lib-development-guide/features.textile
```diff
@@ -2141,6 +2145,10 @@ class VCDiffDecoder
 class DeltaExtras
   from: String // the id of the message the delta was generated from
   format: String //the delta format. Only vcdiff is supported as at API version 1.2
+
+class InteractionExtras
```
owenpearson, line 2149: We should also update `TM2i` with something similar to the sentence about `DeltaExtras`

owenpearson (changes requested):

content/client-lib-development-guide/features.textile
```diff
@@ -2141,6 +2152,10 @@ class VCDiffDecoder
 class DeltaExtras
   from: String // the id of the message the delta was generated from
   format: String //the delta format. Only vcdiff is supported as at API version 1.2
+
+class InteractionExtras:
```
owenpearson, line 2156: I mentioned before - we should also update `TM2i` with something similar to the sentence about `DeltaExtras`

owenpearson (changes requested):

content/client-lib-development-guide/features.textile
```diff
@@ -630,6 +630,10 @@ h3(#realtime-channel). Channel
 ** @(RTL19c)@ In the case of a delta message with a @vcdiff@ @encoding@ step, the @vcdiff@ decoder must be used to decode the base payload of the of delta message, applying that delta to the stored base payload. The direct result of that vcdiff delta application, before performing any further decoding steps, is stored as the updated base payload.
 * @(RTL20)@ The @id@ of the last received message on each channel must be stored along with the base payload. When processing a delta message (i.e. one whose @encoding@ contains @vcdiff@ step) the stored last message @id@ must be compared against the delta reference @id@, indicated in the @Message.extras.delta.from@ field of the delta message. If the delta reference @id@ of the received delta message does not equal the stored @id@ corresponding to the base payload, the message decoding must fail. The recovery procedure from "RTL18":#RTL18 must be executed.
 * @(RTL21)@ The messages in the @messages@ array of a @ProtocolMessage@ should each be decoded in ascending order of their index in the array.
+* @(RTL22)@ A method must be provided for creating a listener which only fires when the message matches a set of criteria
```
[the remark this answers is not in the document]
Peter-Maguire: The intention is to allow SDKs to choose what pattern they go with for this, hence why it is generic and not specifically attached to the subscribe function

content/client-lib-development-guide/features.textile
Peter-Maguire: TM2i definitely needs to be reworked into something more readable but for now I've added it.

owenpearson: Yeah agreed, but obvs beyond the scope of this work

content/client-lib-development-guide/features.textile
```diff
@@ -630,6 +630,10 @@ h3(#realtime-channel). Channel
...
 * @(RTL20)@ The @id@ of the last received message on each channel must be stored along with the base payload. When processing a delta message (i.e. one whose @encoding@ contains @vcdiff@ step) the stored last message @id@ must be compared against the delta reference @id@, indicated in the @Message.extras.delta.from@ field of the delta message. If the delta reference @id@ of the received delta message does not equal the stored @id@ corresponding to the base payload, the message decoding must fail. The recovery procedure from "RTL18":#RTL18 must be executed.
 * @(RTL21)@ The messages in the @messages@ array of a @ProtocolMessage@ should each be decoded in ascending order of their index in the array.
+* @(RTL22)@ Methods must be provided for creating and destroying a listener which only fires when the message matches a set of criteria
+** @(RTL22a)@ The method must allow for filters matching one or more of: @extras.reference.id@, @extras.reference.type@ or @name@
+** @(RTL22b)@ The method must allow for matching only messages which contain no @extras.reference.type@
```
lmars, line 635: This requirement seems a bit odd, why would a reference not have a type, and why would I want to filter specifically for references which have no type?

content/client-lib-development-guide/features.textile
```diff
@@ -1730,11 +1734,19 @@ class RealtimeChannel:
...
   subscribe((Message) ->) => io // RTL7a
   subscribe(String, (Message) ->) => io // RTL7b
+  subscribe(MessageFilterObject, (Message) ->) // RTL22
   unsubscribe() // RTL8a, RTE5
   unsubscribe((Message) ->) // RTL8a
   unsubscribe(String, (Message) ->) // RTL8a
+  unsubscribe(MessageFilterObject, (Message) ->) // RTL22
   setOptions(options: ChannelOptions) => io // RTL16
 
+class MessageFilterObject:
```
lmars, line 1744: Could this just be `MessageFilter`?

content/client-lib-development-guide/features.textile
```diff
@@ -1730,11 +1734,19 @@ class RealtimeChannel:
...
+  hasRef: bool // RTL22b
+  refId: string // RTL22a
```
lmars, line 1746: Given we have called it `reference` in the extras rather than `ref` (which was my original suggestion), perhaps this should be `referenceId` and `referenceType`?

content/client-lib-development-guide/features.textile
```diff
@@ -2141,6 +2153,10 @@ class VCDiffDecoder
 class DeltaExtras
   from: String // the id of the message the delta was generated from
   format: String //the delta format. Only vcdiff is supported as at API version 1.2
+
+class InteractionExtras:
```
lmars, line 2157: I think this should be called `Reference` (or `MessageReference` if we think `Reference` is too generic), since references aren't always going to be interactions, nor do I think the type name needs to be tied to the fact that it appears in `message.extras`

content/client-lib-development-guide/features.textile
```diff
@@ -1730,11 +1734,19 @@ class RealtimeChannel:
...
   subscribe(String, (Message) ->) => io // RTL7b
+  subscribe(MessageFilterObject, (Message) ->) // RTL22
   unsubscribe() // RTL8a, RTE5
   unsubscribe((Message) ->) // RTL8a
   unsubscribe(String, (Message) ->) // RTL8a
+  unsubscribe(MessageFilterObject, (Message) ->) // RTL22
   setOptions(options: ChannelOptions) => io // RTL16
 
+class MessageFilterObject:
+  hasRef: bool // RTL22b
```
lmars, line 1745: I think `hasRef` (or `hasReference`) is a little ambiguous, since it reads as "this message has references" which could be interpreted as "other messages exist which reference this message" (i.e. it is the target of some other message which references it).

I think really, the requirement is to filter for messages which either are or are not a reference to another message, so "is" is probably better than "has" (i.e. `isRef` / `isReference`).

WDYT?

content/client-lib-development-guide/features.textile
Peter-Maguire: My thinking here was to keep it consistent with `DeltaExtras`, I agree that it should have a name more related to MessageReference, though

content/client-lib-development-guide/features.textile
Peter-Maguire: I think `referenceId`/`referenceType`/`hasReference` would be needlessly verbose, I don't think you could confuse ref with any other property in this context

content/client-lib-development-guide/features.textile
Peter-Maguire: [Done](https://github.com/ably/docs/pull/1444/commits/5b4d8a9b652d63d6bbc388ee022a3ac6dceebbab). This was just a mistype as the way it was always meant to be implemented was to check for the presence of a reference.

content/client-lib-development-guide/features.textile
Peter-Maguire: [Done](https://github.com/ably/docs/pull/1444/commits/9bbc7c471ce1828decf3127b4a48f3275388e747)

content/client-lib-development-guide/features.textile
Peter-Maguire: Good point, from a users perspective isRef makes much more sense. [Done](https://github.com/ably/docs/pull/1444/commits/9bbc7c471ce1828decf3127b4a48f3275388e747).

content/client-lib-development-guide/features.textile
lmars: Ok thanks, so do you know why we didn't use the same logic for the extras field, and used `extras.ref`? Just seems a bit inconsistent

content/client-lib-development-guide/features.textile
lmars: Ok sure, so how about `ReferenceExtras`?

Peter-Maguire: [Done](https://github.com/ably/docs/pull/1444/commits/31b19f0d93c6ec344a98e300a1e90c27cae9e572)

content/client-lib-development-guide/features.textile
Peter-Maguire: I think @Jmgr made the call on that change

Jmgr: I can't remember the whole discussion, but I think we had `refs` at some point and then we decided to have only one, so I just called it reference. Anyway, I'm fine with renaming it `ref` if it's more coherent.

Peter-Maguire: I thought we had settled on `id`, I will change it to `timeserial`

Peter-Maguire: [Done](https://github.com/ably/docs/pull/1444/commits/82364b3dcf09bb0d93e3664687ba20ecc4079a28)

content/client-lib-development-guide/features.textile
```diff
@@ -630,6 +630,10 @@ h3(#realtime-channel). Channel
 ** @(RTL19c)@ In the case of a delta message with a @vcdiff@ @encoding@ step, the @vcdiff@ decoder must be used to decode the base payload of the of delta message, applying that delta to the stored base payload. The direct result of that vcdiff delta application, before performing any further decoding steps, is stored as the updated base payload.
 * @(RTL20)@ The @id@ of the last received message on each channel must be stored along with the base payload. When processing a delta message (i.e. one whose @encoding@ contains @vcdiff@ step) the stored last message @id@ must be compared against the delta reference @id@, indicated in the @Message.extras.delta.from@ field of the delta message. If the delta reference @id@ of the received delta message does not equal the stored @id@ corresponding to the base payload, the message decoding must fail. The recovery procedure from "RTL18":#RTL18 must be executed.
 * @(RTL21)@ The messages in the @messages@ array of a @ProtocolMessage@ should each be decoded in ascending order of their index in the array.
+* @(RTL22)@ Methods must be provided for creating and destroying a listener which only fires when the message matches a set of criteria
```
owenpearson, line 633:
```suggestion
* @(RTL22)@ Methods must be provided for attaching and removing a listener which only executes when the message matches a set of criteria
```

content/client-lib-development-guide/features.textile
```diff
@@ -630,6 +630,10 @@ h3(#realtime-channel). Channel
...
+** @(RTL22a)@ The method must allow for filters matching one or more of: @extras.ref.timeserial@, @extras.ref.type@ or @name@
+** @(RTL22b)@ The method must allow for matching only messages which do not have @extras.ref@
+** @(RTL22c)@ If multiple filters are supplied, each must be satisfied to fire the listener
```
owenpearson, line 636: I think we could word this one a bit better too, maybe something like
> The listener must only be executed if all of the criteria expressed by the provided `MessageFilter` are met

owenpearson (changes requested):

cruickshankpg (approved):

content/client-lib-development-guide/features.textile
Peter-Maguire: [Done](https://github.com/ably/docs/pull/1444/commits/a80ef9d6939950df39bcb0c01996d2f25bfdee3b)

content/client-lib-development-guide/features.textile
```diff
@@ -630,6 +630,11 @@ h3(#realtime-channel). RealtimeChannel
...
 * @(RTL20)@ The @id@ of the last received message on each channel must be stored along with the base payload. When processing a delta message (i.e. one whose @encoding@ contains @vcdiff@ step) the stored last message @id@ must be compared against the delta reference @id@, indicated in the @Message.extras.delta.from@ field of the delta message. If the delta reference @id@ of the received delta message does not equal the stored @id@ corresponding to the base payload, the message decoding must fail. The recovery procedure from "RTL18":#RTL18 must be executed.
 * @(RTL21)@ The messages in the @messages@ array of a @ProtocolMessage@ should each be decoded in ascending order of their index in the array.
+* @(RTL22)@ Methods must be provided for attaching and removing a listener which only executes when the message matches a set of criteria.
+** @(RTL22a)@ The method must allow for filters matching one or more of: @extras.ref.timeserial@, @extras.ref.type@ or @name@.
+** @(RTL22b)@ The method must allow for matching only messages which do not have @extras.ref@.
+** @(RTL22c)@ The listener must only execute if all criteria defined in the provided @MessageFilter@ object are met.
```
owenpearson, line 636:
```suggestion
** @(RTL22c)@ The listener must only execute if all provided criteria are met.
```

owenpearson (changes requested):

owenpearson (approved):
LGTM, thanks

content/client-lib-development-guide/features.textile
```diff
@@ -1174,7 +1179,7 @@ h4. Message
...
+** @(TM2i)@ @extras@ JSON-encodable object, used to contain any arbitrary key value pairs which may also contain other primitive JSON types, JSON-encodable objects or JSON-encodable arrays. The @extras@ field is provided to contain message metadata and/or ancillary payloads in support of specific functionality, e.g. push. Each of these supported extensions is documented separately; for 1.1 the only supported extension is @push@, via the @extras.push@ member; 1.2 adds the @delta@ extension which is of type @DeltaExtras@, the @headers@ extension, which contains arbitrary @string->string@ key-value pairs, settable at publish time, and @reference@ which is of type @ReferenceExtras@. Unless otherwise specified, the client library should not attempt to do any filtering or validation of the @extras@ field itself, but should treat it opaquely, encoding it and passing it to realtime unaltered.
```
lmars, line 1182:
```suggestion
** @(TM2i)@ @extras@ JSON-encodable object, used to contain any arbitrary key value pairs which may also contain other primitive JSON types, JSON-encodable objects or JSON-encodable arrays. The @extras@ field is provided to contain message metadata and/or ancillary payloads in support of specific functionality, e.g. push. Each of these supported extensions is documented separately; for 1.1 the only supported extension is @push@, via the @extras.push@ member; 1.2 adds the @delta@ extension which is of type @DeltaExtras@, the @headers@ extension, which contains arbitrary @string->string@ key-value pairs, settable at publish time, and @ref@ which is of type @ReferenceExtras@. Unless otherwise specified, the client library should not attempt to do any filtering or validation of the @extras@ field itself, but should treat it opaquely, encoding it and passing it to realtime unaltered.
```

lmars (commented):
Inline with the effort being done in https://github.com/ably/docs/pull/1477/files, I think `ReferenceExtras` and `MessageFilters` need a spec point which documents the fields, and then a reference to those spec points in the IDL

content/client-lib-development-guide/features.textile
Peter-Maguire: [Done](https://github.com/ably/docs/pull/1444/commits/7cab8a2ee08d6f63d428fa1582565c67578a2c49) and [Done](https://github.com/ably/docs/pull/1444/commits/bb8796913fee28740dae8c99b9d96dd80949a0a6)

PR #1445, opened 2022-06-01
Feature spec: Add spec point for incremental backoff and jitter

### LINKED ISSUE #832: Revisit backoff strategy for client libraries
Currently a library will retry a request, or attempt to reconnect, if it receives a 5xx error. The retry sequence is simply a periodic retry with a fixed interval, which causes a couple of issues:

- as there is no backoff, this can lead to a very high sustained rate of retries, if there is a large number of failing clients;

- since all clients implement the same backoff time, they can all retry simultaneously if the original failure occurred simultaneously (as has happened in some recent incidents).

A more intelligent retry approach, including some form of backoff and jitter, would address those problems.

Separately, a library will **not** retry if it receives a rate-limiting error with a 429 error code. However, many applications want to retry these operations, and so they perform retry within their own code. This can cause problems because, especially in the rate-limiting case, some backoff would be highly desirable. If the library implemented an intelligent retry on a 429 error, then we could minimise the degree to which applications use a less sympathetic retry strategy themselves.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/EDU-56) by [Unito](https://www.unito.io)

owenpearson:
Resolves #832

### RTB1a

The backoff coefficient described in `RTB1a` can be calculated in javascript like so:
```ts
function getBackoffCoefficient(n) {
  return Math.min((n + 2) / 3, 2);
}
```

The idea of this is modelled off of the '15 -> 20 -> 25 -> 30' sequence suggested by paddy in #832. For the default retry client options this will give that exact sequence, and if users configure their client to use a different retry time this will give an equivalent sequence of longer retry times proportionate to their configured retry time.

### RTB1b

The jitter coefficient described in `RTB1b` can be calculated in javascript like so:
```ts
function getJitterCoefficient() {
  return 1 - Math.random() * 0.2;
}
```
The reason I chose not to allow for the jitter coefficient to be a value higher than 1 is that this could result in a retry length longer than 30s for the 4th retry and beyond. A side effect of this is that all initial retry attempts will be slightly shorter than they would have been prior to this change. It could also work to allow for a jitter coefficient between 0.8 and 1.2 and just say that the retry time is the minimum of that calculation and 30s.

### RTB1

The overall retry time calculation would look like:
```ts
function getRetryTime(initValue, n) {
  return initValue * getBackoffCoefficient(n) * getJitterCoefficient();
}
```

The following code can be used to test it:
```ts
[1, 2, 3, 4, 5].forEach((n) => {
  console.log(getRetryTime(initValue, n));
});
```
Which, with an `initValue` of 15 (seconds) will print something like:
```
13.917470451245713
18.415226855678757
20.444851032898747
26.650729887092275
27.803382948778786
```


content/client-lib-development-guide/features.textile
```diff
@@ -732,6 +732,11 @@ h3(#eventemitter). EventEmitter mixin / interface
...
 
+h3(#backoff-jitter). Incremental backoff and jitter
+* @(RTB1)@ For connections in the @DISCONNECTED@ state and realtime channels in the @SUSPENDED@ state the time until retry is calculated as the product of the initial retry timeout (for @DISCONNECTED@ connections this is the @disconnectedRetryTimeout@, for @SUSPENDED@ channels this is the @channelRetryTimeout@), the backoff coefficient as defined by "@RTB1a@":#RTB1a, and the jitter coefficient as defined by "@RTB1b@":#RTB1b.
+** @(RTB1a)@ The backoff coefficient for the nth retry is calculated as the minimum of @(n + 2) / 3@ and @2@.
+** @(RTB1b)@ The jitter coefficient is a random number between 0.9 and 1. The randomness of this number doesn't need to be cryptographically secure but should be approximately uniformly distributed.
```
paddybyers, line 738: I think we should consider a wider range, eg (0.8, 1)

owenpearson: Ok cool, I've changed it to this in [b7534ec](https://github.com/ably/docs/pull/1445/commits/b7534ec3120c990febdd0fed1ad9a716d817ea97) and updated the description.

content/client-lib-development-guide/features.textile
```diff
@@ -732,6 +732,11 @@ h3(#eventemitter). EventEmitter mixin / interface
...
 ** @(RTE6a)@ The set of listeners called by @emit@ must not change over the course of the @emit@. That is: If a listener being called by @emit@ registers another listener, that second listener should not be called by that invocation of @emit@ (even if it would have been called had it already been present); and if a listener being called by @emit@ removes other listeners, but those other listeners would otherwise have been called during that @emit@ invocation, they should still be called. Tests should exist for both adding and removing. See "https://goo.gl/OVTtjO":https://goo.gl/OVTtjO
 
+h3(#backoff-jitter). Incremental backoff and jitter
+* @(RTB1)@ For connections in the @DISCONNECTED@ state and realtime channels in the @SUSPENDED@ state the time until retry is calculated as the product of the initial retry timeout (for @DISCONNECTED@ connections this is the @disconnectedRetryTimeout@, for @SUSPENDED@ channels this is the @channelRetryTimeout@), the backoff coefficient as defined by "@RTB1a@":#RTB1a, and the jitter coefficient as defined by "@RTB1b@":#RTB1b.
+** @(RTB1a)@ The backoff coefficient for the nth retry is calculated as the minimum of @(n + 2) / 3@ and @2@.
```
lmars, line 737: For the avoidance of doubt, and to make this a little clearer, perhaps we should include what the resulting sequence actually is (i.e, `[1, 4/3, 5/3, 2, 2, 2, ...]`, or `[1, 1.33, 1.67, 2, 2, 2, ...]`)?

lmars:
> with an initValue of 15 (seconds) will print something like:

I wonder if it's worth putting this example in the spec to avoid any ambiguity, or perhaps even specifying it as a test fixture?

content/client-lib-development-guide/features.textile
owenpearson: Good idea, I've added this in [b175078](https://github.com/ably/docs/pull/1445/commits/b1750781dbe4ea316d73ca49c096d4b3b4cb214f).

owenpearson:
> I wonder if it's worth putting this example in the spec to avoid any ambiguity, or perhaps even specifying it as a test fixture?

@lmars I can't think of a way to do this without kind of bloating the spec point, and personally I think it would be a bit too much detail (I will include that example in the GitHub issues to implement this feature if that helps?). If you disagree would you mind making a suggested change?

lmars (approved):

lmars:
I think we should put some examples in the `test-resources` directory in ably-common:

https://github.com/ably/ably-common/tree/main/test-resources

Then SDKs can use those as test fixtures, so we know all SDKs implement the feature in exactly the same way, much like we do for encryption.

That doesn't stop us merging this PR though, so I've approved.

sacOO7:
Expression to calculate upper and lower bounds for generated values using `getRetryTime`

```
upper = min((retryCount + 2) / 3, 2) *initialTimeout
lower = 0.8 * upper
```

PR #1450, opened 2022-06-08
Batch operations improvements

owenpearson:
PR to address the retrospective comments on https://github.com/ably/docs/pull/1421/files

### DIFF
--- a/content/api/rest-api.textile
+++ b/content/api/rest-api.textile
@@ -999,9 +999,9 @@ h3(#batch-publish). Batch publish
 
 h6. POST rest.ably.io/messages
 
-Using this endpoint, you can publish messages to channels in parallel, by passing either a single @BatchSpec@ object or an array of @BatchSpec@ objects in the request body.
+Using this endpoint, you can publish messages to channels in parallel, by passing either a single @BatchPublishRequest@ object or an array of @BatchPublishRequest@ objects in the request body.
 
-A single @BatchSpec@ is an object of the form:
+A single @BatchPublishRequest@ is an object of the form:
 
 bc[text]. {
   channels: <channel names>,
@@ -1013,7 +1013,7 @@ bc[text]. {
 
 The maximum size of each request body must be less than 2MiB and the total size of all messages in a @messages@ array must be less than the "message size limit":/general/limits#message-size (64kiB by default for paid accounts, 16kiB for free accounts).
 
-Sample request, containing multiple @BatchSpec@ objects in an array:
+Sample request, containing multiple @BatchPublishRequest@ objects in an array:
 
 bc[json]. [
   {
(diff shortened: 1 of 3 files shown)


paddybyers:
I support this idea. @SimonWoolf wdyt?

owenpearson:
Just had another thought about the SDK design: we could just add a success field to `BatchSuccess` and `BatchError` which would mean you wouldn't need the `isBatchSuccess` typeguard in the above usage example. The interfaces would be:

```ts
interface BatchSuccess {
  success: true;
  channel: string;
  messageId: string;
}

interface BatchError {
  success: false;
  channel: string;
  error: ErrorInfo;
}
```

Then you could write code like:
```ts
if (result.success) {
  // In this scope `result` is a `BatchSuccess`
} else {
  // In this scope `result` is a `BatchError`
}
```

SimonWoolf:
> For multiple BatchSpecs, however, the server only tells us the messageId for successful messages and the ErrorInfo for unsuccessful ones. If you were to publish to the same channel in two BatchSpecs in the same request, if one succeeded and the other failed there's no practical way to correlate these results to the provided BatchSpecs.

> the SDK could process the REST response into an array of BatchResponses in the same order as the provided BatchSpec array.

I might be misunderstanding -- I'm not sure I understand why that case is ambiguous. Like: the proposal to add IDs is just internal, right; for exposing to the user you're supplying the result in the same order the user supplied the batchspecs, so there's no ambiguity. But that's just what realtime does anyway, it also preserves order. So the sdk should never need to reorder anything (based on ids or otherwise) -- "process the REST response into an array of BatchResponses in the same order as the provided BatchSpec array"should always be a no-op.

(Whether it's intuitive to have the results flattened out into a flat array, rather than a nested array with one entry per batchspec, is another question. Would it make your life easier if we did the latter instead?)

SimonWoolf:
```
if (result.success) {
```
No strong objection but I don't think it's particularly necessary, `if('messageId' in result) {` serves the purpose fine without needing a new field

tbedford:
@owenpearson - where are we on this one please? Thanks.

owenpearson:
@tbedford I think we're broadly agreed on the changes I suggested above. The realtime team have made a change to the backend which unblocks us, but the work to finish this off hasn't been picked up by SDK team yet so this is on hold indefinitely for now.

SimonWoolf:
(per ADR61, we're planning on making realtime return the new format when x-ably-version=2.0 is specified, so client libs will have to implement it as part of protocol v2.0)

QuintinWillison:
This pull request shall remain open in this repository (`ably/docs`) for visibility only, until the features spec has moved over to `ably/specification` (see https://github.com/ably/specification/issues/1). At that point, part of it will be re-formed in that repository, targeting a v2 integration branch there.

QuintinWillison:
This changes proposed in this pull request has been [moved](https://github.com/ably/docs/pull/1450#ref-issue-1393770811) to the new home of the features spec, in [the `ably/specification` repository](https://github.com/ably/specification).

PR #1453, opened 2022-06-09
[EDX-175] Document the stats types and add to the IDL

lawrence-forooghian:
## Description

This adds documentation for all the stats-related types, instead of relying on links to the Ruby documentation.

## Review

Please review commit-by-commit and see commit messages for motivation and approach.

## Outstanding

If anybody would like to suggest descriptions for the things that say `TODO description`, I'd be very grateful. Else I'll leave the TODOs as is to be addressed some other time and will make an issue for it.

### DIFF
--- a/content/client-lib-development-guide/features.textile
+++ b/content/client-lib-development-guide/features.textile
@@ -2050,6 +2050,59 @@ enum StatsIntervalGranularity:
   DAY
   MONTH
 
+class Stats.ConnectionTypes // TS4
+  tls: Stats.ResourceCount // TS4a
+  plain: Stats.ResourceCount // TS4b
+  all: Stats.ResourceCount // TS4c
+
+class Stats.MessageCount // TS5
+  count: Int // TS5a
+  data: Int // TS5b
+
+class Stats.MessageTypes // TS6
+  messages: Stats.MessageCount // TS6a
+  presence: Stats.MessageCount // TS6b
+  all: Stats.MessageCount // TS6c
+
+class Stats.MessageTraffic // TS7
+  realtime: Stats.MessageTypes // TS7a
+  rest: Stats.MessageTypes // TS7b
+  webhook: Stats.MessageTypes // TS7c
+  all: Stats.MessageTypes // TS7d
+
+class Stats.RequestCount // TS8
+  succeeded: Int // TS8a
+  failed: Int // TS8b
+  refused: Int // TS8c
+
+class Stats.ResourceCount // TS9
+  opened: Int // TS9a
+  peak: Int // TS9b
+  mean: Int // TS9c
+  min: Int // TS9d
+  refused: Int // TS9e
+
+class Stats.PushStats // TS10
+  messages: Int // TS10a
+  directPublishes: Int // TS10b
+  notifications: Stats.PushNotificationCount // TS10c
+
+class Stats.PushNotificationCount // TS13
+  invalid: Int // TS13a
+  attempted: Int // TS13b
+  successful: Int // TS13c
+  failed: Int // TS13d
+
+class Stats.XchgMessages // TS11
+  all: Stats.MessageTypes // TS11a
+  producerPaid: Stats.MessageDirections // TS11b
+  consumerPaid: Stats.MessageDirections // TS11c
+
+class Stats.MessageDirections // TS14
+  all: MessageTypes // TS14a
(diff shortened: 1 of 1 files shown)


owenpearson:
Hey @lawrence-forooghian my review for this is still in progress, just posting here for visibility.

I don't feel too comfortable with adding a bunch of TODOs to the spec, so would like to get descriptions added for all of the stats fields if possible. I haven't been find any existing descriptions so far (the realtime codebase doesn't have any), but worse case scenario is I'll end up writing some myself.

lawrence-forooghian:
@owenpearson I've had a go at writing some, based on pretty shaky sources of information / my guesses. There's also a couple left that need documenting. Do you have any thoughts on how we might be able to move this forwards now?

lawrence-forooghian:
@owenpearson I've had a go at writing some, based on pretty shaky sources of information / my guesses (see the commit messages for more details). There's also a couple left that need documenting. Do you have any thoughts on how we might be able to move this forwards now?

owenpearson:
@lawrence-forooghian I've looked at the realtime code and have come up with these suggestions to fill in the gaps:
- TS10c: The count of push notifications.
- TS13: Contains a count of push notifications published broken down by outcome.
- TS13a: Total number of attempted push notifications which were rejected due to invalid request data.
- TS13b: Total number of attempted push notifications including notifications which were rejected as invalid or failed to publish.

content/client-lib-development-guide/features.textile
```diff
@@ -2015,6 +2050,59 @@ enum StatsIntervalGranularity:
   DAY
   MONTH
 
+class Stats.ConnectionTypes // TS4
```
owenpearson, line 2053: I'm not sure about this so feel free to dismiss: for consistency with the rest of the IDL this should probably just be `class ConnectionTypes`, although that may be a bit confusing so maybe we do something like `class ConnectionTypesStats`, wdyt?

content/client-lib-development-guide/features.textile
```diff
@@ -1281,15 +1281,50 @@ h4. Stats
...
+** @(TS8b)@ @failed@ - integer. The number of requests that failed.
+** @(TS8c)@ @refused@ - integer. The number of requests that were refused, typically as a result of permissions or a limit being exceeded.
+* @(TS9)@ @Stats.ResourceCount@ - contains aggregate data for usage of a resource in a specific scope. Has the following attributes:
+** @(TS9a)@ @opened@ - integer. The total number of resources opened of this type.
+** @(TS9b)@ @peak@ - integer. The peak number of resources of this type used for this period.
+** @(TS9c)@ @mean@ - float. The average number of resources of this type used for this period.
+** @(TS9d)@ @min@ - integer. The minimum total resources of this type used for this period.
+** @(TS9e)@ @refused@ - integer. The number of resource requests refused within this period.
+* @(TS10)@ @Stats.PushStats@ - detailed stats on push notifications. Has the following attributes:
+** @(TS10a)@ @messages@ - integer. Total number of Push channel messages.
```
owenpearson, line 1312: Going purely on the wikipedia page for 'push technology' I'd say we should use lowercase 'p' for 'push'

owenpearson (changes requested):

lawrence-forooghian:
Thanks @owenpearson for your help here! I'm going to hold off on making any further changes here until I've got a better idea on whether we want to proceed with documenting the stats types ([internal Slack discussion](https://ably-real-time.slack.com/archives/C030C5YLY/p1655473324276499)).

content/client-lib-development-guide/features.textile
lawrence-forooghian: I think that I'd prefer to keep the naming consistent with that used in the feature spec, which uses these namespaced `Stats.` types. I think if we wanted to change the naming we should do it in the feature spec too, which I don't really want to do in this PR.

lawrence-forooghian:
@owenpearson I've added your suggestions. Would you mind taking another look at this one, please?

content/client-lib-development-guide/features.textile
owenpearson: Yeah sure, I do think they should both be changed but it's very minor and I'm happy to let it pass so we can land this PR 👍

owenpearson (approved):
LGTM, thanks

PR #1462, opened 2022-06-15
[EDX-146] Address some differences between the IDL and the feature spec

lawrence-forooghian:
## Description

This addresses some differences between the IDL and the feature spec - sometimes by changing the IDL, sometimes by changing the feature spec. **Please review commit-by-commit – the commit messages contain plenty of details.**


content/client-lib-development-guide/features.textile
```diff
@@ -668,8 +668,8 @@ h3(#realtime-presence). Presence
 ** @(RTP16b)@ If the channel is @ATTACHING@ or @INITIALIZED@, then if @ClientOptions.queueMessages@ has not been explicitly set to @false@, the presence messages should be queued at a channel level, to be handled per @RTP16a@ once the channel becomes @ATTACHED@, or failed if the channel state becomes @SUSPENDED@, @FAILED@, or @DETACHED@ first
 ** @(RTP16c)@ In any other case the operation should result in an error
 * @(RTP6)@ @Presence#subscribe@ function:
-** @(RTP6a)@ Subscribe with no arguments subscribes a listener to all presence messages
-** @(RTP6b)@ Subscribe with a single action argument - such as @ENTER@, @LEAVE@, @UPDATE@ or @PRESENT@ - subscribes a listener to receive only presence messages with that action
+** @(RTP6a)@ Subscribe with a single listener argument subscribes a listener to all presence messages
+** @(RTP6b)@ Subscribe with a listener argument and an action argument - such as @ENTER@, @LEAVE@, @UPDATE@ or @PRESENT@ - subscribes a listener to receive only presence messages with that action
```
Peter-Maguire, line 672: The listener would be after the action argument in the code so it makes sense to write them that way here

```suggestion
** @(RTP6b)@ Subscribe with an action argument and a listener argument - such as @ENTER@, @LEAVE@, @UPDATE@ or @PRESENT@ - subscribes a listener to receive only presence messages with that action
```

content/client-lib-development-guide/features.textile
```diff
@@ -593,15 +593,15 @@ h3(#realtime-channel). Channel
 ** @(RTL6h)@ The @publish(name, data)@ form should not take any arguments other than those two. If a client library has supported additional arguments to the @(name, data)@ form (e.g. separate arguments for @clientId@ and @extras@, or a single @attributes@ argument) in any 1.x version, it should continue to do so until version 2.0.
 ** @(RTL6f)@ @Message#connectionId@ should match the current @Connection#id@ for all published messages, a test should exist to ensure the @connectionId@ for received messages matches that of the publisher
 * @(RTL7)@ @Channel#subscribe@ function:
-** @(RTL7a)@ Subscribe with no arguments subscribes a listener to all messages
-** @(RTL7b)@ Subscribe with a single name argument subscribes a listener to only messages whose @name@ member matches the string name
+** @(RTL7a)@ Subscribe with a single listener argument subscribes a listener to all messages
+** @(RTL7b)@ Subscribe with a listener argument and a name argument subscribes a listener to only messages whose @name@ member matches the string name
```
Peter-Maguire, line 597:
```suggestion
** @(RTL7b)@ Subscribe with a name argument and a listener argument subscribes a listener to only messages whose @name@ member matches the string name
```

Peter-Maguire (changes requested):

content/client-lib-development-guide/features.textile
lawrence-forooghian: Thanks, done!

content/client-lib-development-guide/features.textile
lawrence-forooghian: Thanks, done!

QuintinWillison (approved):

content/client-lib-development-guide/features.textile
```diff
@@ -2117,7 +2117,7 @@ class EventEmitter<Event, Data>:
   off() // RTE5
   off((Data...) ->) // RTE5
   off(Event, (Data...) ->) // RTE5
-  emit(Event, Data...)  // RTE6
+  emit(Event, Data...)  // internal, RTE6
```
maratal, line 2126: Maybe to make _internal_ specifier a part of the IDL? Just thoughts, can be ignored.

maratal (approved):

content/client-lib-development-guide/features.textile
lawrence-forooghian: I think it's a good idea but not one I really want to address here - this PR is quite big already and I'm just following the existing habit of marking internal things with a comment. Personally I'm not sure whether we should even have internal methods in the IDL, but that's for another day 😄

Peter-Maguire (approved):

PR #1471, opened 2022-06-22
removal of breaking : from api ref

Sapples:

## Description

Stats API page incorrectly said parameters should be prefixed by a colon in Ruby. 

* [Quick wins](https://ably.atlassian.net/browse/EDU-609).
* [Stats API options parameters appear to be incorrect](https://ably.atlassian.net/browse/EDU-822).

## Review

* [API Ref - Statistics](https://ably-docs-edu-822-stats-2ewh6k.herokuapp.com/api/rest-sdk/statistics/)


tbedford (changes requested):
I'm not sure this is quite right. Looking at the diff it appears the colon (`:`) has been removed from Ruby parameters, but I think this is not correct - the colon in Ruby indicates a symbol. 

Looking at the ticket description, it seems to me the actual issue is in the Curl command provided:

```
curl -u "api-key" https://rest.ably.io/stats?unit=:minute\&:start=1655107425000\&:end=1655197425000\&:limit=1000
```

If you look carefully there is a colon before end and limit and I think this is what needs to be removed.
 
However, I think the page linked to in the ticket is for the Ruby SDK, not the Curl command, so I think the issue is in the original ticket - there is in fact no error in the documentation for the link provided in the ticket. Worth following up with @richiejarvis . Thanks.

Sapples:
Closing as not needed, see JIRA ticket: https://ably.atlassian.net/browse/EDU-822

PR #1473, opened 2022-06-24
[EDX-177] Extract script tags

MymmiJ:

## Description

Moving scripts from inline <script> tags in textile into their own files and preparing them for execution.

The next step is, during node creation, to check if there exists a valid static JS file for a given slug and, if there is such a file, to load it as an asset using the static asset URL.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-177).


DuncanEdwards (approved):
Hey, it is what it is.

PR #1477, opened 2022-06-29
[EDX-146] Make sure every type, method, and attribute in IDL mentions a feature spec point

lawrence-forooghian:
## Description

This makes sure that every type, method, and attribute in the IDL mentions a feature point spec. It adds feature spec points where necessary, and also checks all of the feature spec points currently mentioned in the IDL for appropriateness, updating the IDL where necessary.

We are doing this as part of our efforts to make the IDL the canonical description of the public interface of our client SDKs. As part of this work, we want to be sure that the IDL is a subset of the information contained in the feature spec points.

## Review

**Please review commit-by-commit, since there are a lot of changes, and a lot of explanation of what I've done in the commit messages.**

## Outstanding work

There is no feature spec point describing the inheritance of `ClientOptions` from `AuthOptions` – this is a large one and I've split it into a separate task https://github.com/ably/docs/issues/1476 to be worked on later.


content/client-lib-development-guide/features.textile
```diff
@@ -2118,22 +2118,22 @@ enum DeviceFormFactor:
...
-class PushChannelSubscription:
+class PushChannelSubscription: // PCS*
   +forDevice(channel: String, deviceId: String) => PushChannelSubscription
   +forClientId(channel: String, clientId: String) => PushChannelSubscription
   deviceId: String? // PCS2, PCS5, PCS6
   clientId: String? // PCS3, PCS6
   channel: String // PCS4
 
-class ErrorInfo:
+class ErrorInfo: // TI1
```
maratal, line 2178: TI*?

content/client-lib-development-guide/features.textile
```diff
@@ -1605,7 +1607,7 @@ class RealtimeClient: // RTC*
   constructor(keyOrTokenStr: String) // RTC12
   constructor(ClientOptions) // RTC12
   auth: Auth // RTC4
-  push: Push
+  push: Push // RTC13
```
maratal, line 1658: Confusing with RSC21 above, maybe to add ref to Push object itself here as well?

content/client-lib-development-guide/features.textile
lawrence-forooghian: Thanks! Changed.

Peter-Maguire (approved):

content/client-lib-development-guide/features.textile
maratal: I've meant that both of them are the same Push object, but have a different spec point, which was confusing. But that's how it should be, so nevermind =)

maratal (approved):

ikbalkaya (approved):
LGTM

content/client-lib-development-guide/features.textile
```diff
@@ -677,6 +705,7 @@ h3(#realtime-presence). RealtimePresence
 ** @(RTP6b)@ Subscribe with an action argument and a listener argument - such as @ENTER@, @LEAVE@, @UPDATE@ or @PRESENT@ - subscribes a listener to receive only presence messages with that action
 ** @(RTP6c)@ Implicitly attaches the @RealtimeChannel@ if the channel is in the @INITIALIZED@ state. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result
 * @(RTP7)@ @RealtimePresence#unsubscribe@ function:
+** @(RTP7c)@ Unsubscribe with a no arguments unsubscribes all listeners
```
lmars, line 708:
```suggestion
** @(RTP7c)@ Unsubscribe with no arguments unsubscribes all listeners
```

content/client-lib-development-guide/features.textile
```diff
@@ -1133,7 +1162,7 @@ h3(#push-device-authentication). Push device authentication
 
 h3(#push-channels). Push channels
 
-* @(RSH7)@ In platforms that support receiving push notifications, @RestChannel@ and @RealtimeChannel@ have an additional @push@ field, as an object with the following interface:
+* @(RSH7)@ In platforms that support receiving push notifications, @RestChannel@ and @RealtimeChannel@ have an additional @push@ field, as a @PushChannel@ object with the following interface:
```
lmars, line 1165:
```suggestion
* @(RSH7)@ In platforms that support receiving push notifications, @RestChannel@ and @RealtimeChannel@ have an additional @push@ field, which is a @PushChannel@ object with the following interface:
```

content/client-lib-development-guide/features.textile
```diff
@@ -358,6 +373,7 @@ The threading and/or asynchronous model for each realtime library will vary by l
 
 h3(#realtimeclient). RealtimeClient
 
+* @(RTC12)@ Has the the same constructors as @RestClient@, as defined in "RSC1":#RSC1
```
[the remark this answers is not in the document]
lawrence-forooghian: Thanks for spotting, fixed.

content/client-lib-development-guide/features.textile
lawrence-forooghian: Thanks for spotting, fixed.

content/client-lib-development-guide/features.textile
lawrence-forooghian: Thanks for spotting, fixed.

content/client-lib-development-guide/features.textile
```diff
@@ -1530,6 +1583,7 @@ h4(#token-params). TokenParams
 ** @(TK2b)@ @capability@ string - Capability requirements JSON stringified for the token. When omitted, Ably will default to the capabilities of the underlying key
 ** @(TK2c)@ @clientId@ string - A @clientId@ string to associate with this token. If @clientId@ is @null@ or omitted, then the token is prohibited from assuming a @clientId@ in any operations, however if @clientId@ is a wildcard string @'*'@, then the token is permitted to assume any @clientId@. Any other string value for @clientId@ implies that the @clientId@ is both enforced and assumed for all operations for this token
 ** @(TK2d)@ @timestamp@ long - The timestamp (in milliseconds since the epoch) of this request. Timestamps, in conjunction with the @nonce@, are used to prevent requests from being replayed. @timestamp@ is a "one-time" value, and is valid in a request, but is not validly a member of any default token params such as @ClientOptions#defaultTokenParams@
+** @(TK2e)@ @nonce@ optional string - Used to prevent against replay attacks. If not provided, then the library will automatically generate one if needed, as described by "RSA9c":#RSA9c.
```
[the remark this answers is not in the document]
lawrence-forooghian: I'm not sure why it's in the IDL, but I wonder if the intention was that it be exposed as an internal API? For example [ably-ruby uses it in a test that repeats a nonce to trigger a specific exception](https://github.com/ably/ably-ruby/blob/65e0a24889bcc356bd0e5259f7ffa0d1133fd046/spec/acceptance/rest/client_spec.rb#L1341-L1354) (to be honest, I'd say that the `it` for that test and actual assertions it makes don't really match, but that's another matter).

It's currently exposed in cocoa, js, dotnet, ruby, and rust. It's not exposed in java or go.

I don't really have a strong opinion on this, but the general approach I've taken elsewhere was to add something (that's already in the IDL) to the spec if it's implemented in the majority of our client SDKs and if it doesn't otherwise seem _too_ controversial.

Do you have a preference for what I should do here?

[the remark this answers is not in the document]
lmars: Hmm yes, I mean that Ruby test doesn't seem to make sense, since surely _not_ allowing the nonce to be set at all by the user removes the need to test that the user doesn't replay one. Also, `TokenParams#nonce` is even [documented in the API reference docs](https://docs.ably.io/api/realtime-sdk/types/#token-params) 🤔 .

So, happy for you to merge this to be consistent with what we've documented and implemented elsewhere, and I can raise the issue separately.

lmars (approved):

PR #1485, opened 2022-07-05
[EDX-178] Load script files

MymmiJ:

## Description

This PR is to load script files into Gatsby from script files; it doesn't necessarily guarantee that the scripts will function as they currently do.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-178).

## Review

Check the below pages to see pages which should respectively load, and not load, a script:

* [Should load a script](http://localhost:8000/docs/quick-start-guide)
* [Should not attempt to load a script](http://localhost:8000/docs/best-practice-guide)

Check Network tab -> JS filter and look for a request of:

`GET /scripts/quick-start-guide.js HTTP/1.1
...
`


data/createPages/index.js
```diff
@@ -45,6 +46,14 @@ const createPages = async ({ graphql, actions: { createPage } }) => {
         edge.node.parentSlug,
         edge.node.version,
       );
+      // Check if script exists for page
```
DuncanEdwards, line 49: This should be split out into a separate method (it can be in this file) that returns a script

src/components/blocks/complex-functionality/Script.tsx
```diff
@@ -0,0 +1,24 @@
...
+  attribs,
+}: {
+  attribs: {
+    src: string;
+    crossorigin?: string;
+    onLoad?: (event: Event) => void;
+    onError?: (event: Event) => void;
+  };
+}) => {
+  if (attribs.src && isString(attribs.src)) {
```
DuncanEdwards, line 16: Inline ?: if statement and lets stop using NullComponent

DuncanEdwards (changes requested):

DuncanEdwards (approved):

PR #1490, opened 2022-07-11
[EDX-133] New tip design

MymmiJ:

## Description

Adds unique 'tip' design to display tooltips.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-133).

## Review

View this page: http://localhost:8000/docs/quick-start-guide

There should be an element quite high on the page with the TIP markup and display.

The same should be true of any page with a `p(tip).` entry in the textile.

* [Page to review](http://localhost:8000/docs/quick-start-guide)


src/components/blocks/typography/P/index.tsx
```diff
@@ -0,0 +1,45 @@
+import GenericHtmlBlock from '../../Html/GenericHtmlBlock';
+import '@ably/ui/core/styles.css';
+import { PendingClassName } from './PendingClassName';
+import { HtmlComponentProps } from '../../../html-component-props';
+import { Tip } from './Tip';
+
+const Paragraph = ({ data, attribs }: HtmlComponentProps<'p'> & { attribs: { ['data-variant']: string } }) => {
+  const isTip = attribs.className && (attribs.className === 'tip' || attribs.className.includes(' tip '));
```
DuncanEdwards, line 8: What about if tip is the last className? IE:
`<div className="test tip"/>`

DuncanEdwards (changes requested):
Also, the border is too think around the tip by one pixel

src/clientside-data-utilities/clientside-data-utilities.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    assert(
+      property(
+        lorem({ maxCount: 1 }), // The class to find
+        lorem(), // The full className
+        nat(2), // Where to place the class within the className
+        (cssClass, className, position) => {
+          const attribs = {
+            className: className.replace(cssClass, `#${cssClass}`),
+          };
+          if (position === 0) {
```
DuncanEdwards, line 49: Aaaargh! This is the same code as above and not refactored.. We need a helper method here to minimize cut + paste test checking. I feel like we also need a quote from Clean Code (Oh, not again):
image

DuncanEdwards (changes requested):
A little test refactoring required.

DuncanEdwards (approved):

PR #1491, opened 2022-07-11
[EDU-863] - Kafka Connector Overview additions

tbedford:

## Description

Improve the Kafka overview:

- [x] Better explanation of use case
- [x] Cross reference Kafka rule with explanation of use
- [x] Add overview diagram

Tickets:

* [Epic: EDU-756](https://ably.atlassian.net/browse/EDU-756)
* [EDU-863](https://ably.atlassian.net/browse/EDU-863)

## Review

* [Kafka Connector Overview](https://ably-docs-edu-863-kafka-zd2rhm.herokuapp.com/general/kafka-connector/)


tbedford:
@ikbalkaya - thanks Ikbal. I've added a note on that with links to further information.

content/general/kafka-connector.textile
```diff
@@ -8,15 +8,74 @@ languages:
...
+You can use the Ably Kafka Connector to send data from one or more "Kafka topics":https://developer.confluent.io/learn-kafka/apache-kafka/topics/ into Ably "channels":/realtime/channels. The following diagram illustrates a typical deployment:
+
+<a href="/images/diagrams/ably-kafka-connector-overview.png" target="_blank">
+  <img src="/images/diagrams/ably-kafka-connector-overview.png" style="width: 100%"
+  title="Click to enlarge" alt="Ably Kafka Connector overview">
+</a>
+
+h2(#data-ably-kafka). Data from Ably to Kafka
+
+If you need to send data from Ably to Kafka, use a "Kafka rule":/general/firehose/kafka-rule.
```
marklewin, line 34: It seems wrong to have separate heading for something that we're saying use a different tool for. IMO this should be at the top of the page, to stop anyone reading further if they're looking to do Ably->Kafka rather than Kafka->Ably.

Previously this was mentioned after the "The Ably Kafka Connector provides a ready-made-integration..." sentence and IMO would be the best place for it. And maybe use some sort of callout to make it clearer?

content/general/kafka-connector.textile
```diff
@@ -8,15 +8,74 @@ languages:
...
+
+The Ably Kafka Connector is a sink connector built on top of Kafka Connect. It can be self-hosted or hosted with a third-party provider, the most common being the Confluent Platform. You can download it from either "GitHub":https://github.com/ably/kafka-connect-ably or "Confluent Hub":https://www.confluent.io/hub/ably/kafka-connect-ably and install it into your Kafka Connect workers.
+ 
+The Ably Kafka Connector provides a ready-made integration between Kafka and Ably via your Ably API key. Once installed, you can configure the connector with your Ably API key to enable data from one or more Kafka topics to be published into a single Ably channel or multiple Ably channels. Events are then distributed in realtime to web, mobile, and IoT clients over feature-rich, multi-protocol pub/sub Ably channels optimized for last-mile delivery.
+
+h2(#mapping-options). Mapping options
+
+The Ably Kafka Connector supports two mapping options:
+
+1. Static
```
marklewin, line 46: If we number these here, we should probably number the headings below too.

content/general/kafka-connector.textile
```diff
@@ -8,15 +8,74 @@ languages:
...
+
+```
+channel = channel_#{topic}
+message.name = message_#{key}
+```
+
+For an @events@ topic, the channel would translate as @channel_events@. For a @remove@ key, the @message.name@ would be configured as @message_remove@.
+
+It is important to note that you can also use static mappings in conjunction with pattern-based mapping. For example, you could define the channel as @channel_#{topic}@, but @message.name@ as @single_message@. In this instance, the channel will be mapped by pattern and the message name will be mapped statically, as shown in the following:
```
marklewin, line 69:
```suggestion
It is important to note that you can also use static mapping in conjunction with pattern-based mapping. For example, you could define the channel as @channel_#{topic}@, but @message.name@ as @single_message@. In this instance, the channel will be mapped by pattern and the message name will be mapped statically, as shown in the following configuration:
```

marklewin (changes requested):
No show-stoppers ... just a couple of suggestions!

content/general/kafka-connector.textile
tbedford: Yes, I think we need to call it out somehow, but perhaps a new section with heading is too much. I did this because I was asked to add this info (even though it was already on the page - it was not standing out enough!)

content/general/kafka-connector.textile
tbedford: As the toolchain adds section numbering automatically, this would result in some odd numbering, so I've just made the list bulleted.

marklewin (approved):
LGTM

PR #1496, opened 2022-07-20
Remove mention of occupancy events on metachannels; add integrations to metadata page

SimonWoolf:
Remove all mention of channel.occupancy events on metachannels: these are long-deprecated in favour of inband occupancy events and occupancy rules, there's no circumstances in which we'd enable this for a customer now, so we shouldn't document it.

Also added integrations to the list of options for obtaining channel metadata and in a couple other places, not sure why were omitted


tbedford (approved):
LGTM, thanks @SimonWoolf!

PR #1497, opened 2022-07-20
[EDU-781] - Update Quickstart code to use promises 

tbedford:
## Description

The purpose of this PR is to update the Quickstart Guide for both JavaScript and Node.js language options, to use the promises interface. 

I have also created a [repo](https://github.com/ably/quickstart-js) to contain complete working source code, as an aid to developers. The repo is linked to from both the JavaScript and Node.js language versions.

Tickets:

* Epic: [EDU-779](https://ably.atlassian.net/browse/EDU-779)
* [EDU-781](https://ably.atlassian.net/browse/EDU-781)

## Review

Review (and test) the Quickstart guide for both JavaScript and Node.js language options:

* [Quickstart guide](https://ably-docs-edu-781-updat-ef7cyq.herokuapp.com/quick-start-guide/)

### DIFF
--- a/content/root/quick-start-guide.textile
+++ b/content/root/quick-start-guide.textile
@@ -511,10 +511,10 @@ blang[default].
 blang[python,php].
   The <span lang="python">Python</span><span lang="php">PHP</span> SDK is only available with a REST interface. You will need to use one of the realtime libraries to create or manage a connection to Ably. Use the language selector above to select a language with realtime support.
 
-bc[javascript]. ably.close(); // resolves synchronously
+bc[javascript]. ably.close(); // runs synchronously
 console.log('Closed the connection to Ably.');
 
-bc[nodejs]. ably.close(); // resolves synchronously
+bc[nodejs]. ably.close(); // runs synchronously
 console.log('Closed the connection to Ably.');
 
 bc[java]. ably.connection.close();
--- a/content/root/quick-start-guide.textile
+++ b/content/root/quick-start-guide.textile
@@ -265,20 +265,12 @@ blang[python,php].
 *Note*: The connection example uses basic authentication to pass the API key from the application to Ably. Basic authentication should not be used client-side in a production environment, such as in a browser, to avoid exposing the API key. "Token authentication":/core-features/authentication#token-authentication should be used instead.
 
 bc[javascript]. const ably = new Ably.Realtime.Promise('{{API_KEY}}');
-try {
-  await ably.connection.once('connected');
-  console.log('Connected to Ably!');
-} catch (error) {
-  console.error(error);
-}
+await ably.connection.once('connected');
+console.log('Connected to Ably!');
 
 bc[nodejs]. const ably = new Ably.Realtime.Promise('{{API_KEY}}');
-try {
-  await ably.connection.once('connected');
-  console.log('Connected to Ably!');
-} catch (error) {
-  console.error(error);
-}
+await ably.connection.once('connected');
+console.log('Connected to Ably!');
 
 bc[java]. ClientOptions options = new ClientOptions("{{API_KEY}}");
 AblyRealtime ably = new AblyRealtime(options);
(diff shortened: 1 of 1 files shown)


content/root/quick-start-guide.textile
```diff
@@ -360,18 +369,28 @@ blang[default].
...
-```[nodejs](code-editor:quick-start-guide/send-message)
-  const channel = ably.channels.get('quickstart');
-  channel.subscribe('greeting', (message) => {
-    console.log(`Received a greeting message in realtime: ${message.data}`);
-  });
+```[nodejs]
+// get the channel to subscribe to
+const channel = ably.channels.get('quickstart');
+
+// Subscribe to a channel
```
marklewin, line 388: Consider using block comments for these multi-line comments? `/* comment ... */`

marklewin (approved):
One minor suggestion about comments for your consideration, but otherwise LGTM.

content/root/quick-start-guide.textile
tbedford: Yes, good idea, reads a little easier in that form.

tbedford:
@owenpearson - if you're happy Owen, I can merge, thanks!

content/root/quick-start-guide.textile
```diff
@@ -489,15 +511,11 @@ blang[default].
 blang[python,php].
   The <span lang="python">Python</span><span lang="php">PHP</span> SDK is only available with a REST interface. You will need to use one of the realtime libraries to create or manage a connection to Ably. Use the language selector above to select a language with realtime support.
 
-bc[javascript](code-editor:quick-start-guide/close). ably.connection.close();
-ably.connection.on('closed', () => {
-  alert('Closed the connection to Ably.');
-});
+bc[javascript]. ably.close(); // resolves synchronously
```
owenpearson, line 514: I think it's a bit confusing to use the word 'resolves' here, that kind of implies that the method returns a `Promise` but in this case it doesn't. I would suggest changing it to either 'runs synchronously' or 'executes synchronously'

content/root/quick-start-guide.textile
```diff
@@ -261,15 +264,21 @@ blang[python,php].
...
-bc[javascript](code-editor:quick-start-guide/connect). const ably = new Ably.Realtime('{{API_KEY}}');
-ably.connection.on('connected', () => {
-  alert('Connected to Ably!');
-});
+bc[javascript]. const ably = new Ably.Realtime.Promise('{{API_KEY}}');
+try {
+  await ably.connection.once('connected');
+  console.log('Connected to Ably!');
+} catch (error) {
+  console.error(error);
```
owenpearson, line 272: This catch block will never execute since the promise returned by `ably.connection.once('connected')` will never reject.

If you wanted this catch block to execute if the client fails to connect then it would have to be something like this:
```ts
await Promise.any([
  ably.connection.once('connected'),
  ably.connection.once('failed'),
]);
 ```

**Review by owenpearson — CHANGES_REQUESTED** [review]
Looks great :) just a couple of small things to fix

File: content/root/quick-start-guide.textile (hunk 3 above)
Comment (owenpearson) [reply content/root/quick-start-guide.textile:272]: I'm not sure whether using `Promise.any` will make the example too complicated for beginner JS devs. It might be worth just not catching the error and doing:
```ts
  await ably.connection.once('connected');
  console.log('Connected to Ably!');
```
If the client fails to connect the promise will never resolve and the library will still log an error message to the console so it's not like the developer wouldn't be able to see what's happening.

content/root/quick-start-guide.textile
tbedford: OK, changed to "runs".

content/root/quick-start-guide.textile
tbedford: I've gone for your latter suggestion. Much cleaner!

owenpearson (approved):
LGTM

PR #1506, opened 2022-07-26
[EDX-230] Replace nested codeblocks within blangs

m-hulbert:
## Description

This PR updates the `/realtime/connection` page (for all versions) to replace occurrences of codeblocks that are nested within `blang` elements. This is to make these easier to handle in a future toolchain.

See [JIRA](https://ably.atlassian.net/browse/EDX-230) for further information.

## Review

To review, visit the [realtime/connection](https://ably-docs-edx-230-neste-b4mjmt.herokuapp.com/realtime/connection) page.

### DIFF
--- a/content/realtime/connection.textile
+++ b/content/realtime/connection.textile
@@ -140,136 +140,127 @@ The @Connection@ object is an @EventEmitter@ and emits an event whose name is th
 
 The @Connection@ object can also emit an event that is not a state change: an @update@ event. This happens when there's a change to connection conditions for which the connection state doesn't change - that is, when the library remains connected, e.g. after a "reauth":/realtime/authentication#auth-url.
 
-blang[jsall].
-  ```[jsall]
-    realtime.connection.on('connected', function(stateChange) {
-      console.log('Ably is connected');
-    });
-  ```
+```[jsall]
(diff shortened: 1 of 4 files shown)


tbedford (changes requested):
Looks good. I've tested under the website view. I checked each code sample and switched between languages and noted that when a language isn't supported (mostly Flutter) the code sample default to JavaScript and displays a warning.

The only slight issue I want to raise is that the warning text is sometimes hidden as can be seen in the following screenshot:

image

Not necessarily a blocker for this PR, but thought I should raise it. Possibly has to be a website style fix?

m-hulbert:
> Looks good. I've tested under the website view. I checked each code sample and switched between languages and noted that when a language isn't supported (mostly Flutter) the code sample default to JavaScript and displays a warning.
> 
> The only slight issue I want to raise is that the warning text is sometimes hidden as can be seen in the following screenshot:
> 
> image
> 
> Not necessarily a blocker for this PR, but thought I should raise it. Possibly has to be a website style fix?

This is an issue that already exists on live as that code block wasn't changed as part of this PR. Worth noting so we can get it fixed though!

tbedford (approved):
LGTM under web view.

PR #1507, opened 2022-07-26
[EDX-153] - Update ToC in new toolchain

Sapples:

## Description

Updated Table of Contents to match latest live Table of Contents.

* [EDX-153 - Update ToC](https://ably.atlassian.net/browse/EDX-153).

## Review

Run locally new toolchain locally and check links work in the side nav table of contents and that it's rendering correctly.

### DIFF
--- a/data/yaml/page-furniture/sidebar.yaml
+++ b/data/yaml/page-furniture/sidebar.yaml
@@ -6,29 +6,33 @@ items:
       - label: "Quickstart guide"
         link: /docs/quick-start-guide
       - label: "Tutorials"
-        link: /tutorials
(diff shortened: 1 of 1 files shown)


m-hulbert (changes requested):
Links all look good! 

It's just inconsistent in a few places with sentence case, e.g. `Quickstart guide` and `Client libraries`. 

I'd also suggest you remove the `new: true` tags for now as we may identify these within the front matter of a page in future.

Note: links attached to a dropdown don't currently work but this will likely be addressed in the redesign for the ToC planned in a future sprint.

m-hulbert (approved):

PR #1508, opened 2022-07-27
Fix broken link in WebRTC screen share tutorial 

stichbury:
## Description

* [Jira epic ticket](https://ably.atlassian.net/browse/EDU-608) 
* [Jira ticket](https://ably.atlassian.net/browse/EDU-866).

## Review
See my notes on the JIRA ticket. This is my first PR in docs BTW, so please be kind 🙏 

> This is going to be harder to fix than originally thought because the chrome extension is no longer available, but the tutorial appears to be predicated on using Temasys libraries. I found a replacement chrome extension [Circuit by Unify](https://chrome.google.com/webstore/detail/circuit-by-unify/mhkbaognlahkdimlfcfhbeihldmjofgg?hl=en)  but I cannot be sure that it’s sufficient to just swap it into Chrome and have the tutorial work as described.
>  
> Given that work to review which tutorials we keep up to date is ongoing, I don’t think it’s sensible for me to spend any time trying to test if this works (as it’s from an external contributor and not a key tutorial anyway, from what I can tell).
> 
>  It’s hacky, horrible, but I’m going to make a preliminary fix by simply removing the broken link and linking to the temasys website instead. It doesn’t make the tutorial work any better than it currently does but it’s not working now either, __and__ the link is broken, so at least it fixes the link. Still really awful. Sorry.

* [Page to review](https://ably.com/tutorials/web-rtc-screen-sharing)

### DIFF
--- a/content/tutorials/web-rtc-screen-sharing.textile
+++ b/content/tutorials/web-rtc-screen-sharing.textile
@@ -339,7 +339,7 @@ Now, let's go ahead and implement the call, receive and end call features.
 Paste the following code at the end of your @ably-screenshare.js@ file:
 
 
-p(tip). Note: Since Chrome requires an extension to share your screen, we will need to install the "temasys chrome plugin":https://temasys.io
+p(tip). Note: Since Chrome requires an extension to share your screen, you will need to install the "temasys chrome plugin":https://temasys.io
 
 ```[javascript]
     function initiateCall(client_id) {
--- a/content/tutorials/web-rtc-screen-sharing.textile
+++ b/content/tutorials/web-rtc-screen-sharing.textile
@@ -339,7 +339,7 @@ Now, let's go ahead and implement the call, receive and end call features.
 Paste the following code at the end of your @ably-screenshare.js@ file:
 
 
-p(tip). Note: Since Chrome requires an extension to share your screen, we will need to install the temasys chrome plugin "here":https://chrome.google.com/webstore/detail/skylink-WebRTC-tools/ljckddiekopnnjoeaiofddfhgnbdoafc
+p(tip). Note: Since Chrome requires an extension to share your screen, we will need to install the "temasys chrome plugin":https://temasys.io
 
 ```[javascript]
     function initiateCall(client_id) {


content/tutorials/web-rtc-screen-sharing.textile
```diff
@@ -339,7 +339,7 @@ Now, let's go ahead and implement the call, receive and end call features.
 Paste the following code at the end of your @ably-screenshare.js@ file:
 
 
-p(tip). Note: Since Chrome requires an extension to share your screen, we will need to install the temasys chrome plugin "here":https://chrome.google.com/webstore/detail/skylink-WebRTC-tools/ljckddiekopnnjoeaiofddfhgnbdoafc
+p(tip). Note: Since Chrome requires an extension to share your screen, we will need to install the "temasys chrome plugin":https://temasys.io
```
tbedford, line 342: > Note: Since Chrome requires an extension to share your screen, we will need to install the "temasys chrome plugin":https://temasys.io

Minor point on style. We try to avoid "we" as then who has the responsibility to install the plugin - the reader or Ably? Sometimes it's not obvious. 

So in this case you can write "you need to install..."

tbedford (changes requested):
Very minor style point, otherwise looks good.

content/tutorials/web-rtc-screen-sharing.textile
stichbury: OK, I've fixed this. The text is scattered with "we will" because it's come from an external author (I think) but at least this one is resolved and probably one of the most confusing if left unaddressed.

Of course, it's still confusing because there's nothing useful behind the link, but I can't really help that. Sigh. Thanks @tbedford

tbedford (approved):
LGTM

stichbury:
Thanks @tbedford -- as this is my first PR, please can I just check -- am I OK to merge with one review? Do we typically squash & merge?

tbedford:
@stichbury - if you're happy you can merge. We normally just use "merge pull request".

PR #1509, opened 2022-07-27
Fix broken link to client options section of API docs

stichbury:
## Description

* [Jira epic ticket](https://ably.atlassian.net/browse/EDU-608) - if applicable.
* [Jira ticket](https://ably.atlassian.net/browse/EDU-868).

## Review

Replaced broken link with new link into API docs

* [Page to review](https://ably-docs-edu-868-fix-b-zdojju.herokuapp.com/rest/usage/)

### DIFF
--- a/content/rest/usage.textile
+++ b/content/rest/usage.textile
@@ -188,7 +188,7 @@ blang[flutter].
 
 "View our client library SDKs feature support matrix":https://ably.com/download/sdk-feature-support-matrix to see the list of all the available features.
 
-The supported "client options are described below":#client-options.
+The supported "client options are described below":/api/rest-sdk/types#client-options.
 
 h2(#tutorials). Tutorials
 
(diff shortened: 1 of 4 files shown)


stichbury:
Query for the reviewer ==> this page has several versions -- I updated the `usage.textile` file but not the equivalent under `/versions` 

SHOULD I HAVE UPDATED THEM ALL?

content/rest/usage.textile
```diff
@@ -188,7 +188,7 @@ blang[flutter].
 
 "View our client library SDKs feature support matrix":https://ably.com/download/sdk-feature-support-matrix to see the list of all the available features.
 
-The supported "client options are described below":#client-options.
+The supported "client options are described below":/api/rest-sdk/types#client-options.
```
tbedford, line 191: I would change the link text. It is true that the client options used to be below, but we split them off so that is no longer true. Perhaps something like: 

The supported client options are described in the API reference documentation.

tbedford (changes requested):
The link works fine. 

Yes, because the API references were split off for all versions, you need to make the change for all versions. Thanks.

content/rest/usage.textile
stichbury: Sorry @tbedford I don't know how I missed that. I've changed it now. Also fixed in the other versions.

tbedford (approved):
Checked link for all versions in review app.

PR #1520, opened 2022-08-04
Update Ably queues diagram

tbedford:
## Description

Update the Ably Queues overview diagram.

* Epic [EDU-719](https://ably.atlassian.net/browse/EDU-719)
* [EDU-929](https://ably.atlassian.net/browse/EDU-929)

## Review

View the updated diagram: 

* [Ably Queues](https://ably-docs-edu-929-updat-0oaksm.herokuapp.com/general/queues/)


m-hulbert (commented):
Minor point... but do you think we should use something other than a van sending lat/long in the example. This has some overlap and ability to cause confusion now that we offer AAT I'd have thought?

marklewin (commented):
I like this, but would suggest putting the AAT stuff in a separate "note"-style callout. A quick read of this as-is might make it look like you _need_ AAT to do this? What we should perhaps be saying is that this is how you would build such a solution from scratch, but if you want all the mapping, etc out of the box, consider AAT.

tbedford:
Yes, I don't want to muddy the waters, and you're right, it may imply it's a requirement (it's not). I will have a rethink.

content/general/queues.textile
```diff
@@ -46,6 +46,8 @@ Ably combines both pub/sub and queueing functionality in its "platform":https://
   <img src="/images/diagrams/ably-queues-overview.png" style="width: 100%" alt="Ably Queues overview">
 </a>
 
+**Note:** It's not a requirement for this scenario, but Ably has the "Asset Tracking SDK":/asset-tracking, which can provide a complete and convenient asset tracking solution. You can read more about what's provided in the "docs":/asset-tracking.
```
m-hulbert, line 49: Not sure you need the second sentence here?

tbedford: I've changed it a bit. I decided it was worth guiding to both the product page and the docs. The product page also has a nice big box linking to the docs too.

marklewin (approved):
LGTM

m-hulbert (approved):

PR #1522, opened 2022-08-05
Edx 124 generate random channel names

MymmiJ:

## Description

Replaces {{RANDOM_CHANNEL_NAME}} tokens with a random channel name; memoized so it's the same per page loaded.

Also some small refactors to the Code.js block.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-124).

## Review

Visit pages, you should see a random channel name each time you reload the page e.g.:

```
var realtime = new Ably.Realtime({
  key: '*********************************************************',
  clientId: 'bob' }
);
var channel = realtime.channels.get('can-guy-pea');
channel.presence.subscribe('enter', function(member) {
  alert('Member ' + member.clientId + ' entered');
});
channel.presence.enter();
```

The API Key text will vary based on whether the app has access to an API key or not. You can fake an API key with:

* GATSBY_DOCS_SIGNED_IN - set to any string to force the application to behave as though you are logged in
* GATSBY_DOCS_API_KEYS - set to any string to force the application to behave as though you have API keys available

When you press the 'Copy' button under these conditions, you should see (for example):

```
var realtime = new Ably.Realtime({
  key: 'dev_api_key',
  clientId: 'bob' }
);
var channel = realtime.channels.get('can-guy-pea');
channel.presence.subscribe('enter', function(member) {
  alert('Member ' + member.clientId + ' entered');
});
channel.presence.enter();
```

* [Page to review](http://localhost:8000/docs/realtime/presence)


src/components/blocks/software/Code/index.js
```diff
@@ -76,20 +88,28 @@ const Code = ({ data, attribs }) => {
    */
   const contentWithObfuscatedKey = useMemo(
     () =>
-      content.replace && content.replace(/{{API_KEY}}/g, '*********************************************************'),
-    [content, activeApiKey],
+      dataContainsKey
+        ? contentWithRandomChannelName.replace(
+            /{{API_KEY}}/g,
+            '*********************************************************',
```
DuncanEdwards, line 94: `Array(n).join("*")`

src/components/blocks/software/Code/MultilineCodeContent.tsx
```diff
@@ -0,0 +1,33 @@
...
+  displayLanguage,
+}: {
+  signedIn: boolean;
+  dataContainsKey: boolean;
+  contentWithObfuscatedKey: string;
+  contentWithKey: string;
+  content: string;
+  displayLanguage: string;
+}) => {
+  const contentMaybeWithKey = dataContainsKey
```
DuncanEdwards, line 19: Refactor into a method and maybe the first one could be an non linline if statement. Multiple nested inlines are a bit much and hurt my head!

DuncanEdwards (changes requested):
I was expected stars here?
image
instead of `loaidng API key`?

Other than that only minor refactors.

MymmiJ:
> *

Hmm if you set those env variables then it's possibly because of how you run the app, I've noticed that npm runs its own terminal session so doesn't always have access to your env variables

src/components/blocks/software/Code/index.js
```diff
@@ -89,10 +89,7 @@ const Code = ({ data, attribs }) => {
   const contentWithObfuscatedKey = useMemo(
     () =>
       dataContainsKey
-        ? contentWithRandomChannelName.replace(
-            /{{API_KEY}}/g,
-            '*********************************************************',
-          )
+        ? contentWithRandomChannelName.replace(/{{API_KEY}}/g, new Array(57).join('*'))
```
DuncanEdwards, line 92: Sorry, last one, that 57 needs to be a constant. Seems like an arbitrary number, so good luck naming that!

DuncanEdwards (changes requested):
Sorry! Looks good but one more TINY change

DuncanEdwards (approved):

PR #1526, opened 2022-08-08
Spec: implement first draft of protocol v2.0: no-connection-serial

SimonWoolf:
As described in [ADR63](https://ably.atlassian.net/wiki/spaces/ENG/pages/2238579598/ADR63+No+More+connection+Time+Serials)

NB: this is the first draft of a new breaking spec version. It describes a significant change in the realtime protocol, and cannot be implemented piecemeal. I've made the base branch `integration/protocol-2.0` per Cf https://github.com/ably/docs/blob/main/README_INTERNAL.md#branch-and-tag-scheme-for-features-spec .

It needed fewer changes than I was expecting.


content/client-lib-development-guide/features.textile
```diff
@@ -503,8 +501,10 @@ h3(#realtime-connection). Connection
...
-** @(RTN16b)@ @Connection#recoveryKey@ is an attribute composed of the @connectionKey@, and the latest @connectionSerial@ received on the connection, and the current @msgSerial@
-** @(RTN16c)@ @Connection#recoveryKey@ becomes @Null@ when a connection is explicitly @CLOSED@ or @CLOSED@ by the server, as connection state is not retained for connections closed intentionally. The @Connection#key@ and @Connection#id@ is set to @Null@
+** @(RTN16b)@ This clause has been removed as of the 2.0 spec and replaced with @RTN16g@.
+** @(RTN16g)@ @Connection#getRecoveryKey@ is function that returns a string which incorporates the @connectionKey@, the current @msgSerial@, and a list of pairs of channel @name@ and current @channelSerial@ for every currently attached channel. This must be serialized in a way which is able to encode any unicode channel name. The SDK may assume that the recovery key will only be consumed by the same type of SDK, so this spec does not specify any particular serialization; however, the format should be forward-compatible through the same major version of the SDK.
```
paddybyers, line 505: suggest replace "list of pairs" with language that is less implementation-specific, eg "set" or "collection"

content/client-lib-development-guide/features.textile
```diff
@@ -694,6 +696,7 @@ h3(#realtime-presence). RealtimePresence
 * @(RTP4)@ Ensure a test exists that enters 250 members using @RealtimePresence#enterClient@ on a single connection, and checks for @PRESENT@ events to be emitted on another connection for each member, and once sync is complete, all 250 members should be present in a @RealtimePresence#get@ request
 * @(RTP5)@ RealtimeChannel state change side effects:
 ** @(RTP5a)@ If the channel enters the @DETACHED@ or @FAILED@ state then all queued presence messages will fail immediately, and the @PresenceMap@ and "internal PresenceMap (see RTP17)":#RTP17 is cleared. The latter ensures members are not automatically re-entered if the @RealtimeChannel@ later becomes attached. Since channels in the @DETACHED@ and @FAILED@ states will not receive any presence updates from Ably, presence events (specifically @LEAVE@) should not be emitted when the @PresenceMap@ is cleared as each presence member's state is unknown
+*** @(RTP5a1)@ A channel entering the @DETACHED@ or @FAILED@ state should also clear any @channelSerial@ it has stored per @RTL25@.
```
paddybyers, line 699: Not also `suspended` ?

SimonWoolf: should have included suspended, thanks

content/client-lib-development-guide/features.textile
SimonWoolf: gone with 'collection'

SimonWoolf:
> I don't feel that mutating features spec points is a good idea, because we don't have any concept of versioning of particular spec points downstream. In this case an example is the change to RTN16c.

Sure, I do appreciate that in general - there's multiple spec points I removed rather than mutating. RTN16c I kept just because I felt like the substance of it wasn't changed, we've just renamed the thing it's about. So before it was `Connection#recoveryKey becomes Null when ...`, and now it's `Connection#getRecoveryKey() returns Null when ...`. But if you implement the other spec item of renaming `recoveryKey` to `getRecoveryKey()` then RTN16c should already be satisfied now if it was before, no?

But if you disagree I'm happy to make that change.

> Also, the addition of RTN16g1 feels like it's starting to lean into a territory that should be outside of scope for the features spec. ... My preference going forwards is that the spec describes the behaviour that's expected, without necessarily needing to go into specifics of API design.

I'm a bit confused here -- afaics the spec is mostly about describing the API design, I don't really understand what it would look like if it didn't do that? I understand you're planning on splitting off the IDL, but the [new IDL](https://github.com/ably/canonical-api-reference-prototyping/blob/main/descriptions.md) also explicitly distinguishes properties from functions just like the old one..?

I do appreciate that the "SDKs which choose not to increment their public major version when implementing protocol v2.0 may continue to implement this as ..." instruction may well not be appropriate for the spec, and I'd appreciate your advice on what to do here. I figured that without RTN16g1, the spec is basically mandating an change to the user-visible API, forcing a major version bump even if one would not otherwise be necessary, so I was trying to walk a line that didn't force that -- would be happy to have guidance on the right line to walk here.

QuintinWillison:
Thanks, @SimonWoolf. 😁 

In response to points raised in [your previous comment](https://github.com/ably/docs/pull/1526#issuecomment-1234555337)...

> **RTN16c I kept just because I felt like the substance of it wasn't changed, we've just renamed the thing it's about.**

I'll take another look as a review the PR in closer detail. However, my gut still says that even a redirection (as result of the rename) is change enough to warrant a new spec point reference. 🤔 

> **afaics the spec is mostly about describing the API design**

There is a lot of opinionated API design built into the features spec currently, I agree. My preference going forwards is that we veer away from mandating API design choices that could conflict with language or platform idiomatic alternatives. I tried to capture that in [this internal comment](https://ably.atlassian.net/wiki/spaces/ENG/pages/2278817920/ADR66%3A+New+home+for+the+features+specification?focusedCommentId=2283831369), replying to @Srushtika, on ADR66 (_New home for the features specification_):

> _Going forwards my ideal situation is that we can write feature spec points in such a way as to be less prescriptive when it comes to user-facing API, to allow for platform/language-specific variance to be decided by SDK engineers for their particular SDK. There’s also the potential for us to replace existing spec points with ones that also follow that principle. So, by keeping concerns discrete, I feel we keep the wriggle room open to do that in due course._

In respect of:

> **I figured that without RTN16g1, the spec is basically mandating an change to the user-visible API, forcing a major version bump even if one would not otherwise be necessary, so I was trying to walk a line that didn't force that**

Again, I need to look closer at this change, however more generally there might be a softer change where we can move the SDK as a `minor` version bump by deprecating the older API and introducing the newer API as a 'new feature'. That assumes there's a way for the older API to have some kind of compatibility bridge within the implementation in order to delegate responsibility to the newer API internally in a meaningful way. That may not be the case here (TBC, once I review this in depth).

SimonWoolf:
@QuintinWillison I've replaced RTN16c: https://github.com/ably/docs/pull/1526/commits/a1e7f06e4dcab8cd8c08e1d7f44bdf6be0b18356

I'll leave RTN16g1 until you take a closer look -- I'm happy to write pretty much what you suggest, but that seems like it'd unavoidably mean without getting into the specifics of what a client lib should do until & after it increments its major sdk version, which it seems like you're saying you want to avoid?

@andydunstall could you review https://github.com/ably/docs/pull/1526/commits/8179de3ec155bcf9bcb6b424b456d0a6038cbbfc for correctness?

SimonWoolf:
@sacOO7 [pointed out](https://github.com/ably/ably-js/pull/1045#discussion_r968143385) that the current spec doesn't actually explicitly say how connection recovery works. It has some info around the edges, but is missing the central bits, like that the recover param should be set to the connectionKey. This seems like a good opportunity to fix that, therefore: https://github.com/ably/docs/pull/1526/commits/2a932d4c9e19a9855060994bc882cc27cf23b9a7

Also moved the new channelserial state to channel.properties as I'd forgotten that existed: https://github.com/ably/docs/pull/1526/commits/5786bfe5abab93081c70f53e2e404f8f3c764a34

QuintinWillison:
This pull request shall remain open in this repository (`ably/docs`) for visibility only, until the features spec has moved over to `ably/specification` (see https://github.com/ably/specification/issues/1). At that point, it will be re-formed in that repository, targeting a v2 integration branch there.

content/client-lib-development-guide/features.textile
```diff
@@ -686,14 +695,15 @@ h3(#realtime-presence). RealtimePresence
 * @(RTP17)@ The RealtimePresence object should also keep a second @PresenceMap@ containing only members that match the current @connectionId@. Any incoming presence message that satisfies @RTP17b@ should be applied to this object in the same way as for the normal @PresenceMap@. This object should be private and is used to maintain a list of members that need to be automatically re-entered by the @RealtimePresence@ object when required to by @RTP17c@.
```
ikbalkaya, line 695: Remaining reference to old spec point
```suggestion
* @(RTP17)@ The RealtimePresence object should also keep a second @PresenceMap@ containing only members that match the current @connectionId@. Any incoming presence message that satisfies @RTP17b@ should be applied to this object in the same way as for the normal @PresenceMap@. This object should be private and is used to maintain a list of members that need to be automatically re-entered by the @RealtimePresence@ object when required to by @RTP17f@.
```

content/client-lib-development-guide/features.textile
```diff
@@ -485,29 +483,38 @@ h3(#realtime-connection). Connection
...
+**** @(RTN15c6)@ A @CONNECTED@ @ProtocolMessage@ with the same @connectionId@ as the current client (and no @error@ property). This indicates that the resume attempt was valid. The client library should move all channels that were in the @ATTACHING@, @ATTACHED@, or @SUSPENDED@ states to the @ATTACHING@ state, and initiate an @RTL4c@ attach sequence for each. The connection should also process any messages queued per @RTLc6@ (there is no need to wait for the attaches to finish before processing queued messages).
```
ikbalkaya, line 491: I think this should be RTL6c
```suggestion
**** @(RTN15c6)@ A @CONNECTED@ @ProtocolMessage@ with the same @connectionId@ as the current client (and no @error@ property). This indicates that the resume attempt was valid. The client library should move all channels that were in the @ATTACHING@, @ATTACHED@, or @SUSPENDED@ states to the @ATTACHING@ state, and initiate an @RTL4c@ attach sequence for each. The connection should also process any messages queued per @RTL6c@ (there is no need to wait for the attaches to finish before processing queued messages).
```

content/client-lib-development-guide/features.textile
```diff
@@ -485,29 +483,38 @@ h3(#realtime-connection). Connection
...
+**** @(RTN15c7)@ @CONNECTED@ @ProtocolMessage@ with a new @connectionId@ and an @ErrorInfo@ in the @error@ field. In this case, the resume was invalid, and the error indicates the cause. The @error@ should be set as the @reason@ in the @CONNECTED@ event, and as the @Connection#errorReason@. The internal @msgSerial@ counter should be reset so that the first message published to Ably will contain a @msgSerial@ of @0@. Otherwise the process is the same as per @RTN16c6@: The client library should move all channels that were in the @ATTACHING@, @ATTACHED@, or @SUSPENDED@ states to the @ATTACHING@ state, and initiate an @RTL4c@ attach sequence for each. The connection should also process any messages queued per @RTLc6@.
```
ikbalkaya, line 492: And this
```suggestion
**** @(RTN15c7)@ @CONNECTED@ @ProtocolMessage@ with a new @connectionId@ and an @ErrorInfo@ in the @error@ field. In this case, the resume was invalid, and the error indicates the cause. The @error@ should be set as the @reason@ in the @CONNECTED@ event, and as the @Connection#errorReason@. The internal @msgSerial@ counter should be reset so that the first message published to Ably will contain a @msgSerial@ of @0@. Otherwise the process is the same as per @RTN16c6@: The client library should move all channels that were in the @ATTACHING@, @ATTACHED@, or @SUSPENDED@ states to the @ATTACHING@ state, and initiate an @RTL4c@ attach sequence for each. The connection should also process any messages queued per @RTL6c@.
```

content/client-lib-development-guide/features.textile
sacOO7, line 491: Can we mention `serverside backlog messages are available and will be received on every freshly attached channel`. Because we are adding explicit integration test for the same.

content/client-lib-development-guide/features.textile
```diff
@@ -683,17 +692,18 @@ h3(#realtime-presence). RealtimePresence
...
-** @(RTP17c)@ The RealtimePresence object should perform automatic re-entry in the following situations:
-*** @(RTP17c1)@ After a @SYNC@ operation has completed, per "@RTP18b@":#RTP18b
-*** @(RTP17c2)@ When an @ATTACHED@ message is received with no "@HAS_PRESENCE@":#TR3a flag (so no @SYNC@ is expected as the server does not believe anyone is currently present)
-** @(RTP17d)@ Automatic re-entry consists of, for each member of the @RTP17@ internal @PresenceMap@ whose @memberKey@ is not also a member of the normal @PresenceMap@, publishing a @PresenceMessage@ with an @ENTER@ action using the @clientId@ and @data@ attributes from that member, and removing that member from the internal @PresenceMap@
+** @(RTP17f)@ The RealtimePresence object should perform automatic re-entry whenever a channel moves into the @ATTACHED@ state. (It does not need to do so for @RTL12@ @ATTACHED@ events received on already-@ATTACHED@ channels).
+** @(RTP17g)@ Automatic re-entry consists of, for each member of the @RTP17@ internal @PresenceMap@, publishing a @PresenceMessage@ with an @ENTER@ action using the @clientId@, @data@, and @id@ attributes from that member.
```
sacOO7, line 699: Don't we need to remove member from internal presence map once `presence` event is published?

sacOO7: Till now, for every `ENTER` presence action, we were passing only `clientId` and `data`. Are we supposed to pass `id` for every other presence `ENTER` now?

content/client-lib-development-guide/features.textile
```diff
@@ -485,29 +483,38 @@ h3(#realtime-connection). Connection
...
+** @(RTN16j)@ When a library is instantiated with the @recover@ client option, for every channel/channelSerial pair in the @recoveryKey@, it should instantiate a corresponding channel and set its "RTL15b":#RTL15b @channelSerial@.
+** @(RTN16k)@ When the library first connects to Ably after being instantiated with a @recover@ client option, it should add an additional @recover@ querystring param to the websocket request, set from the @connectionKey@ component of the @recoveryKey@. Once the library has successfully connected to Ably, it should never again supply a @recover@ querystring param.
+** @(RTN16g)@ @Connection#getRecoveryKey@ is function that returns a string which incorporates the @connectionKey@, the current @msgSerial@, and a collection of pairs of channel @name@ and current @channelSerial@ for every currently attached channel. This must be serialized in a way which is able to encode any unicode channel name. The SDK may assume that the recovery key will only be consumed by the same type of SDK, so this spec does not specify any particular serialization; however, the format should be forward-compatible through the same major version of the SDK.
```
sacOO7, line 509: Can we add, if decoding fails when `recover` value is supplied externally, it should raise some error? Currenly, we silently catch the error and then return null for `recover` value.

content/client-lib-development-guide/features.textile
```diff
@@ -485,29 +483,38 @@ h3(#realtime-connection). Connection
...
-*** @(RTN15g3)@ When a connection attempt succeeds after the connection state has been cleared in this way, channels that were previously @ATTACHED@, @ATTACHING@, or @SUSPENDED@ must be automatically reattached, just as if the connection was a resume attempt which failed per "RTN15c3":#RTN15c3
-** @(RTN15b)@ In order for a connection to be resumed and connection state to be recovered, the client must have received a @CONNECTED@ ProtocolMessage which will include a private connection key. To resume that connection, the library reconnects to the "websocket":https://ably.com/topic/websockets endpoint with two additional querystring params:
+*** @(RTN15g3)@ When a connection attempt succeeds after the connection state has been cleared in this way, channels that were previously @ATTACHED@, @ATTACHING@, or @SUSPENDED@ must be automatically reattached, just as if the connection was a resume attempt which failed per "RTN15c7":#RTN15c7
```
sacOO7, line 486: Do we need even this spec point, it says the same thing as `RTN15c7`. Feels more like an incomplete duplication, and confusing at the same time. We are saying the connection attempt succeeds yet the resume fails. I mean it doesn't cover a case for successful resume.

content/client-lib-development-guide/features.textile
```diff
@@ -485,29 +483,38 @@ h3(#realtime-connection). Connection
...
+*** @(RTN16g1)@ SDKs which choose not to increment their public major version when implementing protocol v2.0 may continue to implement this as a @Connection#recoveryKey@ attribute until the next major version change, using a getter function that's backwards-compatible with attribute access where possible, else by recomputing the recovery key string on every message.
+** @(RTN16h)@ @Connection#getRecoveryKey@ should return @Null@ when a the SDK is in the @CLOSED@, @CLOSING@, @FAILED@, or @SUSPENDED@ states, or when it does not have a @connectionKey@ (for example, it has not yet become connected). The @Connection#key@ and @Connection#id@ should also be @Null@ at these times.
```
sacOO7, line 511: Do we need to check if `connectionId` is null too? Currently I am only checking if `connectionKey` is null

sacOO7: I am currently only checking 
```
                if (Key.IsEmpty() || InnerState.State == Realtime.ConnectionState.Closing
                                  || InnerState.State == Realtime.ConnectionState.Closed
                                  || InnerState.State == Realtime.ConnectionState.Failed
                                  || InnerState.State == Realtime.ConnectionState.Suspended)
                {
                    return null;
                }
```
While in cocoa, we also check if `id` should be empty

content/client-lib-development-guide/features.textile
SimonWoolf: > I mean it doesn't cover a case for a successful resume
A successful resume when the library is not attempting to resume is not a possible case
> We are saying the connection attempt succeeds yet the resume fails
the resume did not 'fail', in the rtn15g case there is no attempt at a resume.

I don't think I agree that it's duplicative. RTN15c7 is a resume attempt which failed. RTN15g has the library not attempt a resume, there is nothing to 'fail'. It's saying that in this case, channels should also be reattached.

content/client-lib-development-guide/features.textile
SimonWoolf: No, getting a CONNECTED with no error does not guarantee that the resume on each channel will succeed. That's kindof the point of this new resume model: it's now handled independently by each channel, the resume for each channel will succeed iff the channelSerial you send maps to something still in memory, irrespective of what the connection is doing.

The connection resume now pretty much just means you can keep using the same connectionId, which is really only relevant for presence members.

content/client-lib-development-guide/features.textile
SimonWoolf: > Don't we need to remove member from internal presence map once presence event is published?

No. That spec item was RTP17d, it was removed for 2.0.

> Till now, for every ENTER presence action, we were passing only clientId and data. Are we supposed to pass id for every other presence ENTER now?

When you send an ENTER becuase the use called presence.enter(), you don't add an id because there isn't one. This spec item is about _re-entry_ specifically; where you do have an id because the internal my-members map is populated by presence messages from the server, which will have ids.

content/client-lib-development-guide/features.textile
SimonWoolf: That looks fine to me

SimonWoolf:
https://github.com/ably/docs/pull/1526/commits/988218b4402baead4231697a967e54581f7ba92e : I have removed RTN15f entirely (it was mostly redundent to RTN19a, and the bit that wasn't redundent was wrong as it implied you should only republish if a resume was successful, which is no longer correct given transient publishing), and instead added a couple small clarifications to RTN19a.

content/client-lib-development-guide/features.textile
sacOO7: You mean I should also check for `id` ?

SimonWoolf: no, what you have there already looks fine

content/client-lib-development-guide/features.textile
sacOO7: As I mentioned currently we don't throw error, rather silently catch it and return `null`

content/client-lib-development-guide/features.textile
sacOO7: `RTN15g3` is talking about resume connection attempts right?

content/client-lib-development-guide/features.textile
sacOO7: Sure, thanks : )

content/client-lib-development-guide/features.textile
```diff
@@ -683,17 +694,18 @@ h3(#realtime-presence). RealtimePresence
...
-*** @(RTP17c1)@ After a @SYNC@ operation has completed, per "@RTP18b@":#RTP18b
-*** @(RTP17c2)@ When an @ATTACHED@ message is received with no "@HAS_PRESENCE@":#TR3a flag (so no @SYNC@ is expected as the server does not believe anyone is currently present)
-** @(RTP17d)@ Automatic re-entry consists of, for each member of the @RTP17@ internal @PresenceMap@ whose @memberKey@ is not also a member of the normal @PresenceMap@, publishing a @PresenceMessage@ with an @ENTER@ action using the @clientId@ and @data@ attributes from that member, and removing that member from the internal @PresenceMap@
+** @(RTP17f)@ The RealtimePresence object should perform automatic re-entry whenever a channel moves into the @ATTACHED@ state. (It does not need to do so for @RTL12@ @ATTACHED@ events received on already-@ATTACHED@ channels).
+** @(RTP17g)@ Automatic re-entry consists of, for each member of the @RTP17@ internal @PresenceMap@, publishing a @PresenceMessage@ with an @ENTER@ action using the @clientId@, @data@, and @id@ attributes from that member.
+** @(RTP17c)@ This clause and its subclauses have been replaced by @RTN17f@ as of the 2.0 spec.
```
sacOO7, line 702: This is supposed to be `RTP17f` instead of `RTN17f`

content/client-lib-development-guide/features.textile
sacOO7: I added a warning log when decoding fails and then it returns null, and then it won't set `recover` param for null value

QuintinWillison:
This change proposed in this pull request has been [moved](https://github.com/ably/docs/pull/1526#ref-issue-1393766551) to the new home of the features spec, in [the `ably/specification` repository](https://github.com/ably/specification).

content/client-lib-development-guide/features.textile
QuintinWillison: @sacOO7 If this conversation needs to continue, then that will need to happen against the new pull request that @SimonWoolf will open against https://github.com/ably/specification/issues/8. Please resolve this conversation _here_ once you have re-raised your concern _there_, or if you are already happy that this matter is resolved. Thanks.

content/client-lib-development-guide/features.textile
QuintinWillison: @sacOO7 If this conversation needs to continue, then that will need to happen against the new pull request that @SimonWoolf will open against https://github.com/ably/specification/issues/8. Please resolve this conversation _here_ once you have re-raised your concern _there_, or if you are already happy that this matter is resolved. Thanks.

content/client-lib-development-guide/features.textile
QuintinWillison: @sacOO7 If this conversation needs to continue, then that will need to happen against the new pull request that @SimonWoolf will open against https://github.com/ably/specification/issues/8. Please resolve this conversation _here_ once you have re-raised your concern _there_, or if you are already happy that this matter is resolved. Thanks.

content/client-lib-development-guide/features.textile
QuintinWillison: @sacOO7 If this conversation needs to continue, then that will need to happen against the new pull request that @SimonWoolf will open against https://github.com/ably/specification/issues/8. Please resolve this conversation _here_ once you have re-raised your concern _there_, or if you are already happy that this matter is resolved. Thanks.

content/client-lib-development-guide/features.textile
QuintinWillison: @sacOO7 If this conversation needs to continue, then that will need to happen against the new pull request that @SimonWoolf will open against https://github.com/ably/specification/issues/8. Please resolve this conversation _here_ once you have re-raised your concern _there_, or if you are already happy that this matter is resolved. Thanks.

content/client-lib-development-guide/features.textile
QuintinWillison: @sacOO7 If this conversation needs to continue, then that will need to happen against the new pull request that @SimonWoolf will open against https://github.com/ably/specification/issues/8. Please resolve this conversation _here_ once you have re-raised your concern _there_, or if you are already happy that this matter is resolved. Thanks.

content/client-lib-development-guide/features.textile
```diff
@@ -54,7 +54,7 @@ h2(#test-guidelines). Test guidelines
...
-* @(G4)@ This spec defines API version 1.2. A client library must identify to Ably the version of the spec it uses in all requests and connections, per "RSC7a":#RSC7a and "RTN2f":#RTN2f. The spec it uses is defined as the latest API version for which the library implements all spec items relating to the wire protocol
+* @(G4)@ This spec defines API version 2.0. A client library must identify to Ably the version of the spec it uses in all requests and connections, per "RSC7a":#RSC7a and "RTN2f":#RTN2f. The spec it uses is defined as the latest API version for which the library implements all spec items relating to the wire protocol
```
sacOO7, line 57: can we highlight `2.0` as a string? Actually, mistakenly in dotnet code it was using `v=2` as a param and hence server was failing to recognize the correct protocol version.

SimonWoolf:
Ported to https://github.com/ably/specification/pull/88

content/client-lib-development-guide/features.textile
SimonWoolf: In https://github.com/ably/specification/pull/88 I've made it an inline code block to make it clearer that it's a literal string.

PR #1531, opened 2022-08-11
[EDX-241] - Styling fixes, refactor

MymmiJ:

## Description

Fixes in accordance with design feedback.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-241).

## Review

Ensure that:

1. Search bar behaves as expected/specified by Lenka - new minimum width and maximum width to be implemented
2. External link icons appear on Resources dropdown menu
3. Reinstate the active-nav styling
4. Apply consistent styling to the menu items from the base menu item component
5. No extra indent for menu items vs heading in resources section

* [Page to review](http://localhost:8000/docs/realtime/usage?lang=swift)


DuncanEdwards:
If I click on API reference is looks like this:
image
Is that right?

DuncanEdwards:
Is this right that the meganav doesn't span the whole width:
image

What determines the width of the meganav?

DuncanEdwards:
These 3 menu items need to be shifted 2 pixels down
image

DuncanEdwards:
This logo needs to be centred vertically:
image

DuncanEdwards:
The bottom padding on the meganav is too small and needs to be larger like the top:
image

MymmiJ:
> Is this right that the meganav doesn't span the whole width: image
> 
> What determines the width of the meganav?

No, sorry, I was pushing changes while you were reviewing; the <aside> that contains the dropdown meganav was not set to w-full after some other CSS changes (making it fixed).

DuncanEdwards:
Okay, sorry, I have to point out one more thing! So when this is light blue:
image
The menu is "toggled on" which means you can move away from it and it will stay open. (Check ably.com Resources menu for an example)

MymmiJ:
> Okay, sorry, I have to point out one more thing! So when this is light blue: image The menu is "toggled on" which means you can move away from it and it will stay open. (Check ably.com Resources menu for an example)

Nice, thanks for pointing this out.

I've solved this by:
* Moving dropdown menu items to be children of their corresponding dropdown button, so they can naturally be tabbed through when the component is active
* Focusing the dropdown button no longer expands the dropdownMenu, you have to click or press Enter to make the dropdownMenu active.
* Locking the dropdownMenu until it's cleared with a specific clearLocked method or replaced with another locked dropdowMenu
* Adding a hook which removes the dropdownMenu when a user clicks outside the entire TopMainNav.

Let me know what you think! I think it's nearly there now :D

I've also pushed some fixes to the AccountPanel that will show up when logged in.

DuncanEdwards:
Looking better but the highlight doesn't work right with non keyboard usage:
image

This is how it should highlight:
image
I guess key differences are:
- It highlights both lines
- It has more padding

DuncanEdwards (changes requested):
We're getting closer!

MymmiJ:
> We're getting closer!

Changes requested have been done, should be ready to merge I think now?

DuncanEdwards:
I hate to be that person but that button left + right margin (8px) needs to be padding:
image
Else the menu kind of flickers around as you move the mouse across.

MymmiJ:
> I hate to be that person but that button left + right margin (8px) needs to be padding: image Else the menu kind of flickers around as you move the mouse across.

Nice one, have sorted this. mx-0 and px-12

DuncanEdwards (approved):
Yes! Finally! Woop woop!

PR #1540, opened 2022-08-16
[WEB-2519] Prevent iframe embedding

denissellu:

## Description

Switch to the nginx buildpack and ensure we serve up `X-Frame-Options` headers to prevent docs from being embedded into iframes.

* [WEB-2519](https://ably.atlassian.net/browse/WEB-2519).

## Review

The review app should work as normal with no visible changes to the user. Embedding docs in an iframe should not be possible anymore.

* [Page to review](link)

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: bin/start-nginx-solo
--- a/app.json
+++ b/app.json
@@ -15,7 +15,7 @@
       "url": "https://github.com/ably-forks/heroku-buildpack-nanoc.git"
     },
     {
-      "url": "https://github.com/heroku/heroku-buildpack-static"
+      "url": "https://github.com/heroku/heroku-buildpack-nginx"
     }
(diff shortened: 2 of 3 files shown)


kennethkalmer:
@denissellu I've reworked the PR a little. The main issue was the conflicting buildpacks, so your nginx file was never used. By switching from the deprecated `heroku-buildpack-static` to `heroku-buildpack-nginx` things seem to work. Can you confirm with your examples?

And when we're ready to merge we first need to update the buildpack configuration in production or we'll break things :)

@tbedford @m-hulbert the changes in this PR would not affect you in any way, it is a security improvement for the deployed docs. Please let me know if you have any questions for us though!

denissellu:
thanks @kennethkalmer, I completely missed that, it's all working now, here is the link of it https://jsbin.ably.com/egihos/1/edit

PR #1545, opened 2022-08-18
EDX-231: Create sitemaps for docs

snikidev:

## Description

Custom config of a gatsby sitemap plugin to include only relevant `docs` files/pages

* [Jira ticket](https://ably.atlassian.net/browse/EDX-231).

## Review

The main `sitemap.xml` file is generated inside the `public/docs` folder on a build that references the` sitemap-0.xml` file, which, in essence, is the actual sitemap.

### DIFF
--- a/gatsby-config.ts
+++ b/gatsby-config.ts
@@ -9,12 +9,41 @@ export const plugins = [
   'gatsby-plugin-postcss',
   'gatsby-plugin-styled-components',
   'gatsby-plugin-image',
-  'gatsby-plugin-sitemap',
   'gatsby-plugin-sharp',
   'gatsby-transformer-yaml',
   'gatsby-transformer-sharp',
   'gatsby-plugin-react-helmet',
   'gatsby-plugin-ts',
+  {
+    resolve: 'gatsby-plugin-sitemap',
+    options: {
+      resolveSiteUrl: () => siteMetadata.siteUrl,
+      excludes: ['/_*'],
+      output: '/docs',
+      filterPages: ({ path }) => {
+        // NOTE: these didn't work in `exclude` field, so we had to use custom regex in filterPage function
+        if (
+          // Legacy versions of pages are excluded from the sitemap (anything with a URL containing .../versions/vx.x/where x is at least one digit, e.g. v1.1, v1.2
+          /\/\b(versions)\b\/(v)[0-9]./.test(path) ||
+          // Urls with /docs/code-
+          /\/\b(docs)\/\b(code-)/.test(path) ||
+          // anything with a URL beginning with /docs/tutorials
+          /\/\b(docs)\b\/(tutorials)\//.test(path) ||
+          // /documentation/
+          /\/\b(documentation)\/$/.test(path) ||
+          // Exclude root domain url
+          path === '/'
+        )
+          return true;
+
+        return false;
+      },
+      serialize: ({ path }) => ({
+        url: path,
+        changefreq: 'monthly',
+      }),
+    },
+  },
   // Images
   {
     resolve: 'gatsby-source-filesystem',
--- a/gatsby-config.ts
+++ b/gatsby-config.ts
@@ -20,24 +20,18 @@ export const plugins = [
       resolveSiteUrl: () => siteMetadata.siteUrl,
       excludes: ['/_*'],
       output: '/docs',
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "develop": "export EDITOR_WARNINGS_OFF='true' && gatsby clean && gatsby develop",
     "start": "npm run develop",
     "edit": "unset EDITOR_WARNINGS_OFF && gatsby clean && gatsby develop",
-    "build": "gatsby build --prefix-paths",
+    "build": "npm run clean && gatsby build --prefix-paths",
```
DuncanEdwards, line 26: Yeah, I should have done this 4 months ago. While we're here, can you also add this to "start"?

gatsby-config.ts
```diff
@@ -9,12 +9,41 @@ export const plugins = [
...
   'gatsby-transformer-sharp',
   'gatsby-plugin-react-helmet',
   'gatsby-plugin-ts',
+  {
+    resolve: 'gatsby-plugin-sitemap',
+    options: {
+      resolveSiteUrl: () => siteMetadata.siteUrl,
+      excludes: ['/_*'],
+      output: '/docs',
+      filterPages: ({ path }) => {
```
DuncanEdwards, line 23: This whole thing can just be an inline return

gatsby-config.ts
```diff
@@ -9,12 +9,41 @@ export const plugins = [
...
+        // NOTE: these didn't work in `exclude` field, so we had to use custom regex in filterPage function
+        if (
+          // Legacy versions of pages are excluded from the sitemap (anything with a URL containing .../versions/vx.x/where x is at least one digit, e.g. v1.1, v1.2
```
DuncanEdwards, line 26: I love this, you are doing every single thing in your power to make this code readable for the next person. Sometimes i make developers turn these comments into methods but I feel like that would make this harder to understand and it is SO easy to understand now.

DuncanEdwards (changes requested):
Awesome! Just two tiny changes.

My rules are generally to use inlines where possible and if you are doing an if statement, to have "squigglies":
IE 
Not preferred
```
if (itIsTrue()) return true;
```

Preferred
```
if (itIsTrue()) {
  return true;
}
```

Actually I should look into updating the linting rule. ...

gatsby-node.ts
```diff
@@ -18,3 +19,11 @@ export const createSchemaCustomization: GatsbyNode['createSchemaCustomization']
   `;
   createTypes(typeDefs);
 };
+
+// Gatsby doesn't allow you to change sitemap filename, so we have to do it manually.
+// https://github.com/gatsbyjs/gatsby/issues/31515
+exports.onPostBuild = () => {
+  if (fs.existsSync('./public/docs/sitemap-index.xml')) {
+    fs.renameSync('./public/docs/sitemap-index.xml', './public/docs/sitemap.xml');
```
dpiatek, line 27: Hey @snikidev just FYI, but this is not necessary. Docs are served through the website Rails app, and we can map routes there (we already do for Voltaire sitemap, for example https://github.com/ably/website/blob/main/config/routes.rb#L665). 

In fact, this sitemap will not be accessible on ably.com without routing additions on the website.

DuncanEdwards: I kind of went through the same thought process but I actually quite like this as it is. If someone comes into the docs area searching for `sitemap.xml` they will find this line and understand exactly how this is setup. But yes, you are right we will need the Ruby route. Shall I create a ticket for it?

dpiatek: Yup, a ticket would be good, otherwise sitemap won't be accessible at all :) I'll leave you guys to decide what you want to do. For Voltaire we even skip the index https://github.com/ably/website/blob/main/app/services/ably/backends/cms.rb#L165-L171

gatsby-node.ts
```diff
@@ -1,3 +1,4 @@
+import fs from 'fs';
```
dpiatek, line 1: This is no longer needed I believe?

dpiatek (approved):

MymmiJ (approved):

gatsby-node.ts
snikidev: ah yes 🤦

PR #1548, opened 2022-08-19
[EDX-93] docs site mobile styles

MymmiJ:

## Description

To implement some mobile styling in line with the Figma doc here:

* [Jira ticket](https://ably.atlassian.net/browse/EDX-93).

## Review

Make the screen bigger and smaller.

Check that the designs match up with the lovely Figma illustrations

Ensure that the nested menus work as you might expect, with allowances for matching the designs above

* [Homepage](http://localhost:8000)
* [A page without language variants](http://localhost:8000/docs/core-features/history?lang=android)
* [A page with language variants](http://localhost:8000/docs/quick-start-guide?lang=android)


DuncanEdwards (changes requested):
- Ably logo needs to be moved DOWN 3 pixels
- Hamburger logo needs to be moved DOWN 2 pixels
- Search icon needs to show border on activation like the ably.com one does:
image

- The highlighting of menu item activation uses a blue border:
image
when it should use a blue highlight:
image

- The body is still scrollable when the menu is open and it shouldn't be. This scrollability needs to be on the menu so we can access deep down menu items:
image

MymmiJ:
> ...

I've implemented these changes, I've not changed the Dropdown Content highlighting behaviour within the DropdownItemContent item because that is already handled in the other PRs relating to style tweaks.

DuncanEdwards:
I hate to be "that guy" but now there are the following glitches:
Search is not centred within the box:
image

And the `API References` and `Resources` menu items are pushed up to high on the non-mobile menu.

PR #1551, opened 2022-08-22
EDX-136: Integrate add search for search functionality no UI/UX

snikidev:

## Description

Implementing search functionality to fetch results + test. A copy-paste from [website](https://github.com/ably/website/pull/4849)

* [Jira ticket](https://ably.atlassian.net/browse/EDX-136) 

## Review

No way to test it now but locally 🤷


src/components/Header/TopMainNav/SearchBar/SearchBar.tsx
```diff
@@ -1,10 +1,32 @@
...
+  const initialSearch = new URLSearchParams(location.search).get('q');
+
+  const [searchTerm, setSearchTerm] = useState(initialSearch?.toString());
   const textInput = useRef<null | HTMLInputElement>(null);
   const focusOnSearchInput = () => textInput.current && textInput.current.focus && textInput.current.focus();
+
+  const {
+    actions: { search },
+  } = useSearch({
+    addsearchApiKey: process.env.GATSBY_ADDSEARCH_API_KEY,
```
dpiatek, line 17: oh, does this now work on the frontend? (process.env is node runtime, not browser runtime so normally would not work?)

src/components/Header/TopMainNav/SearchBar/SearchBar.tsx
```diff
@@ -1,10 +1,32 @@
...
+import { useLocation } from '@reach/router';
+import { useSearch } from 'src/hooks';
 import { SearchIcon } from '.';
 
 export const SearchBar = () => {
-  const [searchTerm, setSearchTerm] = useState('');
+  const location = useLocation();
+  const initialSearch = new URLSearchParams(location.search).get('q');
+
+  const [searchTerm, setSearchTerm] = useState(initialSearch?.toString());
```
dpiatek, line 10: I don't think you need this, `useSearch` will return `state` which has `query`?

dpiatek (changes requested):

DuncanEdwards (commented):
Can you please fix the linting issues without changing the linting settings :-)

src/components/Header/TopMainNav/SearchBar/SearchBar.tsx
snikidev: no, it's fine. We have Gatsby. it does it's own magic of grabbing env variables and converting them through webpack. 👍

src/components/Header/TopMainNav/SearchBar/SearchBar.tsx
snikidev: We have a controlled component, so we need that for our input field. I'll have a look if that'll work. 🤔 👍

snikidev: Our `useSearch` hook doesn't return a `setSearch` function that would work with a `query` the same way as `useState` hook would, since it's a controlled component, I don't think we can go without `useState`.

I can do it a different way with a `useRef` tho. So we'll grab an input value with `inputRef.current.value` when `onKeyDown` and trigger `search()` function then, if you prefer. 🙂

dpiatek: hey @snikidev , I must admit, I don't follow :) perhaps we can get on a call?

To summerise what I mean - state is internally held by the hook, so `setSearch` -> `search` and `searchTerm` is `query`

dpiatek: There should be no need for `useRef` or `useState` here, the hook abstracts away the actions and state of a "search", just without providing a component.

src/hooks/use-search/use-search.ts
```diff
@@ -0,0 +1,247 @@
...
+  total_hits: number;
+  processing_time_ms: number;
+  hits: Hit[];
+  facets: null;
+  rangeFacets: null;
+};
+
+export type State = {
+  client: string | null;
+  query?: string;
```
dpiatek, line 41: query, by design, should be always a string (empty one by default) - was there code that made it not so?

src/hooks/use-search/use-search.ts
```diff
@@ -0,0 +1,247 @@
...
+  page?: Result['page'];
```
dpiatek, line 42: page should be always set, by default to `1` - was the compiler saying otherwise?

dpiatek (approved):

src/hooks/use-search/use-search.ts
snikidev: yep, my bad 👍

src/hooks/use-search/use-search.ts
snikidev: it does default to `1` but we don't always pass a page in our `search()` function so that's why it's optional. I just reused `State` type for our `search` action payload 
```
const searchAction = (dispatch: Dispatch<Action>) => (payload: Pick<State, 'query' | 'page'>) => dispatch({ type: 'search', payload });
```

I just didn't want to create another type, but I can if it's preferred. 👍 🙂

dpiatek: right, yeah that's fine. I was thinking of https://github.com/ably/website/pull/4849/files#diff-b580f33881698284e8a407553ae825a97c41f2f21a19b2699b548a8501c496b1R58. I think it's to keep types non-polymorpic where possible :)

src/hooks/use-search/use-search.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+      );
+
+      const searchMock = jest.fn((_, cb) => cb({ error: { message: 'Error' } }));
+      jest.spyOn(result.current.state.client, 'search').mockImplementation(searchMock);
+
+      act(() => {
+        result.current.actions.search({ query: 'kafka' });
+      });
+
+      expect(result.current.state.loading).toBe(false);
```
DuncanEdwards, line 127: Can this block of 4 expects be refactored into a helper method where the 4 expected values are passed into it.

DuncanEdwards (changes requested):
Minor refactor

PR #1564, opened 2022-09-01
[EDX-202] Render unwrapped root level text in p elements

MymmiJ:

## Description

Wraps root level text elements in `<p>` elements. A simple change, tracking down the best place and way to do this without affecting anything else was what took most time so I've added a line to the postParser folder.

**Unrelated Changes:**
- Removes `husky` githooks package from the repository, as it did not work particularly well.
- Adds optional new .githooks folder that you can activate or deactivate using scripts.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-202).

## Review

Run the branch locally, check:

* [Page to review](http://localhost:8000/docs/rest-api?lang=python)

F3 or Cmd+F to find 'HATEOS':

Screenshot 2022-09-01 at 16 49 14

Check that nothing's wrong, see how it doesn't display currently on the [staging site](https://ably-docs-next-staging.herokuapp.com/docs/rest-api?lang=csharp)

### DIFF
--- a/data/transform/post-parser/root-level-text-wrapper.ts
+++ b/data/transform/post-parser/root-level-text-wrapper.ts
@@ -1,1 +1,14 @@
-export const rootLevelTextWrapper = (content: string) => content.replace(/\n([A-Z].*)\n/g, '\n<p>$1</p>\n');
+/**
+ * This regex finds:
+ * (<\/[A-Za-z]+>)\n - closing HTML tag, followed by newline.
+ * ([A-Z].*)\n - a capital letter preceding any information, up to a newline.
+ *
+ * The overall effect is to find capitalised text that immediately follows on
+ * from a closing tag at the end of a line, which will always be plain text
+ * that is supposed to be wrapped in <p> elements but has been missed by the
+ * textile-js parser.
+ *
+ * Regex101: https://regex101.com/r/j4BLqX/1
+ */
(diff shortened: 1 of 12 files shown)


data/transform/post-parser/root-level-text-wrapper.ts
```diff
@@ -0,0 +1,2 @@
+export const rootLevelTextWrapper = (content: string) =>
+  content.replace(/(<\/[A-Za-z]+>)\n([A-Z].*)\n/g, '$1\n<p>$2</p>\n');
```
artismarti, line 2: Could we please add a comment here explaining the regex? :)

artismarti (approved):
looks good!

Would recommend squashing the hooks commit & the `p` tag related commits separately. (Or creating separate PRs for them - but that might be something for next time)

🚢

data/transform/post-parser/root-level-text-wrapper.ts
MymmiJ: Added, explained 👍
    /**
     * This regex finds:
     * (<\/[A-Za-z]+>)\n - closing HTML tag, followed by newline.
     * ([A-Z].*)\n - a capital letter preceding any information, up to a newline.
     *
     * The overall effect is to find capitalised text that immediately follows on
     * from a closing tag at the end of a line, which will always be plain text
     * that is supposed to be wrapped in <p> elements but has been missed by the
     * textile-js parser.
     *
     * Regex101: https://regex101.com/r/j4BLqX/1
     */

PR #1569, opened 2022-09-07
[EDX-135] Update UI for API references

MymmiJ:

## Description

Allows different components to be used in order to change the display of API Reference pages.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-135).
* [Figma file](https://www.figma.com/file/vyr9ZPZ15eUWtCKSEktdEL/Docs-v1?node-id=742%3A12196)

**Note:** - it's not possible to exactly match the font-weight, as there seems to be no light weight version of Source Code Pro installed; I'm not sure how to remedy this as I don't know where or how to get access to this font.

## Review

Make sure the page looks like the Figma page.

* [API reference page](http://localhost:8000/docs/api/realtime-sdk/channels?lang=ruby)
* [Regular documentation page](http://localhost:8000/docs/realtime/channels?lang=ruby)


src/templates/apiReference.js
```diff
@@ -0,0 +1,139 @@
...
+import PageTitle from '../components/PageTitle';
+import { DOCUMENTATION_PATH } from '../../data/transform/constants';
+import { safeWindow } from '../utilities/browser/safe-window';
+import { PREFERRED_LANGUAGE_KEY } from '../utilities/language/constants';
+import { createLanguageHrefFromDefaults, getLanguageDefaults } from '../components/common/language-defaults';
+
+const getMetaDataDetails = (document, prop, alternative = '') =>
+  document && document.meta && document.meta[prop] ? document.meta[prop] : alternative;
+
+const CANONICAL_ROOT = `https://www.ably.com${DOCUMENTATION_PATH}`;
```
dpiatek: `ably.com` should be an environment variable, ideally

src/templates/apiReference.js
```diff
@@ -0,0 +1,139 @@
...
+const META_DESCRIPTION_FALLBACK = `Ably provides a suite of APIs to build, extend, and deliver powerful digital experiences in realtime. Organizations like Toyota, Bloomberg, HubSpot, and Hopin depend on Ably’s platform to offload the growing complexity of business-critical realtime data synchronization at global scale.`;
+const NO_LANGUAGE = 'none';
+
+const ApiReference = ({
+  location: { search },
+  pageContext: { contentOrderedList, languages, version, contentMenu, slug, script },
+  data: { document, versions },
+}) => {
+  const params = new URLSearchParams(search);
```
dpiatek: Should `search` not be just `params`?

src/templates/apiReference.js
```diff
@@ -0,0 +1,139 @@
...
+  const language = params.get('lang') ?? DEFAULT_LANGUAGE;
+  useEffect(() => {
+    const preferredLanguage = safeWindow.localStorage.getItem(PREFERRED_LANGUAGE_KEY);
```
dpiatek: I've looked at how `safeWindow` is implemented, I think it could use some changes. We shouldn't conflate accessing the window with accessing localStorage - those are 2, very different concerns.

Instead, this should be something like:
```
... = storage.getItem(PREFERRED_LANGUAGE_KEY)
```

And this figures out the window, and returns nothing if none is set or we're not in a browser environment.

src/templates/apiReference.js
```diff
@@ -0,0 +1,139 @@
...
+      <PageLanguagesContext.Provider value={languages}>
+        <Helmet>
+          <title>{title}</title>
+          <meta property="og:title" content={title} />
+          <meta property="twitter:title" content={title} />
+          <link rel="canonical" href={canonical} />
+          <meta name="description" content={description} />
+          <meta property="og:description" content={description} />
+          <meta name="twitter:description" content={description} />
+        </Helmet>
```
dpiatek: Ideally this would be its own component, similarly to Voltaire https://github.com/ably/voltaire/blob/main/src/components/head/head.js, so we don't get discrepancies on what pages can output certain meta tags and which can't

src/components/blocks/api-reference/external-references/ApiReferenceBlockquote.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { ComponentProps, ReactNode } from 'react';
+import { HtmlComponentProps } from '../../../html-component-props';
+import { Blockquote } from '../../external-references';
+import Html from '../../Html';
+
+const StyledApiReferenceBlockquote = ({
+  attribs,
```
dpiatek, line 7: I'm a bit concerned with naming `props` `attribs`. Reason being, it's a very non-standard and also HTML already has attributes, which work very differently to props (and in fact, there are differences between using the in HTML and JavaScript). 

I'm worried this would create confusion both for people working between projects and more junior developers.

Why not do `props` or `...props` ?

src/components/blocks/api-reference/headings/ApiReferenceH6.tsx
```diff
@@ -0,0 +1,27 @@
...
+  </h6>
+);
+styled.h6`
+  align-items: center;
+  padding: 4px 6px;
+  background: #ffe6dc;
+  margin-top: 10px;
+  margin-bottom: 24px;
+  border: 1px solid #ff9e7a;
+  border-radius: 2px;
```
dpiatek, line 23: We generally use `rem`s everywhere for zoom/tailwind compatiblity

src/components/blocks/api-reference/headings/ApiReferenceH6.tsx
```diff
@@ -0,0 +1,27 @@
...
+    items-center w-min pb-4 px-8 mt-10 mb-24
+    rounded-sm bg-api-reference-attribute-highlight border border-api-reference-attribute-border`}
+  >
+    {children}
+  </h6>
+);
+styled.h6`
+  align-items: center;
+  padding: 4px 6px;
+  background: #ffe6dc;
```
dpiatek, line 19: if this color is the style guide, you can use css vars here `var(--color-x)`

tailwind.config.js
```diff
@@ -4,6 +4,8 @@ const periodicTableOfRealtimeColors = {
   'low-level-transport': 'var(--color-jazzy-pink)',
   'application-transport': '#e40060',
   'realtime-transport': '#ff2739',
+  'api-reference-attribute-highlight': '#ffe6dc',
+  'api-reference-attribute-border': '#ff9e7a',
```
dpiatek, line 8: This should be added in a seperate hash, these colors are very specific to the periodic realtime table ... which is in voltaire, not here ??? :)

src/templates/template-prop-types.js
```diff
@@ -0,0 +1,10 @@
+import PropTypes from 'prop-types';
+// Temporary: These are to be replaced with TypeScript types or removed as soon as possible
+
+export const templatePropTypes = {
+  location: PropTypes.shape({
+    search: PropTypes.string,
+  }),
+  pageContext: PropTypes.object,
+  data: PropTypes.object,
+};
```
dpiatek, line 10: I'm a bit confused why we are adding `PropTypes` while using TypeScript? Unless I'm missing something, everything touched should now be migrated to TS.

src/templates/base-template.js
```diff
@@ -0,0 +1,100 @@
...
+import RightSidebar from '../components/Sidebar/RightSidebar';
+import PageTitle from '../components/PageTitle';
+import { safeWindow } from '../utilities/browser/safe-window';
+import { PREFERRED_LANGUAGE_KEY } from '../utilities/language/constants';
+import { createLanguageHrefFromDefaults, getLanguageDefaults } from '../components/common/language-defaults';
+import { templatePropTypes } from './template-prop-types';
+import { DOCUMENTATION_PATH } from '../../data/transform/constants';
+
+const getMetaDataDetails = (document, prop, alternative = '') =>
+  document && document.meta && document.meta[prop] ? document.meta[prop] : alternative;
```
dpiatek, line 19: `document?.meta?.[prop]` ?

src/components/blocks/block-component-maps/component-map.js
```diff
@@ -92,7 +95,15 @@ const HtmlTypeComponentMap = Object.freeze({
   [HtmlDataTypes.wbr]: Wbr,
 });
 
-const componentMap = propOr(IS_TEXT, curry.placeholder, HtmlTypeComponentMap);
+const ApiReferenceHtmlTypeComponentMap = Object.freeze({
```
dpiatek, line 98: Why do we need a `freeze` here?

data/createPages/index.js
```diff
@@ -37,62 +41,86 @@ const createPages = async ({ graphql, actions: { createPage, createRedirect } })
       }
     }
   `);
-  if (result.data.allError.nodes && result.data.allError.nodes.length > 0) {
+  if (documentResult.data.allError.nodes && documentResult.data.allError.nodes.length > 0) {
     process.exit(1);
```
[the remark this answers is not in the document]
MymmiJ: We could log here, the reason we don't right now is that the error should already have been caught and logged.
Because Gatsby spawns a bunch of `async` processes during the `onCreateNode` step, though, and errors don't prevent the onCreateNode processes from continuing, the workaround is to create Error nodes and then quit when we get to the next synchronous stage in the Gatsby pipeline if any Error nodes exist.
It's an awkward solution and an alternative would be welcome, I'm not sure what we would do instead though. We could log only during createPages and never during onCreateNode, but then if there's an error that _does_ manage to crash Gatsby somehow or the Error Node is broken, we wouldn't see the output.

For context:
The original ticket here was [EDX-21](https://ably.atlassian.net/browse/EDX-21). When I originally detected missing meta descriptions, I wanted to console.warn and continue just so editors could see the error. However it was decided that it should be a requirement to always have a meta description, so the app had to avoid building while still throwing & logging an error when a page didn't have a meta description.

src/components/blocks/api-reference/external-references/ApiReferenceBlockquote.tsx
MymmiJ: These are just the attributes taken from the HTML as rendered via. textile, so they are intended to work the same way as attributes; for the basic styled blockquote component they happen to be the only props provided, but there could be others.

src/components/blocks/api-reference/headings/ApiReferenceH6.tsx
MymmiJ: Sadly I don't think it is, I searched Ably UI and it wasn't there; I believe the API Reference design was intended to be a quick update of the old API Reference stylings to be more in keeping with the overall new design, and hasn't been integrated fully with the linear design system yet.

src/templates/template-prop-types.js
MymmiJ: Feeling time pressure on this one, these are the existing PropTypes that we already have and I'm not sure if I can add TypeScript types for the innermost components/templates and have time to do everything else

tailwind.config.js
MymmiJ: Some of those colours are (or maybe only were) used a lot elsewhere in Voltaire, so when I was taking components from Voltaire early in docs development I thought this was a metaphorical periodic table of colours to be used to style anything; happy to make a separate hash for this.

data/createPages/index.js
```diff
@@ -11,15 +11,19 @@ const { identity } = require('lodash');
 const { safeFileExists } = require('./safeFileExists');
```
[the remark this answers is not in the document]
MymmiJ: This will add a lot of changes to the repo if I add it to linting rules now, can I create a ticket and address separately?

tailwind.config.js
dpiatek: yes please, most of these are one of colors for https://ably.com/periodic-table-of-realtime. You can just rename the hash the leave the colors you want.

data/createPages/createPageVariants.js
```diff
@@ -22,13 +22,17 @@ const addLanguagesToSet =
     }
   };
 
+const getLanguagesFromContent = (contentOrderedList) => {
+  const languageSet = new Set();
+  contentOrderedList.forEach(addLanguagesToSet(languageSet));
```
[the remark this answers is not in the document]
MymmiJ: Sorry, I forgot to respond here:

Short answer is we get these languages from the HTML attributes on the page. This provides a substitute set of languages to choose from for users in case the languages specified in the frontmatter meta are not set.

Long answer, in the interest of providing more insight into the full pipeline:

addLanguagesToSet is a recursive function that travels down the content data tree that looks very much like this:

`type ContentData = { data: ContentData[] | string | null; attribs?: { ...htmlAttributes }; type: HtmlElementName; }`

so the HTML would look like:

`<p lang="javascript">Some content</p><p lang="csharp">Some content</p><p lang="javascript">Some different content</p>`

and the data would look like:

`[{ data: "Some content", type: "p", attribs: { lang: "javascript" } }, { data: "Some content", type: "p", attribs: { lang: "csharp" } }, { data: "Some different content", type: "p", attribs: { lang: "javascript" } }]`

so in this case the unduplicated array would look like:

`['javascript', 'csharp', 'javascript']`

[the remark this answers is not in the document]
dpiatek: thanks for the explanation @MymmiJ :+1:

data/createPages/index.js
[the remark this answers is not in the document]
dpiatek: thanks @MymmiJ Could we add your explanation as a comment? I think it will be quite helpful in the future.

src/components/blocks/block-component-maps/component-map.js
dpiatek: Discussed offline (revist when this file is in TS)

dpiatek (approved):

src/components/ButtonWithTooltip/ButtonWithTooltip.module.css
```diff
@@ -12,7 +12,7 @@
 
 .tooltipClass span,
 .notificationClass span {
-  @apply text-white font-light bg-proprietary-realtime-software rounded-md text-center inline-block;
+  @apply text-white font-light rounded-md text-center inline-block;
```
snikidev, line 15: 🤔

PR #1570, opened 2022-09-08
[Suggestion] - EDX 125 add unsupported prism code snippets

MymmiJ:

## Description

Adds and updates support for languages we don't already support in Prism

## Review

Decide if this is desirable or necessary for your branch


snikidev (approved):
looks good 🔥

PR #1580, opened 2022-09-13
Redesign API Key indicator

MymmiJ:

## Description

Redesigning the API key indicator to show a custom designed tooltip on hover and display in the appropriate location.

## Review

Run docs locally and view any page.

Ask me if you have any issues!

Compare and contrast for example:

http://localhost:8000/docs/quick-start-guide?lang=flutter

https://ably-docs-next-staging.herokuapp.com/docs/quick-start-guide?lang=javascript

Paying attention to the API key indicator:

Screenshot 2022-09-13 at 15 31 04

Screenshot 2022-09-13 at 15 31 18

### DIFF
--- a/src/components/blocks/software/Code/ApiKeyIndicator.tsx
+++ b/src/components/blocks/software/Code/ApiKeyIndicator.tsx
@@ -1,12 +1,29 @@
-import React from 'react';
+import React, { useState } from 'react';
+import Icon from '@ably/ui/core/Icon';
 
-const tooltip =
-  'This code example uses a temporary key that is rate limited and expires in 4 hours. Sign in to Ably to use your API keys instead.';
-
-const APIKeyIndicator = () => (
-  <div className="docs-api-key-label" title={tooltip}>
-    Api Key: Demo Only
-  </div>
-);
+const APIKeyIndicator = ({ tooltip }: { tooltip: string }) => {
+  const [tooltipHover, setTooltipHover] = useState<boolean>(false);
+  return (
+    <div
+      className="flex flex-row w-full justify-start mt-2"
+      onMouseOver={() => setTooltipHover(true)}
+      onMouseOut={() => setTooltipHover(false)}
+    >
+      <div className="docs-api-key-label" title={tooltip}>
+        Demo Only
+      </div>
+      <Icon additionalCSS="ml-4 cursor-help" name="icon-gui-info" size="1rem" />
+      {tooltipHover ? (
+        <aside
+          className="w-240 max-w-240 absolute -mt-140 box-border
+          whitespace-pre-wrap bg-white shadow-tooltip rounded border border-light-grey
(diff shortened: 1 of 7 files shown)


src/components/blocks/software/Code/ApiKeyIndicator.tsx
```diff
@@ -1,12 +1,29 @@
...
 
-const tooltip =
-  'This code example uses a temporary key that is rate limited and expires in 4 hours. Sign in to Ably to use your API keys instead.';
-
-const APIKeyIndicator = () => (
-  <div className="docs-api-key-label" title={tooltip}>
-    Api Key: Demo Only
-  </div>
-);
+const APIKeyIndicator = ({ tooltip }: { tooltip: string }) => {
```
ferdipret, line 4: We should pull this type out into it's own object or interface, not sure what the convention is for the project.

src/components/blocks/software/Code/ApiKeyIndicator.tsx
```diff
@@ -1,12 +1,29 @@
...
+  const [tooltipHover, setTooltipHover] = useState<boolean>(false);
```
ferdipret, line 5: TS can infer the type here, no need to add the generic.

src/components/blocks/software/Code/ApiKeyIndicator.tsx
```diff
@@ -1,12 +1,29 @@
...
+  return (
+    <div
+      className="flex flex-row w-full justify-start mt-2"
+      onMouseOver={() => setTooltipHover(true)}
+      onMouseOut={() => setTooltipHover(false)}
```
ferdipret, line 10: The inline handlers would cause an extra render that we can avoid by passing function references.

src/components/blocks/software/styles.css
```diff
@@ -1,15 +1,16 @@
...
 @layer components {
   .docs-label {
-    @apply absolute bg-charcoal-grey p-4 ui-text-menu3;
+    @apply bg-charcoal-grey p-4 ui-text-menu3;
     color: #d9d9da;
   }
 
   .docs-language-label {
-    @apply top-4 right-4 docs-label cursor-default;
+    @apply absolute top-4 right-4 docs-label cursor-default;
```
ferdipret, line 9: This `absolute` feels a little weird to me. Without the component also specifying the relative element, I think this is a bit error prone.

ferdipret (changes requested):
Looks good, just a few things for consideration.

src/components/blocks/software/styles.css
MymmiJ: This is part of the process of removing the `absolute` from another, more general class, we don't use this class any more so we can simply remove it from the code base.

ferdipret (approved):
LGTM :shipit:

PR #1597, opened 2022-09-27
[EDU-904] - Add location animator section

tbedford:
## Description

Adds location animator section to docs for Android.

* [EDU-904](https://ably.atlassian.net/browse/EDU-904)

## Review

* [Location animator](https://ably-docs-edu-904-add-l-jhqoec.herokuapp.com/asset-tracking/using-the-sdks/#location-animator)

### DIFF
--- a/content/asset-tracking/using-the-sdks.textile
+++ b/content/asset-tracking/using-the-sdks.textile
@@ -17,6 +17,7 @@ jump_to:
     - Authentication#authentication
     - Using the Publishing SDK#publishing-sdk
     - Using the Subscribing SDK#subscribing-sdk
+    - Location animator#location-animator
     - See also#see-also
 ---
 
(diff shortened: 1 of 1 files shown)


content/asset-tracking/using-the-sdks.textile
```diff
@@ -600,8 +601,82 @@ subscriber.nextLocationUpdateIntervals
 
 See also the SDK reference docs for more "Subscriber properties":https://sdk.ably.com/builds/ably/ably-asset-tracking-android/main/dokka/subscribing-sdk/subscribing-sdk/com.ably.tracking.subscriber/-subscriber/index.html.
 
+h2(#location-animator). Location animator
+
+The location animator is an extension for the subscriber that provides smooth location updates animation. 
+
+The location animator receives location updates from the publisher and creates animation steps from them. These steps are then applied to an ongoing animation.
```
KacperKluka, line 608: "The location animator receives location updates from the publisher" - doesn't it imply that the location animator somehow communicates directly with the Publisher? 🤔 I wonder whether we should change it to show that you as the user have to pass the locations from the subscriber to the animator?
```suggestion
The location animator creates animation steps from location updates that are provided by the Subscriber SDK. These steps are then applied to an ongoing animation. 
```

content/asset-tracking/using-the-sdks.textile
```diff
@@ -600,8 +601,82 @@ subscriber.nextLocationUpdateIntervals
...
+
+The location animator is an extension for the subscriber that provides smooth location updates animation. 
+
+The location animator receives location updates from the publisher and creates animation steps from them. These steps are then applied to an ongoing animation. 
+
+The animation duration is then recalculated as the sum of location update interval (the time to next location update from the publisher) and the intentional delay. The intentional delay allows for network issues or delays in the location updates. For example, if the intentional delay was set to two seconds, then a two second buffer will help the animation remain smooth and continuous, even if the publisher did not send a location update within the specified time, with a two second margin.
+
+The location animator then recalculates each step duration to evenly take up the animation duration, so the overall animation of the trackable is smooth.
+
+Each step is animated at 60 FPS, and this applies to both the map marker and the camera position.
```
KacperKluka, line 614: "this applies to [...] the camera position." - not true 😅 

The camera positions are not animated on their own, we simply publish a single camera position whenever it needs to be updated. We expect the users to use their map provider SDK to animate the camera from the current position to the new one. Otherwise, the camera would not be responsive for user input if we were to update its position 60 times per second 😉 
```suggestion
Each step is animated at 60 FPS, and this applies to the map marker.
```

content/asset-tracking/using-the-sdks.textile
```diff
@@ -600,8 +601,82 @@ subscriber.nextLocationUpdateIntervals
...
+val locationAnimator: LocationAnimator = CoreLocationAnimator(
+  intentionalAnimationDelayInMilliseconds,
+  animationStepsBetweenCameraUpdates
+) 
+```
+
+There are two optional parameters:
+
+- intentionalAnimationDelayInMilliseconds := The higher the delay, the more buffer there is to allow for unexpected delays of location updates from the publisher. Also moves the actual position of the trackable backwards in time. The higher the setting the more network issues and delays are allowed for, but the position updates will be less realtime in nature. By default this is set to two seconds, but you should test what value works best for your use case.
+- animationStepsBetweenCameraUpdates := By default, when a step is being animation, the camera position is also updated. But this can be changed so that the camera position updates, for example, on every five location updates.
```
KacperKluka, line 632: Is it a typo? 🤔 
```suggestion
- animationStepsBetweenCameraUpdates := By default, when a step is being animated, the camera position is also updated. But this can be changed so that the camera position updates, for example, on every five location updates.  
```

content/asset-tracking/using-the-sdks.textile
```diff
@@ -600,8 +601,82 @@ subscriber.nextLocationUpdateIntervals
...
+
+The location animator exposes a positions flow:
+
+```[kotlin]
+locationAnimator.positionsFlow
+  .onEach { trackablePosition -> }
+  .launchIn(scope)
+```
+
+This provides data you can use to move your map marker on the map, or when you want to move your trackable in the screen.
```
KacperKluka, line 654: We could add info here that this flow is updated 60 times per second when the animation is ongoing 🤔

content/asset-tracking/using-the-sdks.textile
```diff
@@ -600,8 +601,82 @@ subscriber.nextLocationUpdateIntervals
...
+
+h3(#location-animator-location-updates-animate). Providing location updates to animate
+
+To use the animator you need to provide it with some information:
+
+```[kotlin]
+locationAnimator.animateLocationUpdate(locationUpdate, expectedIntervalBetweenLocationUpdatesInMilliseconds)
+```
+
+- locationUpdate := The location update from the publisher.
```
KacperKluka, line 674: Location animator users only have access to the subscriber SDK, so perhaps we should not mention the publisher here? 🤔 
```suggestion
- locationUpdate := The location update from the subscriber.
```

content/asset-tracking/using-the-sdks.textile
```diff
@@ -600,8 +601,82 @@ subscriber.nextLocationUpdateIntervals
...
+- expectedIntervalBetweenLocationUpdatesInMilliseconds := Obtained from subscriber API, expected time within which the publisher is expected to send a location update.
+
 h2(#see-also). See also
 
 * "Overview":/asset-tracking
 * "Using the example apps":/asset-tracking/example-apps
+* "Asset Tracking Android training course":https://github.com/ably/asset-tracking-android-training
```
KacperKluka, line 681: I'm not sure if we should call it a `course` 🤔 Perhaps we can call it "playground" or "workshop project"?

KacperKluka (changes requested):
I've left a few suggestions 😁

content/asset-tracking/using-the-sdks.textile
tbedford: Yes, I wasn't sure. I was guessing the videos were available to customers. I think workshop project is good.

content/asset-tracking/using-the-sdks.textile
tbedford: Done.

KacperKluka (approved):
LGTM, thanks 👍

PR #1602, opened 2022-09-29
[EDX-294] API Reference language display

MymmiJ:

## Description

This ticket is a collection of very closely-related/interdependent fixes:

* API Reference has to display languages that appear on their own when they are inside divs, spans or code elements.
* Pre-wrapped code elements (`<pre><code></code></pre>`) have no non-text child elements
* When pre-wrapped code elements are rendered, all child nodes are checked to see if there are multiple text nodes, as there can be multiple sequential text nodes.
* Fix issues with root-level-text-wrapper failing to wrap all root-level-text in <p></p> tags, add tests for that case.
* Add new dedicated API Reference elements for Spans, Divs, dds and dts.

Minor changes:
* post-parser files simply renamed to .ts files
* Remove unused styled-component code from ApiReferenceH6, likely a rebase error
* Fix table display by enabling StyledComponents to understand colspan/rowspan directives
* Fix console.error from use-search.ts hook test

* [Jira ticket](https://ably.atlassian.net/browse/EDX-294).


data/transform/post-parser/blang.ts
```diff
@@ -1,4 +1,4 @@
-const { TAG_OPEN_REGEX_STRING, TAG_CLOSE_REGEX_STRING, TAG_OPEN_OR_CLOSE_REGEX_STRINGS } = require('./constants');
+import { TAG_OPEN_REGEX_STRING, TAG_CLOSE_REGEX_STRING, TAG_OPEN_OR_CLOSE_REGEX_STRINGS } from './constants';
```
snikidev, line 1: 🔥 👍

src/components/blocks/Html/LinkableHtmlBlock.js
```diff
@@ -1,16 +1,25 @@
-import React from 'react';
+import React, { useContext } from 'react';
```
snikidev, line 1: do you fee like making it a TS file as well? 😄

src/components/blocks/wrappers/ConditionalChildrenLanguageDisplay.js
```diff
@@ -8,11 +8,11 @@ const ConditionalChildrenLanguageDisplay = ({ children }) => {
   let currentGroup = false;
   const childLanguageGroups = [];
   const toFilter = [];
-  Children.forEach(children, ({ props, props: { attribs = null } }, index) => {
```
snikidev, line 11: could also become TS, if you're in the mood for it 😅

snikidev (approved):
Looks good, but I'd move all JS files here to TS! 🙃  👍

src/components/blocks/wrappers/ConditionalChildrenLanguageDisplay.js
MymmiJ: I've investigated and it causes a lot more issues; TypeScript doesn't support unions as big as would need to be created here.

PR #1604, opened 2022-09-29
[EDX-260] Tables UI/ UX

snikidev:

## Description

Updating Table UI

* [Jira ticket](https://ably.atlassian.net/browse/EDX-260).

## Review

There is a table on `/docs/core-features/authentication?lang=go` page.


src/components/blocks/table/Table.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { ComponentProps } from 'react';
+import cn from 'classnames';
+import Html from 'src/components/blocks/Html';
+import { HtmlComponentProps, ValidReactElement } from 'src/components/html-component-props';
+import HtmlDataTypes from '../../../../data/types/html';
+
+import { table, tableContainer } from './Table.module.css';
+
+type Props = {
```
MymmiJ, line 9: I think this can just be replaced with: `HtmlComponentProps<'table'>`

MymmiJ: I see that this causes a TS error, which we should handle imo; if data is null or `isString` (using lodash or lodash/fp), we should return 'null' as this would be an invalid table.

src/components/blocks/table/Tbody.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { ComponentProps } from 'react';
+import { HtmlComponentProps, ValidReactElement } from 'src/components/html-component-props';
+import Html from 'src/components/blocks/Html';
+import HtmlDataTypes from '../../../../data/types/html';
+
+type Props = {
```
MymmiJ, line 6: As per Table, I think this should be `HtmlComponentProps<'tbody'>` and if data is null or isString, return null.

src/components/blocks/table/Td.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { ComponentProps } from 'react';
+import { HtmlComponentProps, ValidReactElement } from 'src/components/html-component-props';
+import GenericHtmlBlock from '../Html/GenericHtmlBlock';
+
+const Td = (props: { data: HtmlComponentProps<ValidReactElement>[]; attribs: ComponentProps<'td'> }) => {
```
MymmiJ, line 5: I think this can just be `HtmlComponentProps<'td'>`

src/components/blocks/table/Th.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { ComponentProps } from 'react';
+import cn from 'classnames';
+import { HtmlComponentProps, ValidReactElement } from 'src/components/html-component-props';
+import GenericHtmlBlock from '../Html/GenericHtmlBlock';
+
+import { theader } from './Th.module.css';
+
+const Th = (props: { data: HtmlComponentProps<ValidReactElement>[]; attribs: ComponentProps<'th'> }) => {
```
MymmiJ, line 8: I think this can just be `HtmlComponentProps<'th'>`

src/components/blocks/table/Tr.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { ComponentProps } from 'react';
+import { HtmlComponentProps, ValidReactElement } from 'src/components/html-component-props';
+import Html from 'src/components/blocks/Html';
+import HtmlDataTypes from '../../../../data/types/html';
+
+type Props = { data: HtmlComponentProps<ValidReactElement>[]; attribs: ComponentProps<'tr'> };
+
+const Tr = ({ data, attribs }: Props) => (
```
MymmiJ, line 8: I think this should be `HtmlComponentProps<'tr'>`, and this time if data isString or is null, return <tr>{data}</tr>

MymmiJ (changes requested):
Nice one. If you can have a look at the TS changes required and change those or come back with a comment I will approve.

MymmiJ (approved):

PR #1608, opened 2022-10-01
Migrate _most_ of the Client Library Development Guide to `ably/specification`

QuintinWillison:
Apologies for the size of this pull request, but it's logical and necessary to perform this content migration atomically.

All 'moved to' (new home) URLs I've added begin with `https://sdk.ably.com/builds/ably/specification/main/` however those links will give you a `403 Access Denied` until https://github.com/ably/specification/pull/4 has landed to the `main` branch of [the `ably/specification` repository](https://github.com/ably/specification), which is the source code base that renders the content to HTML and pushes out to `sdk.ably.com`.


tbedford (approved):
Epic!

stmoreau (approved):
LGTM - nice one!

QuintinWillison:
Thanks for the review input so far. ❤️ ... Just a quick reminder that we'll not merge this pull request until after https://github.com/ably/specification/pull/4 has merged.

lmars (approved):

m-hulbert (approved):
Approving as this all looks good and I have no objections. 

Two comments I'd like your opinion on though @QuintinWillison:

* I think the previous versions and diffs should be preserved, and in my opinion this makes sense to be stored with the latest spec (even if just in raw textile).
* Any objections if we get redirects set up with the web team in the future to remove these pages from docs entirely? Do we need to preserve the deep links of the features spec for anything?

QuintinWillison:
Thanks, @m-hulbert. I have thoughts on both:

> I think the previous versions and diffs should be preserved, and in my opinion this makes sense to be stored with the latest spec (even if just in raw textile).

I disagree. The change history has to, logically, be discernible using `git` tools based on commit history - which I've carried over to the new home. These static snapshots of diffs were always a bad idea, in my opinion, as they duplicate that information and could conceivably drift out of date or be wrong in ways that tools cannot spot. I have, however, created https://github.com/ably/specification/issues/94 to work on at my end to bridge this discoverability gap. Feel free to carry the conversation on in that issue, perhaps expanding on what you would like to see visibility of.

> Any objections if we get redirects set up with the web team in the future to remove these pages from docs entirely? Do we need to preserve the deep links of the features spec for anything?

I would love for that to happen and I had presumed that this would be the next logical step, after this pull request has landed. My intention here was to get the 'soft redirects' in place. What I would want to see from future redirects, of course, is that they also carry over the HTML element id anchors too. Some of these documents are quite long and that was one of my goals, to keep that rich, deep linking carrying forwards.

PR #1616, opened 2022-10-07
[EDX-295] Definition list indentation fix

MymmiJ:

## Description

Remove bug that means that when a .textile file is used as a partial, definition lists are not parsed correctly by textile.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-295).

## Review

The bug mostly appeared on this page:

* [Page to review](http://localhost:8000/docs/api/realtime-sdk/channels?lang=java)

Here's a screenshot of the problem in the current toolchain, right at the bottom you can see:

    Members
    - key := The key valueType: String - value := The value associated with the keyType: String

It should be rendered as a definition list.

Screenshot 2022-10-07 at 10 25 22

### DIFF
--- a/content/tutorials/reactor-event-azure.textile
+++ b/content/tutorials/reactor-event-azure.textile
@@ -573,20 +573,8 @@ Congratulations, you're done!
 
 h2. Next Steps
 
-<<<<<<< HEAD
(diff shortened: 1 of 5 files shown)


snikidev (approved):
looks neat! 🙌

PR #1617, opened 2022-10-07
[EDX-274] Extra horizontal line is appearing below top nav

snikidev:

## Description

TopNav had an extra border at the bottom on the content side. Now it's gone

* [Jira ticket](https://ably.atlassian.net/browse/EDX-274).

## Review

Look at the page without language selector nav, e.g. `/docs/core-features/pubsub?lang=javascript`


MymmiJ (changes requested):
This works when the language menu doesn't appear, but the language menu is required to have a bottom border, see https://www.figma.com/file/vyr9ZPZ15eUWtCKSEktdEL/Docs-v1?node-id=1301%3A15868 for example; I think the fix should be to move `showCodeMenu` so that the `<div>` is not returned at all if `showCodeMenu` is false.

PR #1618, opened 2022-10-10
Fix incorrect key in JWT authentication example (user claims)

QuintinWillison:
I noticed this in passing, when I was reading this documentation and cross-verifying with the Realtime implementation (internal codebase).

The correct key is used elsewhere in the docs, so this was the only location that I could find this incorrect (pluralised) key being used in this codebase.


tbedford (approved):

tbedford:
Thanks @QuintinWillison !

PR #1626, opened 2022-10-12
[EDX-324]: Update the labels in the language selector to be in the correct format 

snikidev:

## Description

Language Dropdown labels for JS and C# updated

* [Jira ticket](https://ably.atlassian.net/browse/EDX-324).

### DIFF
--- a/src/maps/language.tsx
+++ b/src/maps/language.tsx
@@ -9,13 +9,13 @@ type LanguageSyntaxHighlighterNames = {
 };
 
 const languageLabels: LanguageLabels = {
-  javascript: 'JS',
+  javascript: 'JavaScript',
   java: 'Java',
   ruby: 'Ruby',
   python: 'Python',
   php: 'PHP',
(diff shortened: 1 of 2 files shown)


src/maps/language.tsx
```diff
@@ -9,13 +9,13 @@ type LanguageSyntaxHighlighterNames = {
 };
 
 const languageLabels: LanguageLabels = {
-  javascript: 'JS',
+  javascript: 'JavaScript',
```
MymmiJ, line 12: Does this also affect the local language buttons? I think we need a longLabel and a shortLabel for language and to select the one we need based on our use case

MymmiJ (changes requested):
Let me know if I've got this question right.

src/maps/language.tsx
snikidev: yeah, you're right. fixed! 🙌

MymmiJ (approved):

PR #1643, opened 2022-10-24
[EDX-376] uncapitalize h4s

MymmiJ:

## Description

h4s have a text-transform uppercase (in ui-text-h4) which is removed by this PR - the transform is not appropriate for docs, as case can matter.

Also fixes an error in SuggestionBox.module.css with a comment that webpack doesn't understand.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-376).

### DIFF
--- a/src/components/Header/TopMainNav/SearchBar/SuggestionBox.module.css
+++ b/src/components/Header/TopMainNav/SearchBar/SuggestionBox.module.css
@@ -7,6 +7,5 @@
 
 .snapshot {
   @apply sticky top-16 left-0 right-0;
-  /* Height of the container minus 16px (x2) top and bottom padding */
-  height: calc(546px);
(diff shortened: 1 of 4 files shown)


src/components/Header/TopMainNav/SearchBar/SuggestionBox.module.css
```diff
@@ -7,5 +7,6 @@
 
 .snapshot {
   @apply sticky top-16 left-0 right-0;
-  height: calc(546px); // Height of the container minus 16px (x2) top and bottom padding
+  /* Height of the container minus 16px (x2) top and bottom padding */
+  height: calc(546px);
```
aralovelace, line 11: I think its more responsive to use rem instead of px

aralovelace (approved):

PR #1656, opened 2022-10-26
[EDX-305] Support default values in definition lists

MymmiJ:

## Description

Adds 'default' values to definition lists.
* [Jira ticket](https://ably.atlassian.net/browse/EDX-305).

## Review

* [Page to review](http://localhost:8000/docs/api/rest-api?lang=javascript)

### DIFF
--- a/src/components/blocks/list/Dl/index.tsx
+++ b/src/components/blocks/list/Dl/index.tsx
@@ -28,7 +28,18 @@ const StyledDl = styled.dl`
   dd {
     ${definitionListCell}
   }
-  dd > em:last-of-type {
+  dd > em:not(.italics):first-of-type {
+    background: #e1e1e1;
+    border-radius: 5px;
+    padding: 2px 5px;
+    ::before {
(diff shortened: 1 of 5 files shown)


snikidev (approved):

src/components/blocks/list/Dl/index.tsx
```diff
@@ -28,7 +28,18 @@ const StyledDl = styled.dl`
   dd {
     ${definitionListCell}
   }
-  dd > em:last-of-type {
+  dd > em:not(.italics):first-of-type {
```
aralovelace, line 31: just wondering if we can separate this into a css file so we can convert it to tailwind?

aralovelace (approved):

src/components/blocks/list/Dl/index.tsx
MymmiJ: I don't think targeting children in this way with tailwind is very easy?

PR #1658, opened 2022-10-27
[EDU-1057] - Unsplit all versions of the Ably REST API reference

tbedford:
## Description

Work done:

* Unsplits all versions of the Ably REST API
* Adds redirects
* Removes split versions of REST API reference
* Removes span tags that were causing errors when switching versions (they weren't really necessary)

*Note:* not sure why, but the redirect from `/rest-api` doesn't work when testing in the review app, but works fine in the web view, and also works fine under `nanoc` locally. 

Ticket:

* [EDU-1057](https://ably.atlassian.net/browse/EDU-1057)

## Review

View combined page: 

* [REST API Reference](https://ably-docs-edu-1057-unsp-yaw0ui.herokuapp.com/)

Website view:

* https://github.com/ably/website/pull/5068


m-hulbert (approved):
This LGTM - one minor comment is whether you think we should create a folder for the rest-api as per the SDKs and `token-request-spec` sits within that?

tbedford:
> This LGTM - one minor comment is whether you think we should create a folder for the rest-api as per the SDKs and token-request-spec sits within that?

Nice idea, but I think this will all need to be restructured anyway, as part of the IA hill.

PR #1664, opened 2022-10-27
[EDX-391] Never render the component label server side to avoid double renders

MymmiJ:

## Description

Sometimes two logos render at once when you build the site (as opposed to running `gatsby develop`), for example here:

9206ae6a-fc32-4d88-871d-0e0e5fa8632c

This happens because the component is loaded with no language selected (no query parameter) at build time, and then loaded again at run time.

This is the simplest possible guard to prevent the component from being loaded at build time.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-391).

## Review

* Run `gatsby clean` followed by `gatsby build --prefix-paths && gatsby serve --prefix-paths`
* [Check that there is only one logo displayed here](http://localhost:9000/docs/quick-start-guide)
* Select a language using the dropdown language selector
* [Check that there is only one logo displayed here](http://localhost:9000/docs/general/push/publish?lang=ruby)


aralovelace (approved):
I run and test and  the logo are visible perfectly.

PR #1667, opened 2022-10-28
[EDX-409] Update the table of contents

m-hulbert:
## Description

This PR removes the sections from the table of contents that are represented in the header navigation.


Srushtika (changes requested):
I would suggest having API References at the top of the further reading list. Other than that it looks good, thanks.

Srushtika (approved):
ltgm, thanks

PR #1668, opened 2022-10-28
[EDX-298] Expand nested spans within types

m-hulbert:
## Description

This PR expands nested span tags found within `type` elements so that they can render correctly in the new toolchain.

## Review

Ensure that all types still display correctly in the current toolchain.


tbedford (approved):
Much better!

PR #1678, opened 2022-11-01
[EDX-396] adopt ably UI code highlight

MymmiJ:

## Description

Replaces PrismJS with highlighting from ably ui.
* [Jira ticket](https://ably.atlassian.net/browse/EDX-396).


src/components/blocks/software/Code/MultilineCodeContent.tsx
```diff
@@ -44,19 +49,20 @@ export const MultilineCodeContent = ({
...
-  useLayoutEffect(() => {
-    Prism.hooks.add('before-highlight', function (env) {
-      env.code = env.element.textContent || '';
-    });
-
-    if (preRef.current) {
-      Prism.highlightElement(preRef.current);
-    }
-  }, [displayLanguage]);
+  const highlightedContent = useMemo(() => highlightSnippet(language, renderedContent), [language, renderedContent]);
```
dpiatek, line 52: Was it not possible to highlight before the rendering stage?

src/components/blocks/software/Code/__snapshots__/Code.test.tsx.snap
```diff
@@ -8,79 +8,11 @@ exports[`<Code /> Renders Code elements with language 1`] = `
...
-      >
-        <span
-          class="token string"
-        >
-          'hello!'
-        </span>
-      </span>
+      function (_a, _b, _c, _d) {
+        return fn.apply(this, arguments);
+      }
```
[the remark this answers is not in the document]
MymmiJ: Not sure there's a reason :+1:

src/components/blocks/software/Code/MultilineCodeContent.tsx
MymmiJ: Sort of, would be a performance impact as we choose between different strings based on data provided

dpiatek: Could you elaborate what kind of performance impact we're talking about here? It's important to remember that impact is now moved to the browser, i.e. the user. We don't track rendering perf (yet) but it would be interesting to see if this is noticeable on snippet heavy pages (maybe with some CPU throttling) and compare. (Ideally we'd have data for this already, but we don't).

(I wouldn't block this PR because of this, but I'd like for us to be confident that even on a device that is less powerful than an M1, the pages render fast-enough :) )

MymmiJ: I think the performance impact would be minimal, just seems wasteful to highlight very similar content multiple times and only display one version to me.

This is an abstract performance impact as in we're choosing which option 1 2 3 to render; if we run highlightSnippet(1) highlightSnippet(2) highlightSnippet(3), it can surely only be slower than choosing the option first and then highlighting it?

MymmiJ: Oh, I see what you mean;

it's not possible to render this without information from the user, so that's not possible.

MymmiJ: Yes, whether they're signed in or not and have API keys

dpiatek: right, but that only means it needs to be interpolated later (realize this is not what's done now and might need rework. again, not advocating for anything like that now)

src/components/blocks/software/Code/MultilineCodeContent.tsx
```diff
@@ -1,8 +1,13 @@
-import React, { useLayoutEffect, useRef } from 'react';
-import Prism from 'prismjs';
-import cn from 'classnames';
+import React, { useMemo, useRef } from 'react';
+// @ts-ignore
+import { highlightSnippet, registerDefaultLanguages } from '@ably/ui/src/core/utils/syntax-highlighter';
+// @ts-ignore
+import languagesRegistry from '@ably/ui/src/core/utils/syntax-highlighter-registry';
```
dpiatek, line 5: see https://github.com/ably/voltaire/blob/main/src/types/ably-ui/index.d.ts for examples of how you could type these

dpiatek (approved):
some minor comments, otherwise looks good to me

src/components/blocks/software/Code/MultilineCodeContent.tsx
MymmiJ: Ticket created for this work

PR #1688, opened 2022-11-02
[EDX-450] using sdks disappearing javascript

MymmiJ:

## Description

Ensures that if there is an empty element in the middle of a language group, it does not throw off the filter count; @snikidev I think you brought this up before but I didn't realise we're not handling empty elements in the middle of a group. This should fix.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-450).

## Review

* [Page to review](http://localhost:8000/docs/asset-tracking/using-the-sdks?lang=kotlin)


snikidev (approved):

src/components/blocks/wrappers/ConditionalChildrenLanguageDisplay.js
```diff
@@ -10,6 +10,9 @@ const ConditionalChildrenLanguageDisplay = ({ children }) => {
   const toFilter = [];
   Children.forEach(children, ({ props, props: { attribs = {} } }, index) => {
     if (isIrrelevantForLanguageDisplay(props.data)) {
+      if (currentGroup?.languages) {
+        currentGroup.languages.push('');
+      }
```
snikidev, line 15: 🤔

PR #1701, opened 2022-11-04
[EDX-485] Search results don't match the specs

snikidev:

## Description

Fixing search issues

* [Jira ticket](https://ably.atlassian.net/browse/EDX-485).

### DIFF
--- a/src/components/Header/TopMainNav/SearchBar/SuggestionBox.tsx
+++ b/src/components/Header/TopMainNav/SearchBar/SuggestionBox.tsx
@@ -42,7 +42,7 @@ export const SuggestionBox = ({ results, isActive }: Props) => {
       <div className="font-light text-dark-grey uppercase text-menu3 px-16 mb-2">Results from docs</div>
       {results &&
         results.map((hit, index) => {
-          const { title, highlight, url, id } = hit;
+          const { title, highlight, meta_description, url, id } = hit;
           const [pageTitle, ...breadcrumbs] = title.split(' / ').filter(
             (item) =>
               // We need to get rid of 'Docs' because it's redundant and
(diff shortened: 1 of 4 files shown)


src/components/Header/TopMainNav/SearchBar/SuggestionBox.tsx
```diff
@@ -38,25 +39,35 @@ export const SuggestionBox = ({ results, isActive }: Props) => {
...
-      <div className="font-light text-dark-grey uppercase text-menu3 px-16">Results from docs</div>
+      <div className="font-light text-dark-grey uppercase text-menu3 px-16 mb-2">Results from docs</div>
       {results &&
         results.map((hit, index) => {
           const { title, highlight, url, id } = hit;
+          const [pageTitle, ...breadcrumbs] = title.split(' / ').filter(
+            (item) =>
+              // We need to get rid of 'Docs' because it's redundant and
+              // 'Root` because there are no root section it gets redirected to '/docs/'
+              !DOCUMENTATION_PATH.toLowerCase().includes(item.trim().toLowerCase()) ||
```
MymmiJ, line 50: item.trim().toLowerCase could be a const here maybe?

MymmiJ: This could also be `!DOCUMENTATION_NAME === item.trim().toLowerCase()` I think

MymmiJ (approved):
Once you've made the changes from Srushtika what's here already looks good.

src/components/Header/TopMainNav/SearchBar/SuggestionBox.tsx
snikidev: I tried, but it needs to be `.includes`, because `DOCUMENTATION_PATH` has `/`s at the end and the start of the string. So I'm pretty much using the string as an array.

aralovelace (approved):

src/components/Header/TopMainNav/SearchBar/SuggestionBox.tsx
MymmiJ: Ah yeah, that's why I said DOCUMENTATION_NAME - it's actually a different constant but doesn't really matter

PR #1704, opened 2022-11-04
Add DEFAULT_LANGUAGE option for rendering languages correctly

m-hulbert:
## Description

Adds `DEFAULT_LANGUAGE` to fix issues where default blocks don't always appear.


MymmiJ (approved):
Awesome! LGTM :D thank you!

PR #1705, opened 2022-11-04
[EDX-496] Terminate nested code block at end of partial

m-hulbert:
## Description

The new toolchain requires codeblocks nested within blangs to be terminated at the end of a partial, otherwise the last language runs on into the parent file.

This only occurs in a single partial and this PR fixes it.

## Review

Ensure the partial still renders correctly in the existing toolchain.


tbedford (approved):
Built review app and checked Presence Action. LGTM.

PR #1728, opened 2022-11-10
[EDX-481] - RHS Padding adjustments

aralovelace:
## Description

* [EDX-481](https://ably.atlassian.net/browse/EDX-481?atlOrigin=eyJpIjoiMjMzYzNkOTNlZTE5NGVjNDgxOTVkNzFlYTBkZThhNGIiLCJwIjoiaiJ9)

## Review

- https://ably-docs-2-edx-481-0sly54gf4x.herokuapp.com/docs/api/realtime-sdk

### DIFF
--- a/src/components/Sidebar/RightSidebar/RightSidebar.module.css
+++ b/src/components/Sidebar/RightSidebar/RightSidebar.module.css
@@ -18,5 +18,5 @@
 
 .withLanguageNavBar {
   /* NOTE: when we have a language bar, the space on top should also be doubled. */
-  top: 11rem;
+  top: calc(var(--top-nav-height) + 4rem);
 }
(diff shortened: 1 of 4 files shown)


src/components/Sidebar/RightSidebar/RightSidebar.module.css
```diff
@@ -18,5 +18,5 @@
 
 .withLanguageNavBar {
   /* NOTE: when we have a language bar, the space on top should also be doubled. */
-  top: calc(var(--top-nav-height) * 2);
+  top: 11rem;
```
snikidev, line 21: Are you sure we don't want to use global variable here? We're alos using some calculations in `Sidebar.module.css` I can't remember the exact reasons, but make sure it's in sync. Because if you're taking out the `var(--top-nav-height)` variable here and hardcoding the value, it might affect styles in the `Sidebar` component. 

Again, not sure how, but just FYI 😅

snikidev (approved):
Just a minor comment to look into, could be nothing 🙃

src/components/Sidebar/RightSidebar/RightSidebar.module.css
aralovelace: @snikidev  you are using, to make it sync I use  the variable back and just + the value of the top that was added. thanks

PR #1753, opened 2022-11-21
[EDX-543] - Incorrect text for language

MymmiJ:

## Description

1. Sorts out EDX-543. The cause was that the HTML parser phase 'duplicate lang attributes' was, due to the way jQuery/Cheerio works, _not_ parsing every item as '*[lang]' if that item was nested e.g. `<div><span lang="javascript,python"></span></div>` => ✅, but `<div lang="default"><span lang="javascript,python"></span></div>` => X
2. Adds a `read-textile` script that allows quicker conversion and step by step debugging of parsing phases through `npm run debug:textile`, also takes a `--diff` flag if you want to view diffs between each step.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-543)

## Review

On the page below, ensure that all <span>s show up correctly when they are duplicate language blocks nested inside other language blocks. Examples appear under #client-options **Returns** heading (linked, Client Options in the sidebar) where `status_code` should show up rather than `statusCode` as it does on the current site.

Try running `npm run debug:textile` and see if there's anything else useful we could add.

* [Page to review](http://localhost:8000/docs/api/rest-sdk?lang=ruby#client-options)


read-textile.ts
```diff
@@ -0,0 +1,206 @@
...
+
+const queryUser = async (query: string, outputHandler: (answer: string) => void) => {
+  rl.question(query, outputHandler);
+};
+
+const executeSequentially = (queries: string[], outputHandlers: ((answer: string) => void)[]) => {
+  queries.forEach(async (query, i) => await queryUser(query, outputHandlers[i]));
+};
+
+// TODO: concat postParser, htmlParser steps.
```
snikidev, line 61: is this TODO for this PR or for later work?

read-textile.ts
```diff
@@ -0,0 +1,206 @@
...
+        process.exit();
+        break;
+      default:
+        process.exit();
+        break;
+    }
+  },
+];
+
+executeSequentially(initialQuery, initialOutputHandler);
```
snikidev, line 206: ha! Nice!

snikidev (approved):

read-textile.ts
MymmiJ: Later work 😬 it's a bit boring. Post-v1, kind of a feature.

PR #1748, opened 2022-11-18
[EDX-444] Support for server-side redirects with nginx

kennethkalmer:
## Description

Support for server-side redirects via nginx, meaning we can send proper `301` redirects instead of rendering pages and doing client-side redirects.

This can be achived by generating a map of redirects in `config/nginx-redirects.conf`, which nginx will load up and use.

* [EDX-444](https://ably.atlassian.net/browse/EDX-444)

## Review

Instructions on how to review the PR. 

* [Page to review](link)

PR #1763, opened 2022-11-24
[EDX-444] Incorrect try_files directive broke nginx

kennethkalmer:
A last minute change in #1748 went undetected and unfortunately resulted in a broken nginx configuration. This fixes it by adding the missing `@404` fallback to a `try_files` directive.


MymmiJ (approved):
Whoop!

PR #1784, opened 2022-12-01
[EDX-500] add mqtt

Sapples:

## Description

* [Jira epic ticket](link) - if applicable.
* [Jira ticket](link).

## Review

* [Page to review](link)

### DIFF
--- a/content/protocols/mqtt.textile
+++ b/content/protocols/mqtt.textile
@@ -1,0 +1,150 @@
+---
+title: MQTT
+meta_description: "Any MQTT-enabled client can communication with the Ably service through the Ably MQTT protocol adapter. This is especially useful where an Ably client library SDK is not available for your language of choice."
+---
+
+The Ably MQTT protocol adapter is able to translate back and forth between "MQTT":https://mqtt.org/ and Ably's own protocol, allowing for seamless integration of any systems you may have. MQTT (MQ Telemetry Transport) is a "publish/subscribe":https://ably.com/topic/pub-sub , lightweight messaging protocol designed for constrained devices and low-bandwidth networks. One of the major uses of MQTT is with IoT (Internet of Things), where these principles are key to having effective communication between various devices.
+
+h2(#when-to-use). When to use the MQTT adapter
+
+The MQTT adapter is our recommended way of interacting with Ably from devices which do not have a native Ably client library, such as Arduino platforms, C/C++ applications, and so on. Anyone who has previously been using Pubnub or Pusher client libraries for this purpose may want to consider switching to MQTT. Compared to the Pubnub protocol, using MQTT will result in better performance with lower latency. Compared to the Pusher protocol, MQTT will give you connection state recovery.
+
+We recommend you use MQTT to interact with Ably when:
+
+* Using devices which do not have a device-specific Ably Client Library SDK, such as Arduino platforms, C/C++ applications, and so on
+* If bandwidth is limited and you want to keep network traffic to a minimum
+* You want to keep network traffic to a minimum.
+
+Note, we do not currently support MQTT over websockets but let us know if you are interested.
+
+h2(#limitations). Known limitations 
+
+Our MQTT adapter only supports features supported by both the MQTT protocol and the Ably platform. So:
+
+* It only supports MQTT 3.1.1 clients only. Connection attempts using earlier protocol versions will be rejected
+* Publishing supports QoS 0 or 1. Subscribing only supports QoS 0
+* Session resumption is supported within the usual "Ably time limit of two minutes":/history-storage/storage
+* It doesn’t support any MQTT features that aren’t normally supported by Ably, such as @WILL@ messages, the @RETAIN@ flag or "wildcard channel subscriptions":/channels
+* It doesn't support any Ably features that aren't supported by the MQTT protocol, such as presence, history and push notifications (though presence may be supported soon). Of course, you can always use the Ably REST API in conjunction with the MQTT adapter to access features available over REST
+
+h2(#config). Configuration
+
+We recommend that you always make use of "Token Authentication over Basic Authentication":/auth when trying to connect from devices you may not trust. In addition, if you're using Basic Authentication you'll be required to use SSL to connect to our adapter to ensure the API key cannot be accessed by someone listening in.
+
+h3(#token-auth). Token Authentication
+
+For Token Authentication you'll be required to "create a token from your own servers":/auth/token. Once you have the token, you can simply pass it through when trying to connect to Ably as the connection's @username@, leaving the @password@ empty.
+
+If using Token Authentication, you can subscribe to a special topic, @[mqtt]tokenevents@, to get a warning when the current connection's token is about to expire. This will be a single message, sent 30 seconds before the token expires, with the 13 byte payload @expirywarning@.
+
+On receiving this, the client is recommended to get a new token, then disconnect and reconnect with the new token themselves. If this is not done, the server will abruptly disconnect the connection once the token expires.
+
+An example of this with the NodeJS "MQTT package":https://www.npmjs.com/package/mqtt would be:
+
+```[nodejs]
+var client = mqtt.connect('mqtts:mqtt.ably.io', options);
+client.subscribe('[mqtt]tokenevents');
+
+client.on('message', function(topic, message) {
+  if(topic === '[mqtt]tokenevents') {
+    /* some function that fetches a new token from an auth server */
+    getNewToken(function(err, newToken) {
+      if(err) {
+        // error handling
+        return;
+      }
+      /* reconnect with the new token */
+      client.end();
+      options.username = newToken;
+      client = mqtt.connect('mqtts:mqtt.ably.io', options);
+    });
+    return;
+  }
+});
+```
+
+h2(#pub-sub). Pub/Sub with MQTT
+
+"Ably's channels":/channels correlate to topics in MQTT. An example of how to publish and subscribe with the NodeJS "MQTT package":https://www.npmjs.com/package/mqtt would be as follows:
+
+```[nodejs]
+const mqtt = require('mqtt');
+const encoding = require('text-encoding');
+var decoder = new encoding.TextDecoder();
+var options = {
+  keepalive: 30,
+  username: '{{API_KEY_NAME}}', /* API key's name */
+  password: '{{API_KEY_SECRET}}', /* API key's secret */
+  port: 8883
+};
+
+var client = mqtt.connect('mqtts:mqtt.ably.io', options);
+
+client.on('message', function(topic, message) {
+  console.log(decoder.decode(message));
+});
+
+client.subscribe('my_channel', function (err) {
+  if (!err) {
+    client.publish('my_channel', 'my_message', { qos: 0 });
+  }
+});
+```
+
+h2(#decode-messages). Decode messages
+
+Any data published through or received by the MQTT adapter will be binary encoded, due to MQTT being a binary protocol. This means that you'll need to interpret the message to get the original contents out. For example, to interpret a message using Ably Realtime with JavaScript you'd need to do the following, using the "text-encoding NPM module's TextDecoder":https://www.npmjs.com/package/text-encoding to decode from binary to text:
+
+```[javascript] 
+var ably = new Ably.Realtime('{{API_KEY}}');
+var decoder = new TextDecoder();
+var channel = ably.channels.get('my_channel');
+
+channel.subscribe(function(message) {
+  var command = decoder.decode(message.data);
+});
+```
+
+Or, if you wish to decode messages received through MQTT, you can use the following in NodeJS:
+
+```[nodejs]
+const encoding = require('text-encoding');
+var decoder = new encoding.TextDecoder();
+var client = mqtt.connect('mqtts:mqtt.ably.io', options);
+
+client.on('connect', function () {
+  client.subscribe('my_channel');
+});
+
+client.on('message', function (topic, message) {
+  console.log(decoder.decode(message));
+});
+```
+
+In the above example, @command@ will now contain the message in its original string form.
+
+h2(#ssl-tls). SSL/TLS
+
+We support both SSL and non-SSL connections (the latter uses a different port, see above), but strongly recommend using SSL wherever possible. If you are not using SSL, note that the same restrictions apply as if you were using an Ably client without SSL. That is, "connection attempts using Basic Authentication (i.e. an API key) are disallowed":https://faqs.ably.com/error-code-40103-invalid-use-of-basic-auth-over-non-tls-transport, and any "namespaces which you've enabled 'require TLS' on":https://faqs.ably.com/what-is-a-channel-namespace-and-how-can-i-use-them will be inaccessible. "Contact us":https://ably.com/contact if this is a problem for you.
+
+To use the Ably MQTT protocol adapter, you'll need to ensure you correctly configure your MQTT client as follows:
+* Set the host to "mqtt.ably.io"
+* Set SSL / TLS to true and the port to 8883. (If your MQTT client does not support SSL, you should instead use port 1883, but in this case we disallow api-key auth—see "SSL usage note":#ssl below)
+* Set the keep alive time to somewhere between 15 and 60 seconds. (60s will maximize battery life, 15s will maximize responsiveness to network issues. It must not be any higher than 60s to avoid our load balancer terminating the TCP socket for inactivity)
+* If using an API key, set the username to the part of the API key before the colon, and the password to the part after the colon
+* If using a token, set the username to the token, and leave the password blank
+
+For example, in the NodeJS "MQTT package":https://www.npmjs.com/package/mqtt, you'd need to specify the following:
+
+```[nodejs]
+var options = {
+  keepalive: 30,
+  username: '{{API_KEY_NAME}}', /* API key's name */
+  password: '{{API_KEY_SECRET}}', /* API key's secret */
+  port: 8883
+};
+
+var client = mqtt.connect('mqtts:mqtt.ably.io', options);
+```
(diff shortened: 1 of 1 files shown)


content/protocols/mqtt.textile
```diff
@@ -0,0 +1,150 @@
...
+
+The MQTT adapter is our recommended way of interacting with Ably from devices which do not have a native Ably client library, such as Arduino platforms, C/C++ applications, and so on. Anyone who has previously been using Pubnub or Pusher client libraries for this purpose may want to consider switching to MQTT. Compared to the Pubnub protocol, using MQTT will result in better performance with lower latency. Compared to the Pusher protocol, MQTT will give you connection state recovery.
+
+We recommend you use MQTT to interact with Ably when:
+
+* Using devices which do not have a device-specific Ably Client Library SDK, such as Arduino platforms, C/C++ applications, and so on
+* If bandwidth is limited and you want to keep network traffic to a minimum
+* You want to keep network traffic to a minimum.
+
+Note, we do not currently support MQTT over websockets but let us know if you are interested.
```
m-hulbert, line 18: I'd remove this line as it's unnecessary.

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,150 @@
...
+
+h2(#limitations). Known limitations
```
m-hulbert, line 20: This section needs to be reviewed for readability/clarity, e.g. "though presence may be supported soon".

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,150 @@
...
+
+* It only supports MQTT 3.1.1 clients only. Connection attempts using earlier protocol versions will be rejected
+* Publishing supports QoS 0 or 1. Subscribing only supports QoS 0
+* Session resumption is supported within the usual "Ably time limit of two minutes":/history-storage/storage
+* It doesn’t support any MQTT features that aren’t normally supported by Ably, such as @WILL@ messages, the @RETAIN@ flag or "wildcard channel subscriptions":/channels
+* It doesn't support any Ably features that aren't supported by the MQTT protocol, such as presence, history and push notifications (though presence may be supported soon). Of course, you can always use the Ably REST API in conjunction with the MQTT adapter to access features available over REST
+
+h2(#config). Configuration
+
+We recommend that you always make use of "Token Authentication over Basic Authentication":/auth when trying to connect from devices you may not trust. In addition, if you're using Basic Authentication you'll be required to use SSL to connect to our adapter to ensure the API key cannot be accessed by someone listening in.
```
m-hulbert, line 32: This section would be better served under "authentication" I think. Also don't need to capitalise token and basic auth.

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,150 @@
...
+  keepalive: 30,
+  username: '{{API_KEY_NAME}}', /* API key's name */
+  password: '{{API_KEY_SECRET}}', /* API key's secret */
+  port: 8883
+};
+
+var client = mqtt.connect('mqtts:mqtt.ably.io', options);
+```
+
+This will connect using TLS through MQTT to Ably.
```
m-hulbert, line 150: This may make more sense under "Configuration".  WDYT?

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,150 @@
+---
+title: MQTT
+meta_description: "Any MQTT-enabled client can communication with the Ably service through the Ably MQTT protocol adapter. This is especially useful where an Ably client library SDK is not available for your language of choice."
+---
```
m-hulbert, line 4: Needs `language` here.

m-hulbert (changes requested):
* Lots of usage of "we" and "us" that should be removed.
* Client library SDKs --> Ably SDKs 
* I think this section is missing from the original page: https://ably.com/docs/mqtt#adapter-info
* Need to update links from existing MQTT page. 
* Similar to SSE, the content specific to MQTT should sit here in a logical place too (e.g. rewind)

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,178 @@
...
+
+MQTT is recommended to interact with Ably when:
+
+* Bandwidth is limited and you want to keep network traffic to a minimum
+* You want to keep network traffic to a minimum.
+
+h2(#limitations). Known limitations 
+
+A protocol adapter inevitably adds some latency. Using the adapter will be a little slower than using Ably client libraries. 
+Typically the impact is in the low milliseconds.
```
m-hulbert, line 23:
```suggestion
Using the MQTT adapter will be a little slower than using an Ably SDK, as an adapter adds some level of latency. Typically the impact is in the low milliseconds.
```

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,178 @@
...
+
+Behind the scenes, the adapter just uses the normal Ably service, so there is no problem with using MQTT and Ably client libraries side by side. You can mix and match as you like; for example, using MQTT on your IoT devices, but using the Ably Realtime API on your servers.
```
m-hulbert, line 25: I don't think this one is a limitation. Also a "client library" here too.

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,178 @@
...
+
+While the adapter can be useful for devices which need to use MQTT, there are many benefits to using the Ably client libraries (e.g. continuity guarantees, fallback host support, history, presence) require the use of the Ably client libraries. As a result, if an Ably client library is available for your platform, we recommend you consider using it instead of the protocol adapter.
```
m-hulbert, line 27:
```suggestion
While the adapter can be useful for devices which need to use MQTT, there are many benefits to using an Ably SDK, such as continuity guarantees, fallback host support, history and presence. As a result, if an Ably SDK is available for your platform, it is recommend you consider using it instead of the protocol adapter.
```

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,178 @@
...
+});
+```
+
+In the above example, @command@ will now contain the message in its original string form.
+
+h2(#ssl-tls). SSL/TLS
+
+Ably supports both SSL and non-SSL connections (the latter uses a different port, see above), but strongly recommend using SSL wherever possible. If you are not using SSL, note that the same restrictions apply as if you were using an Ably client without SSL. That is, "connection attempts using Basic Authentication (i.e. an API key) are disallowed":/auth/basic, and any "namespaces which you've enabled 'require TLS' on":/channels will be inaccessible. "Contact us":https://ably.com/contact if this is a problem for you.
+
+h2.(#rewind). Channel rewind
```
m-hulbert, line 159: Same comment here as for the SSE PR: need to include other items such as deltas here and look at wording this is channel options rather than parameters. We can catch up based on the channels section if that helps.

m-hulbert (changes requested):

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,236 @@
...
+
+h2(#channel-options). Channel options
+
+In an MQTT connection, it is also possible to specify "channel options":/channels/options as a query string in the connection URL, instead of as a qualifier on an individual channel name. In this case, the given channel options apply to all channel attachments associated with that connection.
+
+Currently both "rewind":/channels/options/rewind and "delta":/channels/options/delta channel options are supported.
+
+h3(#delta-mqtt). Delta with MQTT
+
+If subscribing to a channel in delta mode using MQTT or one of the protocol adaptors such as "SSE":/protocols/sse, then you will need to decode any received delta messages yourself. There are decoder libraries available to do this for several platforms; see the "download":LINK-ME section for details.
```
m-hulbert, line 166: Probably best to remove mention of SSE here now that they are in dedicated pages (as this content is likely applicable there also). 

I'd also suggest we remove the sentence about decoder libraries for now, as I can't see anything specific on the downloads page about deltas - only the list of MQTT libraries.

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,236 @@
...
+
+When subscribing without an Ably library, the channel @delta@ parameter must be specified using a "qualified channel name":/channels/options.
```
m-hulbert, line 168: The relevant info about channel names can be taken from [here](https://ably.com/docs/realtime/channels/channel-parameters/overview?lang=javascript#using-parameters-non-ably) rather than linking out, as this should be comprehensive for MQTT.

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,236 @@
...
+
+When subscribing without an Ably library, the channel @delta@ parameter must be specified using a "qualified channel name":/channels/options. 
+
+Some transports provide raw message payloads - that is, the content of the @data@ attribute of a @Message@ - without the accompanying metadata. That means that the recipient of the message does not have access to the @extras@ or @encoding@ attributes of the message that would ordinarily be used to decode delta message payloads.
+
+Examples of such transports are MQTT, and SSE in non-enveloped mode. In order to assist applications that use these transports, the @vcdiff@ decoder libraries can check for the @vcdiff@ header at the start of the message payload as an inexact method of determining whether or not the message is a regular message or a delta. Note that, in order to rely on that check, you need to know that that header will not be present in any valid (uncompressed) message in your app. No valid JSON value, for example, will match the @vcdiff@ header check, so it is safe to perform this sniffing on JSON message payloads.
```
m-hulbert, line 172: Suggest we make this one specific to MQTT again too, please. The info will need to be repeated in SSE where applicable.

m-hulbert (changes requested):

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,240 @@
...
+
+In the above example, @command@ will now contain the message in its original string form.
+
+h2(#ssl-tls). SSL/TLS
+
+Ably supports both SSL and non-SSL connections (the latter uses a different port, see above), but strongly recommend using SSL wherever possible. If you are not using SSL, note that the same restrictions apply as if you were using an Ably client without SSL. That is, "connection attempts using Basic Authentication (i.e. an API key) are disallowed":/auth/basic, and any "namespaces which you've enabled 'require TLS' on":/channels will be inaccessible. "Contact us":https://ably.com/contact if this is a problem for you.
+
+h2(#channel-options). Channel options
+
+In an MQTT connection, it is also possible to specify "channel options":/channels/options as a query string in the connection URL, instead of as a qualifier on an individual channel name. In this case, the given channel options apply to all channel attachments associated with that connection.
```
m-hulbert, line 160: I think this is mixing MQTT and SSE a little here. 

My understanding is that for MQTT and SSE you can specify options using a query string in the channel name qualifier. SSE then also has an additional option whereby you can specify the options as part of the connection URL, however that would apply the same options to every channel you attach to with that connection. 

Might be worth double checking this!

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,240 @@
...
+
+When subscribing without an Ably library, the channel @delta@ parameter must be specified using a qualified channel name. For the client libraries that do not currently expose the API, or transports that do not involve using an Ably library, a set of channel parameters can be expressed by including a query string with standard URL query syntax and encoding, within the qualifier part of a channel name. The qualifier part is in square brackets at the start of the channel name.
+
+To specify the parameter @foo@ with value @bar@ on channel @baz@, the qualified channel name would be @[?foo=bar]baz@. If the channel name already has a qualifier, such as @[meta]log@, then the query string follows the existing qualifier, as in @[meta?foo=bar]log@.
+
+Using this syntax with a non-supported Ably library means that channel parameters are specified for the lifetime of the Channel instance; in order to reference the same channel, but with different channel parameters, it is necessary to get a new Channel instance, using a qualified name that includes the new channel parameters.
```
m-hulbert, line 172: I think this part can be dropped from MQTT as it's referring to using an SDK that doesn't yet have full channel options support (it'll be covered in the channel options page instead).

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,240 @@
...
+
+In an MQTT connection, it is also possible to specify "channel options":/channels/options as a query string in the connection URL, instead of as a qualifier on an individual channel name. In this case, the given channel options apply to all channel attachments associated with that connection.
+
+Currently both "rewind":/channels/options/rewind and "delta":/channels/options/delta channel options are supported.
+
+h3(#delta-mqtt). Delta with MQTT
+
+If subscribing to a channel in delta mode using MQTT then you will need to decode any received delta messages yourself.
+
+When subscribing without an Ably library, the channel @delta@ parameter must be specified using a qualified channel name. For the client libraries that do not currently expose the API, or transports that do not involve using an Ably library, a set of channel parameters can be expressed by including a query string with standard URL query syntax and encoding, within the qualifier part of a channel name. The qualifier part is in square brackets at the start of the channel name.
```
m-hulbert, line 168: Might be worth reading and checking this paragraph. I think this is also trying to cover instances of channel options (parameters in this context) being used in SDKs that didn't yet support them. That adds confusion to this section trying to focus on MQTT.

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,240 @@
...
+    /* Process decoded data */
+    console.log(data);
+  });
+
+  client.subscribe(`[?delta=vcdiff]${channelName}`);
+```
+
+h3(#rewind-mqtt). Rewind with MQTT
+
+To use Rewind with MQTT specify the @rewind@ channel option with the value @"1"@ using a querystring parameter, where it will apply to all channels.
```
m-hulbert, line 219: "Rewind" --> "rewind"

This also reads like you can only use rewind with `1` which isn't true. 

For MQTT this also wouldn't apply to all channels (that's the SSE connection string option) - it would only apply to the channels where you qualify the names with a query string.

m-hulbert (changes requested):
Few more bits on the channel options section that I think are conflating a few different transports and this one is good!

content/protocols/mqtt.textile
Sapples: Confirmed with Simpson Woolf.

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,238 @@
...
+
+In the above example, @command@ will now contain the message in its original string form.
+
+h2(#ssl-tls). SSL/TLS
+
+Ably supports both SSL and non-SSL connections (the latter uses a different port, see above), but strongly recommends using SSL wherever possible. If you are not using SSL, note that the same restrictions apply as if you were using an Ably client without SSL. That is, "connection attempts using Basic Authentication (i.e. an API key) are disallowed":/auth/basic, and any "namespaces which you've enabled 'require TLS' on":/channels will be inaccessible. "Contact us":https://ably.com/contact if this is a problem for you.
+
+h2(#channel-options). Channel options
+
+In an MQTT connection you can specify "channel options":/channels/options with a query string in the channel name qualifier.
```
m-hulbert, line 160:
```suggestion
In an MQTT connection you can specify "channel options":/channels/options with a query string in the channel name qualifier. 

The qualifier part is in square brackets at the start of the channel name. To specify the channel option @foo@ with value @bar@ on channel @baz@, the qualified channel name would be @[?foo=bar]baz@. If the channel name already has a qualifier, such as @[meta]log@, then the query string follows the existing qualifier, as in @[meta?foo=bar]log@.
```

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,238 @@
...
+
+h2(#ssl-tls). SSL/TLS
+
+Ably supports both SSL and non-SSL connections (the latter uses a different port, see above), but strongly recommends using SSL wherever possible. If you are not using SSL, note that the same restrictions apply as if you were using an Ably client without SSL. That is, "connection attempts using Basic Authentication (i.e. an API key) are disallowed":/auth/basic, and any "namespaces which you've enabled 'require TLS' on":/channels will be inaccessible. "Contact us":https://ably.com/contact if this is a problem for you.
+
+h2(#channel-options). Channel options
+
+In an MQTT connection you can specify "channel options":/channels/options with a query string in the channel name qualifier. 
+
+Currently both "rewind":/channels/options/rewind and "delta":/channels/options/delta channel options are supported.
```
m-hulbert, line 162:
```suggestion
The "rewind":/channels/options/rewind and "delta":/channels/options/delta channel options are supported with MQTT.
```

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,238 @@
...
+
+h3(#delta-mqtt). Delta with MQTT
+
+If subscribing to a channel in delta mode using MQTT then you will need to decode any received delta messages yourself.
+
+When subscribing without an Ably library, the channel @delta@ parameter must be specified using a qualified channel name. The qualifier part is in square brackets at the start of the channel name.
+
+To specify the parameter @foo@ with value @bar@ on channel @baz@, the qualified channel name would be @[?foo=bar]baz@. If the channel name already has a qualifier, such as @[meta]log@, then the query string follows the existing qualifier, as in @[meta?foo=bar]log@.
```
m-hulbert, line 170:
```suggestion
```

content/protocols/mqtt.textile
```diff
@@ -0,0 +1,238 @@
...
+    /* Process decoded data */
+    console.log(data);
+  });
+
+  client.subscribe(`[?delta=vcdiff]${channelName}`);
+```
+
+h3(#rewind-mqtt). Rewind with MQTT
+
+To use rewind with MQTT you specify the @rewind@ channel option with a numerical value equal to the position you want to attach to, for example @"1"@ to rewind 1 message previous. This will apply to all channels specified in the query string.
```
m-hulbert, line 217:
```suggestion
The @rewind@ channel option enables a client to specify where to start an attachment from. This can be a point in time in the past, or a given number of messages.
```

m-hulbert (changes requested):
Couple of final suggestions for the options section 🙂

m-hulbert (approved):
LGTM

PR #1785, opened 2022-12-01
[EDX-501] add sse

Sapples:

## Description

Add a page covering the SSE protocol.

* [Jira ticket](https://ably.atlassian.net/browse/EDX-501).

## Review

Check /protocols/sse in the new toolchain.

### DIFF
--- a/content/protocols/sse.textile
+++ b/content/protocols/sse.textile
@@ -1,32 +1,32 @@
 ---
 title: SSE
-meta_description: "Ably provides support for Server-Sent Events (SSE). This is useful for where browser clients support SSE, and the use case does not require or support the resources used by the Ably client library SDK."
+meta_description: "Ably provides support for Server-Sent Events (SSE). This is useful for where browser clients support SSE, and the use case does not require or support the resources used by the Ably SDK."
+languages:
+ - javascript
 ---
 
-The Ably SSE and raw HTTP streaming API provides a way to get a realtime stream of events from Ably in circumstances where using a full Ably Realtime client library, or even an "MQTT":/protocols/mqtt library, is impractical.
+The Ably SSE and raw HTTP streaming API provides a way to get a realtime stream of events from Ably in circumstances where using a full Ably Realtime SDK, or even an "MQTT":/protocols/mqtt library, is impractical.
 
 HTTP streaming allows for a request from a client to be held by a server, allowing it to push data to the client without further requests. This, much like WebSockets, help avoid the overhead involved in normal HTTP requests. "Server-sent events":https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events (SSE) provide a thin layer on top of HTTP streaming. A common use of SSE is through the use of "the EventSource API":https://developer.mozilla.org/en-US/docs/Web/API/EventSource in all modern web browsers.
 
 It is subscribe-only: you can not interact with the channel, including to publish, enter presence, query the presence set, attach and detach from channels (without closing and re-opening the stream), or anything else.
 
-Customers who do not want to use a client library on platforms that support SSE, and only require simple subscribe-only streams, may choose to use SSE because it's an open standard, simple, and requires no SDKs on the client-side. HTTP Streaming may be considered on platforms without an SSE client. However, where possible, we strongly recommend the use of one of our "Realtime client libraries":https://ably.com/download, which provide "more features and higher reliability":/basics/use-ably, and the full use of our normal realtime messaging API.
+Customers who do not want to use an SDK on platforms that support SSE, and only require simple subscribe-only streams, may choose to use SSE because it's an open standard, simple, and requires no SDKs on the client-side. HTTP Streaming may be considered on platforms without an SSE client. However, where possible, it is strongly recommend to use of one of our "Realtime SDKs":https://ably.com/download, which provide "more features and higher reliability":/basics/use-ably, and the full use of our normal realtime messaging API.
 
 h2(#when-to-use). When to use the SSE adapter
 
-At Ably we believe in protocol interoperability, and whilst we believe our own native realtime protocol in a lot of cases offers the right balance between performance, portability and quality of service, we recognize that developers should be given the choice to choose the right protocol for their applications based on their own use case. As such, we offer support for SSE in addition to a number of other protocols.
-
 SSE allows for a request from a client to be held by a server, allowing it to push data to the client without further requests. This, much like WebSockets, help avoid the overhead involved in normal HTTP requests. Server-sent events (SSE) is another method of achieving this for web pages, usually done through the use of "EventSource":/https://developer.mozilla.org/en-US/docs/Web/API/EventSource.
 
-This is useful when dealing with devices with extremely limited memory, where using one of the Ably client libraries would not be possible. This is possible largely due to it being a subscribe-only method, without the many additional features Ably provides.
+This is useful when dealing with devices with extremely limited memory, where using one of the Ably SDKs would not be possible. This is possible largely due to it being a subscribe-only method, without the many additional features Ably provides.
 
-We recommend you consider SSE if:
+SSe is recommended if:
 
 * Ably does not offer a native library that supports your target platform
-* Ably offers a REST-only client library for your target platform, but you need a realtime client
+* Ably offers a REST-only SDK for your target platform, but you need a realtime client
 * You have stringent memory restrictions
 * You only wish to subscribe to events on channels
 
-We recommend you use Ably client libraries and our realtime protocol if:
+The Ably SDKs and realtime protocol are recommended if:
 
 * You want a high quality of service and high availability even during significant events such DNS failure or network partitions
 * You want access to features such as Publishing, Presence, History, Push notifications, automatic payload encoding, symmetric encryption
@@ -34,7 +34,7 @@ We recommend you use Ably client libraries and our realtime protocol if:
 
 h2(#config). Configuration
 
-SSE is incredibly simple to get started with. The code sample below provides an example of how to use it with Ably.
+SSE is quick to get started with, the following code sample provides an example of how to use it with Ably:
 
 ```[javascript](code-editor:sse/sse)
 var apiKey ='{{API_KEY}}';
@@ -49,7 +49,7 @@ eventSource.onmessage = function(event) {
 
 h2(#auth). Authentication
 
-It is possible to use either "basic auth":/auth/basic (with an "API key":https://faqs.ably.com/what-is-an-app-api-key) or "token auth":/auth/token (using a "token issued from your server":/auth/token) with SSE. We recommend token auth on the client side for "security reasons":/auth, so you have control over who can connect. Basic auth, while lacking this control, is simpler (it doesn't require you to run an auth server), and you don't have to worry about the client obtaining a new token when the old one expires.
+It is possible to use either "basic auth":/auth/basic (with an "API key":/getting-started/setup) or "token auth":/auth/token (using a "token issued from your server":/auth/token) with SSE. It's recommended to use token auth on the client side for "security reasons":/auth, so you have control over who can connect. Basic auth, while lacking this control, is simpler (it doesn't require you to run an auth server), and you don't have to worry about the client obtaining a new token when the old one expires.
 
 If using basic auth, you can use a querystring parameter of @key@ or an @Authorization: Basic <base64-encoded key>@ header. If using token auth, you can use an @accessToken@ querystring parameter or an @Authorization: Bearer <base64-encoded token>@ header. See "REST API authentication":/auth for more information.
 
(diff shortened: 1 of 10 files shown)


content/protocols/sse.textile
```diff
@@ -0,0 +1,113 @@
...
+
+Customers who do not want to use a client library on platforms that support SSE, and only require simple subscribe-only streams, may choose to use SSE because it's an open standard, simple, and requires no SDKs on the client-side. HTTP Streaming may be considered on platforms without an SSE client. However, where possible, we strongly recommend the use of one of our "Realtime client libraries":https://ably.com/download, which provide "more features and higher reliability":/basics/use-ably, and the full use of our normal realtime messaging API.
+
+h2(#when-to-use). When to use the SSE adapter
+
+At Ably we believe in protocol interoperability, and whilst we believe our own native realtime protocol in a lot of cases offers the right balance between performance, portability and quality of service, we recognize that developers should be given the choice to choose the right protocol for their applications based on their own use case. As such, we offer support for SSE in addition to a number of other protocols.
```
m-hulbert, line 16: This paragraph has a bit of a marketing spin on it. Is it necessary or can we drop it do you think?

content/protocols/sse.textile
```diff
@@ -0,0 +1,113 @@
+---
+title: SSE
+meta_description: "Ably provides support for Server-Sent Events (SSE). This is useful for where browser clients support SSE, and the use case does not require or support the resources used by the Ably client library SDK."
```
m-hulbert, line 3: Need a `language` tag here.

content/protocols/sse.textile
```diff
@@ -0,0 +1,113 @@
...
+
+We recommend you use Ably client libraries and our realtime protocol if:
+
+* You want a high quality of service and high availability even during significant events such DNS failure or network partitions
+* You want access to features such as Publishing, Presence, History, Push notifications, automatic payload encoding, symmetric encryption
+* You want first class browser support with WebSockets
+
+h2(#config). Configuration
+
+SSE is incredibly simple to get started with. The code sample below provides an example of how to use it with Ably.
```
m-hulbert, line 37: Avoid subjective language here please: "incredibly simple".

m-hulbert (changes requested):
Just a few generic pointers for the whole page:

* Avoid using "we" in the text please.
* "Client library SDK" etc. should change to "Ably SDK"
* I think you need to check for links to the existing page and replace those.
* There's a note about adding feature-specific content that applies to SSE to this page, such as rewind which is missing still.

content/protocols/sse.textile
```diff
@@ -0,0 +1,134 @@
...
+}
+
+connectToAbly();
+```
+
+An important thing to note here is that the EventSource API tries to auto-reconnect and re-subscribe to the SSE endpoint when any error occurs, even the token expiry error like in this case. This means that upon manually re-subscribing to the SSE endpoint with a new token, there will be two active subscriptions to the endpoint - one with the old token which would continue to throw an error due to expired credentials and another with the new token. Hence, it is important to close the previous @EventSource@ subscription using @eventSource.close()@ before re-subscribing with the new token as shown in the snippet above.
+
+You can take a look at a "demo app":https://sse-token-auth.glitch.me and "a complete code example for implementing message continuity in an SSE subscription when using token auth":https://glitch.com/edit/#!/sse-token-auth.
+
+h2(#channel-parameters). Channel parameters
```
m-hulbert, line 115: Can we call this one "Channel options" please? 

I think you can use Deltas with SSE too, so can you check all the other features?

m-hulbert (changes requested):

content/protocols/sse.textile
```diff
@@ -0,0 +1,134 @@
...
+
+It is subscribe-only: you can not interact with the channel, including to publish, enter presence, query the presence set, attach and detach from channels (without closing and re-opening the stream), or anything else.
+
+Customers who do not want to use an SDK on platforms that support SSE, and only require simple subscribe-only streams, may choose to use SSE because it's an open standard, simple, and requires no SDKs on the client-side. HTTP Streaming may be considered on platforms without an SSE client. However, where possible, it is strongly recommend to use an "Ably SDK":https://ably.com/download, as they provide "more features and higher reliability":/basics/use-ably, and the full use of our normal realtime messaging API.
+
+h2(#when-to-use). When to use the SSE adapter
+
+This is useful when dealing with devices with extremely limited memory, where using one of the Ably SDKs would not be possible. This is largely due to it being a subscribe-only method, without the many additional features Ably provides.
```
m-hulbert, line 18: I think this paragraph has been orphaned by the removal of the duplicate one, so it now looks a little odd starting "this is useful" - what is useful? :-)

content/protocols/sse.textile
```diff
@@ -0,0 +1,134 @@
...
+
+When a connection is closed as a result of any error (that is, it's not just a dropped connection), then the @error@ event will occur on the @EventSource@ instance, and the data attribute of the event will contain an Ably error body with the information about the nature of error. In the case of a token error - that is an error arising from a problem with the auth token - the code in the error body will indicate that. Token errors have a code in the range @40140 <= code < 40150@. In such cases, the authentication can be retried with a new @accessToken@.
+
+In the future we plan to send an event on the connection that indicates that the token will expire imminently, which will allow a new connection to be established prior to the closure of the previous connection.
+
+h4(#specifying-lasteventid). Specifying the lastEventId
+
+Each message received will have a @lastEventId@ attribute containing the last id of any message received on the connection. When constructing a new connection, this value can be specified as a @lastEvent@ param in the URL.
+
+The following example of implementing message continuity with token auth:
```
m-hulbert, line 74:
```suggestion
The following is an example of implementing message continuity with token auth:
```

content/protocols/sse.textile
```diff
@@ -0,0 +1,134 @@
...
+}
+
+connectToAbly();
+```
+
+An important thing to note here is that the EventSource API tries to auto-reconnect and re-subscribe to the SSE endpoint when any error occurs, even the token expiry error like in this case. This means that upon manually re-subscribing to the SSE endpoint with a new token, there will be two active subscriptions to the endpoint - one with the old token which would continue to throw an error due to expired credentials and another with the new token. Hence, it is important to close the previous @EventSource@ subscription using @eventSource.close()@ before re-subscribing with the new token as shown in the snippet above.
+
+You can take a look at a "demo app":https://sse-token-auth.glitch.me and "a complete code example for implementing message continuity in an SSE subscription when using token auth":https://glitch.com/edit/#!/sse-token-auth.
+
+h2(#channel-options). Channel options
```
m-hulbert, line 113: I think this section is still missing some of the information from [this](https://ably.com/docs/realtime/channels/channel-parameters/deltas?lang=javascript#using-deltas-non-ably) page relevant to SSE, as well as [this one](https://ably.com/docs/realtime/channels/channel-parameters/overview?lang=javascript#using-parameters-non-ably).

m-hulbert (changes requested):
Just a couple more points on this one, similar to the delta comments on MQTT.

content/protocols/sse.textile
```diff
@@ -0,0 +1,230 @@
...
+
+In an SSE connection, it is also possible to specify "channel options":/channels/options as a query string in the connection URL, instead of as a qualifier on an individual channel name. In this case, the given channel options apply to all channel attachments associated with that connection.
```
m-hulbert, line 115: I think this is facing a similar issue to my comment [here](https://github.com/ably/docs/pull/1784#discussion_r1081566456). Once you've checked that what I wrote is correct then we can update the intro here for SSE too.

content/protocols/sse.textile
```diff
@@ -0,0 +1,230 @@
...
+
+When subscribing without an Ably library, the channel @delta@ parameter must be specified using a qualified channel name. For the client libraries that do not currently expose the API, or transports that do not involve using an Ably library, a set of channel parameters can be expressed by including a query string with standard URL query syntax and encoding, within the qualifier part of a channel name. The qualifier part is in square brackets at the start of the channel name.
+
+To specify the parameter @foo@ with value @bar@ on channel @baz@, the qualified channel name would be @[?foo=bar]baz@. If the channel name already has a qualifier, such as @[meta]log@, then the query string follows the existing qualifier, as in @[meta?foo=bar]log@.
+
+Using this syntax with a non-supported Ably library means that channel parameters are specified for the lifetime of the Channel instance; in order to reference the same channel, but with different channel parameters, it is necessary to get a new Channel instance, using a qualified name that includes the new channel parameters.
```
m-hulbert, line 127: Same comment as [here](https://github.com/ably/docs/pull/1784#discussion_r1081569586) if that turns out to be correct 🙂

content/protocols/sse.textile
```diff
@@ -0,0 +1,230 @@
...
+
+In an SSE connection, it is also possible to specify "channel options":/channels/options as a query string in the connection URL, instead of as a qualifier on an individual channel name. In this case, the given channel options apply to all channel attachments associated with that connection.
+
+Currently both "rewind":/channels/options/rewind and "delta":/channels/options/delta channel options are supported.
+
+h3(#delta-sse). Delta with SSE
+
+If subscribing to a channel in delta mode using SSE then you will need to decode any received delta messages yourself.
+
+When subscribing without an Ably library, the channel @delta@ parameter must be specified using a qualified channel name. For the client libraries that do not currently expose the API, or transports that do not involve using an Ably library, a set of channel parameters can be expressed by including a query string with standard URL query syntax and encoding, within the qualifier part of a channel name. The qualifier part is in square brackets at the start of the channel name.
```
m-hulbert, line 123: Same comment as [here](https://github.com/ably/docs/pull/1784#discussion_r1081573392).

content/protocols/sse.textile
```diff
@@ -0,0 +1,230 @@
...
+      }
+
+      /* Process decoded data */
+      console.log(data);
+  };
+```
+
+h3(#rewind-sse). Rewind with SSE
+
+To use Rewind with SSE specify the @rewind@ channel option with the value @"1"@ using a querystring parameter, where it will apply to all channels.
```
m-hulbert, line 209: This makes it appear that you can only use rewind with `1` which isn't true. This is also contradictory as the following two examples show that you can apply it to all channels, or just one depending on how you specify the parameters. Also don't need to capitalise "rewind" here.

m-hulbert (changes requested):
Just a few points to tidy up on the channel options here, please.

content/protocols/sse.textile
```diff
@@ -0,0 +1,228 @@
...
+
+You can take a look at a "demo app":https://sse-token-auth.glitch.me and "a complete code example for implementing message continuity in an SSE subscription when using token auth":https://glitch.com/edit/#!/sse-token-auth.
+
+h2(#channel-options). Channel options
+
+In an SSE connection you can specify "channel options":/channels/options with a query string in the channel name qualifier. You can also specify channel options as a query string in the connection URL, instead of as a qualifier on an individual channel name. In this case, the given channel options apply to all channel attachments associated with that connection.
```
m-hulbert, line 115:
```suggestion
In an SSE connection you can specify "channel options":/channels/options in two different ways:

* with a query string in the channel name qualifier
* as a query string in the connection URL

If specified as part of the connection URL the options will apply to all channels that connection attaches to. Using a channel name qualifier enables channel options to be applied to individual channels. Setting options using a channel name qualifier can also be used to override the options set in the connection URL for specific channels.

A channel name qualifier is the square brackets at the start of the channel name. To specify the channel option @foo@ with value @bar@ on channel @baz@, the qualified channel name would be @[?foo=bar]baz@. If the channel name already has a qualifier, such as @[meta]log@, then the query string follows the existing qualifier, as in @[meta?foo=bar]log@.
```

content/protocols/sse.textile
```diff
@@ -0,0 +1,228 @@
...
+
+Currently both "rewind":/channels/options/rewind and "delta":/channels/options/delta channel options are supported.
```
m-hulbert, line 117:
```suggestion
The "rewind":/channels/options/rewind and "delta":/channels/options/delta channel options are supported with SSE.
```

content/protocols/sse.textile
```diff
@@ -0,0 +1,228 @@
...
+
+h3(#delta-sse). Delta with SSE
+
+If subscribing to a channel in delta mode using SSE then you will need to decode any received delta messages yourself.
+
+When subscribing without an Ably library, the channel @delta@ parameter must be specified using a qualified channel name. The qualifier part is in square brackets at the start of the channel name.
+
+To specify the parameter @foo@ with value @bar@ on channel @baz@, the qualified channel name would be @[?foo=bar]baz@. If the channel name already has a qualifier, such as @[meta]log@, then the query string follows the existing qualifier, as in @[meta?foo=bar]log@.
```
m-hulbert, line 125:
```suggestion
```

content/protocols/sse.textile
```diff
@@ -0,0 +1,228 @@
...
+      }
+
+      /* Process decoded data */
+      console.log(data);
+  };
+```
+
+h3(#rewind-sse). Rewind with SSE
+
+To use rewind with SSE you specify the @rewind@ channel option with a numerical value equal to the position you want to attach to, for example @"1"@ to rewind 1 message previous.This will apply to all channels specified in the query string.
```
m-hulbert, line 207:
```suggestion
The @rewind@ channel option enables a client to specify where to start an attachment from. This can be a point in time in the past, or a given number of messages.
```

m-hulbert (changes requested):
Just a few suggestions to highlight the channel options methods for SSE and note they can be used together.

m-hulbert (approved):
Nice one 🙂

PR #1797, opened 2022-12-06
[EDX-614] Redirect tutorials and client lib development references

kennethkalmer:

## Description

Now that the new toolchain is available on docs.ably.com as well, we have some old nanoc toolchain artefacts to deal with as well:

* Tutorials should be accessed from ably.com/tutorials
* Client library development guides have moved to sdk.ably.com in #1608

See EDX-614

This also addresses [these comments](https://github.com/ably/docs/pull/1608#issuecomment-1266837897) from @m-hulbert & @QuintinWillison in #1608 

>> Any objections if we get redirects set up with the web team in the future to remove these pages from docs entirely? Do we need to preserve the deep links of the features spec for anything?

> I would love for that to happen and I had presumed that this would be the next logical step, after this pull request has landed. My intention here was to get the 'soft redirects' in place. What I would want to see from future redirects, of course, is that they also carry over the HTML element id anchors too. Some of these documents are quite long and that was one of my goals, to keep that rich, deep linking carrying forwards.

## Review

Testing is quite simple on the review app

Visiting https://ably-docs-edx-614-redir-vohrir.herokuapp.com/tutorials should redirect to ably.com/tutorials
Visiting https://ably-docs-edx-614-redir-vohrir.herokuapp.com/client-lib-development-guide should redirect to sdk.ably.com (as per #1608)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -151,6 +151,15 @@ jobs:
             ./bin/assert-success.sh /docs/quick-start-guide
             ./bin/assert-redirect.sh /docs/quick-start-guide/ \
                                      https://localhost/docs/quick-start-guide
+      - run:
+          name: Test website redirects
+          command: |
+            ./bin/assert-redirect.sh /tutorials https://ably.com/tutorials
+      - run:
+          name: Test client library redirects
+          command: |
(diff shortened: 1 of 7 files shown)


MymmiJ (approved):
Nice.

PR #1814, opened 2022-12-12
DRAFT: Revert "Revert "EDX-606 - Add final touches from the UX/UI feedback (…

MymmiJ:
@snikidev Can you take a look at this please? Currently makes the language dropdown menu disappear. If you can fix that and still match EDX-606 we can merge.


snikidev:
reopened in #1816

PR #1821, opened 2022-12-15
hotfix: language bar bug

snikidev:

## Description

Fixes language select dropdown position bug.


MymmiJ (approved):
We REALLY need to revisit these variants, I think they're doing completely different things now

PR #1826, opened 2022-12-22
[EDX-595] In page nav should scroll along with the content

snikidev:

## Description

* [Jira ticket](https://ably.atlassian.net/browse/EDX-595).

### DIFF
--- a/setupTests.js
+++ b/setupTests.js
@@ -9,6 +9,8 @@ class ResizeObserver {
 
 window.ResizeObserver = ResizeObserver;
 
+window.HTMLElement.prototype.scrollIntoView = jest.fn();
+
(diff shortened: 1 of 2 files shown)


MymmiJ (commented):
This can't be done in this way; we need a different approach completely.

If it works (on Chrome), it will drag the main article scroll down as though it were a consecutive element. Currently it doesn't work on Chrome.

get-current-header-id also can't be where we put this logic, as it does work but drags the main article scroll down also.

It can't be done by replaceState because this doesn't work in Chrome.

This will need to be a functionality of the RightSidebar and it will need to be a bit more complicated than this I think.

MymmiJ:
Sadly I don't think this totally fulfils the ticket requirements, closing!

PR #1831, opened 2022-12-29
[EDX-539] CSS alteration to prevent multiple definition list lines showing alongside each other

MymmiJ:

## Description

* [Jira ticket](https://ably.atlassian.net/browse/EDX-539).

## Review

Check this page:

* [Page to review](https://ably-docs-next-staging.herokuapp.com/docs/api/rest-sdk?lang=php#client-options)

It looks like this live, note the gap between definition lists (may need to click to expand):

Screenshot 2022-12-29 at 15 14 14

Locally it should look like this:

Screenshot 2022-12-29 at 15 15 28

See comments for how this fixes the issue and why.

### DIFF
--- a/src/components/blocks/list/Dl/index.tsx
+++ b/src/components/blocks/list/Dl/index.tsx
@@ -20,6 +20,29 @@ const StyledDl = styled.dl`
   margin-bottom: 20px;
   font-size: 16px;
 
+  /**
+  * This :only-child directive is present to ensure that multiple lines are not shown in the situation
+  * where there is a <dl> inside a <div> that is selected based on language type in between two other divs, i.e.
+  * <dl>
+  *   <dt></dt>
+  *   <dd></dd>
+  * </dl>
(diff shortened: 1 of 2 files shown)


kennethkalmer:
I can't help but wonder, and this isn't for this PR, that if we had the `lang` attribute on the `dl` rather than a parent `div` we could sort out this grid rendering issue as well:

Constructor 2022-12-29 17-31-08

kennethkalmer (approved):

MymmiJ:
> I can't help but wonder, and this isn't for this PR, that if we had the `lang` attribute on the `dl` rather than a parent `div` we could sort out this grid rendering issue as well:
> 
> Constructor 2022-12-29 17-31-08

This does make the most sense! I think the other thing we would have to change would be the part of the codebase that merges consecutive <dl> elements together, but presumably just check to see if the <dl> has a lang.

PR #1832, opened 2022-12-29
[EDX-562] Make the on this page selector fixed

MymmiJ:

## Description

Because the `overflow-x` of the main article container was set to `hidden` at some point, `position: sticky;` no longer worked. Removing `overflow-x-hidden` from the main article container fixes, but has some complications. Hopefully one day we can have sticky-y and sticky-x separately, but until then we just have to use `fixed` for this purpose. `overflow-x-clip` might also work but is Tailwind v3.

Fixed actually works quite a bit nicer so suggest we stick with this.

The changes to `CopyLink` and to `Dl/index.tsx` are to restrict the width of the article to resolve some closely related issues, but are not essential.

Ticket: EDX-562

## Review

* [Page to review](http://localhost:8000/docs/api/rest-sdk?lang=php#constructor)


aralovelace (approved):

aralovelace (approved):
code LG
I checked the page it looks great as well

PR #1833, opened 2023-01-04
[CHORE]: Remove need for importing React in every .tsx or .jsx file

MymmiJ:
## Description

Since we are using React 18, there is no longer a need to `import React from "react";` at the start of every .jsx/.tsx file. This change is to the configuration to allow babel and typescript to compile correctly without `import React from "react";`.

There is also:
* a minor typo/link correction in `tools.textile`.
* PrismJS (a code highlighting tool that we no longer use) has been uninstalled.

## Review

Ensure that docs still builds, make sure PageTitle component isn't broken.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -41,6 +41,7 @@
       "plugins": ["@typescript-eslint"],
       "rules": {
         // disable the rule for all files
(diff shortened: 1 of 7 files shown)


aralovelace (approved):
LGTM

PR #1836, opened 2023-01-05
DRAFT: Markdown support demo

MymmiJ:
__DO NOT MERGE__

## Description

Demo of markdown support. Working for existing .md files:
http://localhost:8000/docs/adapters/pubnub
http://localhost:8000/docs/adapters/pusher

Fully functioning, but some small things to tweak if we want to push through with this:
* Define what needs to be included, and what should never be included (e.g. exclude all READMEs, maybe pubnub/pusher)
* Implement a strategy for resolving markdown over textile
* Fix TypeScript errors that occur with `npm run debug:textile` to allow markdown translation.
* Anything else?

### DIFF
--- a/content/adapters/markdown-demo.md
+++ b/content/adapters/markdown-demo.md
@@ -93,11 +93,13 @@ Does [this](#definition-lists) link to definition lists?
 
 ### Complex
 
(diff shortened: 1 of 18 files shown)


m-hulbert:
Closing until Markdown is revisited.

PR #1838, opened 2023-01-13
Token revocation: add warning that issuedBefore does not work with normal tokens

SimonWoolf:
core-feature/auth and rest-api -- add warnings for a bug we have 

context: https://ably-real-time.slack.com/archives/C8SPU4589/p1673614895790359?thread_ts=1673356100.438359&cid=C8SPU4589


SimonWoolf:
Sure.

(The point of asking people to contact support is to get a sense when someone is affected, and be able to tell them as and when we fix it, but if you'd rather not have that, shrug)

PR #1888, opened 2023-04-04
[EDU-1363] Show SDK Version in Language Dropdown

artismarti:
## Description
This PR adds the SDK version label to the language dropdown.

[EDU-1363](https://ably.atlassian.net/browse/EDU-1363)

image

[EDU-1363]: https://ably.atlassian.net/browse/EDU-1363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/maps/language.tsx
```diff
@@ -9,51 +9,51 @@ type LanguageSyntaxHighlighterNames = {
...
-  nodejs: 'Node.js',
-  realtime_nodejs: 'Node.js',
-  rest_nodejs: 'Node.js',
-  flutter: 'Flutter',
-  realtime_flutter: 'Flutter',
-  rest_flutter: 'Flutter',
-  android: 'Android',
-  realtime_android: 'Android',
-  rest_android: 'Android',
+  javascript: 'JavaScript v1.2',
```
aralovelace, line 12: so this is means when there is a new version of the code from the texttile they need to update here as well?

aralovelace (changes requested):
the infobox displays the version number as well is this okay? or does it need to be splitted in the infobox so it will only display the language?

Screenshot 2023-04-05 at 12 09 16

src/maps/language.tsx
artismarti: Yeah. The ticket said they were not fussed where the labels were, so I added it to the most convenient place I could think of.

artismarti:
> the infobox displays the version number as well is this okay? or does it need to be splitted in the infobox so it will only display the language?
> 
Good point! @m-hulbert wdyt? I think it's good for it to remain, but it's pretty straightforward to remove.

aralovelace (approved):

PR #1915, opened 2023-05-23
Fix token limits link

cruickshankpg:
the link was blindly copy pasted probably by me and refers to two places in the file


ben-pickering (approved):
LGTM

PR #1919, opened 2023-05-23
Update table of contents for pre-launch

m-hulbert:
## Description

This PR update the table of contents in preparation for launch. The main points are that it:

* Removes sections left in to test against
* Splits the 'Products' section into component parts

### DIFF
--- a/data/yaml/page-furniture/sidebar.yaml
+++ b/data/yaml/page-furniture/sidebar.yaml
@@ -70,6 +70,17 @@ items:
         link: /docs/storage-history/storage
       - label: "History"
         link: /docs/storage-history/history
+  - label: "Push Notifications"
+    link: ""
+    items:
+      - label: "Overview"
+        link: /docs/general/push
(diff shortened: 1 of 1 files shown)


data/yaml/page-furniture/sidebar.yaml
```diff
@@ -100,175 +157,17 @@ items:
         link: /docs/asset-tracking/example-apps
       - label: "Using the SDKs"
         link: /docs/asset-tracking/using-the-sdks
-  - label: "Ably accounts and apps"
-    link: ""
-    items:
-      - label: "Programmatic management with Control API"
```
tbuntel, line 106: Do we have content for Terraform? Might that go here?

m-hulbert: We don't have anything in the docs yet, no. But agreed that this would likely be the best place for it when we do.

tbuntel (approved):

PR #1935, opened 2023-06-21
[EDU-1403] Add XML as a code block option

m-hulbert:
## Description

This PR fixes an issue reported with the page `general/push/activate-subscribe. It was failing to load when trying to render an XML code block. I've added XML as a language option for code blocks.


artismarti (approved):
🐑 it!

PR #1942, opened 2023-06-28
Rendering and console fixes

dpiatek:
Fix multiple issues when rendering docs, see each commit for details.


kennethkalmer (commented):
Code looks good, doing some manual testing (and rebuilding the review app configured against `https://ably-dev.com`)

kennethkalmer (approved):
Looks good from what I can see, learned a lot from the code review. I would get another final opinion before merging though, just to be safe :)

artismarti (approved):
This was great! :)

PR #1947, opened 2023-07-04
Remove gatsby-plugin-manifest

dpiatek:
This plugin enables a manifest for PWA, native-app like shortcuts on mobile.

However, past including this plugin no work has been done to route correctly both the manifest and the icons included, hence this is broken.

We should remove the plugin for now (so the manifest is not generated) and implement correctly in the future. (as just having this in will log errors in browsers)

### DIFF
--- a/gatsby-config.ts
+++ b/gatsby-config.ts
@@ -65,12 +65,5 @@ export const plugins = [
     },
     __key: 'yaml-page-content',
   },
-  // Meta Data & Environment variables
-  {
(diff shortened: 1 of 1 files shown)


kennethkalmer:
Nice one @dpiatek, I've wondered about routing these manifests before but we could have manifests on voltaire as well, and Rails even, and so it would have needed some thought!

kennethkalmer (approved):
No manifest errors on the review app :shipit:

PR #1948, opened 2023-07-05
[EDU-1413] Add subscription filters content

m-hulbert:
## Description

This PR adds documentation for the subscription filters feature (preview). 

It includes code samples for JS. See [this ticket]() for adding instructions on using a channel qualifier for SDKs that don't currently support `getDerived()`. 

## Review

See the [channels]() page.

### DIFF
--- a/content/channels/index.textile
+++ b/content/channels/index.textile
@@ -999,6 +999,148 @@ channelMessageSubscription.cancel();
 <p>Calling "@subscribe()@":/api/realtime-sdk/channels#subscribe for a channel "implicitly attaches":#subscribe the client to the channel as well. It is important to note that if you call @subscribe()@ followed by @unsubscribe()@, the client remains attached to the channel.</p>
 </aside>
 
+h3(#subscription-filters). Subscription filters
+
+Subscription filters enable you to subscribe to a channel and only receive messages that satisfy a filter expression.
+
+Messages are immediately streamed to clients as soon as they "attach":#attach if they have subscribe "capabilities":/auth/capabilities for that channel. Subscription filters apply server-side filtering to messages, meaning that a client will only ever receive be sent the messages that they subscribe to.
+
+Subscription filters are currently in preview status.
+
+<aside data-type="note">
+<p>Be aware that subscription filters have an additional channel and message cost:</p>
+<ul><li>Every unique subscription filter counts as an additional channel.</li>
+<li>Each message published will count an additional time for each unique subscription filter it satisfies. A message that does not pass the filter will not be charged for.</li></ul>
+<p>Normal "limits":/general/limits still apply when using subscription filters. As such, it is not recommended to publish all data to a single channel and rely solely on subscription filters. A level of partitioning at the channel level is still required for the majority of use cases.</p>
+</aside>
+
+h4(#filter-create). Create a filter expression
+
+Filter expressions should be written using "JMESPath.":https://jmespath.org/ They can be constructed using the message name and "@message.extras.headers@":/channels/messages#properties fields.
+
+@message.extras.headers@ optionally provides ancillary metadata to a message, as Ably can't inspect message payloads themselves. Adding suitable key-value pairs to messages will enable more complicated filter expressions to be constructed resulting in more effective message filtering.
+
+The following is an example of publishing a message with additional metadata:
+
+```[realtime_javascript]
+realtime.channels.get('scoops-kiosk').publish({
+    name: 'ice-cream',
+    data: '...',
+    extras: {
+        headers: {
+            flavor: "strawberry",
+            cost: 35,
+            temp: 3
+        }
+    }
+});
+```
+
+```[rest_javascript]
+rest.channels.get('scoops-kiosk').publish({
+    name: 'ice-cream',
+    data: '...',
+    extras: {
+        headers: {
+            flavor: "strawberry",
+            cost: 35,
+            temp: 3
+        }
+    }
+});
+```
+
+Be aware that @message.extras.headers@ must be a flat object. It can't contain any further nesting or arrays.
+
+The following is an example of a filter expression subscribing to messages with the name “ice-cream”, a flavor of “strawberry” and a cost of less than 50:
+
+```[text]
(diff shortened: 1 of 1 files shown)


content/channels/index.textile
```diff
@@ -999,6 +999,148 @@ channelMessageSubscription.cancel();
 <p>Calling "@subscribe()@":/api/realtime-sdk/channels#subscribe for a channel "implicitly attaches":#subscribe the client to the channel as well. It is important to note that if you call @subscribe()@ followed by @unsubscribe()@, the client remains attached to the channel.</p>
 </aside>
 
+h3(#subscription-filters). Subscription filters
+
+Subscription filters enable you to subscribe to a channel and only receive messages that satisfy a filter expression.
+
+Messages are immediately streamed to clients as soon as they "attach":#attach if they have subscribe "capabilities":/auth/capabilities for that channel. Subscription filters apply server-side filtering to messages, meaning that a client will only ever receive be sent the messages that they subscribe to.
```
mschristensen, line 1006: I think: "receive be sent" should just be either "received" or "be sent"

content/channels/index.textile
```diff
@@ -999,6 +999,148 @@ channelMessageSubscription.cancel();
...
+
+Subscription filters are currently in preview status.
+
+<aside data-type="note">
+<p>Be aware that subscription filters have an additional channel and message cost:</p>
```
mschristensen, line 1011: I think we can remove this. This is not currently true in the preview version. Originally, we imagined we would need to do this for the GA/non-preview release, but in recent discussions we have decided not to bill customers for suppressed messages or the additional channel, and instead Ably will swallow the cost of the additional compute.

content/channels/index.textile
```diff
@@ -999,6 +999,148 @@ channelMessageSubscription.cancel();
...
+<ul><li>Every unique subscription filter counts as an additional channel.</li>
```
mschristensen, line 1012: Same here

content/channels/index.textile
```diff
@@ -999,6 +999,148 @@ channelMessageSubscription.cancel();
...
+<li>Each message published will count an additional time for each unique subscription filter it satisfies. A message that does not pass the filter will not be charged for.</li></ul>
```
mschristensen, line 1013: Same here

content/channels/index.textile
```diff
@@ -999,6 +999,148 @@ channelMessageSubscription.cancel();
...
+
+The following is an example of a filter expression subscribing to messages with a flavor of either “strawberry” or “chocolate”:
+
+```[text]
+headers.flavor == `"strawberry"` || headers.flavor == `"chocolate"`
+```
+
+h4(#filter-subscribe). Subscribe with a filter
+
+In order to use a subscribe to a channel with a filter expression, you obtain a channel instance using the @getDerived()@ method. This accepts a filter expression as a parameter.
```
mschristensen, line 1069: I think we can remove "use a"

content/channels/index.textile
```diff
@@ -999,6 +999,148 @@ channelMessageSubscription.cancel();
...
+
+Clients require the subscribe "capability":/auth/capabilities for one of the following resources in order to receive messages from a subscription filter:
+
+* @[filter]<channel name>@
+* @[*]<channel name>@
+* @[*]*@
+
+It is important to set the correct capabilities for clients that will be carrying out additional operations on the same channel they are subscribed to using a filter. If clients have the subscribe capability for the channel itself and attach to it, such as when entering the presence set, they will be streamed all messages.
+
+The following is an example of a request for a token with only the subscribe capability for the subscription filter, and only the publish capability for the unfiltered channel. This avoids the client from being streamed all messages, even if they attach to the unfiltered channel:
```
mschristensen, line 1126: I don't think this is strictly true. If you specify a filter that matches all messages (i.e. suppresses nothing), then you still receive all messages. The only impact of the capability is that it is required in order to make a filtered subscription, but itself does not determine the set of messages received.

mschristensen (changes requested):
Thanks for this! Added a few comments

content/channels/index.textile
m-hulbert: Maybe I've misunderstood here (and I should write like the following where you specify the actual filter in the capability):

`[<filter>]<channel-name>`

In which case, if you then have `subscribe` capabilities to the channel separately (unfiltered), for example to receive presence updates, then you'd be streamed all messages at the point of attaching to the channel, no?

mschristensen: So currently we don't have support in capabilities for a specific filter expression. This is something we have done some design & thinking around (see [here](https://ably.atlassian.net/wiki/spaces/product/pages/2607546413/DSDR3+Long+Form+Capabilities) and [here](https://ably.atlassian.net/wiki/spaces/product/pages/2607087679/DSDR4+Filtered+Channel+Capabilities)), and we intend to support it as part of the GA implementation, but it's not currently supported in the preview. For now, the only thing that's necessary is the `[filter]*` capability, where `filter` refers to the `qualifier`, not the `qualifierParam` (i.e. the filter expression). More info here: https://ably.atlassian.net/wiki/spaces/ENG/pages/2508619954/ADR82+Filtered+Subscriptions#Capabilities

m-hulbert: I think I may need to reword the paragraph and add a better example, as I was trying to approach this from a 'messages saved' point of view. 

Say for example I have the following capabilities: 

```plaintext
   "[filter]scoops-kiosk": ["subscribe"],
   "scoops-kiosk": ["subscribe"]
```

If I subscribe with a filter to the channel `scoops-kiosk`, that attaches me to the derived channel. As it's a derived channel I won't be sent all messages, as we apply server-side filtering (unless my filter happens to match all messages).

If I then subscribe to `scoops-kiosk` (underived), for example to register a listener for presence events. Would I be streamed all regular message events from the channel? If so, I'm filtering the content I receive, but I'm not saving on message counts.

m-hulbert: Rewritten this section to be simpler, and hopefully clearer.

mschristensen (approved):

PR #1961, opened 2023-08-11
[MMB-244] Initial avatar stack docs

m-hulbert:
## Description

Draft initial docs for avatar stack.

### DIFF
--- a/content/spaces/avatar.textile
+++ b/content/spaces/avatar.textile
@@ -1,0 +1,213 @@
+---
+title: Avatar stack
+meta_description: "Avatar stacks display the online status of members in a space."
+languages:
+  - javascript
+---
+
+Avatar stacks are the most common way of showing the status of members in an application by displaying an avatar for each. This will include whether they are online, or have recently disconnected. Additional information can also be provided, such as a profile picture and email address.
+
+An avatar stack is a reflection of who is present in a space for members to see who they are collaborating with. Events are emitted whenever a member enters or leaves a space, or updates their profile data. Whilst member events are published at the space-level, such as by entering a space, or updating a member's profile data, they can be specifically subscribed to at the member-level to keep an avatar stack updated in realtime.
+
+<aside data-type='further-reading'>
+<p>Avatar stacks build upon the functionality of the "presence":/presence-occupancy/presence feature. Members are entered into the presence set when they "enter the space.":/link</p>
+</aside>
+
+The following four event types are emitted by members:
+
+- Enter := A new member has entered the space. The member has either entered explicitly by calling "@space.enter()@":/link, or has attempted to update their profile data before entering a space, which will instead emit an enter event.
+- Update := A member has updated their profile data by calling "@space.updateProfileData()@":/link.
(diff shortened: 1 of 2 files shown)


content/spaces/avatar.textile
```diff
@@ -0,0 +1,213 @@
+---
+title: Avatar stack
+meta_description: "Avatar stacks display the online status of members in a space."
+languages:
+  - javascript
+---
+
+Avatar stacks are the most common way of showing the status of members in an application by displaying an avatar for each. This will include whether they are online, or have recently disconnected. Additional information can also be provided, such as a profile picture and email address.
+
+An avatar stack is a reflection of who is present in a space for members to see who they are collaborating with. Events are emitted whenever a member enters or leaves a space, or updates their profile data. Whilst member events are published at the space-level, such as by entering a space, or updating a member's profile data, they can be specifically subscribed to at the member-level to keep an avatar stack updated in realtime.
```
Srushtika, line 10: I think the first two paragraphs sound repetitive. Can we merge into one?

Srushtika, line 10: Not sure if we should say space-level and member-level. We should probably refer to the namespaces/ objects instead?

content/spaces/avatar.textile
```diff
@@ -0,0 +1,213 @@
...
+
+<aside data-type='further-reading'>
+<p>Avatar stacks build upon the functionality of the "presence":/presence-occupancy/presence feature. Members are entered into the presence set when they "enter the space.":/link</p>
```
Srushtika, line 13: Given we now call the core 'Pub/Sub Channels' - we should probably say presence feature of the pub/sub channels product?

content/spaces/avatar.textile
```diff
@@ -0,0 +1,213 @@
...
+</aside>
+
+The following four event types are emitted by members:
+
+- Enter := A new member has entered the space. The member has either entered explicitly by calling "@space.enter()@":/link, or has attempted to update their profile data before entering a space, which will instead emit an enter event.
```
Srushtika, line 18: Should these be codified like so: https://ably-docs-mmb-244-avata-kkmeak.herokuapp.com/docs/channels/options#subscribe-occupancy. Aren't they case sensitive like so https://ably-docs-mmb-244-avata-kkmeak.herokuapp.com/docs/connect/states#connection-states?

content/spaces/avatar.textile
```diff
@@ -0,0 +1,213 @@
...
+- Update := A member has updated their profile data by calling "@space.updateProfileData()@":/link.
+- Leave := A member has left the space. The member has either left explicitly by calling "@space.leave()@":/link, or has abruptly disconnected and not re-established a connection within 15 seconds.
+- Remove := A member has been removed from the members list after the "@offlineTimeout@":/link period has elapsed. This enables members to appear greyed out in the avatar stack to indicate that they recently left for the period of time between their leave and remove events.
+
+h2(#profile-data). Profile data
```
Srushtika, line 23: I know we discussed this already but it feels like we shouldn't talk about setting and updating stuff here. Perhaps we can talk about subscriptions first and then add setting and updating as further reading boxes?

PR #1970, opened 2023-08-23
[SDK-3781] Add React documentation

stmoreau:

## Description

Add React documentation

* [Jira epic ticket: React Hooks Product Adoption](https://ably.atlassian.net/browse/SDK-3798)
* [Jira ticket: Create React Documentation on the doc site](https://ably.atlassian.net/browse/SDK-3781).

## Review

We need to achieve the following with this work. I have crossed the points that are intended to be achieved with this PR, so please feel free to review any item that has a cross:
- [x]  Add React as a language (version 1.2 and check Brandfolder for icon image)
- [x]  Add React to table here [Available SDKs](https://ably.com/docs/getting-started/sdks)
- [x]  Update [Quickstart guide](https://ably.com/docs/getting-started/quickstart) with React (useChannel only)
- [x]  Add note when JS is selected to [Quickstart guide](https://ably.com/docs/getting-started/quickstart) and [SDK setup](https://ably.com/docs/getting-started/setup)
- [x]  Add a new ‘Frameworks’ and ‘React’ section in the homepage
- [x]  Add a new ‘Frameworks’ and ‘React’ section in the side nav
- [x]  Add content for the React page
- [x]  Remove react-hooks from tools page


content/frameworks/react.textile
```diff
@@ -0,0 +1,177 @@
...
+meta_description: "The React submodule enables you to use React Hooks to connect to Ably."
+language:
+  - react
+---
+
+Leverage the power of Ably in your React applications using idiomatic, easy-to-use React Hooks. This package allows you to:
+
+* Interact with Ably channels using a react hook.
+* Send messages via Ably using the channel instances the hooks provide.
+* Get notifications of user presence on channels.
```
owenpearson, line 12:
```suggestion
* Get notifications of user presence updates on channels.
```

content/frameworks/react.textile
```diff
@@ -0,0 +1,177 @@
+---
+title: React
+meta_description: "The React submodule enables you to use React Hooks to connect to Ably."
+language:
+  - react
+---
+
+Leverage the power of Ably in your React applications using idiomatic, easy-to-use React Hooks. This package allows you to:
+
+* Interact with Ably channels using a react hook.
```
owenpearson, line 10:
```suggestion
* Subscribe to messages on Ably channels.
```

content/frameworks/react.textile
```diff
@@ -0,0 +1,177 @@
...
+* Send messages via Ably using the channel instances the hooks provide.
```
owenpearson, line 11:
```suggestion
* Send messages over Ably using the channel instances provided by the hooks.
```

content/frameworks/react.textile
```diff
@@ -0,0 +1,177 @@
...
+import * as Ably from 'ably';
+import { AblyProvider, useChannel, usePresence } from 'ably/react';
+
+const client = {
+  key: "your-ably-api-key",
+  clientId: "your-ably-client-id",
+}
+
+root.render(
+  <AblyProvider client={client}>
```
owenpearson, line 63: This example is slightly wrong - you can't pass a `ClientOptions` in as the `client` prop, the prop (and the variable) should be called `options` instead. I also think the primary way the provider should be used is by passing in a client instance so I think that should be the first example and then `options` should be below it as an alternative.

owenpearson (changes requested):
Looks good! A few comments, I know most of them are on text that you copied over from `root.textile` but I guess now's a good time to address them 😄

content/frameworks/react.textile
stmoreau: Done (hopefully correctly) ✔️

content/frameworks/react.textile
stmoreau: Done ✔️

content/frameworks/react.textile
stmoreau: Done ✔️

content/frameworks/react.textile
```diff
@@ -0,0 +1,179 @@
...
+- "useAbly":#useAbly := The @useAbly@ hook grants access to the Ably client instance provided by the AblyProvider context.
+
+These hooks manage the lifecycle of the Ably SDK instances for you, ensuring that you subscribe and unsubscribe to channels and events when your React components re-render.
+
+h2(#installation). Installation
+
+To install the Ably React Hooks, use the following command:
+
+```[sh]
+npm install --save ably
```
ttypic, line 31: Do we plan to release ably + react-hooks with latest tag? Also maybe it worth to mention that react hooks available starting from 1.x.x?

content/frameworks/react.textile
```diff
@@ -0,0 +1,179 @@
...
+  <AblyProvider client={client}>
+    <App />
+  </AblyProvider>
+);
+```
+
+Alternatively, you can create your options and pass them to the context provider:
+
+```[react]
+import * as Ably from 'ably';
```
ttypic, line 70: It looks like unnecessary line, let's delete it?

content/frameworks/react.textile
```diff
@@ -0,0 +1,179 @@
...
+    console.log(message);
+});
+```
+
+h3(#usePresence). usePresence
+
+The @usePresence@ hook allows you to subscribe to presence events on a channel, notifying you when a user joins or leaves.
+
+```[react]
+const { presenceData, updateStatus } = usePresence("your-channel-name");
```
ttypic, line 109: maybe add example with `updateStatus`? maybe with `useEffect`

ttypic (commented):
Looks very nice, like it!

content/frameworks/react.textile
stmoreau: Do we want people to start using our hooks from earlier versions? I was thinking no, that's why I haven't explicitly mentioned it 😬
How were you thinking we could phrase this?

ttypic: "Ably >= 1.x.x includes React hooks functionality out of the box. You don't need to install any additional packages." instead of  "To install the Ably React Hooks, use the following command:". Does it make sense for you? @stmoreau

content/frameworks/react.textile
stmoreau: That's a good point! I could maybe convert this to something simple like:

```
const { presenceData, updateStatus } = usePresence("your-channel-name", "initial state");

// Convert presence data to list items to render    
const peers = presenceData.map((msg, index) => <li key={index}>{msg.clientId}: {msg.data}</li>);

// The `updateStatus` function can be used to update the presence data for the current client
updateStatus("new status");
```

Or did you have something else in mind?

PR #1982, opened 2023-09-14
[MMB-296] Update async code examples

m-hulbert:
## Description

This PR fixes the Spaces code samples to ensure functions are asynchronous where required and corrects the component locking example to correctly set attributes.

### DIFF
--- a/content/spaces/cursors.textile
+++ b/content/spaces/cursors.textile
@@ -41,7 +41,7 @@ The following is an example of a member setting their cursor position by adding 
 
 ```[javascript]
 window.addEventListener('mousemove', ({ clientX, clientY }) => {
-  await space.cursors.set({ position: { x: clientX, y: clientY }, data: { color: 'red' } });
+  space.cursors.set({ position: { x: clientX, y: clientY }, data: { color: 'red' } });
 });
 ```
 
@@ -232,13 +232,13 @@ const client = new Realtime.Promise({ authUrl: '<authEndpoint>', clientId: '<cli
 const spaces = new Spaces(client);
 
 // Create a new space
-const space = await spaces.get('slideshowPresentation');
+const space = await spaces.get('board-presentation');
 
 // Enter the space to become a member, passing a nickname
 await space.enter({ name: 'Helmut' });
 
 // Listen for cursor updates from members
-space.cursors.subscribe('update', (cursorUpdate) => {
+space.cursors.subscribe('update', async (cursorUpdate) => {
   // Use getAll() and filter by the member that moved their cursor to only update the position of that member's cursor
   const members = await space.members.getAll();
   const member = members.find((member) => member.connectionId === cursorUpdate.connectionId);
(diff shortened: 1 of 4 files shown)


content/spaces/cursors.textile
```diff
@@ -40,7 +40,7 @@ The @set()@ method takes the following parameters:
 The following is an example of a member setting their cursor position by adding an event listener to obtain their cursor coordinates and then publishing their position using the @set()@ method:
 
 ```[javascript]
-window.addEventListener('mousemove', ({ clientX, clientY }) => {
+window.addEventListener('mousemove', async ({ clientX, clientY }) => {
   await space.cursors.set({ position: { x: clientX, y: clientY }, data: { color: 'red' } });
```
dpiatek, line 44: I'm not sure we should present this example like this. The `await` here is correct but not needed. It would only be needed, if in the same function, there would be another operation that requires that the cursor is set and confirmed so. So although this is an async operation we don't need to add async/await keywords here.

content/spaces/cursors.textile
```diff
@@ -232,21 +232,21 @@ const client = new Realtime.Promise({ authUrl: '<authEndpoint>', clientId: '<cli
...
+space.cursors.subscribe('update', async (cursorUpdate) => {
   // Use getAll() and filter by the member that moved their cursor to only update the position of that member's cursor
   const members = await space.members.getAll();
   const member = members.find((member) => member.connectionId === cursorUpdate.connectionId);
   renderCursor(cursorUpdate, member);
 });
 
 // Publish the member's cursor position
-window.addEventListener('mousemove', ({ clientX, clientY }) => {
+window.addEventListener('mousemove', async ({ clientX, clientY }) => {
```
dpiatek, line 249: same comment as above

dpiatek (changes requested):

content/spaces/cursors.textile
m-hulbert: Ah thanks Dom - I've fixed both of those.

content/spaces/cursors.textile
```diff
@@ -232,22 +232,22 @@ const client = new Realtime.Promise({ authUrl: '<authEndpoint>', clientId: '<cli
...
   // Use getAll() and filter by the member that moved their cursor to only update the position of that member's cursor
   const members = await space.members.getAll();
   const member = members.find((member) => member.connectionId === cursorUpdate.connectionId);
   renderCursor(cursorUpdate, member);
 });
 
 // Publish the member's cursor position
-window.addEventListener('mousemove', ({ clientX, clientY }) => {
-  await space.cursors.set({ position: { x: clientX, y: clientY } });
+window.addEventListener('mousemove' ({ clientX, clientY }) => {
```
dpiatek, line 249: missing comma

content/spaces/cursors.textile
```diff
@@ -40,8 +40,8 @@ The @set()@ method takes the following parameters:
 The following is an example of a member setting their cursor position by adding an event listener to obtain their cursor coordinates and then publishing their position using the @set()@ method:
 
 ```[javascript]
-window.addEventListener('mousemove', ({ clientX, clientY }) => {
-  await space.cursors.set({ position: { x: clientX, y: clientY }, data: { color: 'red' } });
+window.addEventListener('mousemove' ({ clientX, clientY }) => {
```
dpiatek, line 43: missing comma

dpiatek (changes requested):

dpiatek (approved):

PR #2036, opened 2023-10-19
[EDU-213] Remove API Streamer content

m-hulbert:
## Description

This PR deletes the docs content for API streamer from the site and removes it from the table of contents.


GregHolmes:
I think we should do some form on redirect here. There will be sources outside of the docs that link to the streamer page. If anyone goes to it, they'd just be shown a 404. Would that be what's best for their experience?

m-hulbert:
> I think we should do some form on redirect here. There will be sources outside of the docs that link to the streamer page. If anyone goes to it, they'd just be shown a 404. Would that be what's best for their experience?

Yep! The redirects for these will need to happen in the website though, so I have a separate PR for that 👍

GregHolmes (approved):

PR #2037, opened 2023-10-23
[EDU-189] - Improve Token Revocation description over enforcement times

GregHolmes:
## Description

Token Revocation was confusing developers over the timeline of the revocation enforcement. This PR adds an improvement to the description to reduce the confusion, in stating that the revocation enforcement is immediate unless `allowReauthMargin` is defined as true in the request. This would delay the enforcement by 30 seconds to allow the client to request a new token without disruption to service.

* [Jira ticket](https://ably.atlassian.net/browse/EDU-189).

## Review

* [Page to review](https://ably-docs-edu-189-make--tg4am4.herokuapp.com/docs/auth/revocation)


m-hulbert (changes requested):
Looks good (and thanks for fixing the image and `keyName`).

One small comment - I think we should state the default in the opening paragraph so that they don't need to scroll to a property to find that out.

GregHolmes:
> Looks good (and thanks for fixing the image and `keyName`).
> 
> One small comment - I think we should state the default in the opening paragraph so that they don't need to scroll to a property to find that out.

https://github.com/ably/docs/pull/2037/commits/26b24ef66997e5e4832e77b0f889c50064738226

@m-hulbert Could you please let me know what you think of this? 

Thank you.

franrob-projects (changes requested):
Just a typo fix! Otherwise LGTM 👍

franrob-projects (approved):
LGTM. Rebase time 🚀

m-hulbert (approved):
Good to go with a rebase 👍

PR #2044, opened 2023-10-24
[ACT-80] Use demo keys in pub/sub how-to

kennethkalmer:

## Description

We want to use demo keys in our interactive how-tos so they just work for our visitors. The pub-sub how-to will replace any occurrences of `import.meta.env.VITE_ABLY_KEY` with a demo key. It will also replace the `BASKETBALL_CHANNEL_NAME` and `FOOTBALL_CHANNEL_NAME` env strings with stable random channel names. As a bonus it will set the `environment` of the client options to the value of `GATSBY_ABLY_ENVIRONMENT` if the value is not `production` ensuring keys from the sandbox environment also works.

This PR also includes stylistic updates to the Pub-Sub how-to just to bring things to life a bit more.

* ACT-68
* ACT-80

## Review

* [Pub/Sub How-To](https://ably-docs-act-80-how-to-fr43q0.herokuapp.com/docs/how-to/pub-sub)


src/components/Markdown/MarkdownProvider.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { FC } from 'react';
+import { MDXProvider } from '@mdx-js/react';
+import { HtmlAttributes } from 'src/components/html-component-props';
```
aleksandar-r, line 3: I noticed that you used `HTMLAttributes` for defining props in this component, which is causing a TypeScript error. According to the MDX-JS  [documentation](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mdx/types.d.ts#L22) it appears that using `JSX.IntrinsicElements['h1']` is the recommended approach.

The correct way to define props for the `h1` element in @mdx-js is to use `JSX.IntrinsicElements['h1']`. This ensures that TypeScript recognizes the correct prop types for HTML elements and provides better type safety.

Here's how you can update the code:
```tsx
const H1: FC<JSX.IntrinsicElements['h1']> = ({ children, ...props }) => (
  <h1 className="ui-text-h1 my-40" {...props}>
    {children}
  </h1>
);

File: src/components/Markdown/MarkdownProvider.tsx [hunk 2]
```diff
@@ -0,0 +1,50 @@
...
+const components = {
+  h1: H1,
+  h2: H2,
+  h3: H3,
+  p: Paragraph,
+  ul: Ul,
+  li: Li,
+};
+
+export const MarkdownProvider = ({ children }) => <MDXProvider components={components}>{children}</MDXProvider>;
```
aleksandar-r, line 50: I noticed that there is a missing `children` prop of type `React.ReactNode` in this component, and this omission is causing a TypeScript error. Including the `children` prop is essential for components that can have child elements in React.

To address this issue, please make sure to include the `children` prop in the component's prop type definition. Here's an example of how to do it:

``` tsx
export const MarkdownProvider = ({ children }: { children: React.ReactNode }) => (
  <MDXProvider components={components}>{children}</MDXProvider>
);

**Review by aleksandar-r — CHANGES_REQUESTED** [review]

File: src/pages/docs/how-to/pub-sub.tsx [hunk 3]
```diff
@@ -5,8 +5,64 @@ import HowTo from 'HowTos/pub-sub/how-to.mdx';
 import { graphql, useStaticQuery } from 'gatsby';
 import { Head } from 'src/components/Head';
 import Layout from 'src/components/Layout';
+import { MarkdownProvider } from 'src/components/Markdown';
 import PageTitle from 'src/components/PageTitle';
 import { SidebarProvider } from 'src/contexts/SidebarContext';
+import UserContext from 'src/contexts/user-context';
+import { UserApiKey } from 'src/components/blocks/software/Code/ApiKeyMenu';
```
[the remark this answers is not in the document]
kennethkalmer: I think this is right from an earlier refactor!

aleksandar-r (approved):
Amazing work, way to go!

denissellu (approved):

PR #2052, opened 2023-11-02
[ACT-77] Update pub/sub how-to instructions

m-hulbert:
## Description

This PR updates the instructions for the pub/sub how-to based on the redesign of the app. 

It also updates the code for example, adds a completed copy of each for 'show solution' and slightly increases the interval between score publishes.


kennethkalmer:
> The code to copy into the football subscriber is also missing a trailing { as it currently breaks the build when included. The full line should be: footballChannel.subscribe(footballLeague, (message) => { when this is fixed (see [JIRA](https://ably.atlassian.net/browse/ACT-96))

@m-hulbert it definitely rendered when I pushed it up. I wonder if it wasn't feeling a bit temperamental.

In MDX we need to escape `{` and `<` so they don't trigger the parsing of expressions or JSX tags. This is mentioned at the end of the *Markdown* section on the [What is MDX](https://mdxjs.com/docs/what-is-mdx/#markdown) page:

> * Unescaped left angle bracket / less than (<) and left curly brace ({) have to be escaped: \\< or \\{ (or use expressions: {'<'}, {'{'})

I'm happy to merge this into the final branch and refine the code sample there if needed

GregHolmes (approved):

denissellu (approved):

PR #2059, opened 2023-11-15
COL-586: Learn LiveSync

franrob-projects:
This PR:

- Adds content that I think will best fit under the heading 'learn LiveSync' or similar
- I will tweak with links once I have addressed the navigation

[COL-586: Learn LiveSync ](https://ably.atlassian.net/browse/COL-586)

### DIFF
--- a/content/livesync/mergefunction.textile
+++ b/content/livesync/mergefunction.textile
@@ -1,0 +1,18 @@
+---
+title: Merge Function
+product: Livesync
+---
+
+Merge functions play a crucial role in updating your model with the latest data version upon receiving a mutation event.
+
+Your backend needs to emit events when any changes are made to the data model. These events should reflect the alterations made and be processed by the Ably Models Realtime SDK. This SDK applies the changes to its local version of the model state, producing an updated version. These changes are definded by an update function:
+
+```ts
+type UpdateFunc<T> = (state: T, event: OptimisticEvent | ConfirmedEvent) => Promise<T>;
+```
+
+The update function takes the current model state and the event as inputs and returns the ensuing model state.
+
+There are two types of events - optimistic and confirmed. Events from your backend are always confirmed since they are the outcome of a mutation validated by your backend. The Ably Realtime Models SDK can also generate local optimistic events indicating that local changes have been made but are still waiting for validation from the backend.
+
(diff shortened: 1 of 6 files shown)


content/livesync/mergefunction.textile
```diff
@@ -0,0 +1,18 @@
...
+
+Merge functions play a crucial role in updating your model with the latest data version upon receiving a mutation event.
+
+Your backend needs to emit events when any changes are made to the data model. These events should reflect the alterations made and be processed by the Ably Models Realtime SDK. This SDK applies the changes to its local version of the model state, producing an updated version. These changes are definded by an update function:
+
+```ts
+type UpdateFunc<T> = (state: T, event: OptimisticEvent | ConfirmedEvent) => Promise<T>;
+```
+
+The update function takes the current model state and the event as inputs and returns the ensuing model state.
```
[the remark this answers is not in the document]
mschristensen: "ensuing" is unclear to me, I would say "returns the next version of the model state" or similar - perhaps we can take wording from the docs linked above?

content/livesync/optimistic.textile
```diff
@@ -0,0 +1,43 @@
...
+})
+
+// apply the changes in your backend
+updatePost('my-mutation-id', 'new post text')
+```
+
+The optimistic event will be processed by your @Merge Function@ and included in the local model state optimistically.
+
+You are responsible for updating the persisted model state, typically through a function that modifies the state in your backend over the network. The updatePost(...) method above is an example of this, interacting with a REST HTTP API endpoint to update post data.
+Upon updating the data, your backend API endpoint should update the database and emit a confirmation event. The @Merge Function@ then receives and processes this event, as discussed in .
```
mschristensen, line 39: Missing link at end of sentence

content/livesync/subscriptions.textile
```diff
@@ -0,0 +1,21 @@
...
+After setting up the necessary registrations, you can subscribe to your data model. This subscription enables real-time updates whenever there's a change in the data:
+
+```
+await model.pause();
+// model paused: new events will not be processed and subscription callbacks will no longer be invoked
+await model.resume();
+// processing of events has resumed and new changes will be made available to subscribers
+```
+
+By default, subscriptions operate on an optimistic model, meaning the callback is triggered by any change in the optimistic state or when confirmed changes occur.
```
mschristensen, line 15: "operate on" sounds plays some active roll, but subscribing to updates is passive, so this sounds odd to me

mschristensen (changes requested):

PR #2060, opened 2023-11-16
Fix docs build

dpiatek:
The main fix here is an update to this part:
```
const context = useContext(SidebarContext);
  if (!context) {
    throw new Error('useSidebar must be used within a SidebarProvider');
  }
  return context;
```

I've modified what the `SidebarContext` initial value is, but removed the above, as it's quite normal for it to be false or incomplete. Interestingly, this would run on `build` and immediately fail but never be an issue in dev.


aleksandar-r (approved):
Keeping it tight and secure. Love it.

kennethkalmer (commented):
This is great, thanks Dom!

PR #2064, opened 2023-11-21
COL-553--Adds SDK setup page

franrob-projects:
This PR:

- Adds the SDK setup page

[COL-553: SDK setup page](https://ably.atlassian.net/browse/COL-553)

### DIFF
--- a/content/livesync/Setup
+++ b/content/livesync/Setup
@@ -1,0 +1,190 @@
+---
+title: SDK setup
+meta_description: "Install, authenticate and instantiate the Ably Realtime Data Models SDK."
+product: LiveSync
+---
+
+Use these instructions to install, authenticate and instantiate the Ably Realtime Data Models SDK.
+
+h2(#authenticate). Authenticate
+
+An "API key":/auth#api-keys is required to authenticate with Ably. API keys are used either to authenticate directly with Ably using "basic authentication":/auth/basic, or to generate tokens for untrusted clients using "token authentication":/auth/token.
+
+<aside data-type='important'>
+<p>The examples use "basic authentication":/auth/basic to demonstrate features for convenience. In your own applications, basic authentication should never be used on the client-side as it exposes your Ably API key. Instead use "token authentication":/auth/token. </p>
+</aside>
+
+"Sign up":https://ably.com/sign-up to Ably to create an API key in the "dashboard":https://ably.com/dashboard or use the "Control API":/account/control-api to create an API programmatically.
+
+API keys and tokens have a set of capabilities assigned to them that specify which operations, such as @subscribe@ or @publish@ can be performed on which resources. To use the Ably Realtime Data Models SDK, the API key requires the following "capabilities":/auth/capabilities#58: 
+
+* Publish
+* Subscribe
+* Presence
+* History
+
+h2(#install). Install
+
+The Realtime Data Models SDK is a JavaScript (TypeScript) library that enables you to create live and observable data models in your frontend application. These models remain synchronized with the realtime state of your database model. You can easily integrate this SDK into your project regardless of your frontend framework preference.
+
+Your backend publishes mutation events to Ably. The Realtime Data Models SDK updates your frontend app's local state. You can also pair the SDK with *Ably's Database Connector* to transmit transactional change events with your database mutations.
+
+The Realtime Data Models SDK is available as an "NPM module":#npm.
+
+h3(#npm1). Use NPM
+
+Install the Ably JavaScript SDK and the Realtime Data Models SDK:
+
+```
+npm install ably @ably-labs/models
+```
+
+
+h3(#npm-link). Use the NPM link
+
+Clone this repository and run @npm link@:
+
+```
+git clone git@github.com:ably-labs/models.git
+cd models
+npm link
+```
+
+From your project, link to the cloned project and build:
+
+```
+cd ./your/project
+npm link @ably-labs/models
+pushd ./node_modules/@ably-labs/models
+npm run build
+popd
+```
+
+You should now be able to import @@ably-labs/models@ in your project.
+
+h2(#instantiate). Instantiate
+
+To instantiate the Realtime Data Models SDK, create an "Ably client":/getting-started/setup?lang=javascript and pass it into the @ModelsClient@ constructor:
+
+```
+import ModelsClient from '@ably-labs/models';
+import { Realtime } from 'ably';
+
+const ably = new Realtime.Promise({ key: 'YOUR_ABLY_API_KEY' });
+const modelsClient = new ModelsClient({ ably });
+```
+
+h2(#create-model). Create a model
+
+A @model@ represents a live, observable data model supported by the database.
+
+To create a model, you need to:
+
+1. Define the model's data structure in the frontend application.
+2. Initialize the model.
+3. Update the model based on events from the backend.
+4. Determine how end-users can modify the model.
+
+h3(#structure). Ably Realtime Data Models structure
+
+The structure of the models data as it is used in the frontend application:
+
+```
+type Post = {
+  id: number;
+  text: string;
+  comments: string[];
+}
+```
+
+h3(#initialise). Initialise
+
+An example function used by the model to initialise with the correct data from your backend:
+
+```
+async function sync() {
+  const result = await fetch('/api/post');
+  return result.json();
+};
+```
+
+h3(#merge). Merge change event
+
+The function used by the model to merge a change event that is received and the existing model state:
+
+```
+async function merge(state: Post, event: OptimisticEvent | ConfirmedEvent) {
+  return {
+    ...state,
+    text: event.data, // replace the previous post text field with the new value
+  }
+};
+```
+
+h3(#mutate). Mutate model data
+
+A function that you can use to mutate the model data in your backend:
+
+```
+async function updatePost(mutationID: string, content: string) {
+  const result = await fetch(`/api/post`, {
+    method: 'PUT',
+    body: JSON.stringify({ mutationID, content }),
+  });
+  return result.json();
+};
+```
+
+h3(#new-model-instance). Create a new model instance
+
+Create a new model instance called @post@ by passing the sync and merge functions:
+
+```
+const model = modelsClient.models.get({
+  name: 'post',
+  channelName: 'models:posts',
+  sync: sync,
+  merge: merge,
+});
+```
+
+h3(#live-changes). Subscribe to live changes
+
+Subscribe to live changes to the model data:
+
+```
+model.subscribe((err, post) => {
+  if (err) {
+    throw err;
+  }
+  console.log('post updated:', post);
+});
+```
+
+h3(#optimistic-update). Apply an optimistic update to the model
+
+@confirmation@ is a promise that resolves when the optimistic update is confirmed by the backend. @cancel@ is a function that can be used to cancel and rollback the optimistic update.
+
+```
+const [confirmation, cancel] = await model.optimistic({
+    mutationID: 'my-mutation-id',
+    name: 'updatePost',
+    data: 'new post text',
+})
+```
+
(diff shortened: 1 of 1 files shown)


content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
+---
+title: SDK setup
+meta_description: "Install, authenticate and instantiate the Ably Realtime Data Models SDK."
+product: LiveSync
+---
```
m-hulbert, line 5: You need a `language` here for the codeblocks and dropdown.

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+title: SDK setup
+meta_description: "Install, authenticate and instantiate the Ably Realtime Data Models SDK."
```
m-hulbert, line 3: I'd suggest just using 'Data Models SDK' throughout rather than having to always append 'Ably Realtime' to it.

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+</aside>
+
+"Sign up":https://ably.com/sign-up to Ably to create an API key in the "dashboard":https://ably.com/dashboard or use the "Control API":/account/control-api to create an API programmatically.
+
+API keys and tokens have a set of capabilities assigned to them that specify which operations, such as @subscribe@ or @publish@ can be performed on which resources. To use the Ably Realtime Data Models SDK, the API key requires the following "capabilities":/auth/capabilities#58: 
+
+* Publish
+* Subscribe
+* Presence
+* History
```
m-hulbert, line 24: The repo only seems to state subscribe is required.

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+
+Your backend publishes mutation events to Ably. The Realtime Data Models SDK updates your frontend app's local state. You can also pair the SDK with *Ably's Database Connector* to transmit transactional change events with your database mutations.
+
+The Realtime Data Models SDK is available as an "NPM module":#npm.
+
+h3(#npm1). Use NPM
+
+Install the Ably JavaScript SDK and the Realtime Data Models SDK:
+
+```
```
m-hulbert, line 38: Your codeblocks all need a language defined for syntax highlighting. For command line items you can use `[sh]`

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+npm install ably @ably-labs/models
+```
+
+
+h3(#npm-link). Use the NPM link
```
m-hulbert, line 43: I don't think you need 'the' here.

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+Your backend publishes mutation events to Ably. The Realtime Data Models SDK updates your frontend app's local state. You can also pair the SDK with *Ably's Database Connector* to transmit transactional change events with your database mutations.
+
+The Realtime Data Models SDK is available as an "NPM module":#npm.
+
+h3(#npm1). Use NPM
+
+Install the Ably JavaScript SDK and the Realtime Data Models SDK:
+
+```
+npm install ably @ably-labs/models
```
m-hulbert, line 39: Just to double check: it's staying in labs for Alpha?

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+
+```
+import ModelsClient from '@ably-labs/models';
+import { Realtime } from 'ably';
+
+const ably = new Realtime.Promise({ key: 'YOUR_ABLY_API_KEY' });
+const modelsClient = new ModelsClient({ ably });
+```
+
+h2(#create-model). Create a model
```
m-hulbert, line 77: I'd say anything below this point is part of using the SDK, rather than its setup. 

May be worth creating a ticket to add a section on the benefits of the underlying "core" client here too - but that can come later.

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
+---
```
m-hulbert, line 1: File name should be lowercase `setup` please 🙂

m-hulbert (changes requested):
A few minor comments and fixes with this one, but it's nearly there.

content/livesync/Setup
mschristensen: Correct

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+title: SDK setup
+meta_description: "Install, authenticate and instantiate the Ably Realtime Data Models SDK."
+product: LiveSync
+---
+
+Use these instructions to install, authenticate and instantiate the Ably Realtime Data Models SDK.
```
mschristensen, line 7: Minor, but if listing these out, I would go install, instantiate and authenticate as thats the order those things happen

Although might be simpler to just say "Use these instructions to get started with" ?

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+API keys and tokens have a set of capabilities assigned to them that specify which operations, such as @subscribe@ or @publish@ can be performed on which resources. To use the Ably Realtime Data Models SDK, the API key requires the following "capabilities":/auth/capabilities#58: 
+
+* Publish
+* Subscribe
+* Presence
+* History
+
+h2(#install). Install
+
+The Realtime Data Models SDK is a JavaScript (TypeScript) library that enables you to create live and observable data models in your frontend application. These models remain synchronized with the realtime state of your database model. You can easily integrate this SDK into your project regardless of your frontend framework preference.
```
mschristensen, line 28: > You can easily integrate

Maybe qualify this by saying that the SDK is vanilla JS so it can be used with any framework?

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+"Sign up":https://ably.com/sign-up to Ably to create an API key in the "dashboard":https://ably.com/dashboard or use the "Control API":/account/control-api to create an API programmatically.
+
+API keys and tokens have a set of capabilities assigned to them that specify which operations, such as @subscribe@ or @publish@ can be performed on which resources. To use the Ably Realtime Data Models SDK, the API key requires the following "capabilities":/auth/capabilities#58: 
+
+* Publish
+* Subscribe
+* Presence
+* History
+
+h2(#install). Install
```
mschristensen, line 26: I think this should be the first block, before auth

content/livesync/Setup
mschristensen: You don't need npm link, this is only needed for the example app in the models repo which links to the local package built from source rather than from the package published to npm

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+
+h2(#authenticate). Authenticate
+
+An "API key":/auth#api-keys is required to authenticate with Ably. API keys are used either to authenticate directly with Ably using "basic authentication":/auth/basic, or to generate tokens for untrusted clients using "token authentication":/auth/token.
+
+<aside data-type='important'>
+<p>The examples use "basic authentication":/auth/basic to demonstrate features for convenience. In your own applications, basic authentication should never be used on the client-side as it exposes your Ably API key. Instead use "token authentication":/auth/token. </p>
+</aside>
+
+"Sign up":https://ably.com/sign-up to Ably to create an API key in the "dashboard":https://ably.com/dashboard or use the "Control API":/account/control-api to create an API programmatically.
```
mschristensen, line 17: Should be "API Key"

content/livesync/Setup
mschristensen: Subscribe capability is the only hard requirement

The Models SDK can optionally be used with history, see https://github.com/ably-labs/models/blob/main/docs/replay.md#history

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+  comments: string[];
+}
+```
+
+h3(#initialise). Initialise
+
+An example function used by the model to initialise with the correct data from your backend:
+
+```
+async function sync() {
```
mschristensen, line 105: Do we need to indicate the return type here? We are currently not using the `Post` type defined above in these docs at all. In reality, the type of the model is inferred from the return type of the `sync` function.

The sync function also needs to return a sequence ID which isn't covered here.

See: https://github.com/ably-labs/models/blob/main/docs/usage.md#sync-function

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+  const result = await fetch('/api/post');
+  return result.json();
+};
+```
+
+h3(#merge). Merge change event
+
+The function used by the model to merge a change event that is received and the existing model state:
```
mschristensen, line 113: This doesn't read very clearly to me. Perhaps we could use wording like "is a function that accepts the previous model state and the event and returns the next model state" ?

See: https://github.com/ably-labs/models/blob/main/docs/usage.md#merge-functions

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+```
+async function merge(state: Post, event: OptimisticEvent | ConfirmedEvent) {
+  return {
+    ...state,
+    text: event.data, // replace the previous post text field with the new value
+  }
+};
+```
+
+h3(#mutate). Mutate model data
```
mschristensen, line 124: This is done entirely out-of-band of the SDK, so not sure we need a specific section for this

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+  merge: merge,
+});
+```
+
+h3(#live-changes). Subscribe to live changes
+
+Subscribe to live changes to the model data:
+
+```
+model.subscribe((err, post) => {
```
mschristensen, line 156: We can also subscribe to confirmed changes only using the subscription params: https://github.com/ably-labs/models/blob/main/docs/usage.md#subscriptions

mschristensen: I know the docs linked about throw the error but might be better to avoid that here as in reality you would want to handle the error. So we can just say e.g.:

```
model.subscribe((err, post) => { /* ... */ });
```

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+    data: 'new post text',
+})
+```
+
+h3(#change-to-backend). Apply change to the backend
+
+Call your backend to apply the actual change:
+
+```
+updatePost('my-mutation-id', 'new post text')
```
mschristensen, line 181: This is where we should define the updatePost function specified above

content/livesync/Setup
```diff
@@ -0,0 +1,190 @@
...
+```
+
+h3(#confirm-changes). Confirm changes
```
mschristensen, line 184: I feel having all these steps in separate headings is a bit piecemeal and hard to grok the overall thing when it's split up like that. Would it be worth considering having a single example of the e2e functionality in a code block? For example, as defined here: https://github.com/ably-labs/models/blob/main/docs/usage.md#optimistic-events

mschristensen (changes requested):
Thanks Fran, added some comments :)

content/livesync/Setup
franrob-projects: You can create an API key....

content/livesync/Setup
franrob-projects: @mschristensen Mike will shorten the code snippets

content/livesync/Setup
franrob-projects: @mschristensen will address as part of the above

content/livesync/Setup
franrob-projects: @mschristensen will address as part of the above

content/livesync/Setup
franrob-projects: @mschristensen will address as part of the above

content/livesync/Setup
franrob-projects: @mschristensen will address as part of the above

content/livesync/Setup
franrob-projects: @mschristensen will address as part of the above

PR #2065, opened 2023-11-22
Fix docs home image

dpiatek:
`srcSet` will not get transformed to get the correct `ASSET_PREFIX` and so will never work on the `ably.com` domain. That, and a rogue `}` in the src 😮‍💨 .


kennethkalmer (approved):
It's always the little things, thanks @dpiatek

PR #2101, opened 2024-01-16
[WEB-3531] Rework internal links

kennethkalmer:
## Description

Driven by the basic requirements of ACT-163, this PR reworks how we generate our links in the project so that we can make use of Gatsby's [`withPrefix`](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#add-the-path-prefix-to-paths-using-withprefix). Prior to this we had a mixed bag of constants and string interpolation that we did ourselves, and Gatsby wasn't aware of any of this, and as a result we experience some strange issues in places like markdown rendering.

* ACT-163
* WEB-3531

## What changed?

While developing locally we no longer have to worry about the `/docs` prefix in our URLs and/or local links. For the site everything is just local. Where we do have _external_ links that are still on our domain, we can annotate those links as _"external"_ either via passing a prop to our `Link` component, or by adding an `external` attribute to our links in YAML files (page data & page furniture).

All the templates/components have been updated to now handle this, along with matching nginx config changes for serving up the files correctly. Several tests have been added to various components to lock this down.

We also need to implement a workaround for a bug in `gatsby-plugin-mdx` which was the root cause of this work to begin with.

## Review

* Browse around the [review app](https://ably-docs-act-163-links-akilcn.herokuapp.com/) and notice the performance improvement of browsing the site, and look around for any potential broken links
* Specifically check the links in the "Further Reading" section of the [Pub/Sub How-to](https://ably-docs-act-163-links-akilcn.herokuapp.com/docs/how-to/pub-sub)
* Review the commits in order or the change will be overwhelming


m-hulbert (commented):
Tested links that already include `/docs` and those that don't and all work correctly. 

The only issues I've found is that when you navigate back to a page, the table of contents loses the active page that you're on and everything appears collapsed. 

Love how much snappier everything is!

m-hulbert (approved):
LGTM thanks @jamiehenson ! I know what the issue is with 'Metadata' and I can fix this in a separate PR.

jamiehenson:
Wizzo 👉 Setting a note on my end to merge this in Monday morn

jamiehenson:
I've resolved the merge conflicts that were present with this branch so it's now ready to merge. In theory at least, it's probably worth a quick look over to make sure there have been no obvious regressions as there as some manual intervention with said merge. @kennethkalmer

Things look fine to me locally, but alas I still can't deploy new Heroku instances (my perms screen attached) so that still needs looking at before I can update the instance for wider testing
Screenshot 2024-03-04 at 14 59 31

kennethkalmer (commented):
@jamiehenson I'm happy with the code, thanks for the finishing touches!

PR #2104, opened 2024-01-19
[EDU-480] Add instructions for user management, SSO and 2FA

m-hulbert:
## Description

This PR adds instructions for the following account management tasks:

* User roles
* SSO
* 2FA

## Review

The new section can be seen [here](https://ably-docs-edu-480-sso-ezd5tm7n.herokuapp.com/docs/account/users).

### DIFF
--- a/content/account/sso.textile
+++ b/content/account/sso.textile
@@ -1,0 +1,98 @@
+---
+title: Single sign-on (SSO)
+meta_description: "Single sign-on enables users to authenticate with Ably using your own identity provider."
+meta_keywords: "sso, saml, strict mode, single sign-on"
+---
+
+Single sign-on (SSO) enables your users to authenticate via any SAML-compatible identity provider.
+
+<aside date-type='note'>
+<p>User provisioning and deprovisioning via SCIM or SAML JIT is not supported.</p>
+</aside>
+
+h2(#configure). Configure
+
+Single sign-on is restricted to Enterprise customers only and must be enabled on a per-account basis by "contacting Ably":https://ably.com/contact. Only "account owners":/account/users can configure SSO for an account.
+
+Any SAML-compatible identity provider can be used to enable SSO.
+
+The following instructions are examples of configuring SSO with "Okta":#okta and "Google Workspace":#google.
+
+h3(#okta). Okta
+
+To enable SSO using "Okta":https://www.okta.com/ as the identity provider, configure the following properties in your Ably account dashboard and your Okta account:
+
+In your Ably account dashboard:
+
+1. Log in to your "account":https://ably.com/accounts/any.
+2. Select *Settings* from the account navigation dropdown.
+3. Toggle *Enable Single Sign-On* under the *Authentication Settings* section.
+4. Note down the *Single sign-on URL* and *Audience URI* values.
+
+In your Okta account:
+
+Use the "Okta guide":https://developer.okta.com/docs/guides/saml-application-setup/overview/ for enabling SSO.
+
+1. Upload the Ably logo.
+2. Select *EmailAddress* for the *Name ID format* field.
+3. Select *Email* for the *Application username* field.
+4. Ably requires users' full names. Okta can share user profile fields as SAML attributes.
+5. Ensure you assign users to the newly created Okta application.
+6. Note down the *Identity Provider metadata* from Okta.
+
+In your Ably account dashboard:
+
+1. Log in to your "account":https://ably.com/accounts/any.
+2. Select *Settings* from the account navigation dropdown.
+3. Complete the SSO fields with the values obtained from Okta:
+** *Identity Provider Single Sign-On URL*
+** *Identity Provider Issuer*
+** *X.509 Certificate*
+4. *Save* the authentication settings.
+
+h3(#google). Google Workspace
+
+To enable SSO using "Google Workspace":https://workspace.google.com/ as the identity provider, configure the following properties in your Ably account dashboard and your Google Workspace:
+
+In your Ably account dashboard:
+
+1. Log in to your "account":https://ably.com/accounts/any.
+2. Select *Settings* from the account navigation dropdown.
+3. Toggle *Enable Single Sign-On* under the *Authentication Settings* section.
+4. Note down the *Single sign-on URL* and *Audience URI* values.
+
+In your Google Workspace account:
+
+Use the "Google Workspace guide":https://support.google.com/a/answer/6087519?hl=en for enabling SSO.
+
+1. Upload the Ably logo.
+2. Copy and paste the metadata configuration into your Ably account:
+** *Identity Provider Single Sign-On URL*
+** Use *Entity ID* from Google Workspace as the *Identity Provider Issuer* in your Ably account.
+** *X.509 Certificate*
+3. *Save* the authentication setting changes in your Ably account.
+4. Copy and paste the SAML settings from your Ably account into Google Workspace:
+** Use *Single sign on URL* from your Ably account as the *ACS URL* in Google Workspace.
+** Use *SP Entity Id* from your Ably account as the *Entity ID* in Google Workspace.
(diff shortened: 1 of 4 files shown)


content/account/sso.textile
```diff
@@ -0,0 +1,97 @@
...
+2. Select *Settings* from the account navigation dropdown.
+3. Complete the SSO fields with the values obtained from Okta:
+** *Identity Provider Single Sign-On URL*
+** *Identity Provider Issuer*
+** *X.509 Certificate*
+4. *Save* the authentication settings.
+
+h3(#google). Google Workspace
+
+To enable SSO using "Google Workspace":https://workspace.google.com/ as the identity provider:
```
GregHolmes, line 55: It might just be me, but this and line 57 don't read right. Would it be better to have them under the same line? You're kind of telling them to do this.. then in the dashboard to this.

`To enable SSO using "Google Workspace":https://workspace.google.com/ as the identity provider, in your Ably account dashboard:` or something similar?

GregHolmes (commented):
I could only find one thing to ask.

content/account/sso.textile
```diff
@@ -0,0 +1,97 @@
...
+In your Okta account:
+
+Use the "Okta guide":https://developer.okta.com/docs/guides/saml-application-setup/overview/ for enabling SSO.
+
+1. Upload the Ably logo.
+2. Select *EmailAddress* for the *Name ID format* field.
+3. Select *Email* for the *Application username* field.
+4. Ably requires users' full names. Okta can share user profile fields as SAML attributes.
+5. Ensure you assign users to the newly created Okta application.
+6. Note down the *Identity provider metadata* from Okta.
```
franrob-projects, line 41: The Okta docs use "Identity Provider metadata". Very small detail. Maybe not worth pursuing. 

```suggestion
6. Note down the *Identity Provider metadata* from Okta.
```

franrob-projects (approved):
Everything looks good! I have checked the Google Workspace and Okta property names and the SSO enablement processes using their docs -- as I need identity provider accounts to check via their dashboard. 

I couldn't include the Ably property names and review the SSO enablement processes because I need access to the Enterprise account. As agreed, @denissellu  will check this part.

content/account/sso.textile
kavalerov: There are two sections, basically, `In your Ably account dashboard` and `In your Google Workspace account`, but I see how it is a bit confusing. 

Maybe then on this line something like `To enable SSO using "Google Workspace":https://workspace.google.com/ as the identity provider, first you need to do the following in your Ably account dashboard:`, and then instead of lines 63 and 65 do something like `Then use the "Google Workspace guide":https://support.google.com/a/answer/6087519?hl=en to enable SSO in your Google Workspace account:`

content/account/sso.textile
```diff
@@ -0,0 +1,97 @@
+---
+title: Single sign-on (SSO)
+meta_description: "Single sign-on enables users to authenticate with Ably using your own identity provider."
+meta_keywords: "sso, saml, strict mode, single sign-on"
+---
+
+Single sign-on (SSO) enables your users to authenticate via any SAML-compatible identity provider.
+
+<aside date-type='note'>
+<p>User provisioning and deprovisioning via SCIM or SAML JIT is not supported.</p>
```
kavalerov, line 10: maybe `is not currently supported`? To hint that we may build this in the future?

content/account/users.textile
```diff
@@ -0,0 +1,72 @@
...
+|_. Permission |_. Standard |_. Admin |_. Owner |
+| View all apps | ✓ | ✓ | ✓ |
+| View app configuration | ✓ | ✓ | ✓ |
+| View app settings | ✓ | ✓ | ✓ |
+| View "app statistics":/metadata-stats/stats#app | ✓ | ✓ | ✓ |
+| View "account statistics":/metadata-stats/stats#account | ✓ | ✓ | ✓ |
+| Configure own "2FA":/account/2fa | ✓ | ✓ | ✓ |
+| "Invite new users":#invite | - | ✓ | ✓ |
+| "Remove existing users":#remove | - | ✓ | ✓ |
+| Manage "API keys":/auth | - | ✓ | ✓ |
```
kavalerov, line 28: Probably not needed yet, but once we add a Billing role it will be useful to show another line for `View API keys`, to show that Standard users can view them, and Billing users can not

content/account/users.textile
```diff
@@ -0,0 +1,72 @@
...
+meta_keywords: "users, user management, admin, owner, permission"
+---
+
+The user that creates an Ably account is assigned the account owner role. An account owner has permission to undertake any action within an account, such as inviting new users. There are two other account roles that inherit a subset of an account owner's permissions. An account may only have a single account owner.
+
+h2(#roles). User roles
+
+Users are assigned one of three roles:
+
+* Standard
```
kavalerov, line 13: I know we call it standard right now (or, at least, that's how we called it on the corresponding page in FAQs, I don't think we have used term `Standard` anywhere else), but I am wondering if we can rename it in this PR, instead of doing it in the next where we will add the `Billing` role? 

The primary reason for to use `developer` is that it actually provides more clarity and information on what the role entails. `Standard` doesn't really mean anything other than "not admin", but in reality this role has access to some technical details (i.e. API keys), that other future roles (i.e. Billing) will not have access to. Therefore, I am keen to start using `Developer` instead of `Standard`

content/account/sso.textile
m-hulbert: Fixed, thanks!

content/account/users.textile
m-hulbert: Yep, I can update that as I only used 'standard' to differentiate it from an admin.

content/account/users.textile
m-hulbert: I think this would be a little confusing at the moment, but we can definitely add an additional line when another role is added.

content/account/sso.textile
m-hulbert: Good point: I've reworded this and the same situation in the Okta instructions.

content/account/users.textile
```diff
@@ -0,0 +1,72 @@
+---
+title: User management
+meta_description: "Learn how to manage users, user roles, and the permissions associated with each role."
+meta_keywords: "users, user management, admin, owner, permission"
+---
+
+The user that creates an Ably account is assigned the account owner role. An account owner has permission to undertake any action within an account, such as inviting new users. There are two other account roles that inherit a subset of an account owner's permissions. An account may only have a single account owner.
+
```
denissellu, line 8:
```suggestion
The user that creates an Ably account is assigned the account owner role. An account owner has permission to undertake any action within an account, such as inviting new users. There are two other account roles that inherit a subset of an account owner's permissions. An account can only have a single account owner.

```

should we make it clear that it's not a `may` as it isn't possible to have multple account owners? or maybe it means the same?

content/account/sso.textile
```diff
@@ -0,0 +1,97 @@
...
+
+h3(#google). Google Workspace
+
+To enable SSO using "Google Workspace":https://workspace.google.com/ as the identity provider, configure the following properties in your Ably account dashboard and your Google Workspace:
+
+In your Ably account dashboard:
+
+1. Log in to your "account":https://ably.com/accounts/any.
+2. Select *Settings* from the account navigation dropdown.
+3. Toggle *Enable Single Sign-On* under the *Authentication Settings* section.
```
denissellu, line 61:
```suggestion
3. Toggle *Enable Single Sign-On* under the *Authentication Settings* section.
4. Note down the *Single sign-on URL* and *Audience URI* values.
```
is it worth having that here too?

denissellu (commented):
This looks good to me, just the two comments i left

content/account/users.textile
m-hulbert: Good point! Fixed.

content/account/sso.textile
m-hulbert: Yep, would make sense so I've copied it across.

denissellu (approved):

PR #2112, opened 2024-01-25
Fix wording in `untilAttach` documentation

lawrence-forooghian:
- "since" → "before" (wording was incorrect)
- "received" → "published" ("received" suggests that it’s something that the client received, which it wasn’t because we’re talking about messages they missed)


m-hulbert (approved):
Thanks @lawrence-forooghian ! 🥇

PR #2127, opened 2024-02-12
Chore update AblyUI from v11.0.0 to v13.2.2

aleksandar-r:
## Description

Docs project has been using AblyUi v11.0.0 while latest is v13.2.2 which adds a lot of changes and improvements.

## Review

Get into preview and check pages and classes, does UI uphold or is it falling apart somewhere.


kennethkalmer (changes requested):
@aleksandar-r I didn't expect this to throw up the linter errors like it did, would you mind just fixing the errors (not the warnings) that came up in a separate commit on this PR?

PR #2132, opened 2024-02-21
chore: update README with openssl workaround instructions

jamiehenson:
## Description

Updating the README to contain a solution to difficulties experienced when installing `ruby` via `asdf`, due to mismatched `openssl` versions.

## Review

Documentation update, no further testing required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,16 @@ This is a static site generated using [Gatsby](https://www.gatsbyjs.com/) and do
 
 Install node & npm.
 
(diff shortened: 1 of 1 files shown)


kennethkalmer (approved):
This looks good to me! Interesting catch, I had a look locally and I have both `openssl@1.1` & `openssl@3` installed.

@m-hulbert good addition to the README?

denissellu (approved):

jamiehenson:
Merging as the README addition is purely additive, and so that this doesn't hang over into next week. Happy to make further corrections if needed.

PR #2134, opened 2024-02-22
COL-673: Amends Models examples

franrob-projects:
This PR:

- Removes `name` from `modelcClient` options throughout LiveSync docs.
- We don't require both `name` and `channelName` properties.

[COL-686: Remove the reference to `name` property from the docs](https://ably.atlassian.net/browse/COL-686)


snikidev (approved):
LGTM 👍

PR #2139, opened 2024-02-29
[FIX]: use-search useEffect dependencies 

aleksandar-r:
## Description

We are not certain all functions in useEffect hooks inside useSearch hook will not change, so in order to prevent any unwanted behaviour (like failing tests because of infinite loop), I've wrapped them all inside useCallback so they will be same by reference.

## Review
In console type:

`npm run test`

and check if all tests are passing or Search.spec.tsx is going indefinitely

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -47,6 +47,15 @@
       "plugins": ["@typescript-eslint"],
       "rules": {
         // disable the rule for all files
+        "@typescript-eslint/no-unused-vars": [
+          "warn",
+          {
+            "argsIgnorePattern": "^_",
+            "varsIgnorePattern": "^_",
+            "destructuredArrayIgnorePattern": "^_",
+            "ignoreRestSiblings": true
+          }
+        ],
         "react/react-in-jsx-scope": "off",
         "@typescript-eslint/explicit-module-boundary-types": "off",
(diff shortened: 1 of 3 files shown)


jamiehenson (commented):
Usage of `useCallback` and deps arrays look good to me (and more importantly eslint 😄) 👍 

Whilst the focus is on this file and since the scope is small, I would also consider swapping the `UseSearchProps` interface for
```
interface ConfigureClientData {
  // NOTE: `client` is the AddSearchClient instance that is not typed
  client: unknown;
  query?: string;
  page: State['page'];
  pageLength: number;
}

interface UseSearchProps {
  addsearchApiKey?: string;
  enableParamsSync?: boolean;
  pageLength?: number;
  configureClient?: (data: ConfigureClientData) => void;
}
```
There's another warning on this file that could also be addressed. `argsIgnorePattern` as an option for the `no-unused-vars` ESLint rule ([link](https://eslint.org/docs/latest/rules/no-unused-vars#argsignorepattern)) would allow underscored params (for example) to be used for when we don't care about destructured elements (i.e. L125). 

The warning on L2 should ideally stand as an incentive to add types for that import or remove it 👉

jamiehenson (approved):
💪

PR #2145, opened 2024-03-05
[WEB-3531] Internal links rework image fix

jamiehenson:
## Description

~~Provides a fix to the broken image links by conditionally appending the path prefix to image srcs, depending on whether it is already present~~

Turns out the work we did in #2101 had a huge impact on our images when we went live, and in turn needed a complete rework of how we handle images in the project (for the better).

As per Gatsby's recommendations we now `import` as many images as possible, and where not the images have been moved to `src/images` (instead of `static/images`) and rendered via `GatsbyImage` dynamically.

The old `StaticImage` component has been refactored to support `assetPrefix`, but it really is a last resort for us and should ideally not be used.

Also bumps the Gatsby deps to the latest minors, and removes a Gatsby config in JS that was conflicting with the TS version.

## Review

Go to review link and see if image assets are broken. Additionally, build `docs` locally with `npm run rebuild` and see if the resulting site has broken images.

The review app has been built with the `ASSET_PREFIX` set, so we can inspect the HTML to see the effects working:

---

Monosnap Push Notifications - Device activation and subscription 2024-03-08 09-59-27

---

Monosnap Push Notifications - Device activation and subscription 2024-03-08 09-58-42


src/components/StaticImage.tsx
```diff
@@ -1,7 +1,11 @@
...
+import { withPrefix } from 'gatsby';
+import { ComponentProps } from 'react';
+import { pathPrefix } from '../../gatsby-config';
 
 type StaticImageProps = ComponentProps<'img'> & {
   src: string;
 };
 
-export const StaticImage = ({ src, ...attribs }: StaticImageProps) => <img src={src} {...attribs} />;
+export const StaticImage = ({ src, ...attribs }: StaticImageProps) => (
```
kennethkalmer, line 9: Other than updating the `src`, I wonder if this component can't be removed completely if we figure out why the assetPrefix isn't being respected. Most of the functionality here got removed in 9a2b9f6.

kennethkalmer:
@jamiehenson I significantly reworked the PR as it turned out to be quite the rabbit hole.

The review app has been configured with an `ASSET_PREFIX`, let's see how it goes

kennethkalmer:
@m-hulbert tagged you on this PR as it affects how images are referenced in the textile files, hope you're okay with the changes (see README updates)

m-hulbert (approved):
No problems with sourcing images in this way 🚀 

Pushed a quick fixup (haven't rebased) for the one missing image I spotted.

jamiehenson:
Just a note that I'm indeed looking at this, there were a number of TypeScript errors when I pulled this down so just doing some fixups on that before reviewing further

jamiehenson:
Since this was initially mine I cannot anoint this PR, but it's looking alright to me now 👍 @kennethkalmer my changes are in the final commit (https://github.com/ably/docs/pull/2145/commits/087dc7335192e30e807f3f3412f61195653b3e62)

aleksandar-r (approved):
@jamiehenson amazing work mate, you set a shining example to us all :1st_place_medal:

jamiehenson:
Ah, this is all @kennethkalmer - I just gave it a tiny push!

kennethkalmer:
@jamiehenson nailed the 200 response code on the homepage 🎉 

Monosnap Docs | Ably Realtime 2024-03-11 23-40-46

Let's get this merged in the morning, review on staging carefully before going live

PR #2147, opened 2024-03-06
[WEB-3580] Kill styled-components and related deps, port to css modules/tailwind

jamiehenson:
## Description

We currently use many styling tools in docs - Tailwind, CSS Modules, and Styled Components. Non-trivial tooling is required to support Styled Components, and as a tool it is fundamentally and operationally divergent from the other tools we use. Given as well that it’s only used in a small number of places, usage of this tool should be migrated, and the tool removed.

* [Jira ticket](https://ably.atlassian.net/browse/WEB-3580).

## Review

Look at the review site, make sure the docs aren't structurally or stylistically borked (this PR is purely a refactor, no functionality should be changed).

* [Page to review](link)


aleksandar-r (approved):
Great work Jamie, you the man

PR #2148, opened 2024-03-06
chore: migrate npm to Yarn 1 for inter-repo package manager consistency

jamiehenson:
## Description

The package manager `yarn` is used across `website` and `voltaire`, but here in `docs` - `npm` is used. This PR migrates from `npm` usage to `yarn 1.x`, which mirrors the other repos. This brings about an increased sense of consistency across the development experience, from which we can make more impactful decisions about package managers in future.

* [Jira ticket](link) - N/A

## Review

No impact on generated pages. Need to check that all build steps and package scripts work as they did before.


aleksandar-r:
I swear I see same prettier changes for the 20th time on a PR. I hope it will stop soon.

jamiehenson:
I getchu @aleksandar-r, but at least this isn't an unexpected change. 

As a consequence of moving to `yarn` here, peer dependencies have to be explicitly added to the deps (whereas npm 7+ can install them automatically), and I opted for the most up to date stable `prettier` package here, going from 3.0.x to 3.2.5. 

In 3.1 ([release notes](https://prettier.io/blog/2023/11/13/3.1.0.html)), we get this new ternary nesting formatting behaviour. I considered overriding it to maintain parity with how it was before, but thought it better not to oppose the default Prettier config in the long run.

aleksandar-r (approved):
Beautifully done mate, :1st_place_medal:

PR #2149, opened 2024-03-12
[FIX] Remove pathPrefix from redirects generated by `redirect_from`

kennethkalmer:
## Description

We had a combination of issues with our nginx config and the changes from setting `pathPrefix` that broke the redirects specified using the `redirect_from` frontmatter.

Our nginx redirects list, generated from the `redirect_from` frontmatter, was missing `/docs` on the destination, like this:

~~~
...
/docs/control-api /account/control-api;
/docs/rest/authentication /auth/;
...
~~~

This made it impossible for nginx to match the directs (the matching happens after the URLs get rewritten). This PR fixes it so the redirects config looks like:

~~~
...
/docs/rest/history /docs/storage-history/history;
/docs/rest/versions/v1.1/history /docs/storage-history/history;
...
~~~

It also fixes issues in the ordering of the directives in the nginx config file that resulted in us serving up the "Gatsby placeholder redirect HTML file" - which is a 200 status code - instead of the 301 redirect from our redirects map. This was because the `try_files` directive matched the `.html` file _before_ we checked the results from the redirect map. 

## Review

Fetching `/how-ably-works` on the review app should redirect to `/key-concepts`, not 404 or 200.

* https://ably-docs-fix-website-r-g46vtu.herokuapp.com/docs/how-ably-works
* https://ably-docs-fix-website-r-g46vtu.herokuapp.com/how-ably-works

~~~
$ curl -I https://ably-docs-fix-website-r-g46vtu.herokuapp.com/docs/how-ably-works
HTTP/1.1 301 Moved Permanently
...
Location: http://ably-docs-fix-website-r-g46vtu.herokuapp.com/docs/key-concepts
...
~~~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,11 +153,18 @@ jobs:
             ./bin/assert-redirect.sh /docs/ https://localhost/docs
             ./bin/assert-success.sh /docs/channels
       - run:
-          name: Smoke test quickstart redirects
+          name: Smoke test trailing slash redirects
           command: |
             ./bin/assert-success.sh /docs/getting-started/quickstart
             ./bin/assert-redirect.sh /docs/getting-started/quickstart/ \
                                      https://localhost/docs/getting-started/quickstart
+      - run:
+          name: Smoke test redirect_from redirects
+          command: |
+            ./bin/assert-redirect.sh /docs/how-ably-works \
+                                     http://localhost/docs/key-concepts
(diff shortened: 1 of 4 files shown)


jamiehenson (approved):
🎯

kennethkalmer:
Thanks @jamiehenson! Noticed a small typo in one of the redirect tests, pushed it now and will merge once done

PR #2151, opened 2024-03-12
[WEB-3644] fix: restore var needed for right sidebar sticky scrolling height calculation

jamiehenson:
## Description

Right sidebar was no longer sticky relative to the user's scrolling. This was because the `height` property on its wrapper element referenced a CSS variable that had dropped out of scope, and so was invalid. This PR moves that variable back into scope, and removes the file it was previously in (nothing used anything inside).

* [Jira ticket](https://ably.atlassian.net/browse/WEB-3644).

## Review

Go on the review link, check for the stickiness of the right hand sidebar.


kennethkalmer (approved):
Looks great @jamiehenson, thanks for the quick response. Are we sure the other styles aren't being used in the site? 🧹

jamiehenson:
Correct, this `global-entries` file wasn't even imported anywhere anyways

PR #2154, opened 2024-03-13
Fix broken links in navigation bar, where an extra /docs is added in Ably Accounts and Apps

GregHolmes:
## Description

Three links in the nav were broken because they had an extra /docs/ to the path.

* [Jira ticket](https://ably.atlassian.net/browse/EDU-1504).

## Review

By clicking any of the three links in the nav below to ensure they don't return a 404.

Screenshot 2024-03-13 at 09 47 41


franrob-projects (approved):
🚀

PR #2155, opened 2024-03-13
[WEB-3567] Increase CircleCI resource class

kennethkalmer:
## Description

Now that we're doing full blown image processing during the build step on CI, we're starting to see an increase in build failures due to running out of memory. This then prevents Heroku from deploying `main` automatically due to false negatives.

This PR bumps the resource class for build from `medium+` to `large`, while dropping the explicit `resource_class` keys from other steps.

It also hands over installation of NPM packages to the circleci/node orb, which should do a better job with caching of dependencies than we can

## Review

Look at this build in CI and notice that:

* The yarn caches are actually used and restored
* The `build` step never fails due to running out of memory

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,7 +67,6 @@ jobs:
   install-dependencies:
     executor:
       name: default
-    resource_class: medium+
     steps:
       - checkout
       - restore-yarn-cache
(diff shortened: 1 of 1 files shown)


kennethkalmer:
@denissellu more of an _fyi_ for you, look how much cleaner the CircleCI config is when we defer dependency installation to the circleci/node orb :)

jamiehenson (approved):
_Praise the orb_

PR #2162, opened 2024-03-19
[FIX] Control API & API reference link issues

kennethkalmer:
## Description

Links to the Control API reference need to be explicitly prefixed with `/docs` because the Control API reference is external to the docs site despite sharing URLs.

The summary links in the headings weren't `withPrefix` aware because they use the `FeatuedLink` component from Gatsby. They are now passed through `withPrefix` if they are local.

## Review

* `/docs/api`


m-hulbert (approved):
🚀

PR #2174, opened 2024-04-08
[DTP-707] Implement new IA for LiveSync docs

GregHolmes:
## Description

A current work in progress of implementing the new IA for the LiveSync documentation

* [Jira ticket](https://ably.atlassian.net/browse/DTP-707).

## Review

* [Page to review](https://ably-docs-dtp-707-imple-ayte6a.herokuapp.com/docs/livesync)


content/livesync/frontend.textile
```diff
@@ -0,0 +1,373 @@
...
+
+Create a model using the @models.get()@ method on the client. If you specify a name that already exists, it will be returned.
+
+Instantiate a model using a unique channel name. A unique channel name identifies the model on the client and corresponds to the "channel":/channels name used to subscribe to state updates from the backend:
+
+```[javascript]
+const model = modelsClient.models.get({
+  name: 'Posts',
+  channelName: 'posts',
+  sync,
```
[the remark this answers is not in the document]
Srushtika: Having read further down, I think we need to provide some high level context around how models works at top level intro paragraph before diving into the details.

content/livesync/frontend.textile
```diff
@@ -0,0 +1,373 @@
...
+
+Having each model on its own channel reduces processing overhead, ensuring models only receive relevant messages. While multiple models can share a channel, it requires careful message filtering. Channels aid scalability, but finding the right balance between channels, models, and message frequency is crucial.
+
+For infrequently updated models, grouping them on the same channel may be viable, while frequently changing models benefit from individual channels. Starting with one channel per model and one model per domain object is a good baseline, adjusting as needed.
+
+h2(#create-sync). Create a Sync function
+
+The sync function instructs your model to initialize with the latest data from the backend and is necessary when creating a model. It can be any function that optionally accepts parameters and returns a promise with the latest state of your data model, along with a "@sequenceId@":#sequence-id.
```
Srushtika, line 127: I think this is confusing and doesn't help me understand what the purpose of a sync function is. I think we should separate what a certain functionality offers to the customers vs what it facilitates towards the working of the library itself.

As a developer reading this, the questions I have in mind are "what is a sync function? why is it needed? what does it help me do? how does it help the models sdk  work correctly?

Srushtika: It's also not clear when we say backend, if we mean their backend or our backend, for example.

content/livesync/frontend.textile
```diff
@@ -0,0 +1,373 @@
...
+
+The following is an example sync function that fetches the latest model state from your database by calling `/api/posts` endpoint in your backend:
```
Srushtika, line 129:
```suggestion
The following is an example sync function that fetches the latest model state from your database by calling the @/api/posts@ endpoint in your backend:
```

content/livesync/frontend.textile
```diff
@@ -0,0 +1,373 @@
...
+
+```[javascript]
+async function sync() {
+  const result = await fetch(`/api/posts`);
+
+  return result.json();
+}
+```
+
+The following example is an example output from the sync function:
```
Srushtika, line 139:
```suggestion
The following is an example output from the sync function i.e. a direct result returned from your endpoint:
```

content/livesync/frontend.textile
Srushtika, line 127: At this point, I don't know what a sequenceId is, how to generate or retrieve it, etc. We need to provide more context here before linking off to a separate section.

content/livesync/frontend.textile
```diff
@@ -0,0 +1,373 @@
...
+```[javascript]
+const model = modelsClient.models.get({
+  name: 'Posts',
+  channelName: 'posts',
+  sync,
+  merge,
+});
+```
+
+Call the sync function directly on the model to bootstrap its state during the initial page load:
```
Srushtika, line 170: When do I call the sync function? We we say 'directly on the model', I think we should be explicitly saying "directly on the model instance"

content/livesync/frontend.textile
```diff
@@ -0,0 +1,373 @@
...
+
+```[javascript]
+await model.sync();
+```
+
+The SDK automatically attempts to re-execute the sync function in the case of an error, as specified by the @syncOptions.retryStrategy@ defined in your "@ClientOptions@":/livesync/models#options.
```
Srushtika, line 176: It's unclear to me what goes into that client option? What are the options I can set for this particular client option? Here it says that I should define it but as a reader, I don't know what to define it as.

Srushtika: The link is incorrect - takes me to the old page as the url of this page is different now.

content/livesync/frontend.textile
```diff
@@ -0,0 +1,373 @@
...
+
+h3(#sequence-id). Sequence ID
```
Srushtika, line 178: What is the thinking behind placing this section at this point in the page, OOI?

Srushtika (changes requested):
Hi @GregHolmes , this is a good start! 
I've reviewed the first half of the models page and added a bunch of comments. I think some of the review comments will help you update some of the other sections too without waiting on my review. I can revisit this again after the changes.

content/livesync/frontend.textile
GregHolmes: I thought adding the sequence ID under the sync section was important because you cannot have the `sync` function without the sequence ID, If you only brief over the sequence ID here, it'll leave the developer somewhat confused over what it is. 

Is there an issue with this or do you think it should be elsewhere?

content/livesync/frontend.textile
```diff
@@ -0,0 +1,373 @@
...
+Instantiate a realtime client using the Ably JavaScript SDK and pass the generated client into the Models constructor:
+
+```[javascript]
+const ably = new Realtime({ key: '...' });
+const modelsClient = new ModelsClient({ ably });
+```
+
+h3(#client-options). ClientOptions
+
+<aside data-type='note'>
```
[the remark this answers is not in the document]
GregHolmes: I've done this, but I do worry that the API reference is too low level to be of interest to anyone, it doesn't really help unless you already understand the Models SDK and how to use it. Even then it's only really relevant when you're looking for arguments or responses on a specific method call I think.

content/livesync/frontend.textile
```diff
@@ -0,0 +1,373 @@
...
+  name: 'Posts',
+  channelName: 'posts',
+  sync,
+  merge,
+});
+```
+
+TODO: Diagram of how the model components work together
+
+h2(#designing). Designing your Channels and Models
```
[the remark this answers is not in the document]
GregHolmes: I think this is an interesting question. It could go one of two ways. If you're trying to get set up or a basic understanding of LiveSync and how to implement it or just getting started do you want to know this first? Or is this a piece of information you'd want to learn as you build your actual intended application? 

This was discussed quite a bit on Wednesday. We have too much information overload on most of our docs pages that may never be needed for the developer, or will only be needed later. May or may not be for this subject. But yeah. It could go either side I think.

content/livesync/models.textile
```diff
@@ -1,82 +1,160 @@
...
+Instantiate a realtime client using the Ably JavaScript SDK and pass the generated client into the Models constructor:
+
+```[javascript]
+const ably = new Realtime({ key: '...' });
+const modelsClient = new ModelsClient({ ably });
+```
+
+h3(#client-options). ClientOptions
+
+The Models SDK uses default @ClientOptions@, customization is not a requirement especially if using this library for the first time.
```
Srushtika, line 58:
```suggestion
The Models SDK uses default set of @ClientOptions@. Customization of these options is not a requirement, especially if using this library for the first time.
```

content/livesync/models.textile
```diff
@@ -1,82 +1,160 @@
...
+
+In addition to the underlying Ably realtime client, you can provide a number of other "@ClientOptions@":/api/realtime-sdk/types#client-options to configure the behavior of the Models SDK:
```
Srushtika, line 60: Is the underlying Ably realtime client one of the client options?

Srushtika: If so, I don't think the link to ably instance's client options is useful as it's easy to confuse the client options of the ably instance vs the client options of the model instance

content/livesync/models.textile
```diff
@@ -1,82 +1,160 @@
...
 
-The following is an example sync function that fetches the latest model state from your database:
+For infrequently updated models, grouping them on the same channel may be viable, while frequently changing models benefit from individual channels. Starting with one channel per model and one model per domain object is a good baseline, adjusting as needed.
+
+h2(#create-sync). Create a Sync function
+
+The sync function is one of the core functions required for the Models SDK to function. This function is called when the model is initialized to retrieve the latest data from your backend.
+
+The function can contain any custom functionality within it, but needs to return a promise with two objects within the response. The first object is a field called a "@sequenceId@":#sequence-id, which is used to identify the point in the stream of change events that corresponds to the current version of the database's state.
```
Srushtika, line 116: We say that it needs to return a promise with two objects within the response but we only describe one of those before showing an example.

content/livesync/models.textile
```diff
@@ -1,82 +1,160 @@
...
+In general, when designing your channels and models, aim for a 1:1:1 relationship between channels, models, and your domain model objects. Channels facilitate data partitioning and access control. For instance, in a task management app, each task can be represented by a model with its own channel, ensuring updates are specific to each task. Ably's "Token Capabilities":/auth/capabilities are applied on a per channel basis allowing fine-grained access control for each data model in your application.
 
-The following is an example sync function that fetches the latest model state from your database:
+For infrequently updated models, grouping them on the same channel may be viable, while frequently changing models benefit from individual channels. Starting with one channel per model and one model per domain object is a good baseline, adjusting as needed.
+
+h2(#create-sync). Create a Sync function
+
+The sync function is one of the core functions required for the Models SDK to function. This function is called when the model is initialized to retrieve the latest data from your backend.
```
Srushtika, line 114:
```suggestion
The sync function is one of the core functions in the Models SDK. This function is called when the model is initialized to retrieve the latest data from your backend.
```

content/livesync/models.textile
```diff
@@ -1,82 +1,160 @@
...
+
+The function can contain any custom functionality within it, but needs to return a promise with two objects within the response. The first object is a field called a "@sequenceId@":#sequence-id, which is used to identify the point in the stream of change events that corresponds to the current version of the database's state.
+
+The following is an example sync function that fetches the latest model state from your database by calling the @/api/posts@ endpoint in your backend:
 
 ```[javascript]
-async function sync(id: number, page: number) {
```
Srushtika, line 38: I think the textile style code highlights in the code snippets comments are confusing. Perhaps use back ticks or double quotes?

content/livesync/models.textile
```diff
@@ -1,82 +1,160 @@
...
 
-h3(#sequence). Sequence ID
+The @sequenceId@ enables the Models SDK to identify the point in the stream of change events that corresponds to the current version of the database's state.
 
-The @sequenceId@ enables the SDK to identify the point in the stream of change events that corresponds to the current version of the database's state.
-
-Internally, the SDK replays change events starting from the point indicated by the @sequenceId@ to update the model state with any realtime changes occurring since the state was read from the database. 
+Internally, the SDK replays change events starting from the point indicated by the @sequenceId@ to update the model state with any realtime changes occurring since the state was read from the database.
 
 The sync function calls your backend endpoint to retrieve the highest @sequenceId@ present in the outbox when reading the state:
```
Srushtika, line 159: Hmm. I'm not sure this is correct. Let's discuss this.

content/livesync/models.textile
```diff
@@ -91,25 +169,24 @@ SELECT COALESCE(MAX(sequence_id), 0) FROM outbox;
...
+After the SDK executes the sync function, it internally connects to the "channel":/channels and begins "subscribing":/channels to live, yet unprocessed, messages.  It then paginates backward through the channel's message history, starting from the attachment point, using "@untilAttach@":/storage-history/history#until-attach. The number of items on each page can be configured via the @syncOptions.historyPageSize@ defined in your "@ClientOptions@":#client-options.
 
 When a message is sent to the @sequenceId@, the model processes subsequent messages to maintain continuity with live messages. If the target @sequenceId@ is unreachable, there is insufficient message history to resume from the correct point to update the model. The SDK will attempt a re-sync. It re-syncs by calling the sync function again to acquire a newer version of the model state.
```
Srushtika, line 180: "When a message is sent to the @sequenceId@" - what does this mean?

content/livesync/models.textile
```diff
@@ -91,25 +169,24 @@ SELECT COALESCE(MAX(sequence_id), 0) FROM outbox;
...
 
 To ensure the model's state is up to date, cache the state obtained from the backend endpoint used by the sync function. The model state returned from this endpoint must not exceed the message retention period set on the channel. If no previous messages exist on the channel, the model is assumed to be a new state without any modifications.
 
-h2(#merge). Merge functions
+h2(#create-merge). Create a Merge function
 
 A merge function is required when you instantiate a model. The merge function instructs your model how to compute the next version of the model state upon receiving a change event from the backend. The merge function takes the previous model state and the change event as inputs and outputs the next version of the model state.
```
Srushtika, line 188:
```suggestion
A merge function is also required when you instantiate a model. The merge function instructs your model how to compute the next version of the model state upon receiving a change event from the backend. The merge function takes the previous model state and the change event as inputs and outputs the next version of the model state.
```

content/livesync/models.textile
```diff
@@ -141,17 +218,17 @@ The event passed to the merge function can be either confirmed or optimistic:
 
 The SDK might call the merge function multiple times with the same state and event; to rebase optimistic events onto the newly confirmed state. It's essential that the merge function, when executed with identical inputs, consistently produces the same output. The merge function should be pure and deterministic, not depending on any external state.
 
-h3(#buffer). Event buffer
+h3(#event-buffer). Event buffer
 
 Change events from your backend are delivered to the SDK as messages over the "channels":/channels. By default, messages are processed by the SDK in the order in which they are received.
 
-The SDK supports a short buffering time buffering for change events, facilitating short-term reordering and de-duplication. By default, his feature is not enabled. To activate it, set the @eventBufferOptions.bufferMS@ in the "@ClientOption@":/livesync/models#options to a non-zero value.
+The SDK supports a short buffering time buffering for change events, facilitating short-term reordering and de-duplication. By default, his feature is not enabled. To activate it, set the @eventBufferOptions.bufferMS@ in the "@ClientOption@":#client-options to a non-zero value.
```
Srushtika, line 225:
```suggestion
The SDK supports a short time buffering for change events, facilitating short-term reordering and de-duplication. By default, this feature is not enabled. To activate it, set the @eventBufferOptions.bufferMS@ in the "@ClientOption@":#client-options to a non-zero value.
```

content/livesync/models.textile
```diff
@@ -141,17 +218,17 @@ The event passed to the merge function can be either confirmed or optimistic:
...
 
 When using the event buffer, change events are de-duplicated and ordered according to their "@message.ID@":/api/realtime-sdk/messages#id, which corresponds to the change event's "@sequenceId@":#sequence.
```
Srushtika, line 227:
```suggestion
When using the event buffer, change events are de-duplicated and ordered according to their "@message.id@":/api/realtime-sdk/messages#id, which corresponds to the change event's "@sequenceId@":#sequence.
```

content/livesync/models.textile
```diff
@@ -141,17 +218,17 @@ The event passed to the merge function can be either confirmed or optimistic:
...
 
-By default, the events in the buffer are ordered numerically if the @message.ID@ can be coerced to a number. Otherwise, events will be ordered lexicographically by their @message.ID@. You can specify a custom ordering based on any part of the message via the @eventBufferOptions.eventOrderer@ "@ClientOption@":/livesync/models#options.
+By default, the events in the buffer are ordered numerically if the @message.ID@ can be coerced to a number. Otherwise, events will be ordered lexicographically by their @message.ID@. You can specify a custom ordering based on any part of the message via the @eventBufferOptions.eventOrderer@ "@ClientOption@":#client-options.
```
Srushtika, line 229:
```suggestion
By default, the events in the buffer are ordered numerically if the @message.id@ can be coerced to a number. Otherwise, events will be ordered lexicographically by their @message.id@. You can specify a custom ordering based on any part of the message via the @eventBufferOptions.eventOrderer@ "client option":#client-options.
```

content/livesync/models.textile
```diff
@@ -141,17 +218,17 @@ The event passed to the merge function can be either confirmed or optimistic:
...
 
-h2(#optimistic). Optimistic updates
+h2(#optimistic-updates). Add Optimistic Updates
```
Srushtika, line 231: "Await the confirmation or rejection of the optimistic update from the backend, or optionally cancel the optimistic update."

content/livesync/models.textile
```diff
@@ -193,8 +270,7 @@ try {
 <p>Be aware that updating the model state in your database happens entirely through your backend endpoint.</p>
```
Srushtika, line 270: I think in the section above this, we need to include a code snippet that shows how the merge function works with an optimistic update vs a confirmed update.

content/livesync/models.textile
```diff
@@ -209,42 +285,25 @@ INSERT INTO outbox (mutation_id, channel, name, ...) VALUES ('my-mutation-id', '
...
 The following example subscribes to model state changes:
 
 ```[javascript]
 model.subscribe((err, post) => { /* model state updated! */ });
 ```
 
-Subscriptions operate on an optimistic model by default. The callback is triggered whenever the optimistic state changes or confirmed changes occur. 
+Subscriptions operate on an optimistic model by default. The callback is triggered whenever the optimistic state changes or confirmed changes occur.
 
 To respond solely to confirmed changes, set the @optimistic@ option to @false@ in the @subscribe()@:
```
Srushtika, line 300:
```suggestion
To respond solely to confirmed changes, set the @optimistic@ option to @false@:
```

content/livesync/models.textile
```diff
@@ -209,42 +285,25 @@ INSERT INTO outbox (mutation_id, channel, name, ...) VALUES ('my-mutation-id', '
...
 
 ```[javascript]
 model.subscribe((err, post) => { /* ... */ }, { optimistic: false });
 ```
 
-h2(#lifecycle). Model lifecycle
+h2(#model-lifecycle). Model lifecycle
 
 A model instance can be in one of the following lifecycle states:
```
Srushtika, line 309: Not sure I understand "the model is attached to the channel and processing realtime." for `ready`

Srushtika: Perhaps it is realtime updates?

content/livesync/models.textile
```diff
@@ -263,7 +322,7 @@ model.on('ready', () => { /* model resumed */ });
 model.on('disposed', () => { /* model disposed */ });
 ```
 
-h3(#pause-resume). Pausing and resuming a model
+h3(#pausing-resuming). Pausing and resuming a model
 
 You can pause a model to prevent the model from consuming realtime updates, while reserving the ability to resume it at some point in the future. Pausing can be useful in instances such as when the UI rendering the model data is temporarily out-of-view.
```
Srushtika, line 328: "The following example un-pauses the model. Event processing will resume and changes will be made available to subscribers:" -->

"The following example resumes the model. Event processing will resume and changes will be made available to subscribers:"

content/livesync/models.textile
```diff
@@ -279,10 +338,14 @@ The following example un-pauses the model. Event processing will resume and chan
 await model.resume();
 ```
 
-h3(#dispose). Disposing of a model
+h3(#disposing). Disposing of a model
```
Srushtika, line 341: Is it disposing off a model?

content/livesync/models.textile
```diff
@@ -279,10 +338,14 @@ The following example un-pauses the model. Event processing will resume and chan
...
 
 When a model is no longer needed, disposed of it to free up resources:
```
Srushtika, line 343:
```suggestion
When a model is no longer needed, you can dispose it to free up resources:
```

content/livesync/models.textile
```diff
@@ -279,10 +338,14 @@ The following example un-pauses the model. Event processing will resume and chan
...
 
 ```[javascript]
 await model.dispose();
 ```
+
+h2(#api-references). API references
```
Srushtika, line 349:
```suggestion
h2(#api-reference). API Reference
```

content/livesync/models.textile
```diff
@@ -279,10 +338,14 @@ The following example un-pauses the model. Event processing will resume and chan
...
+
+For a breakdown of all the API calls, along with parameters and responses in the Models SDK view the "API references":https://sdk.ably.com/builds/ably-labs/models/main/typedoc/
```
Srushtika, line 351:
```suggestion
For a breakdown of all the API calls, along with parameters and responses in the Models SDK view the "API reference":https://sdk.ably.com/builds/ably-labs/models/main/typedoc/.
```

Srushtika (changes requested):
Great updates, @GregHolmes! I've added some minor comments for the data models page. Will review other pages in separate reviews.

content/livesync/database-connector.textile
```diff
@@ -150,11 +282,11 @@ event: message
 data: {"id":"1","connectionId":"CaqkrZ2N_0","timestamp":1700069266050,"encoding":"json","extras":{"headers":{"x-ably-models-event-uuid":"1"}},"channel":"foo","data":"{}","name":"test"}
 ```
 
-h2(#new-db). Run a new database
+h3(#new-database). Run with a new database
 
 Use "Railway":https://railway.app/ with a PostgreSQL plugin instead of using your "existing database":#existing-db.
```
Srushtika, line 287: I think it's important to specify what is Railway and why we are asking to use it when they choose a new db option.

content/livesync/database-connector.textile
```diff
@@ -251,116 +383,76 @@ event: message
 data: {"id":"1","connectionId":"CaqkrZ2N_0","timestamp":1700069266050,"encoding":"json","extras":{"headers":{"x-ably-models-event-uuid":"1"}},"channel":"foo","data":"{}","name":"test"}
 ```
 
-h2(#options). Configuration options
+h2(#monitoring). Monitoring
 
-<aside data-type='note'>
-<p>To correctly configure the connector, you need the Ably "API key":/auth#api-keys and the connection details of your database.</p>
-</aside>
+The Database Connector exposes an HTTP server that can be used to monitor instance health and metrics.
```
Srushtika, line 388:
```suggestion
The Database Connector exposes an HTTP server that can be used to monitor its health and metrics.
```

content/livesync/models.textile
GregHolmes: I'm sorry, I'm not sure what this comment is related to with the header.

content/livesync/models.textile
```diff
@@ -0,0 +1,357 @@
...
+}
+```
+
+<aside data-type='important'>
+<p>Be aware that updating the model state in your database happens entirely through your backend endpoint.</p>
+</aside>
+
+h3(#mutation-id). Mutation ID
+
+The SDK requires a mechanism to identify when a change event received from the backend matches an optimistic event that has already been applied locally.
```
zknill, line 277: > The SDK needs to identify when a change event received from the backend matches an optimistic event that has already been applied locally. To match confirmed and optimistic events the SDK uses a 'mutationId'. This ID can be any string, though it's commonly a "UUID":https://www.npmjs.com/package/uuid . The 'mutationId' is generated in your frontend code and assigned to the optimistic event when you call 'model.optimistic'. You then need to send this 'mutationId' to your backend as part of your API calls to mutate the model. Finally, the mutationId should be included in the confirmed event that's written to the outbox table. 

> If your backend decides to reject the the mutation, you should write the mutationId to the outbox with the 'rejected' column set to 'true'. This will cause the optimistic update to automatically rollback in the frontend client that triggered it. This rollback will be reflected in a 'rejected' `confirmation` promise returned from the call to `model.optimistic`.

> Optimistic mutations that are not confirmed or rejected within the timeout `optimisticEventOptions.timeout` defined in `ClientOptions` (default 5 seconds) will be automatically rolled back in the Model state.

content/livesync/models.textile
```diff
@@ -0,0 +1,357 @@
...
+-- mutate your data, e.g.:
+INSERT INTO comments (comment) VALUES ('New comment!');
+-- write change event to outbox, e.g.:
+INSERT INTO outbox (mutation_id, channel, name, ...) VALUES ('my-mutation-id', 'posts:123', 'addComment', ...);
+COMMIT;
+```
+
+h2(#subscribe). Subscribe to changes in the model
+
+To receive realtime notifications when there are changes in the state, you can subscribe to changes to the changes.
```
zknill, line 292: > ... you can subscribe to the model

The subscribe callback will be passed the updated model state when it changes. This is useful for updating elements in your application, for example the UI.

content/livesync/outbox-node-tables.textile
```diff
@@ -1,39 +1,139 @@
...
+INSERT INTO outbox (mutation_id, channel, name, data) VALUES ('my-mutation-id', 'posts:123', 'addComment', 'New comment!');
+COMMIT;
+```
+
+The following describes the columns in the outbox table:
+
+- @mutation_id@ := is the identifier for the mutation.
+- @channel@ := is the channel to deliver the mutation, in this case, @posts:123@.
+- @name@ := indicates a comment is added.
+- @data@ := is the actual content or details of the mutation.
```
zknill, line 33: Data _has_ to be a json encodable object. (not a single JSON value or a JSON array)

zknill (commented):
Nice work, I've added some wording, clarification and suggestions 🧠 ✍️

content/livesync/models.textile
```diff
@@ -0,0 +1,357 @@
...
+
+h3(#replay-from-history). Replay from history
+
+The SDK must initialize the model's state using the sync function and then precisely apply a sequence of change events from your backend, beginning at the correct point in the stream. This starting point is the @sequenceId@ your backend endpoint returns.
+
+After the SDK executes the sync function, it internally connects to the "channel":/channels and begins "subscribing":/channels to live, yet unprocessed, messages.  It then paginates backward through the channel's message history, starting from the attachment point, using "@untilAttach@":/storage-history/history#until-attach. The number of items on each page can be configured via the @syncOptions.historyPageSize@ defined in your "@ClientOptions@":#client-options.
+
+When a message is published to the channel with a new @sequenceId@, the model processes subsequent messages to maintain continuity with live messages. If the target @sequenceId@ is unreachable, there is insufficient message history to resume from the correct point to update the model. The SDK will attempt a re-sync. It re-syncs by calling the sync function again to acquire a newer version of the model state.
```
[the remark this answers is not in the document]
GregHolmes: @zknill  is this suggestion for all 3 paragraphs to be replaced?

```The SDK will initialise model state using the sync function, and capture the @sequenceId@ you return from your sync function. The SDK then connects to the model's channel, and will rewind backwards though history until it finds the correct point in the stream of channel messages (based on the sequenceId). You can configure how many messages are requested in each page of history when rewinding using the parameter @syncOptions.historyPageSize@ defined in your @ClientOptions@.

After the SDK executes the sync function, it internally connects to the "channel":/channels and begins "subscribing":/channels to live, yet unprocessed, messages.  It then paginates backward through the channel's message history, starting from the attachment point, using "@untilAttach@":/storage-history/history#until-attach. The number of items on each page can be configured via the @syncOptions.historyPageSize@ defined in your "@ClientOptions@":#client-options.

When a message is published to the channel with a new @sequenceId@, the model processes subsequent messages to maintain continuity with live messages. If the target @sequenceId@ is unreachable, there is insufficient message history to resume from the correct point to update the model. The SDK will attempt a re-sync. It re-syncs by calling the sync function again to acquire a newer version of the model state.```


These ones?

File: content/livesync/how-livesync-works.textile [hunk 36]
```diff
@@ -0,0 +1,55 @@
+---
+title: How LiveSync works
+meta_description: "LiveSync enables you to synchronize changes in your database to frontend clients at scale."
+product: livesync
+---
+
+LiveSync is a powerful real-time data synchronization feature designed to facilitate seamless collaboration and data consistency across distributed systems. Leveraging Pub/Sub Channels, LiveSync ensures that data updates are propagated instantly and reliably to all connected clients in real-time. By providing bidirectional synchronization capabilities, LiveSync enables developers to effortlessly synchronize data between clients and servers, as well as between different clients, regardless of their location or network conditions.
```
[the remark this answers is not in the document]
mschristensen: "seamless collaboration and data consistency across distributed systems" - our primary use case is really for synchronising changes from backend databases to frontend clients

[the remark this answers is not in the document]
mschristensen: "instantly" sounds a bit like an over-promise to me, nothing can be instant :)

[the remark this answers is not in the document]
mschristensen: "bidirectional synchronization capabilities" - livesync itself doesn't handle syncing writes from the client to the database, that must be handled by the developer through their existing write paths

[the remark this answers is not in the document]
mschristensen: "as well as between different clients" - this sounds like livesync is peer-to-peer, which it isn't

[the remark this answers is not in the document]
mschristensen: "regardless of their location or network conditions" - not quite true, if the client has no connectivity there's obviously nothing we can do about that, and we don't provide offline support. What we do do is provide connection recovery in case of an intermittent connectivity failure

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,95 @@
...
+
+LiveSync benefits applications requiring real-time collaboration, such as collaborative editing tools, multiplayer games, or financial trading platforms, where maintaining data consistency and low latency are paramount. Developers can build highly responsive and interactive applications that deliver an exceptional real-time user experience.
```
mschristensen, line 9: We should take the use cases from here: https://docs.google.com/document/d/1Fj8Pm_qrC96RVA6rfycLF9p8c95b7yBoFuJWaPIDxcU/edit#heading=h.st644l97ntg3

Collaborative editing does not include text editing, but does include form editing. Multiplayer gaming is suitable for turn-based games rather than high fps games. Not sure about the fit for financial trading platforms, the data volume there are likely too large.

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,95 @@
...
+
+h2(#channel-based-broadcasting). Channel-based broadcasting
+
+The mechanism for synchronizing data updates across devices using Ably's LiveSync closely mirrors the delivery model of real-time messages to subscribers via Ably's "pub/sub channels":https://ably.com/docs/channels. When a message is published on an Ably channel, it is immediately broadcasted to all subscribers of that channel. Similarly, with LiveSync, data updates added to the outbox table are then published to all subscribed clients in real-time, ensuring seamless data synchronization between the database and all clients.
+
+h2(#frontend). Frontend
+
+For all clients to receive realtime updates of the database table using LiveSync, the frontend data models are used. This uses the Models SDK to carry out the following steps:
```
mschristensen, line 17: "For all clients to receive realtime updates of the database table using LiveSync, the frontend data models are used." - not sure about this sentence, I would perhaps say e.g. "Clients consume realtime updates to the database table through the Models SDK, which carries out the following steps"

content/livesync/frontend.textile
GregHolmes: Ok. That makes sense. I've added a sentence that explains what where the sequence id comes from at the beginning.

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,95 @@
...
+
+In the Models SDK, incoming messages on a channel trigger the @merge@ function, which integrates the message content into the existing state of your application.
+
+h3(#subscribing). Subscribing to changes
+
+In the Models SDK, you can subscribe to a model to receive its updated state whenever changes occur. This enables you to seamlessly integrate the updated model state into your user interface or react to changes in various ways.
+
+h3(#optimistic-updates). Optimistic updates
+
+@Optimistic@ updates to enhance UI responsiveness, allowing updates to be applied optimistically in the UI before being confirmed in the database. Each optimistic update is associated with a mutation ID, which enables tracking and matching with events received on the channel. This ID should be included in the mutation sent to your backend and recorded in the outbox table to confirm or reject the update. Unconfirmed optimistic updates will be rolled back by the Models SDK automatically.
```
mschristensen, line 33: Optimistic updates ~to~ enhance UI responsiveness

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,95 @@
...
+
+h2(#use-cases). Use-cases
+
+There are many possible problems that LiveSync can solve or simplify for projects. Below is a list of some of these:
+
+h3(#crm-systems). Customer relationship management systems
+
+A customer relationship management (CRM) system helps organistations to build customer relationships and improve processes to increase sales, improve customer service and manage customer data.
+
+A problem LiveSync can resolve within a customer relationship management (CRM) system, is on the individual customer page rendering their details or other models related to the customer on this page. An example of another model is orders by this customer. When any piece of this detail or their order is updated, this can be automatically picked up by LiveSync and update all frontend clients subscribed to this customer.
```
mschristensen, line 57: This sounds a bit like it's describing the bits that livesync is more challenging to work with, i.e. having the same backend data represented in different models on the client, which requires multiple outbox writes (for each model).

See: https://docs.google.com/document/d/1Fj8Pm_qrC96RVA6rfycLF9p8c95b7yBoFuJWaPIDxcU/edit

>Models would not work well for the list pages or orders or customers, and would not work well for the relational aspects of the CRM (for example, it’d be hard to update all the individual ‘order’ models when a ‘customer’ changed their name, as each order would likely be its own model with it’s own channel, so the database would need to publish an event per-order-model).

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,95 @@
...
+
+h3(#task-management). Productivity applications / Task management apps
+
+Similarly to "CRM systems":#crm-systems above, LiveSync would assist on a specific task within a task management application, rendering the details and updating when a change has been persisted to the database by another person or part of the system.
+
+h3(#online-auction). Online auction
+
+An online auction application is the ability to list products on the platform and have bidders make bids on this product where at the end of the given time, the winner is the person with the highest bid.
+
+LiveSync can make it easier for bidders to view a single auction in realtime with changes in price reflected in every frontend clients view. Each auction item can be represented as a single data Model, and optimistic updates can make it possible for a bidder to submit a bit and see this immediately whether it is accepted by the server as the highest bid, or rejected because another user's bid is higher. This functionality means that the database can resolve two competing bids and store the outcome, which is reflected transactionally into Ably via the outbox table. Ably events and the database are never out of sync.
```
mschristensen, line 71: ~bit~ bid

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,95 @@
...
+
+h3(#collaborative). Collaborative form editing and building
+
+In collaborative form editing applications, LiveSync integrates seamlessly with tools like "Spaces":/products/spaces. Individual form elements, such as survey questions or presentation slides, can be locked for editing by users, with updates propagated to all users via channel events. Optimistic updates can be utilized for real-time validation during editing. It's crucial to use locking in such scenarios, as LiveSync follows a last-write-wins mechanism. Simultaneous editing of the same element by multiple users may lead to conflicts, resulting in the discarding of one user's updates.
```
mschristensen, line 75: >Simultaneous editing of the same element by multiple users may lead to conflicts, resulting in the discarding of one user's updates.

I'm not sure this is a selling point, more a disadvantage of last-write wins. It's not livesync per-se that has this property, it's just a reflection of the conflict resolution semantics of the backend & database.

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,95 @@
...
+
+h3(#ecommerce). Inventory availability in ecommerce
+
+In a shopping application, LiveSync can be employed to manage inventory tracking for items. For instance, a single Model can represent available sizes and the corresponding item counts. When a user makes a purchase, the server processes the order and transactionally publishes an event with the updated size and inventory counts using the outbox. While optimistic updates could be applied to the inventory count, they may be unnecessary since users typically do not concern themselves with inventory availability post-purchase.
```
mschristensen, line 79: >While optimistic updates could be applied to the inventory count, they may be unnecessary since users typically do not concern themselves with inventory availability post-purchase.

Fair, but not really necessary information to surface here imo - isn't the purpose of this section to highlight where livesync is strong?

mschristensen (changes requested):
Thanks Greg, have added some comments. I wonder if we should keep the use-cases to strictly describe where livesync is a good fit, rather than highlight where it isnt.

content/livesync/how-livesync-works.textile
GregHolmes: I think the problem here is though, that LiveSync for CRM is very specific on what it can help with. I've tried to not make it very clear but instead focusing on how on that part it can help. Do you have any recommendations on how to promote LiveSync for a CRM?

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,95 @@
+---
+title: How LiveSync works
+meta_description: "LiveSync enables you to synchronize changes in your database to frontend clients at scale."
+product: livesync
+---
+
+LiveSync is a powerful realtime data synchronization feature designed to facilitate the synchronising of changes from backend databases to frontend clients. Leveraging Pub/Sub Channels, LiveSync ensures that data updates are propagated reliably to all connected clients in realtime.
+
+LiveSync benefits applications requiring realtime collaboration, such as collaborative form editing, online Auction, task management applications, or simple multiplayer turn-based games, where maintaining data consistency and low latency are paramount. Developers can build highly responsive and interactive applications that deliver an exceptional realtime user experience.
```
Srushtika, line 9: I think this reads like marketing speak. Additionally, I think we are missing some context in terms of the context flow before jumping into channel based broadcasting.

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,95 @@
...
+
+h2(#channel-based-broadcasting). Channel-based broadcasting
+
+The mechanism for synchronizing data updates across devices using Ably's LiveSync closely mirrors the delivery model of realtime messages to subscribers via Ably's "pub/sub channels":https://ably.com/docs/channels. When a message is published on an Ably channel, it is immediately broadcasted to all subscribers of that channel. Similarly, with LiveSync, data updates added to the outbox table are then published to all subscribed clients in realtime, ensuring seamless data synchronization between the database and all clients.
```
Srushtika, line 13: I think the first sentence is confusing and unnecessary. It doesn't mirror it, it actually uses the channels to broadcast updates from the connector component.

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,95 @@
...
+
+h2(#frontend). Frontend
```
Srushtika, line 15: I think we may need to revamp the page structure for this in general. It's Step 1 Client makes changes and sees optimistic updates, Step 2 you make an entry in the DB, Step 3 connector watches and publishes update on channels, Step 4 Channels broadcast at scale and Step 5, Models again, merges the confirmed updates with frontend state

Srushtika: Actually having read further, I think we need to give a full overview first, then get into each topic in that order.

Srushtika (commented):
Hey Greg, I've added a few comments to the How LiveSync works page. I think we need to discuss this page, will ping on Slack. Separately, we'll have to update the home page for LiveSync as it still says 'early access' in the label and copy. Plus, it has "real-time", which is inconsistent with our tone of voice.

content/livesync/how-livesync-works.textile
kaschula: I agree there is bit of jump from this paragraph to the next part. Perhaps I high level overview of the the pattern. A write to an outbox table, this is picked up by our ADBC and published to the ably and then onto the subscribers.

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,95 @@
...
+
+Clients consume realtime updates to the database table through the Models SDK, which carries out the following steps:
```
kaschula, line 17: I think maybe this should be something like. "Clients consume realtime updates reflecting database table through....."

content/livesync/models.textile
```diff
@@ -0,0 +1,356 @@
...
+* @history@ if you intend to "sync from historical messages":#replay-from-history.
+
+h2(#install). Install
+
+The Models SDK requires a realtime client created using the "Ably JavaScript SDK":/getting-started/setup to interact with the Ably service.
+
+Install the Ably JavaScript SDK and the Models SDK from "NPM:":https://www.npmjs.com
+
+```[sh]
+npm install ably @ably-labs/models
```
kaschula, line 41: The user will have to install the ably package as well if working in a project that doesn't have ably already.

content/livesync/models.textile
```diff
@@ -0,0 +1,356 @@
...
+```[javascript]
+const ably = new Realtime({ key: '...' });
+const modelsClient = new ModelsClient({ ably });
+```
+
+h3(#client-options). ClientOptions
+
+The Models SDK uses a default set of @ClientOptions@. Customization of these options is not a requirement, especially if using this library for the first time.
+
+In addition to the underlying Ably realtime client, you can provide a number of other @ClientOptions@ to configure the behavior of the Models SDK:
```
kaschula, line 62: I feel like this paragraph and the one above kind of saying the same thing. Maybe we could merge them with something like: 

In addition to the underlying Ably realtime client, you can provide a number of other @ClientOptions@ to configure the default behavior of the Models SDK. Customization of these options is not a requirement, especially if using this library for the first time.

content/livesync/how-livesync-works.textile
mschristensen: Maybe focus on the collaborative form building aspect, rather than making it CRM-specific. E.g. if you have a screen open with a bunch of fields, you can see changes made by other users to those fields as they happen.

content/livesync/models.textile
```diff
@@ -0,0 +1,356 @@
...
+      'comments': []
+    }
+}
+```
+
+The SDK automatically attempts to re-execute the sync function in the case of an error, as specified by the @syncOptions.retryStrategy@ defined in your "@ClientOptions@":#client-options. To configure a @retryStrategy@ set an integer value of milliseconds for the time between retry attempts. Set this value to @-1@ to stop automatic retries.
+
+h3(#sequence-id). Sequence ID
+
+The @sequenceId@ is returned by the Database Connector along with the data that has been updated. This lets the Models SDK to identify the point in the stream of change events that corresponds to the current version of the database's state.
```
kaschula, line 150: I think this wording is wrong. The `sequenceId `is returned by the users sync function, not the Database Connector. The sync function will likely be calling some sort of API in which the sequenceId comes from the users backend.

It is returned along with the latest state at the point in time the sequenceId references.

The database connector doesn't expose any apis and is never interacted with by the models SDK or clients app.

mschristensen: Correct - and the sequence ID is obtained from the user's backend sync endpoint by querying the outbox table

content/livesync/how-livesync-works.textile
GregHolmes: I've updated this by adding another paragraph following this that gives an overview of how LiveSync works each step of the way.

content/livesync/models.textile
GregHolmes: The ably package is included in the install.

content/livesync/models.textile
```diff
@@ -0,0 +1,354 @@
...
+SELECT COALESCE(MAX(sequence_id), 0) FROM outbox;
+-- Commit the database transaction
+COMMIT;
+```
+
+Use the *read committed transaction isolation level* for the correct function, when reading the database state. "PostgreSQL uses this":https://www.postgresql.org/docs/7.2/xact-read-committed.html as default.
+
+h3(#replay-from-history). Replay from history
+
+The SDK will initialise model state using the sync function, and capture the @sequenceId@ you return from your sync function. The SDK then connects to the model's channel, and will rewind backwards though history until it finds the correct point in the stream of channel messages (based on the sequenceId). You can configure how many messages are requested in each page of history when rewinding using the parameter @syncOptions.historyPageSize@ defined in your @ClientOptions@.
```
m-hulbert, line 169:
```suggestion
The SDK will initialise model state using the sync function, and capture the @sequenceId@ you return from your sync function. The SDK then connects to the model's channel, and will rewind backwards though history until it finds the correct point in the stream of channel messages (based on the @sequenceId@). You can configure how many messages are requested in each page of history when rewinding using the parameter @syncOptions.historyPageSize@ defined in your @ClientOptions@.
```

content/livesync/models.textile
```diff
@@ -0,0 +1,354 @@
...
+
+The amount of history available to query on the channel is determined by your "message storage":/storage-history/storage configuration on the channel. This configuration must match the @syncOptions.messageRetentionPeriod@ defined in your "@ClientOptions@":#client-options. The SDK uses this configuration option as a hint as to whether to resynchronize via the sync function and skip paginating through history when messages are expected to have expired.
+
+To ensure the model's state is up to date, cache the state obtained from the backend endpoint used by the sync function. The model state returned from this endpoint must not exceed the message retention period set on the channel. If no previous messages exist on the channel, the model is assumed to be a new state without any modifications.
+
+h2(#create-merge). Create a Merge function
```
m-hulbert, line 179: "merge"

content/livesync/models.textile
```diff
@@ -0,0 +1,354 @@
...
+
+A merge function is required when you instantiate a model.
+
+When a message is received on the channel, the merge function is called to merge that new message into the existing Model state. The merge function is called with the existing Model state and the newly received channel message, and it should return the updated Model state (with the new message merged in).
```
m-hulbert, line 183:
```suggestion
When a message is received on the channel, the merge function is called to merge that new message into the existing model state. The merge function is called with the existing model state and the newly received channel message, and it should return the updated model state (with the new message merged in).
```

content/livesync/models.textile
```diff
@@ -0,0 +1,354 @@
...
+
+It's important that the merge function is pure and deterministic, and not depending on any external state. This is because the merge will often be called multiple times and in different orders (for example when replaying events after rewinding to a sequenceId, or when rebasing optimistic events on top of confirmed events).
```
m-hulbert, line 185:
```suggestion
It's important that the merge function is pure and deterministic, and doesn't depend on any external state. This is because the merge will often be called multiple times and in different orders (for example when replaying events after rewinding to a @sequenceId@, or when rebasing optimistic events on top of confirmed events).
```

content/livesync/models.textile
```diff
@@ -0,0 +1,354 @@
...
+  merge,
+});
+```
+
+The event passed to the merge function can be either confirmed or optimistic:
+
+- Confirmed := change events received from your backend. They describe the result of a change to the data which has been committed to your database.
+- "Optimistic":#optimistic := events that describe mutations that have happened locally, but have not yet been confirmed by your backend.
+
+The SDK might call the merge function multiple times with the same state and event; to rebase optimistic events onto the newly confirmed state. It's essential that the merge function, when executed with identical inputs, consistently produces the same output. The merge function should be pure and deterministic, not depending on any external state.
```
m-hulbert, line 216: A lot of this paragraph is repeated from before the examples.

content/livesync/models.textile
```diff
@@ -0,0 +1,354 @@
...
+
+h3(#event-buffer). Event buffer
+
+Change events from your backend are delivered to the SDK as messages over the "channels":/channels. By default, messages are processed by the SDK in the order in which they are received.
+
+The SDK supports a short time buffering for change events, facilitating short-term reordering and de-duplication. By default, this feature is not enabled. To activate it, set the @eventBufferOptions.bufferMS@ in the "@ClientOption@":#client-options to a non-zero value.
+
+When using the event buffer, change events are de-duplicated and ordered according to their "@message.id@":/api/realtime-sdk/messages#id, which corresponds to the change event's "@sequenceId@":#sequence.
+
+By default, the events in the buffer are ordered numerically if the @message.id@ can be coerced to a number. Otherwise, events will be ordered lexicographically by their @message.id@. You can specify a custom ordering based on any part of the message via the @eventBufferOptions.eventOrderer@ "client option":#client-options.
```
m-hulbert, line 226:
```suggestion
By default, the events in the buffer are ordered numerically if the @message.id@ can be coerced to a number. Otherwise, events will be ordered lexicographically by their @message.id@. You can specify a custom ordering based on any part of the message via the @eventBufferOptions.eventOrderer@ "@ClientOption@":#client-options.
```

content/livesync/models.textile
```diff
@@ -0,0 +1,354 @@
...
+
+h2(#optimistic-updates). Add Optimistic Updates
```
m-hulbert, line 228: "optimistic updates"

content/livesync/models.textile
```diff
@@ -0,0 +1,354 @@
...
+1. Call @model.optimistic()@ on your model with the optimistic event.
+2. Apply the corresponding change to your backend.
+3. Await the confirmation of the optimistic update from the backend, or optionally cancel the optimistic update.
+
+The @model.optimistic()@ function returns a promise resolving to two values:
+
+* A confirmation promise that resolves when the backend confirms or rejects the optimistic update.
+* A function to explicitly cancel the optimistic update.
+
+The following demonstrates an optimistic update to implementing changes in the model:
```
m-hulbert, line 245:
```suggestion
The following demonstrates an optimistic update to implement changes in the model:
```

content/livesync/models.textile
```diff
@@ -0,0 +1,354 @@
...
+    // something went wrong, cancel the optimistic update
+    cancel();
+}
+```
+
+<aside data-type='important'>
+<p>Be aware that updating the model state in your database happens entirely through your backend endpoint.</p>
+</aside>
+
+h3(#mutation-id). Mutation ID
```
m-hulbert, line 270: This should match the case of the parameter.

content/livesync/models.textile
```diff
@@ -0,0 +1,354 @@
...
+await model.dispose();
+```
+
+h2(#api-reference). API Reference
+
+For a breakdown of all the API calls, along with parameters and responses in the Models SDK view the "API reference":https://sdk.ably.com/builds/ably-labs/models/main/typedoc/.
+
+<aside data-type='note'>
+<p>Please "reach out":https://docs.google.com/forms/d/e/1FAIpQLSd00n1uxgXWPGvMjKwMVL1UDhFKMeh3bSrP52j9AfXifoU-Pg/viewform if you have any feedback.</p>
+</aside>
```
m-hulbert, line 354: Same comment as for the connector page.

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,69 @@
+---
+title: How LiveSync works
+meta_description: "LiveSync enables you to synchronize changes in your database to frontend clients at scale."
+product: livesync
+---
+
+LiveSync is a powerful realtime data synchronization feature designed to facilitate broadcasting realtime updates from backend databases to frontend clients at scale. Leveraging Pub/Sub Channels, LiveSync ensures that data updates are propagated reliably and in order to all connected clients in realtime.
```
m-hulbert, line 7: Feature or product?

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,69 @@
...
+
+LiveSync can be used for realtime collaboration, including collaborative form editing, online auctions, task management applications, or simple multiplayer turn-based games, where maintaining data consistency and low latency are important. Furthermore, developers can also build highly responsive and interactive applications with the optimistic updates design pattern offered out of the box.
+
+LiveSync has three key components, 1) a Models SDK that facilitates optimistic updates for the publisher and helps merge confirmed changes from the backend with the overall frontend state 2) a Database Connector at the backend that can watch changes in your database table using the outbox pattern and broadcast those updates and 3) the default Ably Pub/Sub channels that form the mechanism behind the scenes to broadcast those updates.
```
m-hulbert, line 11: I think this should be an unordered list as they aren't steps and also displayed as bullets rather than within the text.

I also don't think we need to refer to them as 'Ably Pub/Sub channels' everytime and can just use 'channels'

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,69 @@
...
+
+<a href="@content/diagrams/what-is-livesync.png" target="_blank">
+  <img src="@content/diagrams/what-is-livesync.png" style="width: 100%" alt="What is LiveSync">
+</a>
+
+h2(#channel-based-broadcasting). Channel-based broadcasting
+
+The mechanism for synchronizing data updates across devices using Ably's LiveSync uses Ably's "pub/sub channels":https://ably.com/docs/channels to publish realtime messages to subscribers. When a message is published on an Ably channel, it is immediately broadcasted to all subscribers of that channel. Similarly, with LiveSync, data updates added to the outbox table are then published to all subscribed clients in realtime, ensuring seamless data synchronization from the database to all clients.
```
m-hulbert, line 19: The first sentence here is a little muddled with 'using' and 'uses'. 

Broadcasted --> broadcast

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,69 @@
...
+
+h2(#frontend). Frontend
+
+Clients consume realtime updates reflecting database tables through the Models SDK, which carries out the following steps:
```
m-hulbert, line 23: Steps suggest these are sequential.

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,69 @@
...
+
+In the Models SDK, incoming messages on a channel trigger the @merge@ function, which integrates the message content into the existing state of your application.
+
+h3(#subscribing). Subscribing to changes
+
+In the Models SDK, you can subscribe to a model to receive its updated state whenever changes occur. This enables you to seamlessly integrate the updated model state into your user interface or react to changes in various ways.
+
+h3(#optimistic-updates). Optimistic updates
+
+@Optimistic@ updates enhance UI responsiveness, allowing updates to be applied optimistically in the UI before being confirmed in the database. Each optimistic update is associated with a mutation ID, which enables tracking and matching with events received on the channel. This ID should be included in the mutation sent to your backend and recorded in the outbox table to confirm or reject the update. Unconfirmed optimistic updates will be rolled back by the Models SDK automatically.
```
m-hulbert, line 39:
```suggestion
@Optimistic@ updates enhance UI responsiveness, enabling updates to be applied optimistically in the UI before being confirmed in the database. Each optimistic update is associated with a mutation ID, which enables tracking and matching with events received on the channel. This ID should be included in the mutation sent to your backend and recorded in the outbox table to confirm or reject the update. Unconfirmed optimistic updates will be rolled back by the Models SDK automatically.
```

I think it's `mutationId` too, right?

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,69 @@
...
+
+h2(#backend). Backend
```
m-hulbert, line 41: Ideally shouldn't have two sequential headers without some text between them.

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,69 @@
...
+
+h3(#outbox-table). Database Outbox table
```
m-hulbert, line 43: "outbox"?

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,69 @@
...
+
+The @outbox table@ serves as the equivalent of publishing to clients subscribed with Ably LiveSync. Automatically reacting to new records written to the outbox, the database connector publishes these records to clients. This approach ensures transactionality between database changes and records published to clients, enhancing data consistency and reliability.
+
+h3(#sequence-id). Sequence ID
+
+The Models SDK utilizes a sequence ID to ensure accurate matching of the model's state with the corresponding point in the change events stream obtained from Ably. This sequence ID, retrieved from your backend during the @sync@ function call, enables the Models SDK to synchronize the model's state effectively during the initial load process.
+
+h2(#use-cases). Use-cases
```
m-hulbert, line 51: "Use cases"?

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,69 @@
...
+
+h3(#optimistic-updates). Optimistic updates
+
+@Optimistic@ updates enhance UI responsiveness, allowing updates to be applied optimistically in the UI before being confirmed in the database. Each optimistic update is associated with a mutation ID, which enables tracking and matching with events received on the channel. This ID should be included in the mutation sent to your backend and recorded in the outbox table to confirm or reject the update. Unconfirmed optimistic updates will be rolled back by the Models SDK automatically.
+
+h2(#backend). Backend
+
+h3(#outbox-table). Database Outbox table
+
+The @outbox table@ serves as the equivalent of publishing to clients subscribed with Ably LiveSync. Automatically reacting to new records written to the outbox, the database connector publishes these records to clients. This approach ensures transactionality between database changes and records published to clients, enhancing data consistency and reliability.
```
m-hulbert, line 45: Mixture of "LiveSync" and "Ably LiveSync" throughout this page.

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,69 @@
...
+
+h3(#sequence-id). Sequence ID
+
+The Models SDK utilizes a sequence ID to ensure accurate matching of the model's state with the corresponding point in the change events stream obtained from Ably. This sequence ID, retrieved from your backend during the @sync@ function call, enables the Models SDK to synchronize the model's state effectively during the initial load process.
+
+h2(#use-cases). Use-cases
+
+There are many possible problems that LiveSync can solve or simplify for projects. Below is a list of some products / applications that LiveSync can help along with some of the data models that would be involved:
+
+Similarly to "CRM systems":#crm-systems above, LiveSync would assist on a specific task within a task management application, rendering the details and updating when a change has been persisted to the database by another person or part of the system.
```
m-hulbert, line 55: I think this paragraph was a row in the table below that can be removed?

m-hulbert (changes requested):
Looking good. Mostly editorial comments, but also a few things to fix:

* You need to add `redirect_from` frontmatter variables to redirect the existing pages that have been removed as part of this refresh.
* The root `index.textile` under `livesync` should be the 'how LiveSync works' page, rather than having to create a page called `how-livesync-works`.

content/livesync/how-livesync-works.textile
Srushtika: Should be product.

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,69 @@
...
+
+The @outbox table@ serves as the equivalent of publishing to clients subscribed with Ably LiveSync. Automatically reacting to new records written to the outbox, the database connector publishes these records to clients. This approach ensures transactionality between database changes and records published to clients, enhancing data consistency and reliability.
+
+h3(#sequence-id). Sequence ID
+
+The Models SDK utilizes a sequence ID to ensure accurate matching of the model's state with the corresponding point in the change events stream obtained from Ably. This sequence ID, retrieved from your backend during the @sync@ function call, enables the Models SDK to synchronize the model's state effectively during the initial load process.
+
+h2(#use-cases). Use-cases
+
+There are many possible problems that LiveSync can solve or simplify for projects. Below is a list of some products / applications that LiveSync can help along with some of the data models that would be involved:
```
Srushtika, line 53:
```suggestion
LiveSync can simplify a wide range of applications. Below is a list of a few examples and how LiveSync can enable a snappy experience in those applications:
```

content/livesync/how-livesync-works.textile
```diff
@@ -0,0 +1,69 @@
...
+
+Similarly to "CRM systems":#crm-systems above, LiveSync would assist on a specific task within a task management application, rendering the details and updating when a change has been persisted to the database by another person or part of the system.
+
+| Customer relationship management (CRM) systems | Within a CRM system, an individual customer page would be represented as a single data Model with its own channel. When a user updates the customer's information, optimistic updates can be used to show in the UI the update as greyed until the server returns a confirmation or rejection. The server will save this change in the database. The Database Connector will pick this change up and publish the changes to the frontend where the Model will update and reflect the changes in the UI for every actively subscribed frontend client. |
```
Srushtika, line 57:
```suggestion
| CRM systems | For a Customer Relationship Management or a CRM system, an individual customer details page would be represented as a single data model with its own channel. When a user updates the customer's information, optimistic updates can be used to show the update in the UI as greyed until the server returns a confirmation (or rejection) for all collaborating users. |
```

content/livesync/outbox-node-tables.textile
```diff
@@ -1,39 +1,139 @@
...
+```
+
+The following describes the columns in the outbox table:
+
+- @mutation_id@ := is the identifier for the mutation.
+- @channel@ := is the channel to deliver the mutation, in this case, @posts:123@.
+- @name@ := indicates a comment is added.
+- @data@ := is the actual content or details of the mutation in a JSON encodable object.
+
+For more information, see "outbox table schema.":#schema
```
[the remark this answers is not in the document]
GregHolmes: If it were immediately the next bit of information I wouldn't have added it, but it isn't that far down. So not necessarily necessary.  Removed it.

content/livesync/outbox-node-tables.textile
```diff
@@ -1,39 +1,139 @@
...
+
+<aside data-type='note'>
+  <p>The Database Connector will always keep at least one outbox record in the table, specifically the one with the largest @sequence_id@. This record is still marked as processed via a tombstone and will be deleted when subsequent records are written to the outbox. This approach ensures that the backend endpoint called by your "sync function":/livesync/frontend#create-sync can always obtain the @sequence_id@ at the time that the state is read from the database.</p>
```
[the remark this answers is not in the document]
GregHolmes: Possibly. I've updated it to reflect latest rather than largest.

content/livesync/index.textile
```diff
@@ -1,48 +1,83 @@
...
 
-The "Models SDK":/livesync/models/models enables you to easily create live and observable data models in your frontend applications, ensuring they remain synchronized with your database's state in realtime. By subscribing to a model, you receive the changes made by other users, enabling the creation of a seamless, multiplayer application experience.
+Let's take the example of a CRM application to understand how LiveSync can enable reliable realtime data sync — In a CRM app, an individual customer details page can be represented as a single data model with its own channel. When a user updates the customer's information, optimistic updates can immediately reflect the changes in the UI while the confirmation goes to a roundtrip through the database. When confirmed, the original user's UI update can be confirmed and all other clients can also see that update without needing to refresh the page.
 
-The Models SDK works as follows:
+h3(#known-limitations). Known limitations
```
Srushtika, line 65: LiveSync's current design might be limiting to data models based on graph data structures. As each data model is meant to work with a channel of it's own, any application changes that apply to multiple data models will need to be handled and published separately to the outbox table.

LiveSync might also be limiting to use-cases requiring complex conflict resolution logic and the "last write to the database wins" strategy isn't enough. Having said that, LiveSync can be easily coupled with the component locking feature in Spaces to make sure the individual UI components can be locked thus avoiding the need to resolve conflicts. For most use-cases like collaborative form building, component locking might actually a better user experience compared to co-editing capabilities.


Can we please replace this section with the above copy?

content/livesync/index.textile
```diff
@@ -1,48 +1,83 @@
...
+- applications where peers are trusted, data can be handled entirely peer-to-peer,
+- when functionality allows two users to collaborate on the same element in a collaborative form editing and building application. Conflict resolution will discard one of the collaborator's updates.
 
-* "Authentication":/auth and "connection management":/connect are efficiently managed by the underlying Ably JavaScript SDK.
+h2(#pricing). Pricing
 
-* The Ably JavaScript SDK enables you to extend your application's functionality with "Pub/Sub Channels":/products/channels.
+LiveSync "pricing":https://ably.com/pricing is based on the message consumption. This means that each update published from the Database Connector to Ably Channels is counted as a single message. The message then received by every frontend client subscribed to that channel is counted as one message. If, for example, one update is published by the Database Connector and there are three frontend clients subscribed, the one update will count as four messages and priced as such.
```
Srushtika, line 79:
```suggestion
LiveSync pricing is mainly based on message consumption (alongside concurrent connections and concurrent channels). This means that each update published from the Database Connector to Ably Channels is counted as a single message. The message then received by every frontend client subscribed to that channel is counted as one message. If, for example, one update is published by the Database Connector and there are three frontend clients subscribed, the one update will lead result in four messages and priced as such. Take a look at the "pricing page":https://ably.com/pricing for further information on how these parameters are charged.
```

content/livesync/index.textile
```diff
@@ -1,48 +1,83 @@
...
 
-h2(#dev-status). Development status
+h2(#development-status). Development status
 
-LiveSync is currently in the early release stage, primarily intended for experimental use and not yet suitable for production environments. Your "feedback":https://docs.google.com/forms/d/e/1FAIpQLSd00n1uxgXWPGvMjKwMVL1UDhFKMeh3bSrP52j9AfXifoU-Pg/viewform will help in prioritizing improvements and fixes in subsequent releases.
+LiveSync is in public alpha so that you can explore its capabilities. Your "feedback":https://docs.google.com/forms/d/e/1FAIpQLSd00n1uxgXWPGvMjKwMVL1UDhFKMeh3bSrP52j9AfXifoU-Pg/viewform will help prioritize improvements and fixes in a subsequent release. The features have been validated for a set of use-cases and are stable and perform well, but API changes may occur that impact functionality. The alpha is implemented to work under real-world situations and load, but you should use caution when deciding under what conditions to use it.
```
Srushtika, line 83:
```suggestion
LiveSync is in public alpha so that you can explore its capabilities. Your "feedback":https://docs.google.com/forms/d/e/1FAIpQLSd00n1uxgXWPGvMjKwMVL1UDhFKMeh3bSrP52j9AfXifoU-Pg/viewform will help prioritize improvements and fixes for subsequent releases. The features in this release have been validated for a set of use-cases and the alpha is implemented to work under real-world situations and load, but you should use caution when deciding under what conditions to use it.
```

Srushtika (changes requested):
Hopefully final change suggestions to the 'How LiveSync works' page. Other than the suggestions, I'm happy for this PR to be 'ready to merge'

Srushtika (approved):
Great work on this, @GregHolmes! This is good to go from my perspective (dependant on the website deployment of the integration rule) with any new improvements progressively added over time as separate PRs.

m-hulbert (approved):
🚀

mschristensen (approved):

PR #2179, opened 2024-04-22
chore: roundtrip latency copy update

jamiehenson:
## Description

Updates the copy for Ably in the roundtrip latency section of the current comparison pages - I wasn't sure which config file applied so I applied the change equally to both.

First commit does the change, the second one applies default `prettier` styling to the rest of the YAML files - as my IDE wanted to do this (keeping it as feasibly other IDEs with `prettier` format on save would want to as well)

## Review

Test `/compare/ably-vs-pubnub`, check the "Global median round trip latencies..." section has been updated.

### DIFF
--- a/data/compare.yaml
+++ b/data/compare.yaml
@@ -284,9 +284,9 @@ AblyCompare:
           pusher: "*No.*"
         extra: 'Encoding messages in binary format is faster as it reduces bandwidth for sending and receiving messages, and streamlines the processing time for clients and servers when encoding and decoding messages.<br/><br/>"Learn about our binary protocol":https://faqs.ably.com/do-you-binary-encode-your-messages-for-greater-efficiency'
       roundtrip:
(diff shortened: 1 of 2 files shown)


kennethkalmer (changes requested):
@jamiehenson thank you! we'll need to merge these changes into the `nanoc-toolchain` branch rather than `main`, and the on the website we create another PR to update the embedded docs version. `./scripts/bump-docs.sh` is your friend in the website

jamiehenson:
@kennethkalmer I did wonder if there would be more to it than "update in docs, it appears on the main site". Though when I change the base to `nanoc-toolchain` - there's 250 commits and loads of merge conflicts 😬

jamiehenson:
I'm guessing I just rebase this off of `nanoc-toolchain` as it seems it and `main` aren't synced

PR #2189, opened 2024-05-21
EDU-926: Adds flutter code examples

franrob-projects:
This PR:

- Adds .NET code examples to the required areas of the recently updated Push notification docs.

[EDU-926: Add Flutter to push notifications](https://ably.atlassian.net/browse/EDU-926)

### DIFF
--- a/content/push/device.textile
+++ b/content/push/device.textile
@@ -6,6 +6,7 @@ languages:
   - android
   - swift
   - objc
+  - flutter
 redirect_from:
   - /realtime/push/activate-subscribe
   - /general/versions/v1.1/push/activate-subscribe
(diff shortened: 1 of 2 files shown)


content/push/device.textile
```diff
@@ -40,6 +49,14 @@ h3. Configure an iOS device
 * Go to *push notifications setup*, click *configure push*.
 * Add your *.p12* file.
 
+h4. Add Ably Flutter to your iOS project
```
m-hulbert, line 52: If there isn't a specific push package required then this is a step that we aren't doing for Swift / Kotlin already and is just generic Flutter SDK setup. Do you think this should be added for all, or excluded?

m-hulbert (commented):
I'd have assumed we need a Flutter code sample in every place we already have Swift + Android. Are there not specific steps required in 'integrating Ably into your app' for example?

PR #2197, opened 2024-05-28
[Web-3426] CSS update font weights

aleksandar-r:
## Description

Updating font weights based on comments in ticket below.

* [Ticket](https://ably.atlassian.net/browse/WEB-3426).

## Review

Check Sidebar and card fonts mostly.

### DIFF
--- a/src/components/ProductPage/BodySection/Card/ExampleCard.tsx
+++ b/src/components/ProductPage/BodySection/Card/ExampleCard.tsx
@@ -9,7 +9,7 @@ export const ExampleCard = ({ title, image, link, external }: CardProps) => (
     className="items-center border border-extra-light-grey rounded-lg bg-extra-light-grey flex flex-col h-full hover:border-mid-grey hover:text-cool-black group"
   >
     <Image image={image} style={{ pointerEvents: 'none' }} />
-    <p className="pb-24 ml-8" style={{ fontSize: '1rem' }}>
+    <p className="pb-24 ml-8 font-medium" style={{ fontSize: '1rem' }}>
       {title}
     </p>
   </Link>
--- a/src/components/ProductPage/BodySection/Card/ExampleCard.tsx
+++ b/src/components/ProductPage/BodySection/Card/ExampleCard.tsx
@@ -9,8 +9,6 @@ export const ExampleCard = ({ title, image, link, external }: CardProps) => (
     className="items-center border border-extra-light-grey rounded-lg bg-extra-light-grey flex flex-col h-full hover:border-mid-grey hover:text-cool-black group"
   >
     <Image image={image} style={{ pointerEvents: 'none' }} />
-    <p className="pb-24 ml-8 font-medium" style={{ fontSize: '1rem' }}>
-      {title}
-    </p>
+    <p className="pb-24 ml-8 font-medium text-base">{title}</p>
   </Link>
 );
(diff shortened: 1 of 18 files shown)


src/components/ProductPage/BodySection/Card/ExampleCard.tsx
```diff
@@ -9,7 +9,7 @@ export const ExampleCard = ({ title, image, link, external }: CardProps) => (
     className="items-center border border-extra-light-grey rounded-lg bg-extra-light-grey flex flex-col h-full hover:border-mid-grey hover:text-cool-black group"
   >
     <Image image={image} style={{ pointerEvents: 'none' }} />
-    <p className="pb-24 ml-8" style={{ fontSize: '1rem' }}>
+    <p className="pb-24 ml-8 font-medium" style={{ fontSize: '1rem' }}>
```
jamiehenson, line 12: Not your change - but not sure why the font size would need to be an inline style here, `text-base` should do it

src/components/ProductPage/BodySection/Card/FeatureCard.tsx
```diff
@@ -7,8 +7,10 @@ export const FeatureCard = ({ title, content, image, links }: CardProps) => (
     <div className="py-24 pl-24">
       <div className="flex flex-col h-full">
         <div>
-          <h2 style={{ fontSize: '1.5rem' }}>{title}</h2>
-          <p style={{ fontSize: '1rem' }} className="text-lg text-dark-grey mt-8 mb-40 leading-6">
+          <h2 style={{ fontSize: '1.5rem' }} className="font-extrabold">
+            {title}
+          </h2>
+          <p style={{ fontSize: '1rem' }} className="text-lg text-dark-grey mt-8 mb-40 leading-6 font-medium">
```
jamiehenson, line 13: Same here, `text-base` and `text-2xl` should work here

src/components/Sidebar/Sidebar.tsx
```diff
@@ -68,7 +68,7 @@ export const Sidebar = ({
       )}
       {data.map(({ label, content }) => (
         <div key={label} className={cn('px-24 mb-32 transition-all', { 'opacity-0': collapsed })}>
-          {label && <SectionTitle className="mb-8">{label}</SectionTitle>}
+          {label && <SectionTitle className="mb-8 !font-bold">{label}</SectionTitle>}
```
jamiehenson, line 71: Do we want this to happen across all `SectionTitle`s? It's also used on the "RightSidebar", seems like both _should_ have the same font weight, but stranger design things happen. If so it would probably be better to make the change [here](https://github.com/ably/docs/blob/main/src/components/Sidebar/SectionTitle.module.css#L2).

jamiehenson (approved):
Looks good, have some questions but they're not blockers

src/components/Sidebar/Sidebar.tsx
aleksandar-r: They did show it on right sidebar in the ticket, I am waiting on designs confirmation still, so we will see.

src/components/ProductPage/BodySection/Card/ExampleCard.tsx
aleksandar-r: It left me confused also, will try it with text base :smile:

src/components/Sidebar/Sidebar.tsx
aleksandar-r: I did do it on on module also, let me see why that did not work.

PR #2198, opened 2024-05-29
[WEB-3834] Fix the heroku buildpack nginx script to a commit

denissellu:
This pull request updates the `install-nginx` build script to fetch the `build_nginx` script from a specific commit, this will make sure that we use a known and stable version of the `build_nginx` script, and not be at the mercy of any update


kennethkalmer (approved):
Nice catch!

PR #2213, opened 2024-06-24
[WEB-3571] chore: remove old compare textile assets

jamiehenson:

## Description

Currently, certain competitor pair combinations do not resolve in showing a new compare page, but an older static version based on Textile assets in `docs`. This PR removes those templates so that the correct asset resolution can occur.

* [Page to review](link)


kennethkalmer (approved):
🧹🧹🧹 ✨✨✨

PR #2225, opened 2024-07-11
[CHAT-398] currentlyTyping is a set not an array

m-hulbert:
## Description

This PR updates `currentlyTyping` to correctly be a set, rather than an array.

### DIFF
--- a/content/chat/rooms/typing.textile
+++ b/content/chat/rooms/typing.textile
@@ -28,19 +28,19 @@ The following is the structure of a typing event:
 
 ```[json]
 {
-  "currentlyTyping": [
+  "currentlyTyping": {
     "clemons",
     "zoranges",
-  ],
+  },
 }
 ```
 
 The following are the properties of a typing event:
 
 |_. Property |_. Description |_. Type |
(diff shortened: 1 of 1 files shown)


content/chat/rooms/typing.textile
```diff
@@ -28,19 +28,19 @@ The following is the structure of a typing event:
...
 ```
 
 The following are the properties of a typing event:
 
 |_. Property |_. Description |_. Type |
-| currentlyTyping | An array of all users currently typing. | Array |
+| currentlyTyping | A set of all users currently typing. | Set |
 
-You can use the length of the @currentlyTyping@ array to decide whether to display individual user names, or that multiple people are typing in your user interface.
+You can use the length of the @currentlyTyping@ set to decide whether to display individual user names, or that multiple people are typing in your user interface.
```
vladvelici, line 43: size of the set. Annoyingly, sets have sizes and arrays have lengths.

m-hulbert: Ah thanks. Fixed!

PR #2227, opened 2024-07-12
EDU-1568: Adds REST coverage to push

franrob-projects:
This PR: 

- Adds REST coverage to the push docs.

[EDU-1568](https://ably.atlassian.net/browse/EDU-1568)

[EDU-1568]: https://ably.atlassian.net/browse/EDU-1568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/push/configure/device.textile
+++ b/content/push/configure/device.textile
@@ -123,7 +123,6 @@ dotnet add package ably.io.push.ios
 // Then, add the SDK to your .cs file:
 
 using IO.Ably;
-using IO.Ably.Realtime;
 ```
 
 ```[flutter]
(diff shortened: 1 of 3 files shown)


m-hulbert (approved):
Looks good to me, but might be worth getting an SDK dev to check over the push code samples 🙂

content/push/configure/device.textile
```diff
@@ -180,41 +162,37 @@ The following diagram demonstrates the direct activation process:
...
 ActivatePush = new Command(() => Ably.Push.Activate());
 ```
 
-```[flutter]
+```[realtime_flutter]
 AblyRealtime ably = getAblyRealtime();
 ably.push.activate();
 ```
 
-```[javascript]
```
owenpearson, line 214: if we're removing javascript examples from this page entirely then we should also remove the javascript row in the table under "Device activation lifecycle"

owenpearson (commented):
looks good, one suggestion below. also - i think this was raised before, but we should really have something on the 'device.textile' page mentioning that web push exists, it would be quite easy to look at that page and just assume we don't support push activation in javascript.

also, unrelated to this PR but i do wonder a bit about having 'REST' and 'Realtime' toggles on the code examples. for the push publishing examples, 'realtime' appears to be selected by default so if you're looking for a PHP example it might not be obvious that we even have one, you'd have to know to click on 'REST' to find it

PR #2231, opened 2024-07-15
Auth docs: tweak js code sample

SimonWoolf:
- semicolons - no ASI in sample javascript code please!
- calling the callback inside the try/catch is an antipattern

page to review: token auth

really I would like an authcallback that returns a promise, but until that's implemented, the code snippet for doing this as-is was annoying me


m-hulbert (approved):
Thanks Simon!

PR #2247, opened 2024-08-22
[WEB-3922] - Move the ControlAPI from website to docs

aralovelace:
## Description
- Use Redoc to display the YAML content and create a reusable component for future use. Just use an HTML element by referencing the Redoc script using a link to the files hosted on a CDN.
- copy the styles and other elements  from the previous page to the new Control API page
- Added a Go to top button and make it as a separate component
- remove the `control API` content file as it won't be used anymore in the `website` 

Note for Future: In the Modernise UI (WEB-3426) this task adds the new Control API page needs to be updated and use the new font styling 

* WEB-3922

## Review

Preview - https://ably-docs-web-3922-gpfi4yshwf6.herokuapp.com/docs/api/control-api
Compare it to the current page - https://ably.com/docs/api/control-api


kennethkalmer (commented):
Could we call the `DisplayRedoc` component `Loader`, or `Render, or event just `Display`? Then in `index.ts` we can re-export it as `Redoc` so it when we then render it we simply call `<Redoc specUrl='' />`?

I also don't think we need an environment variable for the spec URL, it can just be provided by the caller. Just thinking ahead if we ever need to render other specs, then naming ENV vars becomes an issue.

m-hulbert:
> Could we call the `DisplayRedoc` component `Loader`, or `Render, or event just `Display`? Then in `index.ts`we can re-export it as`Redoc`so it when we then render it we simply call``?
> 
> I also don't think we need an environment variable for the spec URL, it can just be provided by the caller. Just thinking ahead if we ever need to render other specs, then naming ENV vars becomes an issue.

What are your thoughts on just moving the spec file into the docs repo, rather than calling out to open specs? @kennethkalmer @aralovelace

kennethkalmer:
> What are your thoughts on just moving the spec file into the docs repo, rather than calling out to open specs? @kennethkalmer @aralovelace

I don't have a strong opinion on this @m-hulbert, because I don't have all the insights about these various open repos and their purpose. The "well groomed" spec needs a home somewhere, and as long as it is accessible over HTTPS we can fetch and render it. 

If we want it in the repo we can simply drop it into `static/open-specs/control-v1.yml` and update the references.

aralovelace:
thanks @kennethkalmer  I have updated the components and remove the env variable
Also as suggested by @m-hulbert the file is now local;

m-hulbert (approved):
LGTM! 🚀

kennethkalmer (approved):
Nice one @aralovelace, thank you!

PR #2259, opened 2024-09-02
chat: point docs to umd Chat bundle

AndyTWF:
## Description

Related to https://github.com/ably/ably-chat-js/pull/333

Currently, only the UMD variant of Chat works with ably-js, so we need to point to that.

Also expands the example of how to use the CDN-served library.

## Review

See Chat Setup page


splindsay-92 (approved):
Looks good, does it need to be two separate commits for this though? Also, the PR body here has some leftover stuff in it. Apart from that, I'll approve :)

m-hulbert:
Yeah, can you squash the commits please @AndyTWF 🙂

AndyTWF:
Thanks both, squashed :)

PR #2264, opened 2024-09-09
Change python example for Channel options without Ably SDK support

cameron-michie:
Change from 'ably_rest' to 'realtime', to fit in with other examples, and reflect that it is a realtime client not a rest client in the example.


m-hulbert:
Addressed in https://github.com/ably/docs/pull/2276 - thanks @cameron-michie !

PR #2265, opened 2024-09-10
Update rest idempotent publishing doc

sacOO7:
## Description
Update channel idempotent publishing doc. Current wording used is a bit confusing, adding more clarity for the same.
Link to internal conversation -> https://ably-real-time.slack.com/archives/C012DA3MCB0/p1725522278686219

### DIFF
--- a/content/channels/index.textile
+++ b/content/channels/index.textile
@@ -755,11 +755,13 @@ h3(#idempotency). Idempotent publish
 
 Idempotency ensures that multiple publishes of the same message cannot result in duplicate messages.
 
-It is possible that a client publishing a message using the REST interface may not receive acknowledgement of receipt from Ably, due to issues such as network failure outside of Ably's control. Clients will automatically attempt to re-publish messages in these instances, which could result in duplicate messages.
+It is possible that a client publishing a message using the REST interface may not receive acknowledgement of receipt from Ably, due to issues such as network failure outside of Ably's control. Clients will internally attempt to re-publish messages in these instances.
 
-If idempotent publishing is enabled using the @idempotentRestPublishing@ "@ClientOptions@":/api/rest-sdk#client-options, the Ably SDK will assign a unique ID to each message which ensures that subsequent retry attempts cannot result in duplicate messages.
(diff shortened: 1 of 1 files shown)


content/channels/index.textile
```diff
@@ -753,13 +753,15 @@ channel.Publish(context.Background(), "action", "boom!")
 
 h3(#idempotency). Idempotent publish
 
-Idempotency ensures that multiple publishes of the same message cannot result in duplicate messages.
+By default idempotentRestPublishing is enabled in all latest ably SDKs. It ensures that publishing of a message doesn't result in duplicate messages.
```
mschristensen, line 756: I think we should be more specific than "in all latest Ably SDKs", can we point to a specific version, or if there are too many to list for different SDKs, we should keep this statement as-is.

content/channels/index.textile
```diff
@@ -753,13 +753,15 @@ channel.Publish(context.Background(), "action", "boom!")
...
+It is possible that a client publishing a message using the REST interface may not receive acknowledgement of receipt from Ably, due to issues such as network failure outside of Ably's control. Clients will internally attempt to re-publish messages in these instances.
 
-If idempotent publishing is enabled using the @idempotentRestPublishing@ "@ClientOptions@":/api/rest-sdk#client-options, the Ably SDK will assign a unique ID to each message which ensures that subsequent retry attempts cannot result in duplicate messages.
+When idempotent publishing is enabled, the Ably SDK will internally assign a unique ID to each message which ensures that subsequent retry attempts cannot result in duplicate messages. Idempotent publishing can be disabled using the @idempotentRestPublishing@ "@ClientOptions@":/api/rest-sdk#client-options.
 
-It is also possible to manually specify message IDs. The following is an example of how you might do this:
+Note that, ably server can only track duplicate message for 2 mins since last message with same id was published. So, message with same id published after 2 mins window will result in duplicate message at receiver end.
```
mschristensen, line 762:
```suggestion
Please note that Ably can only detect duplicate messages within a 2-minute window after the original message with the same ID is published. If a message with the same ID is published after this 2-minute window, it will be treated as a new message.
```

mschristensen (changes requested):

mschristensen (approved):
Small comment, once addressed this lgtm, thanks

m-hulbert:
@sacOO7 thanks for this! I've pushed a fixup to change some of the ordering, which also addresses @mschristensen 's comments. If you're happy with my changes then I'll rebase and merge.

mschristensen (approved):

PR #2266, opened 2024-09-10
[WEB-3920] Add Inkeep search widget

matt423:
## Description

If Inkeep is enabled show the Inkeep search widget instead of the existing search. 
___

_Notes_: 
* Current behaviour of adding the search query to the URL query parameter is not added, whilst awaiting for feedback from Inkeep
* No independent feature flag has been added as no there's no current business requirement to be able to turn on/off search independently from Inkeep chat.

* WEB-3920

## Review

* [Preview App](https://ably-docs-web-3920-add--75xwwf.herokuapp.com/docs/) can be used to interact with the Inkeep search widget
  * [Homepage](https://ably-docs-web-3920-add--75xwwf.herokuapp.com/docs/)
  * [Example page with search in header](https://ably-docs-web-3920-add--75xwwf.herokuapp.com/docs/products/channels) 
* Locally, set the `INKEEP_...` variables to enable Inkeep and the search widget

### DIFF
--- a/src/components/Header/SearchBar/InkeepSearchBar.tsx
+++ b/src/components/Header/SearchBar/InkeepSearchBar.tsx
@@ -1,0 +1,11 @@
+import { CSSProperties } from 'react';
+
+export const InkeepSearchBar = ({
+  className,
+  extraInputStyle,
+}: {
+  className: string;
+  extraInputStyle: CSSProperties;
+}) => {
+  return <div id="inkeep-search" className={className} style={extraInputStyle}></div>;
+};
--- a/src/components/Header/SearchBar/SearchBar.tsx
+++ b/src/components/Header/SearchBar/SearchBar.tsx
@@ -9,9 +9,10 @@ import { useSearch } from 'src/hooks';
 import { isMac } from 'src/utilities';
 
 import { SuggestionBox } from './SuggestionBox';
+import { InkeepSearchBar } from './InkeepSearchBar';
 import { KeyIcon } from './KeyIcon';
 
-import { searchInput } from './SearchBar.module.css';
+import { searchInput, searchInputNoPadding } from './SearchBar.module.css';
 
 export enum DisplayMode {
   FULL_SCREEN = 'FULL_SCREEN',
@@ -44,6 +45,7 @@ export const SearchBar = ({
         siteMetadata {
           externalScriptsData {
             addsearchSiteKey
+            inkeepEnabled
           }
         }
       }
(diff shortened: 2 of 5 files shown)


src/components/Header/SearchBar/SearchBar.tsx
```diff
@@ -103,30 +105,39 @@ export const SearchBar = ({
...
-        <Icon name="icon-gui-search" size="24px" additionalCSS="absolute left-16 top-12" />
-        {!isInFocus && (
-          <div className="absolute right-16 top-12 hidden lg:flex items-center justify-end">
-            <KeyIcon className="mr-4">{keyIcon}</KeyIcon>
-            <KeyIcon className="pt-2">K</KeyIcon>
-          </div>
+        {externalScriptsData.inkeepEnabled ? (
+          <InkeepSearchBar
+            className={`${cn(searchInput)} ${cn(searchInputNoPadding)}`}
+            extraInputStyle={{ ...extraInputStyle }}
```
jamiehenson, line 111: Why is `extraInputStyle` being spread within an object literal here? Seems like you could just pass it in without doing that

src/components/Header/SearchBar/SearchBar.tsx
```diff
@@ -103,30 +105,39 @@ export const SearchBar = ({
...
+        ) : (
+          <>
+            <input
+              type="text"
+              ref={textInput}
+              placeholder="Search"
+              className={cn(searchInput)}
+              value={query}
+              onChange={handleSearch}
+              style={{ ...extraInputStyle }}
```
jamiehenson, line 122: Same query as above

src/components/Header/SearchBar/SearchBar.tsx
```diff
@@ -103,30 +105,39 @@ export const SearchBar = ({
...
+            />
+            <Icon name="icon-gui-search" size="24px" additionalCSS="absolute left-16 top-12" />
+            {!isInFocus && (
```
jamiehenson, line 125: We should present these conditional renderings as ternaries with `null` on the other end, as passing `null` is more beneficial to the renderer than `false`.

src/components/Header/SearchBar/SearchBar.tsx
```diff
@@ -103,30 +105,39 @@ export const SearchBar = ({
...
-        />
-        <Icon name="icon-gui-search" size="24px" additionalCSS="absolute left-16 top-12" />
-        {!isInFocus && (
-          <div className="absolute right-16 top-12 hidden lg:flex items-center justify-end">
-            <KeyIcon className="mr-4">{keyIcon}</KeyIcon>
-            <KeyIcon className="pt-2">K</KeyIcon>
-          </div>
+        {externalScriptsData.inkeepEnabled ? (
+          <InkeepSearchBar
+            className={`${cn(searchInput)} ${cn(searchInputNoPadding)}`}
```
jamiehenson, line 110: Do we need more of this `cn` package? From what I can see it simply concatenates strings together, and the usages here only have one argument each so limited package usefulness aside, it wouldn't do anything

src/components/Header/SearchBar/SearchBar.tsx
```diff
@@ -103,30 +105,39 @@ export const SearchBar = ({
...
+                <KeyIcon className="mr-4">{keyIcon}</KeyIcon>
+                <KeyIcon className="pt-2">K</KeyIcon>
+              </div>
+            )}
+            <SuggestionBox
+              results={results}
+              error={error}
+              query={query}
+              isActive={isInFocus}
+              displayLocation={String(displayLocation)}
```
jamiehenson, line 136: We shouldn't cast a potential `undefined` as a `String` here to satisfy the existing types of `SuggestionBox`, it undermines the benefit of what TS is providing, changes should be made elsewhere (I appreciate this isn't your code)

src/components/Header/SearchBar/SearchBar.tsx
jamiehenson: I would advocate for removing all usages of this package where possible

jamiehenson (changes requested):

src/components/Header/SearchBar/SearchBar.tsx
matt423: It's existing logic, it's just been shifted down into the new ternary. Minimising blast radius by only changing what is necessary really

src/components/Header/SearchBar/SearchBar.tsx
matt423: It was mainly from just copying the existing syntax used and it did work, so didn't question it too much. Will update :+1:

src/components/Header/SearchBar/SearchBar.tsx
matt423: Was following convention within this file, have removed :+1:

src/components/Header/SearchBar/InkeepSearchBar.tsx
```diff
@@ -0,0 +1,3 @@
+export const InkeepSearchBar = ({ className, extraInputStyle }: { className: string; extraInputStyle: string }) => {
+  return <div id="inkeep-search" className={className} style={{ ...extraInputStyle }}></div>;
+};
```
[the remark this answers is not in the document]
matt423: Thanks, have changed to using `CSSProperties`. 

The original(albeit strange) implementation did work and the compiler did not throw any errors for the type. Iit does now if I pass incorrect type to `extraInputStyle` though which is good

src/components/Header/SearchBar/SearchBar.tsx
matt423: This is the existing behaviour, just shows in diff as moved within ternary condition.

src/components/Header/SearchBar/SearchBar.tsx
matt423: Existing behaviour. Happy to look into outside of this PR but keen to minimise blast radius of changes

jamiehenson (approved):

PR #2267, opened 2024-09-12
Pubnub: tweak message continuity wording

SimonWoolf:
- we no longer stay connected for 5 minutes, only 2
- we have never provided 5-20 minutes. The original docs said: "In Pubnub’s model, message continuity is available for ‘5-20 minutes’ and up to 100 messages". At some point this was changed to "PubNub message continuity is available for 5-20 minutes for up to 100 messages". The former version makes it clear that this is talking about pubnub itself. The latter sounds like it could be talking about the pubnub adapter.


splindsay-92 (approved):
LGTM

m-hulbert (approved):
Thanks Simon 👌

PR #2269, opened 2024-09-13
Integrations: Add note around webhook data envelope support

splindsay-92:

## Description

When creating a new Webhook integration rule, it wasn't obvious that only message and presence events can be sent as raw payload in webhooks. Trying to do the same with occupancy or lifecycle events would generate an error when trying to create the rule. I have added a small section to the docs to make this clearer, in the hopes we will avoid future confusion.

* [REA-2071](https://ably.atlassian.net/browse/REA-2071)

## Review

* [Page to review](https://ably.com/docs/general/webhooks?)

[REA-2071]: https://ably.atlassian.net/browse/REA-2071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/general/webhooks.textile
+++ b/content/general/webhooks.textile
@@ -143,6 +143,10 @@ When you configure a rule using "single requests":#batching, you are given the o
 
 However, if you don't need anything besides the payload of each message, or the endpoint expects a very restricted data structure, you may choose not to envelope messages and instead have only the message payload (@data@ element) published. This has the advantage of requiring one less parsing step, however decoding of the raw payload in the published message will be your responsibility.
 
+<aside data-type='note'>
+<p>When utilizing webhook-based integrations, such as Zapier or Google Functions, only data from the @channel.message@ and @channel.presence@ sources can be received as raw payloads. @channel.lifecycle@ and @channel.occupancy@ events will always be enveloped.</p>
+</aside>
+
 Check out examples of "enveloped":#envelope-examples and "non-enveloped":#no-envelope-examples examples down below.
 
 h3(#channel-filter). Channel filter


mclark-ably:
It is probably worth mentioning that when creating the rule on the integrations dashboard unselecting enveloped will therefore result in an error

I think this probably needs a review from deved team also

m-hulbert:
> It is probably worth mentioning that when creating the rule on the integrations dashboard unselecting enveloped will therefore result in an error
> 
> This should have a review from deved team

@mclark-ably I think a better UX would be if we disabled the functionality in the dashboard based on which source you select. I've pushed a few minor changes, if everyone is happy with those then I'll merge 🙂

mclark-ably:
> > It is probably worth mentioning that when creating the rule on the integrations dashboard unselecting enveloped will therefore result in an error
> > This should have a review from deved team
> 
> @mclark-ably I think a better UX would be if we disabled the functionality in the dashboard based on which source you select. I've pushed a few minor changes, if everyone is happy with those then I'll merge 🙂

Thanks @m-hulbert , I agree but I don't know what the complexities are around this and would need to be a web team task (see internal comment https://ably-real-time.slack.com/archives/C8SPU4589/p1725986843113469?thread_ts=1725965046.845339&cid=C8SPU4589)

m-hulbert:
> > > It is probably worth mentioning that when creating the rule on the integrations dashboard unselecting enveloped will therefore result in an error
> > > This should have a review from deved team
> > 
> > 
> > @mclark-ably I think a better UX would be if we disabled the functionality in the dashboard based on which source you select. I've pushed a few minor changes, if everyone is happy with those then I'll merge 🙂
> 
> Thanks @m-hulbert , I agree but I don't know what the complexities are around this and would need to be a web team task (see internal comment https://ably-real-time.slack.com/archives/C8SPU4589/p1725986843113469?thread_ts=1725965046.845339&cid=C8SPU4589)

I'm going to merge this and let's pick that up separately in Slack.

m-hulbert (approved):

PR #2270, opened 2024-09-13
EDU-1660: Adds CDN activation copy

franrob-projects:
This PR:

- Updates the Web Push Activation section:
  - Adds instructions for activating via CDN as an alternative to a package manager.
  - Refines and rephrases the existing instructions for direct activation of web push.
  
  [EDU-1660](https://ably.atlassian.net/browse/EDU-1660)

[EDU-1660]: https://ably.atlassian.net/browse/EDU-1660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/push/configure/web.textile
+++ b/content/push/configure/web.textile
@@ -83,11 +83,40 @@ The following diagram demonstrates the direct activation process:
   <img src="@content/diagrams/push-direct-activation-web-push.png" style="width: 100%" alt="push notifications in Ably">
 </a>
 
+h4(#package-man). Activate via package manager
+
+If you have installed the Ably SDK using a package manager (such as npm), you can use the @activate@ method to enable push notifications. You can then initialize the push plugin with the Ably client instance.
+
 The following example initializes an instance of the Ably realtime service and then activates the push notifications feature _for that instance_:
 
 ```[javascript]
-await ably.push.activate();
+await client.push.activate();
 ```
+
+h4(#no-package-man). Activate via Content Delivery Network (CDN)
+
+If you can't use a package manager, you can activate it via CDN by directly loading the push plugin in your HTML using a @<script>@ tag.
+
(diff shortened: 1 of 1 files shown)


content/push/configure/web.textile
```diff
@@ -83,12 +83,41 @@ The following diagram demonstrates the direct activation process:
...
+
+The following example is loading the Ably push notifications library from a CDN:
+
+```[html]
+<script src="https://cdn.ably.com/lib/push.umd.min-2.js"></script>
+```
+
+When loaded, the push plugin becomes available on the global object via the @AblyPushPlugin@ property. You can then initialize the push plugin with the Ably client instance.
+
+The following example initializes an instance of the Ably realtime service and then activates the push notifications feature _for that instance_:
```
mclark-ably, line 108: I'm not sure this line is entirely accurate as the example shows instantiating a REST client. 

I think generally most browser clients are going to be using realtime however so I'm not sure why the examples in the GitHub repo show using REST. Probably worth checking with the ecosystems team

VeskeR: it will work both for Rest and Realtime clients. we can change the example here to be:
```
const client = new Ably.Realtime({
  ...options,
  plugins: { Push: AblyPushPlugin },
});
```
so that the language above matches the example

franrob-projects: Thanks for the feedback guys! @mclark-ably please see fixup 9357e3a10697242dfafccd3a049521c64f97d8e7

mclark-ably (approved):

m-hulbert (approved):
Needs a rebase then good to go 👍

PR #2279, opened 2024-09-20
Change Chat SDK React usePresence hook to use enterWithData / leaveWithData instead of dataToEnterWith

cameron-michie:
Update the Chat SDK React hooks to include the correct usePresenceParams as outlined here

https://sdk.ably.com/builds/ably/ably-chat-js/main/typedoc/interfaces/chat_react.UsePresenceParams.html#enterWithData


m-hulbert (approved):
Thanks @cameron-michie !

PR #2305, opened 2024-11-08
History: document how to retrieve persist-last message with the rest api

SimonWoolf:
REA-1622

as well as persist-last retrieval, also fixed the section about exceptions to canonical global order, which was confused (enabling persistence doesn't change the order you get for messages from live history, which is always global unless you're using untilAttach. also you don't need to enable persistence to "prevent duplication and missing messages", none of our history apis return duplicate or missing messages given any settings. or if they ever do, that's a bug)

### DIFF
--- a/content/storage-history/history.textile
+++ b/content/storage-history/history.textile
@@ -793,7 +793,4 @@ It is important to note that this is not necessarily the order that messages wer
 
 Ably preserves ordering for a specific publisher on a specific channel but, for example, if two publishers in regions A and B publish _message-one_ and _message-two_ simultaneously, then it is very possible that a subscriber in region A will receive _message-one_ before _message-two_, but that a subscriber in region B will receive _message-two_ before _message-one_.
 
-There are instances where messages will not be in canonical global order:
-
-* Recent messages (less than two minutes old) are retrieved from live ephemeral storage and are still ordered by region. They only appear in the canonical global order if you enable "message persistence":/storage-history/storage, which also prevents duplication and missing messages.
-* You choose to retrieve historical messages only up to the point at which a client attaches to a channel. You would typically use this approach to bring a subscriber up to date as part of connection state recovery.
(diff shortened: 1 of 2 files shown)


content/storage-history/history.textile
```diff
@@ -792,7 +793,4 @@ It is important to note that this is not necessarily the order that messages wer
...
-
-* Recent messages (less than two minutes old) are retrieved from live ephemeral storage and are still ordered by region. They only appear in the canonical global order if you enable "message persistence":/storage-history/storage, which also prevents duplication and missing messages.
-* You choose to retrieve historical messages only up to the point at which a client attaches to a channel. You would typically use this approach to bring a subscriber up to date as part of connection state recovery.
+There are some instances where messages will not be in canonical global order. If you are using continuous history (that is, either a REST history call with @untilAttach=true@, or rewind on a realtime connection), then recent messages (less than two minutes old) are retrieved from live ephemeral storage and are still ordered by region, in order to correctly mesh with the live stream of messages in that region. Older messages retrieved from persisted history (if that is enabled) will be in canonical global order, and Ably will ensure that the transition between the two happens with no duplicates or missed messages.
```
m-hulbert, line 796:
```suggestion
There are some instances where messages will not be in canonical global order when using "continuous history":#continuous-history. Continuous history utilizes either a REST history call with the @untilAttach@ parameter set to @true@, or the "rewind":/channels/options/rewind feature on a realtime connection. Messages that are less than two minutes old are retrieved from live ephemeral storage and are ordered only by region. This is to ensure that they correctly mesh with the live stream of messages currently being published in that region. Messages older than two minutes are retrieved from "persisted history":/storage-history/storage#all-message-persistence (if it is enabled). These messages will be in canonical global order. Ably ensures that the transition between those messages retrieved from ephemeral storage, and those from persisted storage, occurs without duplication or missed messages.
```

Just a suggestion to re-word this one slightly, but let me know if I've misunderstood it.

m-hulbert (approved):
Minor rewording suggestion, but LGTM other than that 👍

PR #2306, opened 2024-11-08
History: reduce max normal retention period from 365 to 30 days

SimonWoolf:
We don't support a year of normal history, we've never supported a year of normal history. if you want a year, that's what persist-last is for

https://ably-real-time.slack.com/archives/C019XP5SYB0/p1731108835200039?thread_ts=1729587824.143959&cid=C019XP5SYB0


GregHolmes (approved):

mattheworiordan:
See https://ably-real-time.slack.com/archives/C019XP5SYB0/p1731326506586359?thread_ts=1729587824.143959&cid=C019XP5SYB0. I am afraid I don't agree, and if we don't let customers/prospect ask us to store data for longer, when we know it's a valid customer ask, then we're not giving ourselves the opportunity to capture interest in longer term storage and to understand the use cases that justify it.

PR #2319, opened 2024-11-29
[ECO-5105] docs: add ably-chat-kotlin docs

ttypic:
## Description

Ably Chat for Android docs


content/chat/setup.textile
```diff
@@ -111,13 +150,48 @@ const App = () => {
...
+    ClientOptions().apply {
+        key = "{{API_KEY}}"
+        clientId = "<clientId>"
+    },
+)
+
+val chatClient = ChatClient(realtimeClient)
+```
+
 A "@ClientOptions@":/api/realtime-sdk#client-options object may be passed to the Pub/Sub SDK instance to further customize the connection, however at a minimum you must set an API key and provide a @clientId@ to ensure that the client is "identified":/auth/identified-clients.
```
sacOO7, line 176: It will be great, if we provide more clarification related to `clientId` usage as per internal discussion -> https://ably-real-time.slack.com/archives/CURL4U2FP/p1733668250689229

sacOO7 (changes requested):

content/chat/setup.textile
ttypic: it's out of the scope of this PR

PR #2323, opened 2024-12-02
[EDU-1669] - Add PubSub example for channels and messages

GregHolmes:
This PR contains the two code samples for the new Examples section we are creating. It contains a Typescript and a React version for channels and messages with Pub/Sub. Which will be rendered in the code sandbox live examples in the future, along with others.

The README within each directory /examples/pub-sub-channel-messages/javascript and /examples/pub-sub-channel-messages/react contains information on how to get these running.

[Jira epic ticket](https://ably.atlassian.net/browse/EDU-1541)
[Jira ticket](https://ably.atlassian.net/browse/EDU-1669).


m-hulbert (changes requested):
Example is looking good, but a few things:

* You're adding all the examples in this PR.
* The `page.md` hasn't been updated and is still referencing Chat.
* After the 10 headlines, we publish `undefined` when you click the button which looks bad. We'd be better off disabling the publish button. 
* Do you think it's worth having an input box for users to enter their own headline too?

m-hulbert (commented):
Also forgot to suggest that we reverse the order of the feed (so newest is at the top), as I would say that's more common in a news feed. WDYT?

GregHolmes:
@m-hulbert  I've updated based on your feedback.
We discussed and agreed on another PR to have custom status updates. But I feel in this instance having custom news headlines wouldn't display any extra to this feature than what's already there.

examples/pub-sub-channel-messages/javascript/src/script.ts
```diff
@@ -0,0 +1,66 @@
...
+  const headlinesRemaining = document.getElementById('headlines-remaining');
+  headlinesRemaining.innerText = `Headlines remaining: ${headlines.length}`;
+}
+
+getHeadlinesRemaining();
+
+const channel = client.channels.get('cut-nil-tie');
+
+channel.subscribe((message) => {
+  const messagesDiv = document.getElementById('headlines');
```
umair-ably, line 33: can this be outside of the callback? e.g. on line 31 instead

umair-ably (approved):
LGTM minus the nitpick... it's still unclear to me which code snippets (if any) will be exposed on the page itself but from a cursory glance, this looks good

ttypic (approved):
Looks good to me

examples/pub-sub-channel-messages/page.md
```diff
@@ -0,0 +1,118 @@
+# Sending and receiving messages with Pub/Sub
+
+Enable clients to send and receive messages in a channel.
+
+Messaging enables instant, scalable communication through publishing and subscribing to channels. Messages represent dynamic data payloads that clients publish and subscribe to, powering collaborative applications across realtime newsfeeds, auction applications or financial websites.
+
+Messaging is implemented using [Ably Pub/Sub](https://ably.com/docs/auth). The Pub/Sub SDK provides a set of flexible APIs capable of building any realtime application and is powered by Ably's reliable and scalable platform.
```
m-hulbert, line 7: This is linking to auth 🙂

examples/pub-sub-channel-messages/page.md
```diff
@@ -0,0 +1,118 @@
+# Sending and receiving messages with Pub/Sub
+
+Enable clients to send and receive messages in a channel.
+
+Messaging enables instant, scalable communication through publishing and subscribing to channels. Messages represent dynamic data payloads that clients publish and subscribe to, powering collaborative applications across realtime newsfeeds, auction applications or financial websites.
```
m-hulbert, line 5:
```suggestion
Messaging enables instant, scalable communication through publishing and subscribing to channels. Messages represent data payloads that clients publish and subscribe to. They are the basis of realtime communication in all applications.
```

m-hulbert: I think it would be worth having a paragraph about channels after this. Maybe something along the lines of:

> Channels are used to separate messages into different topics. They are the building block of creating a realtime application using the publish-subscribe pattern. Channels are also the unit of security and scalability. Whilst billions of messages may be delivered by Ably, clients receive only the messages on the channels they subscribe to.

m-hulbert (changes requested):
A few minor copy suggestions. Example looks good as-is!

PR #2325, opened 2024-12-03
[EDU-1655] - Pub/Sub Presence example

GregHolmes:
This PR contains the two code samples for the new Examples section we are creating. It contains a Typescript and a React version for presence with Pub/Sub. Which will be rendered in the code sandbox live examples in the future, along with others.

The README within each directory /examples/pub-sub-presence/javascript and /examples/pub-sub-presence/react contains information on how to get these running.

[Jira epic ticket](https://ably.atlassian.net/browse/EDU-1541)
[Jira ticket](https://ably.atlassian.net/browse/EDU-1655).


NickLittman:
In the React example, the Presence set isn't hydrating on page load. Any user who joins will only see themselves and anyone who joins after them, not anyone who has joined before them.

The `usePresenceListener` hook exposes the `presenceData` [state object](https://ably.com/docs/getting-started/react#usePresenceListener) which you could use instead of storing the viewer info in your own `viewers` state.

examples/pub-sub-presence/page.md
```diff
@@ -0,0 +1,121 @@
+# Presence updates with Pub/Sub
+
+Presence enables users to see who else is online within an application.
+
+Use presence to display which users are currently online in a channel. Users can also manually update their status, such as to set themselves as 'Away', or write a custom message.
+
+Presence enables users to quickly understand the availability of others to communicate more effectively. For example, on a live video stream users can see who else is currently viewing.
+
+Presence is implemented using [Ably Pub/Sub](https://ably.com/docs/products/channels). The Ably SDK provides features such as presence, publishing and subscribing to messages within channels in a high performance and scaling capability.
```
m-hulbert, line 9: Suggest we replace the 2nd sentence with the "agreed" upon one for Pub/Sub 🙂

examples/pub-sub-presence/page.md
```diff
@@ -0,0 +1,121 @@
+# Presence updates with Pub/Sub
+
+Presence enables users to see who else is online within an application.
+
+Use presence to display which users are currently online in a channel. Users can also manually update their status, such as to set themselves as 'Away', or write a custom message.
```
m-hulbert, line 5:
```suggestion
Use presence to display which users are currently 'present' on a channel. Users can also manually update their status, such as to set themselves as 'Away', or write a custom message.
```

m-hulbert (commented):
2 minor wording suggestions. 

Also still think a manual status would be useful in order to show off `presence.update()` here (and then add that to the page.md)

GregHolmes:
> In the React example, the Presence set isn't hydrating on page load. Any user who joins will only see themselves and anyone who joins after them, not anyone who has joined before them.
> 
> The `usePresenceListener` hook exposes the `presenceData` [state object](https://ably.com/docs/getting-started/react#usePresenceListener) which you could use instead of storing the viewer info in your own `viewers` state.

 Good spot. Thank you. I think this was a problem caused by a refactor and I never added it back in.

examples/pub-sub-presence/javascript/src/script.ts
```diff
@@ -0,0 +1,71 @@
...
+  clientId: faker.person.firstName(),
+  key: import.meta.env.VITE_PUBLIC_ABLY_KEY as string,
+});
+const channel = client.channels.get('viewer-presence');
+
+async function initializePresence() {
+  await channel.presence.subscribe((member) => {
+    const { action, clientId, data } = member;
+
+    switch (action) {
```
ttypic, line 15: I believe we don't recommend this, it can lead to some issues, it's better to refetch presence set here (we are going to provide presence set as one of listener's argument later to make it more useful). Action can be used to show toasters or other notification, not for rebuilding the set (because getting the initial set is not atomic)

examples/pub-sub-presence/react/src/app/page.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+export default function Home() {
+  const [viewers, setViewers] = useState<Viewer[]>([]);
+
+  const ably = useAbly();
+  const currentClientId = ably?.auth.clientId;
+
+  const { presenceData } = usePresenceListener('viewer-presence', (presence) => {
+    switch (presence.action) {
```
ttypic, line 18: no need to do this, `presenceData` is always represent actual presence set, so it can be replaced with:

```ts
const { presenceData } = usePresenceListener('viewer-presence')
const viewers: Viewer[] = useMemo(() => presenceData.map(member => ({
      clientId: member.clientId,
      isOnline: true
},  presenceData);

```

ttypic (changes requested):

ttypic (approved):
Looks good to me!

m-hulbert (approved):
Good to go with a rebase 🙂

PR #2332, opened 2024-12-10
Chat: Clarify the message listener input.

splindsay-92:

## Description

In chat, the object passed to a listener is called the MessageEventPayload, this contains both a type (describes the event, message.created..) and a message (contains the actual chat message).
We should be clear here in the docs too.

* [messages](http://localhost:8000/chat/rooms/messages)


AndyTWF (approved):

m-hulbert:
I think we perhaps need some context around this, e.g. when would I use `event.type` over `message.action`. 

If we believe that's an important distinction to make, then I think we should also update the example message payload to show both, or add it as a separate example relating to the first question (i.e. when would I use it).

splindsay-92:
> I think we perhaps need some context around this, e.g. when would I use `event.type` over `message.action`.
> 
> If we believe that's an important distinction to make, then I think we should also update the example message payload to show both, or add it as a separate example relating to the first question (i.e. when would I use it).

@m-hulbert So, event.type should be used for listeners as listeners receive event payloads.

m-hulbert (approved):

PR #2356, opened 2024-12-20
Added examples directory to be ignored in tests

GregHolmes:
Tests is having issues due to there being multiple instances of React (examples has many).


jamiehenson (approved):
📈

PR #2368, opened 2025-01-06
[EDU-1700] Add Chat to SDK tabs

m-hulbert:
## Description

This PR adds a tab for Chat to the list of SDKs. It also fixes an issue with the font weight for active tabs.

### DIFF
--- a/src/components/SDKsPage/MainSection/index.tsx
+++ b/src/components/SDKsPage/MainSection/index.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { container, active_tab } from '../sdks.module.css';
+import { container } from '../sdks.module.css';
 import CardGrid from '../Card/CardGrid';
 import { data } from '../data';
 import Link from 'src/components/Link';
@@ -7,10 +7,11 @@ import Link from 'src/components/Link';
 export enum Tab {
   CHANNELS = 'channels',
   SPACES = 'spaces',
+  CHAT = 'chat',
 }
 
 const MainSection = ({ tab }: { tab: Tab }) => {
-  const activeTab = tab !== Tab.SPACES ? Tab.CHANNELS : Tab.SPACES;
+  const activeTab = Object.values(Tab).includes(tab as Tab) ? tab : Tab.CHANNELS;
 
   return (
     <div>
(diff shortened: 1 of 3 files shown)


GregHolmes (approved):

src/components/SDKsPage/MainSection/index.tsx
```diff
@@ -7,27 +7,34 @@ import Link from 'src/components/Link';
 export enum Tab {
   CHANNELS = 'channels',
   SPACES = 'spaces',
+  CHAT = 'chat',
 }
 
 const MainSection = ({ tab }: { tab: Tab }) => {
-  const activeTab = tab !== Tab.SPACES ? Tab.CHANNELS : Tab.SPACES;
+  const activeTab = tab === Tab.CHAT ? Tab.CHAT : tab === Tab.SPACES ? Tab.SPACES : Tab.CHANNELS;
```
jamiehenson, line 14: If the goal here is to "set activeTab to tab if it's in the enum, otherwise channels" then we can simplify this nested ternary statement by examining the enum as a whole i.e.

```ts
const activeTab = Object.values(Tab).includes(tab as Tab) ? tab : Tab.CHANNELS
```

src/components/SDKsPage/MainSection/index.tsx
```diff
@@ -7,27 +7,34 @@ import Link from 'src/components/Link';
...
             to="/sdks"
-            className={`ui-text-h3 mr-16 px-8 py-16 inline-block ${activeTab === Tab.CHANNELS ? active_tab : null}`}
+            className={`text-h3 mr-16 px-8 py-16 inline-block ${activeTab === Tab.CHANNELS ? active_tab : null}`}
           >
             Pub/Sub
           </Link>
           <Link
             to="/sdks?tab=spaces"
-            className={`text-h3 font-normal px-8 py-16 inline-block ${activeTab === Tab.SPACES ? active_tab : null}`}
+            className={`text-h3 mr-16 px-8 py-16 inline-block ${activeTab === Tab.SPACES ? active_tab : null}`}
```
jamiehenson, line 28: We can retire the usage of this CSS module here and stick with Tailwind which integrates in with our design system a little better. In Tailwindish:

- `font-weight: 800` = `font-extrabold`
- `color: #ff5416` = `text-orange-600`
- I have no clue what role `relative` would play here, it can probably be removed, but for parity that would just be `relative`

Ideally we can then also delete the `.active_tab` class.

Further, not needed here but [take a look at the `cn` helper we use in `docs`](https://github.com/ably/docs/blob/main/src/components/Layout/Layout.tsx#L96-L98) to make constructing class strings a bit more streamlined instead of interpolating.

jamiehenson (commented):
Additions look good, but a couple of recommendations for how to tidy up a little around these changes

src/components/SDKsPage/MainSection/index.tsx
m-hulbert: Thanks for the tips and the info - looks a lot cleaner 🙂

jamiehenson: I see that `active_tab` is still being imported and referenced but it's been deleted from the imported CSS module file, I'd imagine this would cause problems - I'd say you'd need to go a bit further and add in the other classes listed above (unless they are no longer needed) and then remove the import to the now-removed `active_tab` class

jamiehenson: i.e. something like
```ts
className={`ui-text-h3 mr-16 px-8 py-16 inline-block ${activeTab === Tab.SPACES ? 'relative font-extrabold text-orange-600' : null}`}
```

m-hulbert: Oops, you're quite right. Shouldn't have been multi-tasking... Tidied this up now.

jamiehenson: Apologies, I didn't see the `::after` part of the now-departed CSS module (which forms the underline under the selected tab) - we would lose the underline without porting that as well. To save more PR tennis I've popped this in a followup commit here. One could argue this isn't a save now at all, but hey, at least it brings the styling approach more towards "standard".

jamiehenson (approved):

PR #2386, opened 2025-01-17
chat: add update/delete comparison docs

AndyTWF:
## Description

Adds the update/delete comparison docs for Chat, as these were only present in the repo README.

## Review

Two sections on Chat/Rooms/Messages

### DIFF
--- a/content/chat/rooms/messages.textile
+++ b/content/chat/rooms/messages.textile
@@ -279,6 +279,10 @@ const MyComponent = () => {
 };
 ```
 
+On rare occasions updates might arrive from the server out of order (in terms of the global order in which these events occur). An example of this is when messages are transiting between regions, and latencies between nearby regions is lower than those further away. To deterministically determine whether an update should supersede the current version, simply compare the @version@ strings using the standard string comparison operators. Alternatively, the @Message@ interface provides convenience methods to compare two instances of the same base message to determine which version is newer: @versionBefore()@, @versionAfter()@, and @versionEqual()@.
+
+The same out-of-order situation can happen between updates received over realtime and HTTP responses (e.g. when updating a message). In the situation where two concurrent updates happen, both might be received via realtime before the HTTP response of the first one arrives. Always compare the message @version@ to determine which instance of a @Message@ is newer.
+
 h3(#update-structure). Message update structure
 
 The following is the structure of an updated message:
@@ -402,6 +406,12 @@ const MyComponent = () => {
   return <div>...</div>;
 };
 ```
+
+On rare occasions, deletes and updates might arrive over realtime out of order. Again, should two concurrent actions happen in disparate regions, you will likely receive the action processed in the region closest to you first, for example.
+
+When the second action arrives, you will need to determine the order of these actions; this is done deterministically by comparing the @version@ field of the messages, using the standard string comparison operators. For convenience, the @Message@ interface provides methods to compare two instances of the same base message to determine which action is newer: @versionBefore()@, @versionAfter()@, and @versionEqual()@.
+
+The same out-of-order situation can happen between deletions received over realtime and HTTP responses, whereby both might be received via realtime before the HTTP response of the first one arrives.
 
 h3(#deletion-structure). Message deletion structure
 
--- a/content/chat/rooms/messages.textile
+++ b/content/chat/rooms/messages.textile
@@ -94,13 +94,7 @@ The following are the properties of a message:
 | | description: Optional description for the action. | String or undefined |
 | | metadata: Optional additional metadata about the action. | Object or undefined |
 
-<aside data-type="note">
-<p>Both @serial@ and @version@ are lexicographically sortable strings. This means they can be used to enforce a deterministic global ordering based on string comparison.</p>
- <p><strong>Global Ordering Rules:</strong></p>
-  <p>For message @serial@: If the @serial@ of one message occurs before another when lexicographically sorted, the first message is considered to have occurred globally before the other. If the @serial@ values are identical, the two messages are considered to be the same message.</p>
-  <p>For @version@: When two messages share the same @serial@, their respective @version@ values can be used to determine the order of actions applied to the message.</p>
-  <p>An action (such as an update) with a @version@ is considered to have occurred before some other action, if the first action's @version@ occurs before another when sorted lexicographically. If the @version@ values are identical, the actions are considered to be the same.</p>
-</aside>
+See "below":#global-ordering for more information on how to apply deterministic global ordering to the chat messages in your application.
 
 h3(#unsubscribe). Unsubscribe from messages
 
@@ -217,7 +211,7 @@ const MyComponent = () => {
   const handleMessageUpdate = (msg: Message) => {
     update(msg, { text: "my updated text" }, { description: "Message update by user" })
     .then((updatedMsg: Message) => {
-    console.log('Message updated:', updatedMsg);
+      console.log('Message updated:', updatedMsg);
     })
     .catch((error) => {
       console.error('Error updating message: ', error);
@@ -247,6 +241,12 @@ const {unsubscribe} = room.messages.subscribe((event) => {
       console.log('Received message: ', event.message);
       break;
     case MessageEvents.Updated:
+      const existing = myMessageList.find(event.message);
+      if (existing && event.message.versionBefore(existing)) {
+        // We've already received a more recent update, so this one can be discarded.
+        return;
+      }
+
       console.log('Message updated: ', event.message);
       break;
     default:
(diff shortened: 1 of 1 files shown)


content/chat/rooms/messages.textile
```diff
@@ -279,6 +279,10 @@ const MyComponent = () => {
 };
 ```
 
+On rare occasions updates might arrive from the server out of order (in terms of the global order in which these events occur). An example of this is when messages are transiting between regions, and latencies between nearby regions is lower than those further away. To deterministically determine whether an update should supersede the current version, simply compare the `version` strings using the standard string comparison operators. Alternatively, the `Message` interface provides convenience methods to compare two instances of the same base message to determine which version is newer: `versionBefore()`, `versionAfter()`, and `versionEqual()`.
```
splindsay-92, line 282: Only question here is, some places I think we refer to it as origin message (maybe that's just me though) and others we call it a base message. I don't mind either way, but lets agree to one and go with that?
@vladvelici @AndyTWF

splindsay-92 (approved):
Just one minor comment, but it shouldn't block :)

content/chat/rooms/messages.textile
AndyTWF: We don't refer to it as the "origin message" anywhere in the website docs - so happy to stick with base for now

splindsay-92: Crazy idea, what about root message? 👀

m-hulbert:
I was in the middle of suggesting some minor copy changes, but I notice that we have a section about global ordering under subscribe already. Is it worth pulling that out of the note and have a section on the page that is dedicated to explaining how ordering works for all operations?

content/chat/rooms/messages.textile
vladvelici: I think we should define terms very clearly. What's a version, what's a message.

- A chat message is uniquely identified by `serial`.
- "Two instances of the same _base_ message" == "Two instances of the same message"
- The version of a chat message is uniquely identified by `version`.

It's fine to use _base message_, I just think (not now but at some point) we should go through the docs and make sure we don't need to actually say that.

vladvelici: Should we make sure we always say **_chat message_** when we refer to chat messages as opposed to ably pubsub messages? That might make things like _base message_ not needed?

AndyTWF:
> I was in the middle of suggesting some minor copy changes, but I notice that we have a section about global ordering under subscribe already. Is it worth pulling that out of the note and have a section on the page that is dedicated to explaining how ordering works for all operations, rather than repeating it in all 3 sections?

Yes I think that's a good idea, will do it now!

AndyTWF:
@splindsay-92 @m-hulbert @vladvelici I've done a little tweak to create a dedicated "how to do ordering" section, then just updated examples and commentary to point at that section. I think this looks a bit cleaner?

content/chat/rooms/messages.textile
```diff
@@ -434,3 +457,23 @@ The deleted message response is identical to the structure of a message, with th
...
+
+h2(#global-ordering). Ordering Chat Message Events
+
+Chat messages and update events are delivered in realtime to clients connected to a particular region in the order in which that region receives them. The order in which a given region receives these events may be different from the "global" order of events, i.e. the true time-based order in which events happened.
+
+Chat messages are uniquely identified by their @serial@ and may have multiple @versions@ as a result of edit and delete operations. Both @serial@ and @version@ are lexicographically sortable strings. This means they can be used to enforce a deterministic global ordering based on string comparison.
+
+h3(#ordering-new). Ordering New Messages
+
+If the @serial@ of one message occurs before another when lexicographically sorted, the first message is considered to have occurred globally before the other. If the @serial@ values are identical, the two messages are considered to be the same message.
```
vladvelici, line 469:
```suggestion
If the @serial@ of one message occurs before another when lexicographically sorted, the first message is considered to have occurred globally before the other. If the @serial@ values are identical, the two messages are, in fact, the same message.
```

They really won't be two different messages in any way, so they _are_ the same, not just _considered_ the same?

content/chat/rooms/messages.textile
```diff
@@ -434,3 +457,23 @@ The deleted message response is identical to the structure of a message, with th
...
+
+The @Message@ object also has convenience methods @before@, @after@ and @equal@ which provide the same comparison.
+
+h3(#ordering-update-delete). Ordering Updates and Deletes
+
+When two messages share the same @serial@, their respective @version@ values can be used to determine the order of actions applied to the message. An action (such as an update) with a @version@ is considered to have occurred before some other action, if the first action's @version@ occurs before another when sorted lexicographically. If the @version@ values are identical, the actions are considered to be the same.
```
vladvelici, line 475:
```suggestion
When two messages instances share the same @serial@, their respective @version@ values can be used to determine the order of actions applied to the message. An action (such as an update) with a @version@ is considered to have occurred before some other action, if the first action's @version@ occurs before another when sorted lexicographically. If the @version@ values are identical, the actions are the same.
```

vladvelici: Or even 
```suggestion
When two message instances share the same @serial@ they represent the same chat message. Their respective @version@ values can be used to determine the order of actions applied to the message. An action (such as an update) with a @version@ is considered to have occurred before some other action, if the first action's @version@ occurs before another when sorted lexicographically. If the @version@ values are identical, the actions are the same.
```

content/chat/rooms/messages.textile
AndyTWF: Done!

content/chat/rooms/messages.textile
AndyTWF: Applied the second!

vladvelici (approved):

content/chat/rooms/messages.textile
```diff
@@ -434,3 +457,23 @@ The deleted message response is identical to the structure of a message, with th
 | version | Set to the serial of the deletion action. |
 | timestamp | Set to the time the message was deleted. |
 | operation | Set to the details the actioning client provided in the request. |
+
+h2(#global-ordering). Ordering Chat Message Events
```
m-hulbert, line 461: Sentence case for headings please 🙂

content/chat/rooms/messages.textile
```diff
@@ -434,3 +457,23 @@ The deleted message response is identical to the structure of a message, with th
...
+
+Chat messages and update events are delivered in realtime to clients connected to a particular region in the order in which that region receives them. The order in which a given region receives these events may be different from the "global" order of events, i.e. the true time-based order in which events happened.
+
+Chat messages are uniquely identified by their @serial@ and may have multiple @versions@ as a result of edit and delete operations. Both @serial@ and @version@ are lexicographically sortable strings. This means they can be used to enforce a deterministic global ordering based on string comparison.
+
+h3(#ordering-new). Ordering New Messages
+
+If the @serial@ of one message occurs before another when lexicographically sorted, the first message is considered to have occurred globally before the other. If the @serial@ values are identical, the two messages are, in fact, the same message.
```
m-hulbert, line 469:
```suggestion
If the @serial@ of one message occurs before another when lexicographically sorted, the first message is considered to have occurred before the other. If the @serial@ values are identical, the messages are the same message.
```

content/chat/rooms/messages.textile
```diff
@@ -434,3 +457,23 @@ The deleted message response is identical to the structure of a message, with th
...
+
+The @Message@ object also has convenience methods @before@, @after@ and @equal@ which provide the same comparison.
```
m-hulbert, line 471: Worth linking out to the API refs for these?

content/chat/rooms/messages.textile
```diff
@@ -434,3 +457,23 @@ The deleted message response is identical to the structure of a message, with th
...
+
+h3(#ordering-update-delete). Ordering Updates and Deletes
+
+Applying an action to a message produces a new version, which is uniquely identified by the @version@ property. When two message instances share the same @serial@ they represent the same chat message, but they can represent different versions. Lexicographically sorting the two message instances by the @version@ property gives the global order of the message versions: the message instance with a greater @version@ is newer, the message instance with a lower @version@ is older, and if their @version@ is equal then they are the same version.
+
+The @Message@ object also has convenience methods @versionBefore@, @versionAfter@ and @versionEqual@ which provide the same comparison.
+
+Update and Delete events provide the full message payload, so may be used to replace the earlier version of the message wholesale.
```
m-hulbert, line 479:
```suggestion
Update and Delete events provide the full message payload, so may be used to replace the entire earlier version of the message.
```

m-hulbert (approved):
A few suggested changes, but nothing blocking. 

I also agree with @vladvelici that it's a bit confusing that we alternate between calling them chat messages and messages, which may lead you to think they're different things.

content/chat/rooms/messages.textile
AndyTWF: Done!

content/chat/rooms/messages.textile
AndyTWF: Done!

content/chat/rooms/messages.textile
AndyTWF: Done!

content/chat/rooms/messages.textile
AndyTWF: Done!

PR #2391, opened 2025-01-21
Control API docs for ingress/mongodb rule

zknill:
Update open api spec for creating and modifying rules to include mongodb change stream ingress rules.


mschristensen (approved):
LGTM

m-hulbert (commented):
Everything looks good and runs without any spectral errors. My only comment would be about naming. We've called these 'MongoDB database connector rules' in the docs, so I think we should keep this consistent rather than adding 'changestream' back into the mix.

m-hulbert (approved):

PR #2395, opened 2025-01-23
[WEB-4192] Fix broken call to action internal links

denissellu:
This PR includes updates to internal link handling and call-to-action (CTA) links in the `BodySection` components for both the homepage and product pages, now we make sure that we check to see if the links are internal or external links privided in the call to action.

This will fix broken links on the pubsub and asset tracking, and improves on this #2381

### DIFF
--- a/data/yaml/page-content/chat.yaml
+++ b/data/yaml/page-content/chat.yaml
@@ -17,7 +17,7 @@ sections:
     releaseStage: ''
     callToAction:
       text: View setup instructions
-      href: '/docs/chat/setup'
+      href: '/chat/setup'
       type: link
   - title: Demo
     description: Take a look at a livestream chat demo for an example of what you can build with Ably Chat.
--- a/src/components/Homepage/BodySection/BodySection.tsx
+++ b/src/components/Homepage/BodySection/BodySection.tsx
@@ -1,4 +1,5 @@
 import cn from '@ably/ui/core/utils/cn';
+import { withPrefix } from 'gatsby';
 import { ImageProps, getImageFromList } from 'src/components/Image';
 import { SectionProps } from '../HomepageContent';
 import { BodySectionDescription } from './BodySectionDescription';
@@ -6,6 +7,7 @@ import { HeroCard } from './Card/HeroCard';
 import { FeatureCard } from './Card/FeatureCard';
 import { SdkCard } from './Card/SdkCard';
 import FeatureLink from '@ably/ui/core/FeaturedLink';
+import { checkLinkIsInternal, normalizeLegacyDocsLink } from 'src/utilities/link-checks';
 
 const cardTypes = {
   hero: HeroCard,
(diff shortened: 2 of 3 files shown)


src/components/Homepage/BodySection/BodySection.tsx
```diff
@@ -39,6 +41,10 @@ export const BodySection = ({ section, images }: { section: SectionProps; images
   const singleColumn = columns == 1;
   const bottomMargin = sectionBottomMarginVariants[section.bottomMargin];
 
+  if (section.callToAction && checkLinkIsInternal(section.callToAction.href)) {
```
aralovelace, line 44: wow this is bizarre

src/components/ProductPage/BodySection/BodySection.tsx
```diff
@@ -63,11 +69,11 @@ export const BodySection = ({ section, images }: { section: SectionProps; images
           })}
         </div>
       )}
-      {section.callToAction ? (
+      {section.callToAction && (
```
aralovelace, line 72: curios why this is change? should it return null if doesnt exist? to prevents potential errors or rendering issues that could occur

aralovelace (commented):
just left one comment with the iteration bit

data/yaml/page-content/chat.yaml
```diff
@@ -17,7 +17,7 @@ sections:
     releaseStage: ''
     callToAction:
       text: View setup instructions
-      href: '/docs/chat/setup'
+      href: '/chat/setup'
```
jamiehenson, line 20: This is all gonna change in a min with all links formatted with `/docs/` in place (the ticket in my in progress queue) but if this fixes it in this moment then sure

src/components/Homepage/BodySection/BodySection.tsx
denissellu: we have to check like this, otherwise the `/docs` links won't work locally, and but will in production or they work locally, and break in production, this look up is used whenever the `<Link>` component is used

src/components/ProductPage/BodySection/BodySection.tsx
jamiehenson: I don't see a downside to this check as `FeatureLink` depends on a lot of stuff from `section.callToAction` - the component may be able to handle that being missing but given the purpose of the PR I'm guessing it doesn't.

As for returning null, I revisited the React docs on this. [The latest on conditional rendering](https://react.dev/learn/conditional-rendering#logical-and-operator-) doesn't seem to have a problem with using this versus a ternary that returns `null` in the false case (which is what I've recommended to do in the past) - equating returning `false` as equivalent to `null` in the JSX rendering tree. In the past, there was a performance/memory benefit to `null` over `false` - maybe there still is.

So basically we can use either.

jamiehenson (approved):

data/yaml/page-content/chat.yaml
denissellu: yeah, this is how it should have been done in the first place!

i'm curious, how will it work going forward, just add `/docs` to all the links? and locally?

aralovelace (approved):

src/components/ProductPage/BodySection/BodySection.tsx
denissellu: it'll do the same thing i believe, but this uses a shorter syntax, if `section.callToAction` is true, render the `FeatureLink` component

(if this part is true) && (this part will render)

denissellu: @jamiehenson thanks for the info!! i was aiming to be more inline with how we have it in other places ie, https://github.com/ably/docs/blob/main/src/components/Homepage/BodySection/BodySection.tsx, lol and i choose the way i liked the look of :D

PR #2421, opened 2025-02-11
Channel enumeration: document that default for v3 is by=id

SimonWoolf:
Page to review:

- https://ably-docs-enumerate-by--jihqua.herokuapp.com/docs/metadata-stats/metadata/rest#enumerate
- https://ably-docs-enumerate-by--jihqua.herokuapp.com/docs/api/rest-api#enumeration-rest

(I'm a bit confused as to we have two separate pages for channel enumeration api docs, but made the change in both)

@coderabbitai ignore


content/metadata-stats/metadata/rest.textile
```diff
@@ -36,7 +36,7 @@ The following parameters are supported:
...
 - prefix := optionally limits the query to only those channels whose name starts with the given prefix<br>__Type: @string@__
-- by := _value_ optionally specifies what to return. Use @by=id@ to return just channel names. Use @by=value@ to return "@ChannelDetails@":/api/realtime-sdk/channel-metadata#channel-details. The default has changed: it used to be @by=value@, but for for protocol v3 or newer (the @version@ parameter passed to @rest#request@ is ≥ 3), it is @by=id@. **You are strongly recommended to use @by=id@ unless you really need the occupancy for every channel**. For one thing it will be much faster; for another, making very regular @by=value@ enumeration queries can prevent channels from closing down from inactivity.
+- by := _id (v3) or value (v2)_ optionally specifies what to return. Use @by=id@ to return just channel names. Use @by=value@ to return "@ChannelDetails@":/api/realtime-sdk/channel-metadata#channel-details. Using @by=id@ is recommended unless occupancy data is needed for every channel. @by=value@ queries can keep channels active but may return results more slowly.
```
SimonWoolf, line 39: > can keep channels active but may return results more slowly.

I don't understand the purpose of this change. Those are both bad things, joining them with "but" makes it sound sarcastic -- "It tastes terrible, but at least the portions are tiny".

I also not sure why you're trying to tone down the recommendation. There's no "may" about it, it _will_ be much slower and generally worse, and we want to discourage people from using it unless they absolutely have to.

content/partials/shared/_channel_enumeration.textile
```diff
@@ -25,4 +25,4 @@ Client libraries do not provide a dedicated API to enumerate channels, but make
...
 * The API does not guarantee that there will be no duplicated results between different pages, especially if a channel is alive in multiple regions. (It does not _currently_ do so, but it may begin to do so with no warning or deprecation period, so your implementation should be able to cope with duplication)
-* If you use @by=value@ (which until recent protocol versions was the default), just enumerating channels can briefly keep them alive, meaning if you do very regular enumeration you can get a situation where channels never close.
+* If you use @by=value@ in previous versions, just enumerating channels can briefly keep them alive, meaning if you do very regular enumeration you can get a situation where channels never close.
```
SimonWoolf, line 28: > If you use `by=value` in previous versions,

This changes the meaning, which is now wrong. by=value doesn't only have this effect in previous versions, it has this effect in every version. the only thing the version changes is that by=value is no longer the default.

content/partials/shared/_channel_enumeration.textile
```diff
@@ -25,4 +25,4 @@ Client libraries do not provide a dedicated API to enumerate channels, but make
...
 * The API does not guarantee that there will be no duplicated results between different pages, especially if a channel is alive in multiple regions. (It does not _currently_ do so, but it may begin to do so with no warning or deprecation period, so your implementation should be able to cope with duplication)
-* If you use @by=value@ in previous versions, just enumerating channels can briefly keep them alive, meaning if you do very regular enumeration you can get a situation where channels never close.
+* If you use @by=value@ (which from v3 protocol versions was the default), just enumerating channels can briefly keep them alive, meaning if you do very regular enumeration you can get a situation where channels never close.
```
SimonWoolf, line 28: > by=value (which from v3 protocol versions was the default)

no, "from" != "until"

GregHolmes (approved):

PR #2431, opened 2025-02-18
[EDU-1831] Remove product pages

m-hulbert:
## Description

This PR removes the product pages and their associated data and images. It also removes the `ProductPage` component and migrates those parts that were utilised by the `Homepage` component.


jamiehenson (approved):
YESS KILL IT ALL

I'd done a bit of this on the routing correction branch (not as much), but you doing this here helps a lot as it separates focuses and simplifies the scope of that. After this, there should basically be no yaml furniture left.

PR #2447, opened 2025-02-26
Update _channel_enumeration.textile, correcting an incorrect description

GregHolmes:
Due to having the description on two files, one update was missed in the previous PR: https://github.com/ably/docs/pull/2421 

This PR corrects the update that was missed.


SimonWoolf (approved):
thanks

PR #2393, opened 2025-01-22
[WEB-4176] - Attempt to Fix the Redocly Nav for the Control APi page

aralovelace:
## Description

A PR description indicating the purpose of the PR.

WEB-4176

## Review

Instructions on how to review the PR. 

* [Page to review](link)

PR #2452, opened 2025-02-26
[WEB-4211] Set layout options at build time, reduce and reexpress layout options

jamiehenson:
Tackles issues with the control api where we get `removeChild` JS errors.
Screenshot 2025-02-27 at 12 54 53

This involves a few things:
- restructures the means of setting the layout options so that it is done at build time instead of via a hook on component mount. This is a general improvement as well as a fix for this particular issue
- since the control API page (or more specifically, Redoc) seems to have a big issue with `useEffect` usage alongside it, language detection and setting has been moved into the base template itself, which is the only page that needs it, sparing the control api template the need to have this logic there as well. Ideally we get rid of this DOM-based language detection as well as part of the MDX work
- re-expresses the layout options to be opt-in instead of opt-out. Since we set them everywhere, it's better to render layout components when we have a positive sentiment (i.e. show sidebar if sidebar is true) versus a negative one (show sidebar if hideSidebar isn't true). No change in functionality, just a logical change.

Review link: https://ably-docs-fix-control-a-qu6lgg.herokuapp.com/


jamiehenson:
@aralovelace since this was on your plate initially, did you have a ticket for this, to attach this to?

kennethkalmer:
@jamiehenson this might negate #2393 🤞

jamiehenson:
@kennethkalmer what's wrong with it? I can't really tell from that PR or the linked ticket what the problem is. It's janky and has basically no responsiveness but it _does_ seem operational

kennethkalmer:
> @kennethkalmer what's wrong with it? I can't really tell from that PR or the linked ticket what the problem is. It's janky and has basically no responsiveness but it _does_ seem operational

@jamiehenson there is an issue with redocly, and this existed before you merged the new navigation, where it does this (look at the account section)

https://github.com/user-attachments/assets/d5dfea0e-f481-4401-b875-8022e626672c

kennethkalmer (approved):
Nice! I like how you reexpressed the layout props as well ➕

jamiehenson:
> there is an issue with redocly, and this existed before you merged the new navigation, where it does this (look at the account section)

@kennethkalmer barf. It seems to open the menu above the click, which renders the top and bottom ones inoperable. I don't think it's to do with the layout, fixed or otherwise. Since it's a closed book there's not much we can do inside, the best I can think of off the top of my head is to have a script that runs on top that, when you click a menu item, overrides the functionality of redoc and sets the `display` property on the correct `ul` element.

Today is a bit of a write-off in terms of sprint work so I can try that after this whilst the topic is still fresh.

Issue #2450, opened 2025-02-26
Change request for: /docs/push/configure/device


korarit:
**Page name**: Devices
  **URL**: [/docs/push/configure/device](https://ably.com/docs//docs/push/configure/device)
  Language: **Java

  **Requested change or enhancement**:

[Activate via server](https://ably.com/docs/push/configure/device?lang=java#server) example in java is not java code is kotlin 

![Image](https://github.com/user-attachments/assets/abfa8b2f-5a4d-4b31-a388-8a92466c4c10)

GregHolmes:
Hi @korarit,

Thank you for flagging this issue with the documentation. I've just released an update, which does indeed render Java and not Kotlin in the Java example for server side activation for push notification.

PR #2454, opened 2025-02-27
[EDU-1836] - Updated java example of server side activation for push

GregHolmes:
The server side activation for Android push was actually Kotlin and not Java: https://github.com/ably/docs/issues/2450

This update adds the Java equivalent.


ttypic (approved):
LGTM

PR #2461, opened 2025-03-02
Update Message object attributes

mattheworiordan:
Previous description for idempotency was confusing. You don't update a message with an ID, you provide an ID before publishing. We also did not link to any idempotent docs.


m-hulbert (approved):
LGTM, thank you!

PR #2464, opened 2025-03-05
Add LiveObjects product to site navigation

VeskeR:
## Description

Adds navigation for LiveObjects product in `src` folder.

This was added by searching for "livesync" in "src" folder and adding corresponding "liveObjects" sections where necessary.

Currently missing:

- [ ] Need a png for main docs page for LiveObjects https://github.com/ably/docs/pull/2464/files#diff-74d9b85d26804605bb64641c8787195382c4dc7674c01c5fd5c1716fe168144dR60
- [x] Need LiveObjects icons for docs navigation https://github.com/ably/docs/pull/2464/files#diff-39bc77ddcd9f663f1666da2e0f9a528358251e8f7285384b8f53c47845b06075R7-R8

### Checklist

### DIFF
--- a/src/data/content/homepage.ts
+++ b/src/data/content/homepage.ts
@@ -54,6 +54,13 @@ export default {
           links: [{ text: 'Get started', href: '/docs/spaces' }],
         },
         {
+          title: 'LiveObjects',
+          type: 'feature',
+          content: 'Seamlessly sync application state between clients.',
+          image: 'liveobjects.png',
+          links: [{ text: 'Get started', href: '/docs/liveobjects' }],
+        },
+        {
           title: 'LiveSync',
           type: 'feature',
           content: 'Seamlessly sync database changes with frontend clients.',
--- a/src/data/languages/languageData.ts
+++ b/src/data/languages/languageData.ts
@@ -3,13 +3,13 @@ import { LanguageData } from './types';
 
 export default {
   platform: {
-    javascript: 2.6,
-    nodejs: 2.6,
+    javascript: 2.7,
+    nodejs: 2.7,
   },
   pubsub: {
-    javascript: 2.6,
-    nodejs: 2.6,
-    react: 2.6,
+    javascript: 2.7,
+    nodejs: 2.7,
+    react: 2.7,
     csharp: 1.2,
     flutter: 1.2,
     java: 1.2,
@@ -30,6 +30,9 @@ export default {
   spaces: {
     javascript: 0.4,
     react: 0.4,
+  },
+  liveObjects: {
+    javascript: 2.7,
   },
   liveSync: {
     javascript: 0.4,
--- a/src/data/nav/liveobjects.ts
+++ b/src/data/nav/liveobjects.ts
@@ -1,0 +1,67 @@
+import { NavProduct } from './types';
+
+export default {
+  name: 'Ably LiveObjects',
+  link: '/docs/liveobjects',
+  icon: {
+    closed: 'icon-product-liveobjects-mono',
+    open: 'icon-product-liveobjects',
+  },
+  content: [
+    {
+      name: 'Introduction',
+      pages: [
+        {
+          name: 'About LiveObjects',
+          link: '/docs/liveobjects',
+        },
+      ],
+    },
+    {
+      name: 'Get started',
+      pages: [
+        {
+          name: 'Quickstart',
+          link: '/docs/liveobjects/quickstart',
+        },
+      ],
+    },
+    {
+      name: 'LiveObjects features',
+      pages: [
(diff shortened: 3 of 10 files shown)


src/data/content/homepage.ts
```diff
@@ -53,6 +53,13 @@ export default {
           image: 'spaces.png',
           links: [{ text: 'Get started', href: '/docs/spaces' }],
         },
+        {
+          title: 'LiveObjects',
+          type: 'feature',
+          content: 'Seamlessly sync application state between clients.',
+          image: 'liveobjects.png',
```
VeskeR, line 60: this needs to be added

src/data/nav/liveobjects.ts
```diff
@@ -0,0 +1,71 @@
+import { NavProduct } from './types';
+
+export default {
+  name: 'Ably LiveObjects',
+  link: '/docs/liveobjects',
+  icon: {
+    closed: 'icon-product-liveobjects-mono',
+    open: 'icon-product-liveobjects',
```
VeskeR, line 8: and these two icons

src/data/languages/languageData.ts
```diff
@@ -31,6 +31,9 @@ export default {
     javascript: 0.4,
     react: 0.4,
   },
+  liveObjects: {
+    javascript: 2.6,
```
VeskeR, line 35: not sure about this one. LiveObjects is not a separate product, so it doesn't have its own versioning.
will it inherit the pub/sub SDKs version then? e.g. 2.6 for `ably-js`

cc @mschristensen

mschristensen: Yes I think we should just specify the Ably JS version

mschristensen:
Looks good, we just need some assets from the design team - I'll follow up with Jamie

src/data/languages/languageData.ts
VeskeR: ok, works for me.

Also, technically, it wouldn't be `2.6`, as with the liveobjects release we will bump the minor version in ably-js to at least `2.7` (or more if other versions released before that).

@ably/team-deved would you guys prefer:
1. this PR to have `2.6` version in this PR, and then open a second PR to update the ably-js versions for all products (not just liveobjects) to higher version once it's published
2. keep this PR open until ably-js is published and change the version here to the actual one, as well as for other products?

VeskeR:
I've removed the Setup page as we decided to drop it in favor of Quickstart: https://github.com/ably/docs/pull/2463#discussion_r1985433477

src/data/languages/languageData.ts
m-hulbert: Assuming that LiveObjects is the only change going into `2.7` of ably-js then I think we can just bump both of them to `2.7` in this PR.

src/data/nav/liveobjects.ts
```diff
@@ -0,0 +1,67 @@
...
+          name: 'Batch Operations',
+          link: '/docs/liveobjects/batch',
+        },
+        {
+          name: 'Lifecycle Events',
+          link: '/docs/liveobjects/lifecycle',
+        },
+        {
+          name: 'Typing',
+          link: '/docs/liveobjects/typing',
```
m-hulbert, line 50: As per my review on the main PR; I wonder if the features should remain only LiveCounter and LiveMap, and then these 3 items fall under a separate heading. Any thoughts?

src/data/nav/liveobjects.ts
m-hulbert: These get pulled in from Ably UI, so you don't need to do anything in this PR with them.

src/data/nav/liveobjects.ts
VeskeR: > then these 3 items fall under a separate heading. Any thoughts?

Not against putting them under a separate heading. Honestly, they're under LiveObjects features heading right now because I couldn't think of another collective name for these three.
Any ideas?

src/data/languages/languageData.ts
VeskeR: sgtm, I've preemptively updated it to `2.7`, but will change it again if we end up releasing liveobjects under another minor ably-js version

src/data/nav/liveobjects.ts
m-hulbert: Maybe something like 'Advanced', or 'Advanced usage'? 

I think they should be separated in all honesty, but also struggling with the correct terminology.

VeskeR: Created https://ably.atlassian.net/browse/PUB-1588 to give it another thought later, don't want to block this PR right now

PR #2472, opened 2025-03-12
Fix bug where support link goes to /docs/support

GregHolmes:
A bug has been raised where the link at the top right of the docs to go to support now links to `/docs/support` which doesn't exist, so a 404 is returned. This will likely have been a side effect to some changes to the navigation of the docs recently released. A screenshot of the link is shown below:

Screenshot 2025-03-12 at 09 54 52


aralovelace (approved):
Thanks for doing this Greg! looks good to me

m-hulbert (approved):

PR #2474, opened 2025-03-12
Refactor markdown files for examples for easier implementation into codesandpack

GregHolmes:
Each example has a `page.md` file within the root feature directory, then a `README.md` file in each language specific directory.  The `page.md` file had instructions for both React and JS, which isn't necessary when viewed on the CodeSandpack. So removing the `page.md` file and updating `README.md` files with the parts relevant to that language/framework with the parts from `page.md`.

### DIFF
--- a/examples/spaces-live-cursors/javascript/README.md
+++ b/examples/spaces-live-cursors/javascript/README.md
@@ -12,9 +12,9 @@ Live cursors are implemented using [Ably Spaces](/docs/products/spaces). The Spa
 
 Use the following components to add live cursors into an application:
 
-* [`spaces.get()`](/docs/spaces/space#create) - creates a new or retrieves an existing `space`.
-* [`space.cursors.subscribe()`](/docs/spaces/cursors#subscribe) - subscribes to members' online status and profile updates by registering a listener.
-* [`space.enter()`](/docs/spaces/space#enter) - Entering a space will register a client as a member and emit an `enter` event to all subscribers.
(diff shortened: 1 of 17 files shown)


jamiehenson (approved):
Looks great, thanks 🙏

examples/spaces-live-cursors/javascript/README.md
```diff
@@ -1,6 +1,23 @@
...
+
+Live cursors are utilized in applications such as online whiteboards to enable users to see where others are pointing or what they are drawing. They are also used in slideshow applications to enable users to see which slide each member is interacting with.
+
+Live cursors are implemented using [Ably Spaces](/docs/products/spaces). The Spaces SDK contains a set of purpose-built APIs that abstract away the complexities involved in managing the state of participants in a collaborative application. It is built on top of Ably's core platform, and so it provides the same performance and scaling guarantees.
+
+## Resources
+
+Use the following components to add live cursors into an application:
+
+* [`spaces.get()`](/docs/spaces/space#create) - creates a new or retrieves an existing `space`.
```
jamiehenson, line 15: Another request if you don't mind - discovered this after my initial review. Having hyphens here within the list items confuses the markdown parsing process as it's the syntax for another list.

To help this, and to be more consistent with the react example, can these be changed to colons?

GregHolmes: Sure. I've updated it.

PR #2480, opened 2025-03-13
add batch push publish docs

owenpearson:
adds REST API reference and conceptual docs for batch push

see implementation here: https://github.com/ably/realtime/pull/7217

[PUB-1319](https://ably.atlassian.net/browse/PUB-1319)

[PUB-1319]: https://ably.atlassian.net/browse/PUB-1319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/api/rest-api.textile
+++ b/content/api/rest-api.textile
@@ -29,6 +29,7 @@ jump_to:
     - list channel subscriptions#list-channel-subscriptions
     - list channels#list-channels
     - publish directly to device#push-publish
+    - publish via batch push API#push-publish-batch
   Authentication API:
     - requestToken#request-token
     - revokeTokens#revoke-tokens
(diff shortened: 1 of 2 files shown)


content/api/rest-api.textile
```diff
@@ -1152,6 +1153,55 @@ A successful request returns an empty response.
 
 An unsuccessful request returns an error.
 
+h3(#push-publish-batch). Publish via batch push API
+
+Convenience endpoint to deliver a push notification payload to multiple devices or browsers in a single request by specifying a list of recipients and corresponding payloads.
```
mschristensen, line 1158:
```suggestion
Convenience endpoint to deliver multiple push notification payloads to multiple devices or browsers in a single request by specifying a list of recipients and corresponding payloads.
```

content/api/rest-api.textile
```diff
@@ -1152,6 +1153,55 @@ A successful request returns an empty response.
...
+
+h6. POST rest.ably.io/push/admin/batch/publish
```
mschristensen, line 1160: and here

mschristensen (changes requested):
Great thanks for this, a few comments. I think we also need to include something about:

- What the response looks like, especially for partially failed requests, similar to https://ably-docs-batch-push-qfok2100v.herokuapp.com/docs/api/rest-api?lang=javascript#batch-publish
- Supported parameters or options as in https://ably-docs-batch-push-qfok2100v.herokuapp.com/docs/api/rest-api?lang=javascript#push-publish
- What the limit is per request

owenpearson:
@mschristensen have resolved all the suggestions. regarding your review comment:
> What the response looks like, especially for partially failed requests, similar to https://ably-docs-batch-push-qfok2100v.herokuapp.com/docs/api/rest-api?lang=javascript#batch-publish

the endpoint responds with 201 on success, and a generic error info upon failure. There's actually no client visible partial failure case here because the entire batch is added to the queue - the rest api doesn't discriminate between items within the batch. It's either bad request, server error, or 201 no content.

>Supported parameters or options as in https://ably-docs-batch-push-qfok2100v.herokuapp.com/docs/api/rest-api?lang=javascript#push-publish

There aren't any public params or options here so nothing to add.

> What the limit is per request

I've added something to both pages about a 10k limit. The idea was to discern the limit from load testing, hence why i've been holding off from updating this, but it's likely to be 10k in the end so will leave it as that.

m-hulbert (commented):
Couple of minor suggestions. Can we also add an item to the main 'publish directly' section bullet points to include batch too?

content/api/rest-api.textile
```diff
@@ -1152,6 +1153,56 @@ A successful request returns an empty response.
 
 An unsuccessful request returns an error.
 
+h3(#push-publish-batch). Publish via batch push API
+
+Convenience endpoint to deliver multiple push notification payloads to multiple devices or browsers in a single request by specifying a list of recipients and corresponding payloads.
+Currently the batch push endpoint allows a maximum of 10,000 recipients per request (recipients are counted per payload, so publishing two payloads to the same recipient counts as two recipients).
```
mschristensen, line 1159: Should we change this wording from "recipients" to "notifications" as the thing which is limited?
E.g.: "Currently the batch push endpoint allows a maximum of 10,000 notifications per request (one notification is counted as one payload delivered to one recipient, so publishing two payloads to the same recipient counts as two notifications)."

mschristensen (approved):
Looks good, a minor suggestion

PR #2489, opened 2025-03-22
Update Stats docs

mattheworiordan:
- Clarified the scope of account statistics
- Removed account stats from the REST API - there is no such API, API keys are scoped to apps only. Not sure how this was ever documented.
- Improved clarity around obtaining account IDs:. Using an <aside>. Is this correct?
- Found it odd that in code samples we refer to, for example, {ACCOUNT_ID} then in the docs we refer to that variable as @accountId@. I don't follow the inconsistency. Happy to change code examples to use accountId if preferred, but I don't see any merit in inconsistency in the same page.
- Adjusted headings, specifically differentiating between “Statistics response payload” and “Statistics response”, clarifying the API response structure. I don't think it was clear before what a statistics response payload was vs statistics metrics. I have made what I think is am improvement
- Improved metric definitions, clearly stating the inclusion of push notifications and integrations in outbound metrics. But also just made things a lot more concise.
- Simplified curl command examples, removing redundant --location, and I have no idea why `--request POST` flags was added for a GET method endpoint.
- Provided clearer context around API endpoints by explicitly linking to the correct documentation references.

---

Note I have some questions around whether we need `<aside>` for these smaller "tip" areas that are possibly more subtle, as opposed to stand out which I think this will do. 

Also, do we not need aside for the product release stages in PDR-057?

### DIFF
--- a/content/account/control-api.textile
+++ b/content/account/control-api.textile
@@ -169,13 +169,10 @@ The only operation available at the account-level is to retrieve account-wide st
 To retrieve account-level statistics:
 
 ```[sh]
-curl --location --request POST 'https://control.ably.net/v1/accounts/${ACCOUNT_ID}/stats' \
---header 'Content-Type: application/json' \
---header 'Authorization: Bearer ${ACCESS_TOKEN}' \
---data-raw '{
-    "unit": "minute",
-    "limit": 2
-}'
+curl --request GET \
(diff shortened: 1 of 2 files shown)


GregHolmes (approved):

GregHolmes:
On the consistency of variable naming I've created a JIRA ticket as I believe this needs to be carried out throughout the docs and not just this page, for various variable names. The JIRA ticket: https://ably.atlassian.net/browse/EDU-1886

PR #2491, opened 2025-03-24
[WEB-4086] Port Next projects to Vite 4

jamiehenson:
Sandpack is a fussy master.

Unfortunately we can't use Next projects with it, at least with the version of Sandpack we have access to, due to its server-rendering foundations. Thus, it's necessary to rewrite the Next projects to something else that's more clientside, and an obvious candidate for that due to its usage elsewhere here and the general deprecation of `create-react-app`, is Vite. Though, naturally, Sandpack demands an older version of Vite (4) instead of anything more modern (5, 6).

This PR is a semi-automated rewrite of those projects, and I've also taken the opportunity to further the initial config de-duplication work to make further use of workspaces and reduce the file count down. It also swaps out `faker` for `minifaker`, which is objectively inferior but works with Sandpack.

I've tested all of the non-auth react projects locally and they work, with the exception of `spaces-live-cursors-javascript` which has an ongoing issue not related to this migration. We'll have to hold back on the auth projects and release them with some backend infrastructure in place at a later date.

For @GregHolmes:
Project-wise the general approach I/the bot took was to consolidate the `app/page.tsx` and `app/layout.tsx` into one `App.tsx` component, and replicate multi-page apps with `react-router`


jamiehenson:
@coderabbitai review

examples/auth-generate-jwt/javascript/src/styles.css
```diff
@@ -1,100 +1 @@
...
-input {
-  padding: 0.5rem;
-  width: 100%;
-  height: 2.5rem;
-  font-size: 0.875rem;
-  border-radius: 0.375rem;
-  outline: none;
-  transition: all 0.2s ease-in-out;
-}
+@import '../../../styles.css';
```
jamiehenson, line 1: This is an indicator of a change common throughout this PR - project stylesheets have been de-duplicated, with `examples/styles.css` housing all the common styles, and these ones only importing and adding unique styles.

In the case of no unique styles, the React apps import from `examples/styles.css` directly. The JS ones have a one line file with an input since `index.html` can't seem to reach back beyond the directory it's in.

examples/auth-generate-jwt/javascript/vite.config.ts
```diff
@@ -1,10 +1,7 @@
...
 import { defineConfig } from 'vite';
-import dotenv from 'dotenv';
-import path from 'path';
-
-dotenv.config({ path: path.resolve(__dirname, '../../.env.local') });
+import baseConfig from '../../vite.config';
 
 export default defineConfig({
   ...baseConfig,
+  envDir: '../../',
```
jamiehenson, line 6: This was a more vite-y approach than dotenv

examples/auth-generate-jwt/react/src/App.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { useState, useEffect } from 'react';
```
jamiehenson, line 1: I wouldn't look too closely at the `auth-*` projects. They're not in scope to support just yet.

examples/chat-online-status/javascript/src/script.ts
```diff
@@ -1,10 +1,11 @@
 import * as Ably from 'ably';
 import { ChatClient, PresenceEvent, PresenceMember, AllFeaturesEnabled } from '@ably/chat';
-import { faker } from '@faker-js/faker';
+import minifaker from 'minifaker';
```
jamiehenson, line 3: `minifaker` was brought in purely because `faker` blows up sandpack for reasons long and arduous

examples/package.json
```diff
@@ -88,36 +88,36 @@
...
+    "minifaker": "^1.34.1",
     "nanoid": "^5.0.7",
-    "next": "^15",
-    "postcss-combine-duplicated-selectors": "^10.0.3",
-    "postcss-sort-media-queries": "^5.2.0",
     "react": "^18",
     "react-dom": "^18",
     "react-icons": "^5.4.0",
-    "ts-node": "^10.9.2",
+    "react-router-dom": "^6.22.2",
```
jamiehenson, line 99: Brought in react router as the most general example of a routing solution for react, nothing fancy

examples/auth-generate-jwt/javascript/src/styles.css
kennethkalmer: @jamiehenson how will Sandpack handle this? maybe we can include the file and rewrite the import statement while passing it in?

jamiehenson: @kennethkalmer I've bitten the bullet and brought all the stylesheets back in for the react projects, also relocating them into `/src` which sandpack more natively expects.

Squashed and cherry picked this into the sandpack branch, it works 👍 (thankfully, the child postcss/vite/tw/ts configs are minimal enough to be handled by sandpack's defaults it seems - so we can maintain a compromise between de-duplication in the repo and stuff that works downstream)

jamiehenson: One dumb thing I was bashing my head against though is that Sandpack hates Tailwind if it's not brought in as an external script - but that's a downstream change, not for here

kennethkalmer (approved):
From my side I think the CSS issue was the only one I spotted so happy to see it is resolved. I think we'll find any other snakes in the grass when packaging them up in sandpack

Excellent conducting @jamiehenson!

examples/spaces-avatar-stack/react/.gitignore
```diff
@@ -36,3 +36,35 @@ yarn-error.log*
 next-env.d.ts
 
 .env.local
+
```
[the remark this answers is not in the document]
jamiehenson: It's not fresh. I can search and replace out the next references in the project gitignores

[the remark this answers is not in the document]
jamiehenson: Hm, even though stale may be better to keep in actually, for the moment. Removing the entries surfaces lots of next stuff that's laying in wait, don't want accidental commits of this stuff elsewhere

GregHolmes (approved):

PR #2492, opened 2025-03-24
Request method version missing

mattheworiordan:
See https://github.com/ably/specification/blob/main/textile/features.textile, the method signature is `request(String method, String path, Int version, Dict<String, String> params?, JsonObject | JsonArray body?, Dict<String, String> headers?)`.

This PR should not be merged, but should serve as the basis to check where else we refer to the request method with an incorrect signature.

In addition, we should we which SDKs are 2.x and thus use this new signature.


m-hulbert:
Investigated and fixed in https://github.com/ably/docs/pull/2805

PR #2497, opened 2025-03-28
LLMs.txt

mattheworiordan:
See https://llmstxt.site/ for context.

Modern AI friendly tools provide llms.txt files for LLMs.

This script simply calls out to [Firecrawl](https://docs.firecrawl.dev/features/alpha/llmstxt#known-limitations-alpha) and generates an LLMs.txt file.  However, it is not working as expected and some additional work is needed:

- Crawlers cannot crawl all languages, we need better discovery, perhaps an index page with all languages, and we need to understand why crawlers are not selecting different languages (should these be hyperlinks).
- llms.txt needs to be served from ably.com, not ably.com/docs (website should be able to easily proxy this)
- There are issues with language selection where if you select Node.js, navigate to a page that does not support Node.js, and then navigate to one that does, it remains on JS. This means we're losing the language selection for crawlers too. See GIF below. 

MO screenshot 2025-03-27 at 23 53 11


GregHolmes (approved):

kennethkalmer (commented):
I get the intent, this is pretty cool, but in practice this won't work well and the reason is it will always be behind the current PR. We cannot update the text file with the updates from the PR because the PR has not been deployed yet...

That said, I wonder if we cannot simply attempt to create a build step that simply slurps up all the HTML files in `public`, extracts the elements under a certain selector (so we don't get the navigation elements) and renders it out as a huge text file (or as mutliple text files), and then we could even solve the language problem because we understand the markup

jamiehenson (commented):
So to get my head into this - Firecrawl takes an entry point, crawls from there, and spits out a markdown file consisting of crawled links and a one-line synopsis of the contents that's generated through some LLM-ish means.

As part of the nav rewrite we made the data foundations of the pages more accessible. To achieve what you ask here re: languages, we could compile a matrix of [pages](https://github.com/ably/docs/blob/llms.txt/src/data/nav/pubsub.ts) (pubsub ones as an example here) and [languages per product](https://github.com/ably/docs/blob/llms.txt/src/data/languages/languageData.ts) into a page and point firecrawl at that. 

To be honest, this would get us most of the way there, by looping in the meta descriptions currently on the textile documents we pretty much have what Firecrawl provides without the need for another service.

Alternatively, lightest approach could be to hide some links on each page pointing to each available language variant, but this would fall apart if Firecrawl assesses the visibility of links in its traversal.

paddybyers:
Is this mergeable?

m-hulbert:
@kennethkalmer did you have a chance to look at proxying this if we merge, and/or the scope of work required to generate the files Jamie mentioned?

paddybyers:
bump...

mattheworiordan:
> That said, I wonder if we cannot simply attempt to create a build step that simply slurps up all the HTML files in public,

Sure, this was not meant to be perfect, it was meant to be an experiment of what we could do to address our lack of having an llms.txt file in under an hour. 

> renders it out as a huge text file (or as mutliple text files), and then we could even solve the language problem because we understand the markup

Sure, but we also have a separate problem that needs addressing clearly if crawlers are not able to see the various language versions. I don't think we can ignore that issue.

> Alternatively, lightest approach could be to hide some links on each page pointing to each available language variant, but this would fall apart if Firecrawl assesses the visibility of links in its traversal.

Sure, I think we need to figure out how crawlers more generally can access language versions and ensure they are, because if we don't, we're going to lose the battle of LLMs being able to unearth code samples etc.

kennethkalmer:
I took a different stab at this in #2554 which generates a basic `llms.txt` as part of the build step, similarly to how we generate sitemaps.

As for crawlers being able to access the different languages, we just need to sit down and think it through and look at our options. It is an important issue.

PR #2490, opened 2025-03-24
Live objects: Add REST API docs

zknill:
## Description

Add new page in the "API References" part of the product docs which details the REST APIs available for interacting with live objects.

### Checklist

- [ ] Commits have been rebased.
- [ ] Linting has been run against the changed file(s).
- [ ] The PR adheres to the [writing style guide](../writing-style-guide.md) and [contribution guide](../CONTRIBUTING.md).

PR #2513, opened 2025-04-08
[EDU-1882] - Migrate live objects rest api to open api

GregHolmes:
There was a previous [PR](https://github.com/ably/docs/pull/2490) which provided specs for the LiveObjects API. However, moving forward we'll be building API specs using Open API. This PR converts that LiveObjects API into OpenAPI. 

To view this, open the staging URL to: `/docs/api/liveobjects-rest`


static/open-specs/liveobjects.yaml
```diff
@@ -0,0 +1,641 @@
...
+                      map:
+                        entries:
+                          counter: { data:{ number: 5 }}
+                          myObjectRef: { data: { objectId: "counter:Nz1ZiNjqsDfkDjA61xarinqpWsqEGAAw2mzWWtvX2b8@1742481614000" }}
+                refArray:
+                  summary: Just object references
+                  value:
+                    - "root"
+                    - "map:YIffJYRAP2k2e7ZP+xzequ9c5kDu1LfI/sEOKoWHvv4@1742479683000"
+                    - "counter:Nz1ZiNjqsDfkDjA61xarinqpWsqEGAAw2mzWWtvX2b8@1742481614000"
```
zknill, line 222: It wasn't super clear to me from the rendered doc that the difference in these two examples was the inclusion or not of `values=true` query param. Maybe we can reference that somewhere?

static/open-specs/liveobjects.yaml
```diff
@@ -0,0 +1,641 @@
...
+                        - objectId
+                      properties:
+                        objectId:
+                          type: string
+                        map:
+                          type: object
+                          properties:
+                            entries:
+                              type: object
+                              additionalProperties:
```
zknill, line 179: perhaps we need some `example: ` keys in here?

static/open-specs/liveobjects.yaml
```diff
@@ -0,0 +1,641 @@
...
+                                properties:
+                                  data:
+                                    type: object
+                                    oneOf:
+                                      - properties:
+                                          string:
+                                            type: string
+                                      - properties:
+                                          objectId:
+                                            type: string
```
zknill, line 190: we're going to have this `oneOf` everywhere like: 
```yaml
oneOf: 
 - properties: 
   string:
     type: string
 - properties: 
   number:
     type: number
 - properties: 
   boolean:
     type: boolean
 - properties: 
   bytes:
     type: string
   encoding: 
     type: string
 - properties: 
   objectId:
     type: string
```

is there a way that we can make it a component and then `$ref` that value in these response specs?

static/open-specs/liveobjects.yaml
```diff
@@ -0,0 +1,641 @@
...
+        example: "reactions.likes"
+    operationParameter:
+      name: operation
+      in: query
+      required: true
+      description: |
+        Specifies the operation to perform. Supported values include:<br><br>
+        - `MAP_CREATE`: Create a map object.
+        - `MAP_SET`: Set a key-value pair in a map object.
+        - `MAP_REMOVE`: Remove a key-value pair from a map object. Provide the `objectId` of the map you wish to modify and the key to remove.
```
zknill, line 585: > Provide the `objectId` of the map you wish to modify and the key to remove.
I don't think this is needed if you are doing path ops

static/open-specs/liveobjects.yaml
```diff
@@ -0,0 +1,641 @@
...
+          schema:
+            type: string
+            example: "my-app:likes:counter"
+        - name: objectId
+          in: path
+          required: true
+          description: The object ID.
+          schema:
+            type: string
+            example: "object-123"
```
zknill, line 369: I think here we need to say that the default that customer can use out of the box is "root"

static/open-specs/liveobjects.yaml
```diff
@@ -0,0 +1,641 @@
...
+
+      Maps have additional metadata attached to their map entries.
+
+      | Operation | Example  |
+      |-----------|----------|
+      | Create a map | `{ "operation": "MAP_CREATE", "data": { "myMapKey": {"string": "myMapValue"}, "myOtherKey": {"boolean": true} } }` |
+      | Set a value on a map | `{ "operation": "MAP_SET", "objectId": "map:99OSLwCFQsrs6GID4M8rfO_0sYmVwRADcECua_@1742207566771", "data": {"key":"foo", "value":{"string":"bar"}} }` |
+      | Remove a value from a map | `{ "operation": "MAP_REMOVE", "objectId": "map:99OSLwCFQsrs6GID4M8rfO_0sYmVwRADcECua_@1742207566771", "data": {"key":"myMapKey"} }` |
+      | Create a counter | `{ "operation": "COUNTER_CREATE", "data": { "number": 3.1415926 } }` |
+      | Increment a counter | `{ "operation": "COUNTER_INC", "objectId": "counter:J7x6mAF8X5Ha60VBZb6GtXSgnKJQagNLgadUlgICjkk@1734628392000", "data": {"number": 2} }` |
```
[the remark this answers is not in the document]
GregHolmes: You're right. I'm not entirely sure how this made its way in here. It seemed to replace a valid table that I was accidentally removed too. I've rectified this.

static/open-specs/liveobjects.yaml
GregHolmes: These won't render anywhere. They're shown in the line 202 of this commit as full block examples.

GregHolmes:
@zknill  I've updated based on your feedback. However the last thing (other than your re-review) is the http status responses that are possible.

static/open-specs/liveobjects.yaml
```diff
@@ -0,0 +1,1386 @@
...
+          { "data": { "string" : "Ably Pub/Sub" }}
+          { "data": { "boolean" : true }}
+          { "data": { "bytes": "TGl2ZU9iamVjdHMgaXMgYXdlc29tZQo=", "encoding": "base64" }}
+          { "data": {
+            "objectId":
+              "counter:Nz1ZiNjqsDfkDjA61xarinqpWsqEGAAw2mzWWtvX2b8@1742481614000"
+          }}
+        ```
+
+        <br>Maps are made of entries, which are a user-defined key (e.g. `myMapKey`) and a data value. Counters are data values containing a number.
```
zknill, line 35: these descriptions of data values, maps, and counters seem like they are the wrong way around. 

IMO it should be; 
- there are two types of objects; maps and counters
- maps have XYZ
- counters have ABC

static/open-specs/liveobjects.yaml
```diff
@@ -0,0 +1,1386 @@
...
+          description: Unauthorized
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: '#/components/schemas/AuthError'
+                  - $ref: '#/components/schemas/ClientRestrictionNotSatisfied'
+                  - $ref: '#/components/schemas/InvalidTokenKey'
+                  - $ref: '#/components/schemas/MalformedCredential'
+                  - $ref: '#/components/schemas/OperationObjectSubscribeUnauthorized'
```
zknill, line 117: some of these errors are not rendering: 
image
image

.. etc

zknill (commented):
I'm really sorry @GregHolmes, but for me, these docs don't work right now. 

They are neither: 
- a spec of the endpoints, paths, query params, types, responses, requests, and data
- NOR; a full description of the API and how to use it. 

The docs right now have landed in a half-way between a simplified spec of the requests and responses, and a description of the LiveObjects features, etc. And I think they do neither of the jobs well. 

My suggestion that we strip the feature explanations from these docs, and simplify them down to just the REST API detail (request / response fields and bodys, etc), and add links to other docs that do the explanation. (cc: @mschristensen)

For me, right now, I don't see these docs as usable by someone new to liveobjects.

GregHolmes:
> I'm really sorry @GregHolmes, but for me, these docs don't work right now.
> 
> They are neither:
> 
> * a spec of the endpoints, paths, query params, types, responses, requests, and data
> * NOR; a full description of the API and how to use it.
> 
> The docs right now have landed in a half-way between a simplified spec of the requests and responses, and a description of the LiveObjects features, etc. And I think they do neither of the jobs well.
> 
> My suggestion that we strip the feature explanations from these docs, and simplify them down to just the REST API detail (request / response fields and bodys, etc), and add links to other docs that do the explanation. (cc: @mschristensen)
> 
> For me, right now, I don't see these docs as usable by someone new to liveobjects.

Thanks @zknill , I agree to your suggestion, I wasn't keen on keeping all of the descriptions from your original spec as it doesn't sit well in Open API structure. Do we have pages that describe these features well in the main docs already?

zknill (approved):
SO much easier to read/follow! thanks a lot for reworking and bearing with me 💚

PR #2543, opened 2025-04-17
Ephemeral messages; _message partial

SimonWoolf:
Adds docs for ephemeral messages to pub-sub-advanced.

I was going to update the `extras` description in message objects to add `ephemeral` but was irked by having to make the change in 3 places, which all seemed to have copied and pasted the `Message` types from the message partial instead of just including the partial. (And then inevitably diverged as people only made fixes to one or two of them, which is why partials are good). I couldn't see any reason given in the commit message that did the duplication, so I've taken the liberty of deduplicating (taking the text from the realtime-api version since that seems to have been kept more up to date than the one in hte partial).

PUB-1317


m-hulbert:
I don't disagree with your logic, I just think it's something we should be mindful of is all. I'll have a think about an example that's in-between the two. However, that shouldn't block getting this feature out - I'll raise a separate PR if I have any inspiration.

m-hulbert (approved):

content/pub-sub/advanced.textile
```diff
@@ -901,6 +901,36 @@ channel := realtime.Channels.Get("chatroom")
...
+
+A message can be marked as *ephemeral* to exempt it from:
+- being stored in "persisted history":/docs/storage-history/storage
+- being sent in "attachment rewinds":/docs/channels/options/rewind
+- being sent to clients "resuming over a period of disconnection":/docs/connect/states
+- being sent to "firehose, webhooks, and queue integrations":/docs/integrations
+
+In other words, it will be exempt from everything except being delivered to currently-connected realtime connections.
+
+This is useful for when you want a channel to have a mixture of messages, some of which you care about more than others. For example, say you have a chat room with ephemeral room reactions, and you are using @rewind=100@ so that new attachments get the last one hundred messages on the channel. Without this feature, you may only get 70 'real' messages in the rewind and 30 stale reactions, which you'd then need to filter out clientside.
```
paddybyers, line 914:
```suggestion
This is useful for when you want a channel to have a mixture of messages, some of which you care about more than others. For example, say you have a chat room with ephemeral room reactions, and you are using @rewind=100@ so that new attachments get the last one hundred messages on the channel. Without this feature, you may only get 70 'real' messages in the rewind and 30 stale reactions, which you'd then need to filter out clientside.
```

This is useful for events that are relevant only at the time they are published, and have no value when stale; examples might be streaming of continuously changing values such as realtime telemetry, position information, etc. Since ephemeral messages can be interspersed with other non-ephemeral messages on a channel, it is possible to use a single channel to convey all relevant events for some entity, including a mix of some that need to be persisted and others that are only ephemeral.

SimonWoolf: sure fine

PR #2545, opened 2025-04-17
[WEB-4327] Fix default language setting on non-JS templates

jamiehenson:
Linked to https://ably.atlassian.net/browse/WEB-4327.

If no lang param is supplied, it fell back to the first language defined at the product level in the data, which doesn't always apply as pages within a product can have differing permutations of applicable languages. Instead, it takes from the first language set on the template itself.

Also removes a duplicate setting of the languages context var from a template, and removes some extra baggage from the LayoutContext that we don't need anymore / can be used from existing data.

Review link: https://ably-docs-web-4327-non--2yikws.herokuapp.com/docs/api/realtime-sdk/push

### DIFF
--- a/src/components/LanguageButton/LanguageButton.tsx
+++ b/src/components/LanguageButton/LanguageButton.tsx
@@ -8,7 +8,7 @@ import { languageInfo } from 'src/data/languages';
 import { LanguageKey } from 'src/data/languages/types';
 import { useLayoutContext } from 'src/contexts/layout-context';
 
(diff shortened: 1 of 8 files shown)


aralovelace (approved):
Nice, also it fixes this issue - https://ably-docs-web-4327-non--2yikws.herokuapp.com/docs/getting-started/react

jamiehenson:
@aralovelace

PR #2564, opened 2025-04-29
Remove use cases filters and update product filters

GregHolmes:
Removed the use case filters from examples page as they're not currently needed.

Updated product filters by removing the products that have no examples for, asset tracking, liveobjects, livesync.


jamiehenson (changes requested):
Looks good but breaks [this](https://github.com/ably/docs/blob/edu-1929-remove-filters/src/components/Examples/ExamplesContent.test.tsx) test, could you update it please

GregHolmes:
Good spot, all updated. Thank you Jamie.

jamiehenson (approved):

jamiehenson:
Nice - will merge in

PR #2572, opened 2025-05-02
Fixed issue where pub-sub-occupancy channel name is hard coded

GregHolmes:
The pub-sub-occupancy examples for both javascript and react were hardcoding the channel name. This PR fixes that.

### DIFF
--- a/examples/pub-sub-occupancy/react/src/App.tsx
+++ b/examples/pub-sub-occupancy/react/src/App.tsx
@@ -5,9 +5,12 @@ import { useEffect, useState } from 'react';
 import { FaEye } from 'react-icons/fa';
 import './styles/styles.css';
 
+const urlParams = new URLSearchParams(typeof window !== 'undefined' ? window.location.search : '');
+const channelName = urlParams.get('name') || 'pub-sub-occupancy';
+
 function Stream() {
   const [occupancySubscribers, setOccupancySubscribers] = useState(0);
-  useChannel('occupancy-example', (message) => {
+  useChannel(channelName, (message) => {
     console.log('occupancy: ', message.data);
     setOccupancySubscribers(message.data.metrics.connections);
   });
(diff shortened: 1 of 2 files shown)


examples/pub-sub-occupancy/react/src/App.tsx
```diff
@@ -5,9 +5,13 @@ import { useEffect, useState } from 'react';
 import { FaEye } from 'react-icons/fa';
 import './styles/styles.css';
 
+const urlParams = new URLSearchParams(typeof window !== 'undefined' ? window.location.search : '');
+const channelName = urlParams.get('name') || 'pub-sub-occupancy';
+
+console.log(channelName);
```
kennethkalmer, line 11: ✂️

examples/pub-sub-occupancy/react/src/App.tsx
```diff
@@ -5,9 +5,13 @@ import { useEffect, useState } from 'react';
 import { FaEye } from 'react-icons/fa';
 import './styles/styles.css';
 
+const urlParams = new URLSearchParams(typeof window !== 'undefined' ? window.location.search : '');
+const channelName = urlParams.get('name') || 'pub-sub-occupancy';
```
jamiehenson, line 9: Does this work okay? Usually you'd access this sort of stuff inside a `useEffect` inside the component that runs on mount (i.e. with a dependency array of []) - that's the Reacty way. But given the simplicity and clientside nature of the app, might be alright.

examples/pub-sub-occupancy/react/src/App.tsx
GregHolmes: Removed!

examples/pub-sub-occupancy/react/src/App.tsx
GregHolmes: It does indeed work.

GregHolmes:
@jamiehenson  @kennethkalmer  when you have a moment could you check over this again please?

jamiehenson (approved):

PR #2575, opened 2025-05-06
[WEB-4259] Fix Changelog description homepage

denissellu:
This PR fixes a bug on the changelog description where the parsing failed, so, i'm now just using the default browser parser
### Before
Screenshot 2025-05-06 at 16 52 51

### After
Screenshot 2025-05-06 at 16 52 37

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
     "@ably/ui": "16.2.1",
     "@codesandbox/sandpack-react": "^2.20.0",
     "@codesandbox/sandpack-themes": "^2.0.21",
-    "@js-bits/dom-parser": "^1.0.3",
     "@mdx-js/react": "^2.3.0",
     "@react-hook/media-query": "^1.1.1",
     "@types/cheerio": "^0.22.31",
--- a/src/components/Homepage/Changelog.tsx
+++ b/src/components/Homepage/Changelog.tsx
@@ -4,7 +4,6 @@ import { Key } from 'react';
 import Badge from '@ably/ui/core/Badge';
 import FeaturedLink from '@ably/ui/core/FeaturedLink';
 import Link from '../Link';
-import parse from '@js-bits/dom-parser';
 
 interface ChangelogFeedItemNode {
   title: string;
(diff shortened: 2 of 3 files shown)


src/components/Homepage/Changelog.tsx
```diff
@@ -84,8 +83,16 @@ const stripHtml = (html: string | null | undefined): string => {
...
     return '';
   }
-  const doc = parse(html);
-  return doc.textContent || '';
+
+  const parser = new DOMParser();
+  const doc = parser.parseFromString(html, 'text/html');
+
+  const firstBodyChild = doc.body.firstChild;
+  if (firstBodyChild && firstBodyChild.nodeName && firstBodyChild.nodeName.toLowerCase() === 'parsererror') {
```
jamiehenson, line 91: `firstBodyChild?.nodeName?.toLowerCase()` ?

src/components/Homepage/Changelog.tsx
```diff
@@ -84,8 +83,16 @@ const stripHtml = (html: string | null | undefined): string => {
   if (!html) {
     return '';
   }
-  const doc = parse(html);
-  return doc.textContent || '';
+
+  const parser = new DOMParser();
+  const doc = parser.parseFromString(html, 'text/html');
+
```
aralovelace, line 89: also the code assumes that doc.body is always defined, but it could be null?

aralovelace (commented):
1 comment and agree with Jamie's comments

package.json
```diff
@@ -43,7 +43,6 @@
     "@ably/ui": "16.2.0",
     "@codesandbox/sandpack-react": "^2.20.0",
     "@codesandbox/sandpack-themes": "^2.0.21",
-    "@js-bits/dom-parser": "^1.0.3",
```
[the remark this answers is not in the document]
denissellu: it was lol

src/components/Homepage/Changelog.tsx
denissellu:
```
  if (!html) {
    return '';
  }
```
I think if `doc.body` is nil, then that code will be nil

aralovelace (approved):
Alt Text

m-hulbert (approved):

PR #2576, opened 2025-05-06
[WEB-4367] Fix column dimensions, default menu expansion behaviour

jamiehenson:
Fixes two things from Matt's docs feedback:
- sorts out the styling classes on the nav outer columns, not sure when this regressed but Tailwind wasn't picking up the width classes (https://ably-real-time.slack.com/archives/C01LR5XFHFX/p1746537594249259)
- left nav doesn't respect and maintain product context when navigating from the homepage to a product page (https://ably-real-time.slack.com/archives/C01LR5XFHFX/p1746540944026369)

Also improves responsive behaviour for the header now that we have a TabMenu section up there taking more space.

Review link: https://ably-docs-web-4367-fix--5jgh9r.herokuapp.com/

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "validate-llms-txt": "ts-node bin/validate-llms.txt.ts"
   },
   "dependencies": {
-    "@ably/ui": "16.2.0",
+    "@ably/ui": "16.2.1",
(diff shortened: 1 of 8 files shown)


kennethkalmer (approved):

m-hulbert (approved):
🕺

PR #2589, opened 2025-05-12
Update kotlin docs : single channel migration

sacOO7:


content/chat/rooms/index.textile
```diff
@@ -91,10 +91,18 @@ blang[javascript,swift,kotlin].
...
-  val options = RoomOptions(presence = presence, typing = typing)
-  val room = chatClient.rooms.get(roomId = "basketball-stream", options = options)
+  val room = chatClient.rooms.get(roomId = "basketball-stream") {
+    typing {
+      heartbeatThrottle = 5.seconds
+    }
+    presence {
+      enableEvents = true
+    }
+    reactions()
```
ttypic, line 101: I think it's better to put some options or remove `reactions`, with new logic, I don't think we should even allow calling `reactions()` without arguments

content/chat/setup.textile
```diff
@@ -243,10 +243,10 @@ val realtimeClient = AblyRealtime(
         clientId = "<clientId>"
     },
 )
-val chatOptions = com.ably.chat.ClientOptions(
-  logHandler = CustomLogHandler(), // Implements com.ably.chat.LogHandler interface
-  logLevel = LogLevel.Debug,
-)
+val chatOptions = buildChatClientOptions {
```
ttypic, line 246: I would've inlined into 

```kotlin
 ChatClient(realtimeClient) {
  logHandler = CustomLogHandler() // Implements com.ably.chat.LogHandler interface
  logLevel = LogLevel.Debug
}
```

ttypic (approved):
Looks good to me

content/chat/rooms/index.textile
```diff
@@ -91,10 +91,17 @@ blang[javascript,swift,kotlin].
   ```
 
   ```[kotlin]
-  val presence = PresenceOptions(enter = false)
-  val typing = TypingOptions(timeoutMs = 5_000) // 5 seconds
-  val options = RoomOptions(presence = presence, typing = typing)
-  val room = chatClient.rooms.get(roomId = "basketball-stream", options = options)
+  val room = chatClient.rooms.get(roomId = "basketball-stream") {
```
splindsay-92, line 94: Are any of these defaults values, do all of them need to be passed to enable functionality? :)

splindsay-92 (changes requested):
It looks like you've missed some updates to various links and text, such as the links in the `Set typing status` section of the typing textile. Please can you update these too to reflect the changes to kotlin? It's probably worth using git blame to see where I made updates, as likely kotlin and swift will need updates there too.

content/chat/rooms/index.textile
sacOO7: Yes, actually we don't need to pass options to enable them. 
If we don't pass anything, defaults are applied

sacOO7:
> It looks like you've missed some updates to various links and text, such as the links in the `Set typing status` section of the typing textile. Please can you update these too to reflect the changes to kotlin? It's probably worth using git blame to see where I made updates, as likely kotlin and swift will need updates there too.

Okay, I will check and update 👍

content/chat/rooms/index.textile
splindsay-92: Sure, but are all of the below default values anyway? If so, would it be okay to just use
`room = chatClient.rooms.get(roomId = "basketball-stream")`? I'm not a kotlin expert, so if you say its best like this, then thats fine :)

examples/chat-typing-indicator/javascript/src/script.ts
```diff
@@ -45,7 +45,7 @@ if (element) {
 }
 
 async function startTyping() {
-  await room.typing.start();
```
splindsay-92, line 48: This change will break the example, they are due to be updated in a separate PR :)

content/chat/rooms/index.textile
splindsay-92: Ah ignore me, it looks like this specific section is dedicated to showing options anyway!

examples/chat-typing-indicator/javascript/src/script.ts
sacOO7: Okay, I will revert this one 👍

PR #2592, opened 2025-05-12
[EDU-1941] Add a guide for building livestream chat applications

m-hulbert:
## Description

This PR adds a guide for building livestream chat application with Ably.

* Adds a new `guides` section within the source. 
* Adds a new 'Guides' section to the navigation in Chat.
* Identifies a few MDX issues that are being resolved in a separate branch. 

Going forwards we should ensure that all guides ask the questions;

- [] Does it have enough visual elements?
- [] Does it include a diagram (if applicable)?
- [x] Does it highlight Ably's USPs around scale etc (and link to appropriate content backing this up when available)

### Checklist

- [x] The PR adheres to the [writing style guide](../writing-style-guide.md) and [contribution guide](../CONTRIBUTING.md).


src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,210 @@
+---
```
AndyTWF, line 1: Not sure if this is just MDX - but seem to be missing the left-side navigation on this page

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+* How best to architect the chat rooms in your application for scale.
+* The core functionality to include in a livestream chat application.
+* Optional additional functionality to enhance your application.
+* Considerations for application monitoring and discontinuity handling.
+* How to keep costs as low as possible when building your application.
+
+## How to architect your rooms <a id="architect"></a>
+
+[Rooms](/docs/chat/rooms) are the core building block for any application you create with Ably Chat. They are used to separate clients into different 'rooms', with users only ever receiving messages from other users in the same room. In the context of livestream chat rooms would be used to separate traffic between different streams, events or games.
```
AndyTWF, line 21: > They are used to separate clients into different 'rooms'

Seems a bit repetitive use of the word "rooms" here (very minor)

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+[Server-side batching](/docs/messages/batch#server-side) means that Ably will group published messages over an interval before publishing them to subscribers.This will help to reduce costs, as streams usually contain a large number of talkative users.
+
+Be aware of the length of the interval you configure, over which messages are batched. If it is too high, then users may notice the delay from messages being sent, however this is normally less of a concern for fast-paced livestreams. An interval of 1-2 seconds is likely sufficient for most livestream chat applications.
+
+#### Message ordering
+
+Messages will be received by users in the order that they are broadcast, by the region that the user is connected to. This may differ slightly to users in other regions, which for a fast-paced livestream is usually acceptable.
+
+If you want to ensure that every user sees every message in exactly the same order, you can use the [global ordering of messages](/docs/chat/rooms/messages#global-ordering) to do this.
```
AndyTWF, line 92: Perhaps say here "... that every user sees messages displayed in exactly the same order"

The code below doesn't affect realtime order, it just changes how you'd order messages in your applications view layer

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+Use the [occupancy](/docs/chat/rooms/occupancy) feature to show the count of users within a stream. You can subscribe each user to the occupancy feature and use the `connections` metric to display the number of users attached to the room.
+
+#### On a stream list
+
+If your application hosts multiple streams, then you likely want to display the number of users watching each one. Use a server to periodically retrieve [occupancy](/docs/chat/rooms/occupancy#retrieve) from all streams via a REST call to display this information on a stream list.
+
+### Send reactions to the room
+
+[Room reactions](/docs/chat/rooms/reactions) are ephemeral reactions that users can send to the room, such as 👍or ♥️. They are not persisted and are used for showing sentiment for what is happening in the stream.
```
AndyTWF, line 117:
```suggestion
[Room reactions](/docs/chat/rooms/reactions) are ephemeral reactions that users can send to the room, such as 👍 or ♥️. They are not persisted and are used for showing sentiment for what is happening in the stream.
```

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+Ably recommends you enable server-side batching if you use room reactions in your stream. This is because the highest volumes of reactions are normally sent in a short space of time, such as when a team scores in a game. Server-side batching will reduce the costs associated with the reactions occurring over that short time period.
+
+### Moderate your stream
+
+Keeping your chat stream free of abuse and toxic behavior is paramount to providing a safe and enjoyable experience for your users.
+
+Use AI models to vet content in realtime and keep your stream free of abusive content. Ably recommends using an 'after-publish' form of moderation for livestream chat. This is where messages are briefly visible to users, and then removed after moderation has been run against them.
```
AndyTWF, line 125:
```suggestion
Use AI models to vet content in realtime and keep your stream free of abusive content. Ably recommends using an 'after-publish' form of moderation for livestream chat. This is where messages are briefly visible to users, and then removed after moderation has been run against them. The reason for this is to avoid the latency penalty of moderating inline: in busy rooms, users likely will not have time to comprehend a message before moderation removes it.
```

(Or some better words to that effect!)

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+Optional functionality includes features that enhance the user experience of livestream chat applications, but aren't required.
+
+### Edit and delete messages
+
+Whether users can [edit](/docs/chat/rooms/messages#edit) or [delete](/docs/chat/rooms/messages#delete) the messages that they've previously sent is an optional feature that you can include in your livestream application. Its usefulness will depend on the rate at which messages are published.
+
+### Message history
+
+It's possible to retrieve previously sent messages from [history](/docs/chat/rooms/history), as Ably stores chat messages for 30 days. There are two uses for history that can be useful to livestream chat:
```
AndyTWF, line 153: Might be worth mentioning 30 days can be extended on request?

src/pages/docs/guides/chat/build-livestream.mdx
m-hulbert: Yes, it should be fine now I've rebased it with the MDX updates.

AndyTWF (approved):

mattheworiordan (changes requested):
Sorry for the slowness in my response to this PR. This is a great step forward and I think we're not far off.

I have thoughts on this in different contexts so break this down below. 

## Conceptual mode

### Purpose and intended outcome

We need to be very clear on what the outcome of a guide is, which I don't think is evident when I read this. For example, the intro paragraph states "This guide explains how to do ...." which is fine, but that's a tutorial. 

Compare this to Neon, and https://neon.tech/docs/use-cases/database-per-user, where the intro paragraph starts with "With its serverless and API-first nature, Neon is an excellent choice for building database-per-user applications (or apps where each user/customer has their own Postgres database). Neon is particularly well-suited for architectures that prioritize maximum database isolation, achieving the equivalent of instance-level isolation. This guide will help you get started with implementing this architecture." and their agent page https://neon.tech/docs/use-cases/ai-agents, which starts with:
"title: Use Neon as the Postgres backend for your agents
"context: AI agents can now provision infrastructure, including databases. With AI agents already creating databases every few seconds, they are poised to manage a significant portion of the web's infrastructure in the future — ..."
"proof point: and, like developers, AI agents love working with Neon: Replit partnered with Neon to back Replit Agents, which are already creating thousands of Postgres databases per day. [Learn more](https://neon.tech/blog/looking-at-how-replit-agent-handles-databases)."
"benefits: What makes Neon a good database for AI Agents... " whihc goes on to list 4 key benefits.

We know that a lot of these pages are leading users from marketing pages on these solutions, so selling users on why Ably is not just able to deliver this, but the right choice, is paramount. I realise this is docs, but that doesn't mean we can't tell people why we're a good solution.

The purpose for me is:
- Be punchy about why Ably is a good fit
- Provide benefits at the get go
- Move onto let us show you how you architect your app for this use case
- Provide sufficient depth to be credible
- Reinforce our USPs, the 4 pillars and cost efficiencies at scale

### Reiterate the benefits & USPs throughout 

In the section on "Ably can scale to handle millions of users in a single room. Theoretically this means that you can use a single room per stream or event and have all of your users connect to it. The main consideration for how to architect your rooms is the end-user experience that is appropriate for your specific livestream chat application.", I don't think we've really tackled the benefits here in a way that is obvious to users.

The benefits are, for example: 

How Ably helps you scale:

- Ably supports a limitless number of concurrent rooms. See [how Ably scales channels horizontally](https://example.com). This means you can have any number of concurrent live streams and one to one side chats running for your event, without any degredation in performance
- Ably supports a limitless number of users in  room. This means you can support millions of users in your livestream, who can be active (post messages and reactions) or passive (just watching) participants. See [how Ably scales subscribers on a channel](htpts://example.com) with zero impact on latency. Ably can deliver updates to millions of your users in milliseconds.

etc...

## Aesthetics and readability

- We need visuals to bring these guides to life. At the least we can get a visual to represent the use case at the start, see https://liveblocks.io/docs/guides/how-to-add-users-to-liveblocks-comments for example. 
- Diagrams help convey concepts quickly. I think simple diagrams where appropriate are essential for readability, such as one that shows how messages are routed through each room.

## Depth and Breadth

I think it's critical that we provide sufficient depth and substance to build credibility with our deeply technical audiences. That however does present a breadth problem because as we go deeper and provide substance, this in turn drives content to be longer, and there is a risk these guides become unmanageable. There is no silver bullet here, but where we can move concepts out into perhaps an Advanced Concept page, or simply move it into a separate guide, we may want to do that. I think that's the easier problem to tackle, the harder problem is the depth and substance. At present, I am afraid I don't think this guide is good enough.

For example:

1) "Accept the message rate -> For certain use cases, it's perfectly acceptable that not every message can be read by every user. In these instances the general sentiment returned from reading a proportion of messages is sufficient."

- Ok, but what does this mean exactly to the end user?
- Why can't batching be used?
- Can we visually show a diagram?
- What message rates are acceptable by Ably (specifics), and what message rates can humans perceive (specifics)?

2) "If it's important for your use case that users can read the majority of messages being written, then you can suppress the number of messages that each user can send within a set time period. This will enable you to control the maximum rate of messages, irrespective of how many users join each stream."

You provide a code sample, but don't specify if that is client-side or server-side.
The code doesn't actually provide a rate limiter, it's an interval limiter
The statement "you can control the max rate irrespective of how many users join each stream" is not valid if this is done client side, and I have no idea how one would do this server side, and this code certainly doesn't tell the user how to do that.

3) "If the message rate only peaks to high levels during exciting moments on stream, such as a point being scored in a sporting event, then this can be acceptable for some use cases. In these instances then features such as room-wide reactions help to provide the general sentiment of the chat when users can't read every message."

I am not sure what this is trying to tell the user.

- If they want to use room-wide reactions, then they should and we should explain this, but the context of this section is about messages.
- How does the rate limit get affected by reactions vs messages?
- Why is batching not mentioned here?
- What is the benefit of them using Ably then, what do we offer that makes Ably the right choice for these problems?

### Don't explain everything, but lead them to an answer

Examples:

1) "Users need to assume an identity when using Ably Chat. If your application requires a login then this can be a name, or identifier, read from your identity provider. If your application allows anonymous users, then use a mechanism to randomly assign users a unique client ID when issuing a JWT."

The only link here takes them to JWT token auth guide at https://ably-docs-edu-1941-live-g8qh3l.herokuapp.com/docs/auth/token#jwt.  I am not sure someone following the guide would now understand what to do next. Do we not have a resource we could take them to that says see how you approach auth at X and how to identify users?

For a new user reading this, I am not sure we can assume they now know what to do next.

2) "If a user is required to log in, use their unique user identifier as the client ID when they authenticate with Ably. This will enable you to easily associate their usage in your application with their activity over Ably."

- Should we not link to docs on how to specify client ID?  `clientId` is not shown as an attribute, so I think for a non-Ably user, this is not obviously an attribute of a token being issued.
- You state it will easily associate their usage in your application with their activity over Ably, however I honestly don't know how?  What is meant by this exactly?

> Be aware of the length of the interval you configure, over which messages are batched. If it is too high, then users may notice the delay from messages being sent, however this is normally less of a concern for fast-paced livestreams. An interval of 1-2 seconds is likely sufficient for most livestream chat applications.

1-2 seconds is a material delay and not something we'd recommend. That's on the highest end of what is recommended and will add meningful UX delays.

However, regardless, we're just not explaining how this works, even at a basic level. And we need specificity for developers to build credibility.

For example, we could say if 1,000 users are posting once per minute, we'd expect a message rate of 16 per second, which results in 16 messages being broadcast to 1k users every second, or 960k per minute.

If batching is enabled with 500ms intervals, then users will at most see a delay of 500ms from publish to receiving messages, but the message rate will half to 2 per second, each with an average batch size of 8 published messages. The message rate will drop to 120k per minute. This will effectively reduce the cost of broadcasting these messages by 8x.  

What's important however, is that Ably provides a means to deliver linear cost increases as rooms get larger with batching. For example, if we take the above example, where we have 1000 users each publishing once per minute, and the number of users increases, the message rate climbs to 32 per second to 2,000 sers which equates to 3.84m messages per minute (4x the 1,000 user rate). With 500ms batching enabled, the message rate grows lienearly with 2 messages published per second to 2,000 users, which equates to 240k per minute, exactly double the number at 1,000 users. Ably's unique batching feature ensures you do not expereince exponentiual growth in costs.

> Messages will be received by users in the order that they are broadcast, by the region that the user is connected to. This may differ slightly to users in other regions, which for a fast-paced livestream is usually acceptable.

- Why are we not linking to docs to show the extreme lengths we have gone to to ensure message ordering?
- What does it mean "this may differ slightly to users in other regions"?  Ordering is guaranteed, by region. This is an incredibly vague statement which definitely does not convey expert confidence.
- Where is the "benefit" statement of Ably?

> If you want to ensure that every user sees messages displayed in exactly the same order, you can use the [global ordering of messages](https://ably-docs-edu-1941-live-g8qh3l.herokuapp.com/docs/chat/rooms/messages#global-ordering) to do this.
> `sortedMessages.sort((a, b) => (a.before(b) ? -1 : 1));`

I am not quite sure where to start with this one, but suggesting they get global orderign by sorting messages, without explaining a) how we have gone to extreme lengths to guarantee ordering, b) why global ordering is not really a thing because you can enforce it retrospectively, but if you try and do it in realtime, it's impossible because you'd have to wait for an indeterminate amount of time for each region to be in sync, c) we're a distributed system and their users are global, so latency between regions is not just expected, but not something any system can fix until we can move bits faster than the speed of light.

This does not hold up to developer scrutiny, opens up so many questions, and undermines our guarantees.

> This can be extended for longer periods upon request

I thought we were trying to tell people we support longer periods without making it sounds like it's a special feature. This should be "The storage period can be configured for longer periods."  When they come to configure it, we can tell them to get in touch, but we don't need to do this here, it just reeks of "this is a special feature, you will need to ask sales and pay X".

....

I'm stopping reviewing all sections as I think the above is sufficiently illustrative of the direction.

## Tone

We have a style guide for our tone of voice, and I think we need to be clear on this when writing. Definitely worth looking at, but I think we need a guide now with at least some bullets, such as:

1) Expert confidence - we have built something great, we should tell people why it's the right choice for them, and explain it to a tedchnical audience.
2) Proof points - we don't just make claims, we support that with evidence i.e. we scale to millions, here is how we do that, we offer X latency, here is how we do that and here are our latency metrics.
3) Accurate - we are accurate in what we say, we are the experts
4) Value focussed - we recognise that it's our job to explain the value of the Ably platform in what we write. The reader should clearly understand how Ably delivers value in the problems they are trying to solve.

## Other stuff (or quick comments on sections where I haven't had time to go into detail)

- We should ensure we end with a panel like Vercel's "Didn't find what you need, get help" etc. I know we say if you need help, contact us, but I think that;s somehwat different to tell us if you need more info etc. Subtle I know.
- We're still linking people to our FAQs and Blog for answers? 
- On occupancy, we tell them to use REST to get the occupancy server-side, but a) this would be interpreted as using a REST HTTP endpoint, not the REST SDK, b) they need to know which channels are active, so shouldn't we also recommend an occupancy integration and explain it's benefit?
- You mention `getPreviousMessages` method, yet do not link to it, so the user has to now trawl all of the APIs to find it.
- Server-side message batching key benefit is linear growth as opposed to exponential growth of message volumes as rooms get busy. This point is entirely lost in that cost optmisation section,.

## How do we apply this thinking to all guides?

We need to codify this feedback, and what makes a good guide, and write that up. I'm very happy to help with this. Until we can clearly explain what the goals of each guide is, what tone should be used, what level of specificity and credibilty is needed for our tech audiences, then I fear getting others to write guides will be hard. I suspect we could write this quite quickly before making any changes to this PR, and then reflect on whether the guide accurately addresses the feedback here and would serve us well for future guides.

In addition, I think we could easily write a prompt that helps authors validate the goals of these guides. I tried this myself and [produced the following](https://chatgpt.com/share/682c2cc0-2f20-8008-9e40-0edfee0f2381). It's not good enough, but I believe with more iterations, we could get this to be way more efficient, objective, and useful, and perhaps it could even help make changes once we develop the prompt as we role it out with each guide, which could save humans doing reviews to a large degree.

Shout if you want to discuss.

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+  - **Sharding** is one approach to this problem - separating users into smaller groups so that the chat is more manageable. This can be done randomly, or using some predefined scheme, e.g. language, geographical region.
+  - You can also simply _drop_  a proportion of messages on the client to keep it readable. If you're not worried about everyone seeing everything, this is a simple approach.
+  - Of course, you can also rate-limit how often users send messages. This controls the rate of messages, but can negatively impact the user experience if they're not able to respond when they want to.
+- Do you need to support both public and private chat? If you do, you might want one big room for public chat and a separate room for private messages.
+
+Ably gives you the flexibility to design for your users, not your infrastructure.
+
+## Message throughput and rate management
+
+Livestream chat is defined by high message rates and unpredictable spikes. Ably is engineered to handle these loads without degradation.
```
m-hulbert, line 56: Good place to link to more architectural content backing this up.

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+- **Proven at scale:** Ably delivers over 500 million messages per day for customers, with individual rooms supporting over 1 million concurrent users.
+- **Server-side batching:** Reduce costs and network overhead by grouping messages before delivery. For example, batching messages over a 500ms interval can reduce message costs by up to 8x in high-traffic streams.
+- **Rate limiting:** Control the flow of messages per user to maintain readability and prevent spam. Choose between per-user rate limits, global throttling, or adaptive algorithms.
+
+CODE HERE: Example of implementing per-user rate limiting logic
+
+PLACEHOLDER: Visual showing message spikes and how batching smooths delivery/costs
+
+**Decision point:**
+- What is the maximum message rate you want users to experience?
```
m-hulbert, line 67: Should this touch on a consistent or average rate? E.g. a reactions spike would influence this heavily.

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+- How will you balance real-time delivery with cost efficiency?
```
m-hulbert, line 68: I'm not sure if this is a "how will you" type question. Is this not more relevant when discussing server-side batching below?

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+How you authenticate is also key. To balance security and experience, you want short-lived tokens that can be easily revoked if a users is misbehaving or needs their permissions changed, but also avoiding the overhead of changing permissions every 2 minutes. In production apps, we strongly recommend you don't use API keys for client-side authentication (server-side is OK) - these are long-lived and much more difficult to revoke in a hurry.
+
+With Ably Chat, authentication is best achieved using JSON Web Tokens (JWTs). These are tied to a particular clientID and come with a set of [capabilities](https://ably.com/docs/chat/setup#authentication) that control what a client can and cannot do. Ably's SDKs handle the timing and process of requesting a new token for your, refreshing it when it expires. All you need to do is provide a server-side endpoint that can generate the JWT for the client. This allows clients to use your existing authentication systems or user sessions to generate their Ably token.
+
```
m-hulbert, line 77: I think a code example of setting a `clientId` using JWTs and using an anonymous one would be useful here, in addition to linking out to the authentication docs.

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+**Decision point:**
+- What is the maximum message rate you want users to experience?
+- How will you balance real-time delivery with cost efficiency?
+
+## Authentication
+
+Authentication in livestream chat is a design decision. In some cases, you'll want to tie participants to their identity in your application. In others, anyone can join and participate. Ably Chat supports both of these scenarios - you don't need to go through the hassle of setting up users in Ably and trying to keep them in sync with your application. Every client has a clientID. If you want random users, just choose a random string. If you want to tie it to your application, then use the UUID of the user or some other well-known identifier.
+
+How you authenticate is also key. To balance security and experience, you want short-lived tokens that can be easily revoked if a users is misbehaving or needs their permissions changed, but also avoiding the overhead of changing permissions every 2 minutes. In production apps, we strongly recommend you don't use API keys for client-side authentication (server-side is OK) - these are long-lived and much more difficult to revoke in a hurry.
```
m-hulbert, line 74: I think we should go above "strongly recommend" and just state don't use them on clients.

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+With Ably Chat, authentication is best achieved using JSON Web Tokens (JWTs). These are tied to a particular clientID and come with a set of [capabilities](https://ably.com/docs/chat/setup#authentication) that control what a client can and cannot do. Ably's SDKs handle the timing and process of requesting a new token for your, refreshing it when it expires. All you need to do is provide a server-side endpoint that can generate the JWT for the client. This allows clients to use your existing authentication systems or user sessions to generate their Ably token.
```
m-hulbert, line 76: `/docs/` rather than fully qualified. 

Similarly, "what a client can and cannot do" should probably be more tailored to livestream here.

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+
+## Moderation: Protecting your community
+
+Effective moderation is essential for livestream chat to maintain a safe and engaging environment, prevent spam and inappropriate content, and ensure compliance with platform guidelines. While human moderators are valuable, they cannot scale to handle the high message volumes of popular livestreams - a single stream can generate thousands of messages per minute. Automated moderation solutions are therefore crucial for real-time content filtering at scale.
+
+Ably supports both inline (before-publish) and after-publish moderation workflows, making it easy to integrate with AI-powered moderation tools.
```
m-hulbert, line 83: Do we need to describe this as both "inline" and "before-publish"? 

"Ably supports moderating messages before they're published to the livestream, and retrospectively after they're published" for e.g.

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+When implementing chat moderation, consider these essential factors:
+
+1. **Community standards**
+   - What level of moderation is appropriate for your audience?
+   - How will you handle different types of content? For example:
+     - **Hate speech and harassment:** Detecting discriminatory language, threats, or targeted abuse
+     - **Inappropriate content:** Flagging adult content, violence, or graphic material
+     - **Toxicity:** Measuring overall message sentiment and hostility
+   - AI moderation systems typically classify text according to categories similar to those above, usually with a severity and confidence level. If you have a provider in mind, what thresholds would you set for your chat?
+   - Do you need different rules for different user groups?
```
m-hulbert, line 99: This is all relevant when building a livestream, but where does Ably fit into any of the solutions to these?

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+2. **Technical requirements**
+   - How latency-sensitive is your chat?
+     - Inline AI moderation can add up to 100ms to message delivery
```
m-hulbert, line 103: Worth mentioning that you'd need to potentially factor in batching latency here as well?

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+   - Will you use a third-party moderation service or build your own?
+
+3. **Community-specific approaches**
+   - **Gaming streams:** More relaxed rules for casual language and humor
+   - **News and current affairs:** Strict moderation of misinformation and hate speech
+   - **Educational content:** Emphasis on respectful discussion and spam prevention
```
m-hulbert, line 109: Are some of these points relevant to the build or technical implementation?

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+### How Ably helps keep your chats safe
+
+Ably's integration with moderation providers allows you to customise the moderation policy on a per-chat basis, allowing you to implement different rules and sensitivity levels for different types of content and communities. You can even configure what happens if the moderation provider can't be reached - for example, publishing the message anyway. Get up and running with **out-of-the box integrations in minutes** and try out many providers for free. Ably takes care of all of the complex bits of invoking the moderation provider, handling the results and failure conditions.
+
+Of course, you may already have built moderation in other parts of your organisation and want to re-use it. Ably's got you covered! "Bring your own" moderation rules allow you to connect your chat rooms to your own infrastructure via popular transports including serverless functions, queues and webhooks. By implementing simple APIs, you can integrate your own moderation solution seamlessly with Ably.
```
m-hulbert, line 115: Happy to suggest some amendments here, but this has gone a bit salesman rather than assisting in implementing IMO.

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+CODE HERE: Example of handling moderation failures in Ably Chat
+
+PLACEHOLDER: Diagram showing moderation flow (user → Ably → moderation → delivery)
+
+## Occupancy: Give volume to your livestream
+
+Ably's occupancy features provide you with realtime statistics of how many users are engaging with your live stream. These can be used to hype up your streams and draw in users to be part of the experience.
```
m-hulbert, line 123: This doesn't really explain to me what occupancy does, or how this would help with hype.

I wonder if an image would help with explaining what occupancy can do?

Also we're describing them as statistics here and metrics below.

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+- **Within a stream:** Show the number of active users in real time.
+- **Across streams:** Aggregate occupancy data for stream lists or dashboards.
+
+CODE HERE: Example of retrieving occupancy metrics
+
+## Reactions, typing indicators, and ephemeral events
```
m-hulbert, line 130: I'm not sure "ephemeral events" is something that would catch my eye, and they're essentially the same thing as reactions and indicators here.

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+title: "Guide: Building livestream chat at scale with Ably"
+meta_description: "Architecting livestream chat with Ably: performance, reliability, and cost at scale. Key decisions, technical depth, and why Ably is the right choice."
+meta_keywords: "livestream, chat, scalability, Ably Chat, realtime messaging, dependability, cost optimisation"
+---
+
+Ably Chat is purpose-built for livestream chat at scale. If you need to deliver a seamless, high-throughput chat experience to thousands or millions of users—without sacrificing performance, reliability, or cost—Ably is the proven choice. This guide explains the architectural decisions, technical challenges, and unique benefits of building livestream chat with Ably, so you can deliver the best possible user experience, every time.
+
+## Why Ably for livestream chat?
+
+Ably is trusted by organisations delivering chat to millions of users in real time. Our platform is engineered around the four pillars of dependability:
```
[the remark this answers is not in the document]
AndyTWF: Have changed this up slightly

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+- **Performance:** Ultra-low latency messaging, even at global scale.
+- **Integrity:** Guaranteed message ordering and delivery, with no duplicates or data loss.
+- **Reliability:** 99.999% uptime SLA, with automatic failover and seamless reconnections.
+- **Availability:** Global edge infrastructure ensures users connect to the closest point for optimal experience.
+- **Cost at scale:** Server-side batching, efficient connection management, and flexible pricing keep costs predictable—even as your audience grows.
+
+PLACEHOLDER: Diagram showing Ably's four pillars of dependability and global infrastructure
```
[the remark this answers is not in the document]
AndyTWF: I think we could take an image from the architecture pages

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+Delivering chat events in real time is key to an engaging experience. Ably's serverless architecture eliminates the need for you to manage websocket servers, automatically scaling to handle millions of concurrent connections without provisioning or maintenance. All of the edge-cases around delivery, failover and scaling - leave that to us and focus on what's important to you and your users.
```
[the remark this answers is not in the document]
AndyTWF: I think I meant "events" in the more broad sense to include room reactions for example, but messages gets the point across

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+## What this guide covers
+
+- Key architectural decisions for livestream chat
+- How to design for scale and user experience
+- Approaches to moderation, rate limiting, and batching
+- How Ably's features solve real-world challenges
+- Proof points and best practices for production
```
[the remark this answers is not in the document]
AndyTWF: Reworded

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+## Architecting your chat rooms: UX first, scale always
+
+The most important decision you can make is the experience you want users to have. It determines the look and feel of the livestream and ultimately the impression your users leave with. The most important architectural decision therefore is how you structure your chat rooms. With Ably, you are not limited by technology - only by the user experience you want to deliver. With Ably you can deliver experiences that are:
```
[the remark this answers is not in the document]
AndyTWF: Removed

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+- **Single massive room:** Ably seamlessly supports millions of users in a single room, by automatically distributing websocket connections across multiple servers to ensure endless connection scalability. This is ideal for global events or streams where everyone participates together.
+- **Many concurrent rooms:** You can run thousands of rooms in parallel, each with their own audience and message flow. Not only do connections scale horizontally, but Ably uses [consistent hashing](https://ably.com/blog/implementing-efficient-consistent-hashing) to evenly distribute the management of chat rooms across instances, allowing the number of rooms to grow without limit and without causing load spikes when you do.
```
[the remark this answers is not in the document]
AndyTWF: Ah yes good shout, done

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+  - Alternatively, you may want to make use of translation services, that can translate messages on the fly to the language of the user.
+- Do you want an environment where everyone has time to read every message, or the chat to go wild with hype when something exciting happens?
+  - At anything more than 5-10 messages per second, the chat becomes unreadable for the audience.
+  - **Sharding** is one approach to this problem - separating users into smaller groups so that the chat is more manageable. This can be done randomly, or using some predefined scheme, e.g. language, geographical region.
+  - You can also simply _drop_  a proportion of messages on the client to keep it readable. If you're not worried about everyone seeing everything, this is a simple approach.
+  - Of course, you can also rate-limit how often users send messages. This controls the rate of messages, but can negatively impact the user experience if they're not able to respond when they want to.
+- Do you need to support both public and private chat? If you do, you might want one big room for public chat and a separate room for private messages.
```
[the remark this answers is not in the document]
AndyTWF: Some fan-engagement experiences have a group chat for the main event but also allow DMs

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+- **Single massive room:** Ably seamlessly supports millions of users in a single room, by automatically distributing websocket connections across multiple servers to ensure endless connection scalability. This is ideal for global events or streams where everyone participates together.
+- **Many concurrent rooms:** You can run thousands of rooms in parallel, each with their own audience and message flow. Not only do connections scale horizontally, but Ably uses [consistent hashing](https://ably.com/blog/implementing-efficient-consistent-hashing) to evenly distribute the management of chat rooms across instances, allowing the number of rooms to grow without limit and without causing load spikes when you do.
+
+PLACEHOLDER: Diagram comparing single-room vs. multi-room architectures
+
+In making this decision, there are some **key considerations**:
```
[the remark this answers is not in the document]
AndyTWF: Added

src/pages/docs/guides/chat/build-livestream.mdx
AndyTWF: Done

src/pages/docs/guides/chat/build-livestream.mdx
AndyTWF: Yes added

src/pages/docs/guides/chat/build-livestream.mdx
AndyTWF: Made stronger

src/pages/docs/guides/chat/build-livestream.mdx
AndyTWF: Yes

src/pages/docs/guides/chat/build-livestream.mdx
AndyTWF: Added

src/pages/docs/guides/chat/build-livestream.mdx
AndyTWF: Have reworded slightly to make it less salesman-y

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+Ably's occupancy features provide you with realtime statistics of how many users are engaging with your live stream. These can be used to hype up your streams and draw in users to be part of the experience.
+
+- **Within a stream:** Show the number of active users in real time.
+- **Across streams:** Aggregate occupancy data for stream lists or dashboards.
+
+CODE HERE: Example of retrieving occupancy metrics
+
+## Reactions, typing indicators, and ephemeral events
+
+Enhance engagement with room reactions, typing indicators, and more. Immortalise the winning goal, the controversial decision and the final whistle by allowing users to express en-masse with a flood of emotions. Ably's Chat SDK provides simple interfaces for adding these features to your livestream.
```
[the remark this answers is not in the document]
AndyTWF: Reworded

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+- **Room reactions:** Broadcast sentiment (e.g., 👍, ❤️) to all users. Not persisted, optimised for scale.
+- **Typing indicators:** Show when users are composing messages.
+
+CODE HERE: Example of sending a room reaction
+
+## Message history and context
+
+Some users may join your livestream late, or miss important information. Ably stores chat history for 30 days by default, with options to extend. Retrieve previous messages to provide context for users joining mid-stream.
```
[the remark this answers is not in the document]
AndyTWF: Discussed the design decision more

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+- **Contextual history:** Show recent messages on join.
+- **Deep history:** Allow users to scroll back through older messages in batches.
+
+CODE HERE: Example of fetching message history
+
+## Handling network disruption
+
+Network disruption happens - mobile internet loses signal or someone drives through a tunnel. Ably's SDKs are designed with this in mind.
```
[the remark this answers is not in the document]
AndyTWF: Update wording and referenced arch content

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+It's incredibly rare, but sometimes a client might lose connection to a particular data center. Ably operates in multiple data centers around the world with multiple fallback regions available. If a client can't reach the nearest data center, it will try the next one until the connection is re-established, ensuring minimal downtime and that network issues don't disrupt the experience that you are trying to build.
+
+## Priced for scale
+
+Livestreams can be a high-churn affair, with many users coming and going throughout. In Monthly Active Users (MAU) pricing models, this often leads to bills that far outweigh the actual usage. Ably offers a competitive pricing model that is based on consumption, with discounted rates for higher usage. So you won't be billed for users that come and go, only for whilst they're engaging with your experience.
```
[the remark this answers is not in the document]
AndyTWF: Yeah, I'm going to add a sentence to that effect

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+A full breakdown of pricing, as well as a cost estimator for your scenario, can be found on the [pricing page](https://ably.com/pricing?chat-type=sales).
+
+## Cost optimisation at scale
```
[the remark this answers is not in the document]
AndyTWF: Done

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+Ably is designed to keep costs predictable, even as your audience grows.
+
+### Server-side message batching
+
+In large-room livestream chats, the majority of the cost comes from delivering messages to all of the participants. It also limits how many messages per second can be processed.
```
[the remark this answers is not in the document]
AndyTWF: Used a slightly smaller spoon

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+That's an **8x reduction** in the number of messages sent out to clients per minute.
+
+#### Cost grows linearly as you do
+
+The key takeaway is what happens once you increase the number of users to 2000, the inbound message rate goes up to 32/second, which unbatched is 4 million outbound messages/minute. With 500ms batching, the outbound message rate grows to 240,000 (double the amount for 1000 users). Therefore, with Ably's unique server-side batching feature, **your costs grow linearly as you scale**.
+
+### Aggressive connection management
+
+When a client abruptly disconnects from Ably, there is a 2 minute delay before the connection is cleaned up on the server, to allow the client to resume the connection from where it left off. When you're finished with an Ably connection, be sure to call the `close()` method on the client library to gracefully shut down the connection.
```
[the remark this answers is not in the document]
AndyTWF: Done

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+All Ably SDKs also perform "heartbeats" with the server to allow detection of dropped or disrupted connections. The default interval for this is 15 seconds. By adjusting the heartbeat interval, you can control how quickly a connection is deemed to have dropped and therefore reduce the amount of time connections remain open for.
+
+- **Server-side batching:** Dramatically reduces message costs in high-throughput scenarios.
+- **Aggressive connection management:** Configure heartbeat intervals and close idle connections to minimise resource usage.
+- **Flexible pricing:** Choose the package that matches your expected usage. See [Ably pricing](/docs/pricing#packages) for details.
```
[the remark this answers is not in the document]
AndyTWF: Think this is leftover from an old version, removing.

AndyTWF (approved):

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,354 @@
...
+  * In this situation, Ably's [horizontal and vertical scalability](/docs/platform/architecture/platform-scalability) means you can have one big room for public chat and many smaller rooms for private conversations.
+
+Ably gives you the flexibility to design for your users, not your infrastructure.
+
+## Message throughput and rate management
+
+Livestream chat is defined by high message rates and unpredictable spikes. Ably [is engineered](/docs/platform/architecture/platform-scalability) to handle these loads without degradation.
+
+* **Proven at scale:** Ably delivers over 500 million messages per day for customers, with individual rooms supporting over 1 million concurrent users.
+* **Server-side batching:** Reduce costs and network overhead by grouping messages before delivery. For example, [batching messages](#server-side-batching) over a 500ms interval can reduce the number of outbound messages by up to 8x in high-traffic streams.
```
mattheworiordan, line 58: How did you get to 8x here?

If you are publishing 100 messages a second to 100 people, that's 10k messages a second.
Batch of 500ms reduces that to 2 messages a second to 100 people, i.e. 200 messages a second, a 50x reduction. 
If users expected 1,000s of messages a second, and batch to 500ms, that's a 500x saving. 
You are way too conservative here IMHO>

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,354 @@
...
+    claims,
+    "{{ API_KEY_SECRET }}",
+    { header: header }
+)
+
+console.log('JWT is: ' + ablyJwt);
+```
+</Code>
+
+How you authenticate is also key. To balance security and experience, you want short-lived tokens that can be easily revoked if a users is misbehaving or needs their permissions changed, but also avoids the overhead of issuing new tokens every 2 minutes. **In production apps, you should not use API keys for client-side authentication**. You can use them server-side, but as they are long-lived they are much more difficult to revoke in a hurry.
```
mattheworiordan, line 110: Why are API keys more difficult to revoke?  You can remove them immediately and that makes them invalid there and then. With tokens, unless you use the revoke fature, the lag can be the renewal time. This statement does not feel correct. The point is client-side tokens are for untrusted clients, and can be revoked or blocked from renewal. Too frequently renewing tokens creates second order load problems with the token issuer, and that needs to be considere

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,354 @@
...
+Ably's [occupancy](/docs/chat/rooms/occupancy) feature provides you with realtime metrics of how many users are engaging with your livestream. The metrics are based on the number of connections to the Ably service and can be delivered to clients in realtime. These can be used to build excitement in your streams, showing users how many others are experiencing it with them.
+
+* **Within a stream:** Show the number of active users in realtime.
+* **Across streams:** Aggregate occupancy data for stream lists or dashboards.
+
+## Reactions and typing indicators
+
+Enhance engagement with [room reactions](/docs/chat/rooms/reactions) and [typing indicators](/docs/chat/rooms/typing). Celebrate the winning goal or protest the controversial decision by enabling users to express sentiment en-masse.
+
+* **Room reactions:** Broadcast sentiment (e.g., 👍, ❤️) to all users.
```
mattheworiordan, line 185: Odd the order of this is he opposite of the example below.

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,354 @@
...
+// This method is self-throttling, to prevent large numbers of messages being sent
+// and keep costs down
+await room.typing.keystroke();
+
+// You can let the typing indicator stop naturally, or explicitly stop it
+await room.typing.stop();
+```
+</Code>
+
+Room reactions are a great way to enhance engagement in a livestream and enable users to quickly express sentiment to key moments in the stream. They are not persisted, and are optimized for scale. Ably's platform is designed to support the spikey traffic of livestream chat, ensuring that you don't encounter any performance issues during pivotal moments.
```
mattheworiordan, line 209: Are we not going to mention message reactions?

mattheworiordan, line 209: But they will be persisted imminently no?  Should we not just build for hat?

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,354 @@
...
+
+## Priced for scale
+
+Livestreams can be a high-churn affair, with many users coming and going throughout. In Monthly Active Users (MAU) pricing models, this often leads to bills that far outweigh the actual usage. Ably offers a competitive pricing model that is based on consumption, with discounted rates for higher usage. You won't be billed for users that come and go, only for the time they spend engaging with your application.
+
+If you have an audience that is relatively consistent, for example a number of people sign up for an event and then stay for the duration, Ably also offers a competitive MAU pricing model in addition to the traditional consumption-based model for these scenarios.
+
+A full breakdown of pricing, as well as a cost estimator for your scenario, can be found on the [pricing page](https://ably.com/pricing?chat-type=sales).
```
mattheworiordan, line 269: Why chat-type=sales?

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,354 @@
...
+* Total outbound message rate: **1,000,000 outbound messages/minute** or 16,667 messages/second
+
+#### With 500ms batching:
+
+* Messages are grouped into 500ms batches, so users will see a maximum of 500ms delay from publish to receiving
+* In 500ms, we expect: (1000 messages/minute) × (0.5 seconds/60 seconds) ≈ 8.3 messages per batch
+* Each batch is delivered once to each user
+* Over the space of a minute, there are now only **120,000 outbound messages/minute** or 2000 messages per second
+
+That's an **8x reduction** in the number of messages sent out to clients per minute.
```
mattheworiordan, line 292: I'd go with a senario that shows a higher numger, and perhaps explain that as the number of participants grows, theis reduction increases (although perhaps that point is covered sufficiently below)

src/pages/docs/guides/chat/build-livestream.mdx
AndyTWF: It was based on the example in the batching section, but I agree that it paints a far too conservative picture - will reword

src/pages/docs/guides/chat/build-livestream.mdx
AndyTWF: Reworded

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    console.error('Message failed to send:', error);
+});
+```
+</Code>
+
+## Occupancy: Give volume to your livestream
+
+Ably's [occupancy](/docs/chat/rooms/occupancy) feature provides you with realtime metrics of how many users are engaging with your livestream. The metrics are based on the number of connections to the Ably service and can be delivered to clients in realtime. These can be used to build excitement in your streams, showing users how many others are experiencing it with them.
```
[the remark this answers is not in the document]
AndyTWF: Added

src/pages/docs/guides/chat/build-livestream.mdx
AndyTWF: Perhaps less useful in the large single-room case, but I've added this as it relates to 1:1 DMs and small side-groups that we see in some FE experiences

AndyTWF: At the moment, we don't persist room reactions - they're ephemeral messages. There's a long-term feature consideration about making them persisted.

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,354 @@
...
+```
+</Code>
+
+## Handling network disruption
+
+Network disruption happens - mobile internet loses signal or someone drives through a tunnel. All of Ably's SDKs are designed with this in mind, so that you don't have to handle complicated reconnection logic.
+
+Every SDK instance keeps track of where it's at in the message stream. If the connection is lost, the library will [automatically attempt to reconnect](/docs/platform/architecture/connection-recovery) to the servers and in doing so, resume its position in the stream. This enables the chat to continue as if the user never left. After extended periods of disconnection, the client can make use of [history](#history) to backfill missing messages.
+
+It's incredibly rare, but sometimes a client might lose connection to a particular data center. Ably operates in multiple data centers around the world with multiple fallback regions available. If a client can't reach the nearest data center, it will try the next one until the connection is re-established, ensuring minimal downtime and that network issues don't disrupt the experience that you are trying to build.
```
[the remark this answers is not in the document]
AndyTWF: Link added to relevant architecture guide

src/pages/docs/guides/chat/build-livestream.mdx
```diff
@@ -0,0 +1,354 @@
...
+
+## AI Integrations
+
+AI chatbots are revolutionizing the way we interact with digital platforms. As livestreaming continues to grow, integrating AI functionality into your chat experience can provide enhanced engagement, automated moderation, and personalized interactions.
+
+Ably's flexible [integration capabilities](/docs/integrations) make it easy to connect your chat system with AI services. Through Ably's integrations, you can automatically route messages to:
+
+* Serverless functions (AWS Lambda, Azure Functions, etc.)
+* Webhooks
+* Custom endpoints
```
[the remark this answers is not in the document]
AndyTWF: Aded

PR #2597, opened 2025-05-14
[WEB-4394] fix: only set overflow-x on layout container when not control api

jamiehenson:
From the MDX work, we set `overflow-x` on the outer container in the layout because it solves a lot of element constraint problems. However, the patchwork quilt of js that is redocly's UI falls apart if you do this.

This PR conditionally sets the overflow property depending on whether you're on the control api or not.


kennethkalmer (approved):
Passes the Safari test 👍

PR #2600, opened 2025-05-14
[WEB-4356] Drop inline TOCs from textile

kennethkalmer:
## Description

In an attempt to speed up the rendering of pages using the `document.tsx` template (textile files) we discovered that the old inline TOC generation could in some cases take north of 500ms to complete and since it isn't used anymore we removed it completely.

The Gatsby packages were all upgraded as well in an effort to resolve a race condition in the image processing, but didn't reliably resolve the issue. 

Some unused types were cleaned up in the process too.

### Checklist

### DIFF
--- a/content/api/realtime-sdk.textile
+++ b/content/api/realtime-sdk.textile
@@ -38,25 +38,6 @@ redirect_from:
   - /docs/api/versions/v0.8/realtime-sdk
 ---
 
-inline-toc.
-  Properties:
-    - auth#auth
-    - push#push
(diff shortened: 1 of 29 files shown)


kennethkalmer:
@m-hulbert part of this cleanup means the `.textile` files no longer require the `inline-toc.` nonsense at the top (and we've removed them all as well)

jamiehenson (approved):
Makes sense and another load of stuff cleaned out.

PR #2605, opened 2025-05-15
Improve getting started project setup instructions

owenpearson:
Fixes a couple of issues with the javascript project setup here:

- Fixes usage of undocumented `tsc` requirement, now this is explicitly installed with NPM and used from the installed package rather than expected to be installed globally.
- Fixes `npm install` before `npm init -y`, both of these commands create `package.json` if it doesn't already exist but generally you're supposed to init the project first.
- General minor technical wording improvements

### DIFF
--- a/content/chat/getting-started/javascript.textile
+++ b/content/chat/getting-started/javascript.textile
@@ -38,20 +38,23 @@ ably apps switch
 ably auth keys switch
 ```
 
-* Install "Node.js":https://nodejs.org/en version 18 or greater.
-* Install "ts-node":https://www.npmjs.com/package/ts-node to execute TypeScript files:
-
-```[sh]
-npm install ts-node
-```
-
-* Create a new project in your IDE and install the Ably Chat JavaScript SDK. This will also install the Ably Pub/Sub SDK as it is a dependency:
+* Install any current LTS version of "Node.js":https://nodejs.org/en and create a new project:
 
 ```[sh]
 npm init -y
-tsc --init
+```
 
-npm install @ably/chat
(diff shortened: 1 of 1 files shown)


AndyTWF (approved):

content/chat/getting-started/javascript.textile
```diff
@@ -38,20 +38,24 @@ ably apps switch
...
-* Install "ts-node":https://www.npmjs.com/package/ts-node to execute TypeScript files:
+* Create a new NodeJS project
 
 ```[sh]
-npm install ts-node
+npm init -y
 ```
 
-* Create a new project in your IDE and install the Ably Chat JavaScript SDK. This will also install the Ably Pub/Sub SDK as it is a dependency:
+* Install "Node.js":https://nodejs.org/en version 18 or greater.
```
vladvelici, line 47: Should we recommend 22 LTS?

On that note, node 18 is now ([as of last month](https://github.com/nodejs/release#release-schedule)) no longer maintained, so we should probably drop support: remove 18 and add 24 to tests.

m-hulbert (approved):
Thanks @owenpearson !

content/chat/getting-started/javascript.textile
owenpearson: ah good spot!

I think i'll change the wording here to "any LTS version of Node.js", wdyt?

I agree we should revisit support in the sdk too 👍

vladvelici: Sounds good 👍

vladvelici (approved):

PR #2650, opened 2025-06-06
[WEB-4440] fix: restrict dl child filtering to active language from context

jamiehenson:
Addresses an issue on api references pages where javascript text content was leaking into docs content for other languages, by using the language set by the context only. Since the introduction of that method of tracking state, we don't have a "default" language per se, so the logic was being unnecessary clouded here.

Also removes a default language reference in the `Pre` block.


m-hulbert (approved):
🦵 🔚

PR #2653, opened 2025-06-06
chat: add react message reactions examples, restructure message reactions page

owenpearson:
- adds react code examples for message reactions
- restructures the page so that reaction types and the default type are explained before 'how to actually send a reaction'.

### DIFF
--- a/src/pages/docs/chat/rooms/message-reactions.mdx
+++ b/src/pages/docs/chat/rooms/message-reactions.mdx
@@ -15,9 +15,17 @@ The reaction `name` can be any string. Summaries are aggregated based on unique 
 
 ## Adding a message reaction <a id="adding-reactions"/>
 
+<If lang="javascript,swift,kotlin">
 To add a message reaction use `room.messages.reactions.send(message, params)`. This method takes the following parameters:
 * `message` - The message to add the reaction to. It can also be an object of format `{serial: "message serial"}`.
 * `params` - Set the `name`, and optionally override the `type` or set a `count`.
+</If>
+
+<If lang="react">
+Use the [`sendReaction()`](https://sdk.ably.com/builds/ably/ably-chat-js/main/typedoc/interfaces/chat_react.UseMessagesResponse.html#sendReaction) method available from the response of the `useMessages` hook to add a reaction to a message. This method takes the following parameters:
+* `message` - The message to add the reaction to. It can also be an object of format `{serial: "message serial"}`.
+* `params` - Set the `name`, and optionally override the `type` or set a `count`.
+</If>
 
 <Code>
 ```javascript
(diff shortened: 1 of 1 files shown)


src/pages/docs/chat/rooms/message-reactions.mdx
```diff
@@ -43,38 +99,123 @@ await room.messages.reactions.add(message, {
...
+    </div>
+  );
+};
+```
 </Code>
 
+
 ## Messages and reactions <a id="messages-and-reactions"/>
 
 The `Message` object contains a `reactions` property which is an object that looks like this:
```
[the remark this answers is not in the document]
owenpearson: well any snippet like this which is just explaining a return type of an API will probably be identical for both, the react APIs are only different from js in terms of how you _use_ them, what they actually return should always be the same. I have no idea how many snippets overall would fit into this category though 🤷

src/pages/docs/chat/rooms/message-reactions.mdx
```diff
@@ -13,11 +13,66 @@ The reaction `name` represents the reaction itself, for example an emoji. Reacti
...
+| `Unique` | A user can react to a message only once, with a reaction of their choice. When a user reacts a second time their reaction is changed. | Can 👍 or ❤️ but not both or more than once. | iMessage, WhatsApp, Facebook Messenger |
+| `Distinct` | A user can react to a message with each reaction at most once. | Can 👍 and ❤️ but each reaction only once. No 👍👍. | Slack |
+| `Multiple` | A user can react to a message with any reactions as many times as they like. Optionally a `count` parameter can be set when reacting. Reacting again adds to the existing count.| Can 👍 10 times and ❤️ 100 times. | Claps on Medium |
+
+Note that if adding two identical reactions of type `Distinct`, the second one will be accepted and broadcast as a raw reaction, but it will be ignored in the summary (aggregate). Similarly, when removing a reaction that doesn't exist (of any type), the operation will be accepted and broadcast as a raw reaction, but it will have no effect on the summary.
+
+### Configure the default reaction type <a id="default-type"/>
+
+The default reaction type can be configured at room-level using the Room Options. If nothing is set, the default is `Distinct`.
```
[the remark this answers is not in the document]
owenpearson: have also improved the wording here. it's a lot easier to just show how room options are used in an example (which we're doing in the snippet below) than to describe exactly what the API is, but i think what i have now is more clear either way

src/pages/docs/chat/rooms/message-reactions.mdx
```diff
@@ -170,9 +403,17 @@ If multiple reactions are added in a short period of time, multiple reactions ma
 
 ### Subscribing to raw reactions <a id="raw-reactions"/>
 
+<If lang="javascript,swift,kotlin">
+Raw individual reactions are published for every reaction, unlike summaries which can be rolled up. Raw reactions are useful for receiving all reaction events, but they are not suitable for the purpose of displaying message reaction counts as their effect on the reactions summary depends on the previous reactions.
+
+Individual reactions are not received by default to save bandwidth and to reduce the number of messages and cost. If you want to receive them, you can configure them via a room option which, in turn, sets the appropriate channel options to enable receiving individual annotations and reactions:
```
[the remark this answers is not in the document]
owenpearson: This was actually already here, my PR just duplicated it for react but i've reverted that change. "raw" events is the correct term because the feature is called rawReactions in the API. And tbh i think the previous section adequately describes how summary reactions work, enough for this section to make sense. I agree that the channel options part is irrelevant so have removed that 👍

src/pages/docs/chat/rooms/message-reactions.mdx
```diff
@@ -13,11 +13,66 @@ The reaction `name` represents the reaction itself, for example an emoji. Reacti
...
+| `Distinct` | Users can add each type of reaction once per message. Multiple different reactions are allowed, but duplicates are not. | A user can add both 👍 and ❤️, but cannot add a second 👍. | Slack |
+| `Multiple` | Users can add unlimited reactions, including duplicates. A count parameter specifies how many reactions to add at once. Each new reaction adds to the total count. | A user can add 10 👍 reactions and 100 ❤️ reactions to the same message. | Claps on Medium |
+
+Note that if adding two identical reactions of type `Distinct`, the second one will be accepted and broadcast as a raw reaction, but it will be ignored in the summary (aggregate). Similarly, when removing a reaction that doesn't exist (of any type), the operation will be accepted and broadcast as a raw reaction, but it will have no effect on the summary.
+
+### Configure the default reaction type <a id="default-type"/>
+
+The default reaction type can be configured at room-level by passing `RoomOptions` when calling `rooms.get`. If nothing is set, the default is `Distinct`.
```
AndyTWF, line 30: This sentence doesn't apply to React - with React it's the options passed to the `ChatRoomProvider`

src/pages/docs/chat/rooms/message-reactions.mdx
```diff
@@ -13,11 +13,66 @@ The reaction `name` represents the reaction itself, for example an emoji. Reacti
...
+    },
+});
+```
+
+```react
+import { MessageReactionType } from '@ably/chat';
+import { ChatRoomProvider } from '@ably/chat/react';
+
+const MyComponent = () => {
+  const roomOptions = {
```
AndyTWF, line 48: This would need to be outside the component or memoised, as the provide doesn't do that

src/pages/docs/chat/rooms/message-reactions.mdx
```diff
@@ -43,38 +98,125 @@ await room.messages.reactions.send(message, {
     count: 100,
 });
 ```
+
+```react
+import { MessageReactionType } from '@ably/chat';
+import { useMessages } from '@ably/chat/react';
+
+const MyComponent = () => {
+  const { addReaction } = useMessages();
```
AndyTWF, line 107:
```suggestion
  const { sendReaction } = useMessages();
```

And other instances

src/pages/docs/chat/rooms/message-reactions.mdx
```diff
@@ -120,14 +292,34 @@ Always call `Message.with(event)` when applying message events and reaction even
...
+const MyComponent = () => {
+  useMessages({
+    reactionsListener: (event) => {
+      console.log("received reactions summary event", event);
+    },
+  });
+
+  return <div>...</div>;
+};
+```
```
AndyTWF, line 322: Not this PR, but the `@` in the timeserial comes out a bit odd in the table below

src/pages/docs/chat/rooms/message-reactions.mdx
```diff
@@ -182,9 +415,35 @@ const room = await ablyChatClient.rooms.get('room1', {
     },
 });
 ```
+
+```react
+import { ChatRoomProvider } from '@ably/chat/react';
+
+const MyComponent = () => {
+  const roomOptions = {
```
AndyTWF, line 423: Ditto re room options memoisation

src/pages/docs/chat/rooms/message-reactions.mdx
owenpearson: ah good spot, have wrapped in an <If> thingy with a case for react

src/pages/docs/chat/rooms/message-reactions.mdx
owenpearson: have moved it outside the component scope. to be fair, it costs basically nothing to redeclare a static object and i don't think this component would ever re-render anyway but agree it's slightly neater 👍

src/pages/docs/chat/rooms/message-reactions.mdx
owenpearson: oops

AndyTWF (approved):

PR #2654, opened 2025-06-09
[WEB-4440] Add Open in ChatGPT button to right sidebar

jamiehenson:
Adds an "Open in ChatGPT" button to the right-hand nav sidebar in docs with the prompt "Tell me more about Ably's X feature from Y for Z" where
- `X` is the name of the feature
- `Y` is the docs page describing the feature (we can update this to a Markdown resource later when that functionality is ready - currently in the works by @kennethkalmer)
- `Z` is the targeted programming language, only appended if applicable (it wouldn't be on the Chat index page, for instance) 

Review link (with example page): https://ably-docs-web-4440-gpt--8g8gqj.herokuapp.com/docs/chat/rooms/message-reactions

Screenshot 2025-06-11 at 17 40 25

## Summary by CodeRabbit

- **New Features**
  - Added an "Open in ChatGPT" external link in the sidebar, allowing users to quickly generate prompts based on the current page.

- **Style**
  - Updated the border styling for sidebar external links for improved visual consistency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "validate-llms-txt": "ts-node bin/validate-llms.txt.ts"
   },
   "dependencies": {
-    "@ably/ui": "17.1.0",
+    "@ably/ui": "17.2.1",
     "@codesandbox/sandpack-react": "^2.20.0",
     "@codesandbox/sandpack-themes": "^2.0.21",
     "@gfx/zopfli": "^1.0.15",
--- a/src/components/Layout/RightSidebar.tsx
+++ b/src/components/Layout/RightSidebar.tsx
@@ -30,26 +30,26 @@ const externalLinks = (
   activePage: ActivePage,
   location: WindowLocation,
 ): { label: string; icon: IconName; link: string; type: string }[] => {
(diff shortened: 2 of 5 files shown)


aralovelace (approved):
yey! more AI-integrated experience 🚀

m-hulbert (commented):
This is a really nice quick addition! 

I saw you added the language that was being viewed, but In the interest of providing additional context, I think it might be worth including the product as well (e.g. `Tell me more about Ably Chat's X`). I had some slightly confusing results viewing history in Chat as it was pulling some info from Pub/Sub for example that wasn't relevant. 

You can also get some weird results if you navigate to a page with a query param set for a language that doesn't support that feature, but I think that's potentially unavoidable.

jamiehenson:
Thanks @m-hulbert - we have access to a page's product so can work that in. 

As for the query param, I think we can do something around that - the query param overrides the language, but we also know the languages supported by a page, so we can adjust the condition so that we only include the language part of the prompt if the param either isn't there or lines up with a valid lang on the page.

PR #2657, opened 2025-06-09
[WEB-4447] Generate markdown copies of each HTML file for agents

kennethkalmer:
## Description

PoC to generate markdown version of our generated HTML content to be more supportive of agentic workflows.

### Checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1497,13 +1497,6 @@
   resolved "https://registry.yarnpkg.com/@codesandbox/sandpack-themes/-/sandpack-themes-2.0.21.tgz#f1970f03537434fff008e9c9c3f6581b0e5940c2"
   integrity sha512-CMH/MO/dh6foPYb/3eSn2Cu/J3+1+/81Fsaj7VggICkCrmRk0qG5dmgjGAearPTnRkOGORIPHuRqwNXgw0E6YQ==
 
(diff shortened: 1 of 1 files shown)


m-hulbert:
Thanks for taking a stab at this, Ken! 

Not sure what level of feedback you're looking for, but; 

* We'd need to enable the output to handle a query param for different languages, as it only supports the default at the moment.
* I think we should signal somewhere in the metadata the language the page was generated for once we sort that.
* Surprisingly the output doesn't handle tables even though we're not doing anything non-standard. 
* We do use a non-standard component for code, so we'll need to something with that. 
* Same for the aside syntax that we use - I wonder if we should just show the raw MDX rather than trying to do anything fancy. 
* The textile output is a bit iffy, but that's to be expected and not something we should worry about. 
* I think we should add a button to each page to view in MD once we nail this.

config/mime.types
```diff
@@ -8,6 +8,7 @@ types {
     application/javascript                           js;
     application/atom+xml                             atom;
     application/rss+xml                              rss;
+    text/markdown                                    md;
```
kennethkalmer, line 11: @mattheworiordan I think we need this line in #2862 as well to help nginx along

m-hulbert:
This has been superseded by https://github.com/ably/docs/pull/3000

PR #2669, opened 2025-06-17
Add Ruby Getting Started guide

GregHolmes:
### Checklist

### DIFF
--- a/content/getting-started/ruby.textile
+++ b/content/getting-started/ruby.textile
@@ -1,0 +1,211 @@
+---
+title: "Getting started: Pub/Sub in Ruby"
+meta_description: "A getting started guide for Ably Pub/Sub Ruby that steps through some of the key features using Ruby."
+meta_keywords: "Pub/Sub Ruby, Ruby PubSub, Ably Ruby SDK, realtime messaging Ruby, publish subscribe Ruby, Ably Pub/Sub guide, Ruby realtime communication, Ably tutorial Ruby, Ruby message history, presence API Ruby, Ably Pub/Sub example, realtime Pub/Sub Ruby, subscribe to channel Ruby, publish message Ruby, Ably CLI Pub/Sub"
+languages:
+  - ruby
+---
+
+This guide will get you started with Ably Pub/Sub in Ruby.
+
+It will take you through the following steps:
+
+* Create a client and establish a realtime connection to Ably.
+* Attach to a channel and subscribe to its messages.
+* Publish a message to the channel for your client to receive.
+* Join and subscribe to the presence set of the channel.
+* Retrieve the messages you sent in the guide from history.
+* Close a connection to Ably when it is no longer needed.
+
+h2(#prerequisites). Prerequisites
+
+* "Sign up":https://ably.com/sign-up for an Ably account.
+** Create a new app and get your API key from the "dashboard":https://ably.com/dashboard.
+** Your API key will need the @publish@, @subscribe@, @presence@ and @history@ "capabilities":/docs/auth/capabilities.
+
+* Install the Ably CLI:
+
+```[sh]
+npm install -g @ably/cli
+```
+
+* Run the following to log in to your Ably account and set the default app and API key:
+
+```[sh]
+ably login
+
+ably apps switch
+ably auth keys switch
+```
+
+* Install "Ruby":"https://www.ruby-lang.org/en/documentation/installation/ version 2.7 or greater.
+* Create a new project and install the Ably "Pub/Sub Ruby SDK":https://github.com/ably/ably-ruby :
+
+```[sh]
+# Create a new Gemfile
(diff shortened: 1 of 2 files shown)


m-hulbert (commented):
This is good! Spotted a couple of things running through it though.

Should we add a mention early on (perhaps in the prerequisites?) regarding usage of EventMachine?

content/getting-started/ruby.textile
```diff
@@ -0,0 +1,220 @@
...
+    # Initialize the Ably Realtime client
+    realtime_client = Ably::Realtime.new(key: '{{API_KEY}}', client_id: 'my-first-client')
+
+    # Wait for the connection to be established
+    realtime_client.connection.on(:connected) do
+      puts "Made my first connection!"
+    end
+
+
+    realtime_client.connection.once(:connected) do
```
lawrence-forooghian, line 82: It's not very clear to me what's going on here — why do we have two different `:connected` handlers?

content/getting-started/ruby.textile
```diff
@@ -0,0 +1,220 @@
...
+```[sh]
+ably channels subscribe my-first-channel
+```
+
+Publish another message using the CLI and you will see that it's received instantly by the client you have running locally, as well as the subscribed terminal instance.
+
+To publish a message in your code, you can add the following line to your @get_started@ method after subscribing to the channel:
+
+```[ruby]
+channel.publish 'example', 'A message sent from my first client!'
```
lawrence-forooghian, line 126: I'm not sure it's clear to the user why there are two arguments here, especially given that the CLI `publish` call above only took one argument?

content/getting-started/ruby.textile
```diff
@@ -0,0 +1,220 @@
...
+# Subscribe to presence events on the channel
+channel.presence.subscribe do |member|
+  puts "Event type: #{member.action} from #{member.client_id} with the data #{member.data.inspect}"
+end
+
+# Enter the presence set
+channel.presence.enter("I'm here!")
+```
+
+* In the "dev console":https://ably.com/accounts/any/apps/any/console of your first app, attach to @my-first-channel@. Enter a @client_id@, such as @my-dev-console@, and then join the presence set of the channel. You'll see that @my-first-client@ is already present in the channel. In your terminal you'll see that an event was received when the dev console client joined the channel.
```
lawrence-forooghian, line 145: Maybe I've missed something, but is there a reason that we use both the dev console and the CLI as the "other" Ably client within the same guide? Wouldn't one tool suffice?

content/getting-started/ruby.textile
```diff
@@ -0,0 +1,220 @@
...
+
+```[sh]
+ably channels presence enter my-first-channel --client-id "my-cli" --data '{"status":"learning about Ably!"}'
+```
+
+h2(#step-4). Step 4: Retrieve message history
+
+You can retrieve previously sent messages using the history feature. Ably stores all messages for 2 minutes by default in the event a client experiences network connectivity issues. This can be extended for longer if required.
+
+If more than 2 minutes has passed since you published a regular message (excluding the presence events), then you can publish some more before trying out history. You can use the Pub/Sub SDK, Ably CLI or the dev console to do this.
```
lawrence-forooghian, line 157:
```suggestion
If more than 2 minutes has passed since you published a regular message (excluding the presence events), then publish some more before trying out history. You can use the Pub/Sub SDK, Ably CLI or the dev console to do this.
```

"you can" makes it sound optional, which it's not — they won't see anything in history!

content/getting-started/ruby.textile
```diff
@@ -0,0 +1,220 @@
...
+
+Note that messages are streamed to clients as soon as they attach to a channel, as long as they have the necessary capabilities. Clients are implicitly attached to a channel when they call @subscribe()@. Detaching from a channel using the @detach()@ method will stop the client from being streamed messages by Ably.
+
+Listeners registered when subscribing to a channel are registered client-side. Unsubscribing by calling @unsubscribe()@ will remove previously registered listeners for that channel. Detaching from a channel has no impact on listeners. As such, if a client reattaches to a channel that they previously registered listeners for, then those listeners will continue to function upon reattachment.
+
+* Add the following to your client to close the connection after a simulated 10 seconds. Run it with @ruby get_started.rb@:
+
+```[ruby]
+# Close the connection after 10 seconds
+Thread.new do
```
lawrence-forooghian, line 198: Perhaps it would be better to use `EventMachine.add_timer 10 do …` instead of mixing concurrency idioms?

content/getting-started/ruby.textile
GregHolmes: I've added `name` to the CLI, data is specified there by default. So to keep things consistent I've added the name flag.

lawrence-forooghian (approved):

PR #2674, opened 2025-06-19
Add Python getting started guide for Pub/Sub

GregHolmes:
## Description

[Getting started guide](https://ably-docs-edu-1979-crea-fd39ju.herokuapp.com/docs/getting-started/python) for Pub/Sub with Python

### DIFF
--- a/content/getting-started/python.textile
+++ b/content/getting-started/python.textile
@@ -1,0 +1,217 @@
+---
+title: "Getting started: Pub/Sub in Python"
+meta_description: "A getting started guide for Ably Pub/Sub Python that steps through some of the key features using Python."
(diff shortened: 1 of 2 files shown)


content/getting-started/python.textile
```diff
@@ -0,0 +1,217 @@
...
+```[sh]
+ably channels subscribe my-first-channel
+```
+
+Publish another message using the CLI and you will see that, in your other Terminal window, it's received instantly by the client you have running locally, as well as the subscribed terminal instance.
+
+To publish a message with the client running in your IDE, you can add the following line to your @get_started@ function after subscribing to the channel:
+
+```[python]
+await channel.publish('example', 'A message sent from my first client!')
```
VeskeR, line 123:
```suggestion
await channel.publish("example", "A message sent from my first client!")
```

double quotes for consistency with the rest of the examples when the string is a natural language

VeskeR (changes requested):
lgtm overall, couple of small suggestions

PR #2687, opened 2025-06-25
[EDU-1993] - Add Chat's React Native getting started guide

GregHolmes:
A new getting started guide for Chat in React Native using Expo.

### DIFF
--- a/src/pages/docs/chat/getting-started/react-native.mdx
+++ b/src/pages/docs/chat/getting-started/react-native.mdx
@@ -1,0 +1,944 @@
+---
+title: "Getting started: Chat with React Native"
+meta_description: "A getting started guide for Ably Chat React Native that steps through some of the key features using React Native."
+meta_keywords: "Ably, realtime, quickstart, getting started, basics, Chat, React Native"
+---
+
+This guide will get you started with Ably Chat on a new React Native application.
+
+It will take you through the following steps:
+
+* Creating a client and establishing a realtime connection to Ably
+* Creating a room and subscribing to its messages
+* Sending messages to the room and editing messages
+* Retrieving historical messages to provide context for new joiners
+* Setting up typing indicators to see which clients are typing
+* Use presence to display the online status of users in the room
+* Subscribing to and sending reactions
+* Disconnecting and resource cleanup
+
+## Prerequisites <a id="prerequisites"/>
+
+### Ably <a id="prerequisites-ably"/>
+
+1. [Sign up](https://ably.com/signup) for an Ably account
+
+2. Create a new app and get your API key. You can use the root API key that is provided by default to get started.
+
+3. Install the Ably CLI:
+
+<Code>
+```shell
+npm install -g @ably/cli
+```
+</Code>
+
+4. Run the following to log in to your Ably account and set the default app and API key:
+
+<Code>
+```shell
+ably login
+
+ably apps switch
+ably auth keys switch
+```
+</Code>
+
+<Aside data-type='note'>
+The code examples in this guide include a demo API key. If you wish to interact with the Ably CLI, ensure that you replace them with your own API key.
+</Aside>
+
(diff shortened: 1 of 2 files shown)


src/pages/docs/chat/getting-started/react-native.mdx
```diff
@@ -0,0 +1,956 @@
...
+});
+
+const chatClient = new ChatClient(realtimeClient, {
+  logLevel: LogLevel.Info,
+});
+
+export const ChatProviders: React.FC<{children: React.ReactNode}> = ({
+  children,
+}) => {
+  return (
```
splindsay-92, line 144: The setup of the providers should be identical for react and react native. Could we either make this look like the base react guide, or update that one to match how you've done things here please? :)

src/pages/docs/chat/getting-started/react-native.mdx
```diff
@@ -0,0 +1,956 @@
...
+// Update ChatApp to include ChatBox
+function ChatApp() {
+  return (
+    <ChatRoomProvider
+      name="my-first-room"
+      release={true}
+      attach={true}
+    >
+      <KeyboardAvoidingView
+        style={{flex: 1}}
```
splindsay-92, line 399: Does KeyboardAvoidingView not take tailwind in its classname props, or was it just not behaving well?

src/pages/docs/chat/getting-started/react-native.mdx
```diff
@@ -0,0 +1,956 @@
...
+    [],
+  );
+
+  const handleUpdate = useCallback(async () => {
+    if (!editingMessage || !editText.trim()) {
+      console.log('Early return - missing data');
+      return;
+    }
+
+    room.messages.update(editingMessage.serial, {text: editText.trim()}, {description: "Message update by user"})
```
splindsay-92, line 471: I'm not quite sure whats happened here, is there a reason you are using room instead of the update func exposed by the useMessage hook? :) Seems the text above suggest it's supposed to be present here?

splindsay-92 (commented):
Just a few comments, but looks good!

GregHolmes:
Thanks @splindsay-92  I've updated the PR based on your feedback. If you could give it another look over, that would be greatly appreciated.

src/pages/docs/chat/getting-started/react-native.mdx
```diff
@@ -0,0 +1,970 @@
...
+
+In your `ChatApp.tsx` file, add a new component called `ReactionComponent`, like so:
+
+<Code>
+```react
+// ChatApp.tsx
+function ReactionComponent() {
+  const reactions = ['👍', '❤️', '💥', '🚀', '👎', '💔'];
+  const [roomReactions, setRoomReactions] = useState<Reaction[]>([]);
+  const {send} = useRoomReactions({
```
splindsay-92, line 840: Same here, this is now sendRoomReaction :)

src/pages/docs/chat/getting-started/react-native.mdx
```diff
@@ -0,0 +1,962 @@
...
+```
+</Code>
+
+## Step 10: Disconnection and release <a id="step-10"/>
+
+The disconnection and release section remains largely the same for React Native:
+
+### Automatic detachment on unmount <a id="auto-detach"/>
+
+For React Native components, when a component using the `ChatRoomProvider` unmounts, the room will automatically detach from the underlying channel and clean up associated resources. This behavior is controlled by the `release` prop on the `ChatRoomProvider`.
```
splindsay-92, line 917: This is no longer the case, release and cleanup are automatic on unmount now :) There is no longer a release prop.

splindsay-92: The only time you will need to manually release a room now, is if you aren't using the room provider :)

PR #2690, opened 2025-06-27
[WEB-4460] fix: preserve lang param in content links

jamiehenson:
Raised by @m-hulbert.

Links in content in MDX pages need to preserve any active lang params so that we get a consistent language experience between pages.

Review: https://ably-docs-web-4440-pres-2duhlf.herokuapp.com/docs/chat/rooms?lang=react


m-hulbert (commented):
Works perfectly for MDX, but we also seem to have introduced a regression to textile whenever we broke this. Happy to approve this and deal with textile separately, or include it in here - whichever is easiest.

m-hulbert (approved):
🚀

PR #2692, opened 2025-06-30
[EDU-1982] - Add PHP Getting started guide

GregHolmes:
## Description

PHP Getting started guide.
As PHP SDK only has RESTful functionality, I've added creating a JWT to the getting started guide. There is no subscribing to messages, nor is there entering the presence set. So I believe PHP will be more focused on the backend side of things, authentication is likely key for this.

### DIFF
--- a/content/getting-started/php.textile
+++ b/content/getting-started/php.textile
@@ -1,0 +1,215 @@
+---
+title: "Getting started: Pub/Sub in PHP"
+meta_description: "A getting started guide for Ably Pub/Sub PHP that steps through some of the key features using PHP."
+meta_keywords: "Pub/Sub PHP, PHP PubSub, Ably PHP SDK, publish PHP, Ably Pub/Sub guide, PHP realtime communication, Ably tutorial PHP, PHP message history, presence API PHP, Ably Pub/Sub example, realtime Pub/Sub PHP, publish message PHP, Ably CLI Pub/Sub"
+languages:
+  - PHP
+---
+
+This guide will get you started with Ably Pub/Sub in PHP.
+
+It will take you through the following steps:
+
+* Create a client and establish a connection to Ably with the Rest SDK.
+* Generate a JWT for authenticating frontend clients.
+* Publish a message to the channel with your client for subscribers to receive.
+* Retrieve the presence set of the channel.
+* Retrieve the messages you sent in the guide from history.
+
+<aside data-type='note'>
+<p>This guide covers the Ably PHP SDK which provides RESTful functionality for publishing messages, retrieving history, and checking presence. If you need realtime features like subscribing to channels and receiving live updates, check out the "Ably Laravel broadcaster":https://github.com/ably/laravel-broadcaster which integrates with Laravel's broadcasting system.</p>
+</aside>
+
+h2(#prerequisites). Prerequisites
+
+* "Sign up":https://ably.com/signup for an Ably account.
+** Create a new app, and create your first API key.
+** Your API key will need the @publish@, @presence@, and @history@ capabilities.
+
+* Install the Ably CLI:
+
+```[sh]
+npm install -g @ably/cli
+```
+
+* Run the following to log in to your Ably account and set the default app and API key:
+
+```[sh]
+ably login
+
+ably apps switch
+ably auth keys switch
+```
+
+* Install "PHP":https://www.php.net/downloads.php version 7.2 or greater.
+* Create a new project and install the Ably Pub/Sub PHP SDK:
+
+```[sh]
+# Create a new directory for your project
+mkdir ably-php-quickstart
+cd ably-php-quickstart
+
+# Initialize Composer (if you don't have a composer.json yet)
+composer init
+
+# Install the Ably PHP SDK
+composer require ably/ably-php
+```
+
+<aside data-type='note'>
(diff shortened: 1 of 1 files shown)


content/getting-started/php.textile
```diff
@@ -0,0 +1,207 @@
...
+    'HS256',
+    $apiKeyCredentials[0]
+);
+
+echo $jwt;
+```
+
+h2(#step-3). Step 3: Publish a message to a channel
+
+Messages contain the data that a client is communicating, such as a short 'hello' from a colleague, or a financial update being broadcast to subscribers from a server. Ably uses channels to separate messages into different topics, so that clients only ever receive messages on the channels they are subscribed to. The PHP SDK and Rest API can only publish messages to channels and cannot subscribe to them. To subscribe to channels, for this guide, you will make use of the Ably CLI.
```
m-hulbert, line 122: I think we should drop mention of the REST API to be honest, as it's just another thing to learn in a getting started. 

(also another nitpick on "for this guide" as it makes it sound like you'd read multiple)

content/getting-started/php.textile
```diff
@@ -0,0 +1,207 @@
...
+
+* Use the Ably CLI to subscribe to a channel a message to your first channel. The message will be received by the client you've subscribed to the channel, and be logged to the console.
```
m-hulbert, line 124:
```suggestion
* Use the Ably CLI to subscribe to a channel. The message that your PHP client sends will be received by CLI instance.
```

content/getting-started/php.textile
```diff
@@ -0,0 +1,207 @@
...
+
+```[sh]
+ably channels subscribe my-first-channel
+```
+
+* Add the following lines to the bottom of your @get_started@ file to create a channel instance and register a listener to publish a message to the channel. Then run it with @php get_started.php@:
```
m-hulbert, line 130: You don't register a listener to publish.

content/getting-started/php.textile
```diff
@@ -0,0 +1,207 @@
...
+
+```[php]
+$channel = $ably->channel('my-first-channel');
+$channel->publish('myEvent', 'Hello!');
+```
+
+If you check the console where you subscribed to the channel with the Ably CLI, you should see the message appear.
```
m-hulbert, line 137:
```suggestion
The Ably CLI will receive the message.
```

content/getting-started/php.textile
```diff
@@ -0,0 +1,207 @@
...
+```
+
+* Add the following lines to your @get_started@ function to retrieve the list of members in the presence set. Then run it with @php get_started.php@:
+
+```[php]
+$membersPage = $channel->presence->get(); // Returns a PaginatedResult
+$memberIds = array_map(function($member) {
+    return $member->clientId;
+}, $membersPage->items);
+echo "Members in presence set: " . implode(', ', $memberIds) . "\n";
```
m-hulbert, line 156:
```suggestion
echo "\nMembers in presence set:\n " . implode(', ', $memberIds) . "\n";
```

content/getting-started/php.textile
```diff
@@ -0,0 +1,207 @@
...
+
+$apiKeyCredentials = explode(':', '{{API_KEY}}');
+$jwt = JWT::encode(
+    $payload,
+    $apiKeyCredentials[1],
+    'HS256',
+    $apiKeyCredentials[0]
+);
+
+echo $jwt;
```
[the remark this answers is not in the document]
m-hulbert: As discussed: maybe some text explaining how to then use this to authenticate other clients.

m-hulbert (commented):
Just a few tidy ups!

content/getting-started/php.textile
```diff
@@ -0,0 +1,215 @@
...
+require 'vendor/autoload.php';
+
+use Ably\AblyRest;
+
+$ably = new AblyRest(['key' => '{{API_KEY}}', 'clientId' => 'my-first-client']);
+```
+
+h2(#step-2). Step 2: Generate a JWT
+
+The Ably PHP SDK provides token authentication for secure client connections. Authentication with a JWT is recommended over using API keys directly in production because they can be scoped to specific capabilities and have expiration times. This is essential when authenticating frontend clients where embedding API keys would expose them publicly. With PHP the API key would be considered securely stored as it is on the server side.
```
m-hulbert, line 82:
```suggestion
The Ably PHP SDK provides token authentication for secure client connections. Authentication with a JWT is recommended over using API keys directly in production because they can be scoped to specific capabilities and have expiration times. This is essential when authenticating frontend clients where embedding API keys would expose them publicly. With PHP the API key would be considered secure as it's stored on the server.
```

content/getting-started/php.textile
```diff
@@ -0,0 +1,215 @@
...
+    'HS256',
+    $apiKeyCredentials[0]
+);
+
+echo $jwt;
+```
+
+To use this JWT for authenticating frontend clients, you would typically expose it through an API endpoint that your frontend application can call. For example, you might create an endpoint like @/auth/token@ that returns this JWT. Your frontend clients can then use this JWT when instantiating their Ably client:
+
+```[javascript]
```
m-hulbert, line 122: Unfortunately this will now add JS to the dropdown of languages to view the page in. 

I think the best way to get around this is to set the language to something like shell, JSON or PHP and then use the `fixed` property to hide the language; `<Code fixed="true">`

m-hulbert (approved):

PR #2712, opened 2025-07-11
Chat: A how to guide for replies

splindsay-92:
## Context

Chat currently doesn't have native support for things like replies and quotes, so we need a short guide to give users an idea as to how they might implement a solution with the Chat SDK in the mean time.

[CHA-1016](https://ably.atlassian.net/browse/CHA-1016)

## Description

This PR adds a guide that covers a basic replies setup, using the Chat SDK, and sending message reply metadata in the metadata field of a chat message.

### DIFF
--- a/src/data/nav/chat.ts
+++ b/src/data/nav/chat.ts
@@ -129,6 +129,10 @@ export default {
           name: 'Livestream chat',
           link: '/docs/guides/chat/build-livestream',
         },
+        {
+          name: 'Replies',
+          link: '/docs/guides/chat/replies',
+        }
       ],
     },
   ],
--- a/src/pages/docs/guides/chat/replies.mdx
+++ b/src/pages/docs/guides/chat/replies.mdx
@@ -10,7 +10,7 @@ As such, we will cover only replies in this guide, but the same principles can b
 
 While cross-room replies are possible, this guide will focus on providing replies within the same room for simplicity.
 
-## Understanding the current limitations
+## Understanding the current limitations <a id="limitations"/>
 
 While Ably Chat doesn't have built-in support for replies or quotes, you can implement this functionality by using the metadata fields that can be sent along with chat messages.
 
@@ -21,7 +21,7 @@ Every message in Ably Chat has:
 
 By storing the serial and creation time of a message you want to quote in the metadata of a new message, you can create a reference between messages that your application can interpret and display accordingly.
 
-## Implementing replies using message metadata
+## Implementing replies using message metadata <a id="implementing-replies"/>
 
 To implement a reply to a message, you'll need to:
 
@@ -29,7 +29,7 @@ To implement a reply to a message, you'll need to:
 2. Include that information in the metadata when sending the reply
 3. Process incoming messages to check for reply metadata
 
-### Step 1: Store information about the message being replied to
+### Step 1: Store information about the message being replied to <a id="store-information"/>
 
 When a user selects a message to reply to, you should store its key information. Extract the following properties from the original message:
 
@@ -42,6 +42,7 @@ Store this information in your application state or pass it to your message inpu
 
 For example, when a user clicks a "Reply" button on a message, you would extract these properties from the message and store them in your application state, perhaps in a variable called `replyingTo`.
 
+<Code>
 ```javascript
 import { Message } from '@ably/chat';
 function prepareReply(originalMessage: Message) {
@@ -57,14 +58,16 @@ function prepareReply(originalMessage: Message) {
   return replyingTo;
 }
 ```
+</Code>
 
-### Step 2: Send a reply message with metadata
+### Step 2: Send a reply message with metadata <a id="send-reply"/>
 
 When sending the reply message, include the original message's information in the metadata.
 Create a metadata object with a "reply" property that contains the original message's serial, timestamp and optionally the clientId and a snippet of the original text.
 
 The structure of your metadata object should look like this:
 
+<Code>
 ```javascript
 // Example metadata structure for a reply message
 const metadata = {
@@ -76,11 +79,13 @@ const metadata = {
   }
 };
 ```
+</Code>
 
-Then use the `room.messages.send()` method to send the message with this metadata. The method takes an object with `text` and `metadata` properties.
+Then use the [`room.messages.send()`](https://sdk.ably.com/builds/ably/ably-chat-js/main/typedoc/interfaces/chat-js.Messages.html#send) method to send the message with this metadata. The method takes an object with `text` and `metadata` properties.
 
 In your `text` property, include the reply text that the user is sending. The `metadata` property should contain the metadata object you created earlier.
 
+<Code>
 ```javascript
 // Function to send a reply message
 import {Message} from '@ably/chat';
@@ -118,9 +123,11 @@ const replyText = "I'm responding to your message!";
 
 await sendReply(messageToReplyTo, replyText);
 ```
+</Code>
 
 To test this out, you can try sending a message to a room, and then sending a reply to that message using the `sendReply` function. Make sure to replace `messageToReplyTo` with the actual message object you want to reply to.
 
+<Code>
 ```ts
 const originalMessage = await room.messages.send({
   text: "Hello, this is the original message!"
@@ -130,12 +137,13 @@ const replyText = "This is a reply to the original message!";
 
 await sendReply(originalMessage, replyText);
 ```
+</Code>
 
-## Handling received messages with reply metadata
+## Handling received messages with reply metadata <a id="handling-received-messages"/>
 
 When receiving messages, you need to check if they contain reply metadata and display them accordingly. Here's how to do it:
 
-1. Subscribe to messages in the room using `room.messages.subscribe()`.
+1. Subscribe to messages in the room using [`room.messages.subscribe()`](https://sdk.ably.com/builds/ably/ably-chat-js/main/typedoc/interfaces/chat-js.Messages.html#subscribe).
 2. In the callback function, examine each incoming message.
 3. Check if the message has a `metadata.reply` property.
 4. If it does, extract the reply information (serial, createdAt, clientId, etc.).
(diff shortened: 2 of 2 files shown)


src/data/nav/chat.ts
```diff
@@ -129,6 +129,11 @@ export default {
           name: 'Livestream chat',
           link: '/docs/guides/chat/build-livestream',
         },
+        {
+          name: 'Replies',
+          link: '/docs/guides/chat/replies',
+          languages: [],
```
jamiehenson, line 135: An unintended use for this, but this'll do the job for suppressing the language selector. Two thoughts from here:
- (more for @m-hulbert) we already have a condition where if there is no language content detected on the page (i.e. [here](https://ably.com/docs/chat)), then it won't show the language selector. Is there something in changing that so we only show it if there are two or more languages? Even with one language, there's nothing to switch. It would mean we hide it on this new page as normal without an override as it only has TS.
- either way, if we would like to support TS as a language, then we need to add it with a version number to the [language data](https://github.com/ably/docs/blob/chat/how-to-replies/src/data/languages/languageData.ts#L24-L29), guessing it would be the same version as the JS lib but not for me to say.

TLDR, this override works but a better path forward would be to either not show the language selector for single language pages, or add TS to the language data so we get the selector working with TS

src/pages/docs/guides/chat/replies.mdx
```diff
@@ -0,0 +1,583 @@
...
+1. Create UI components for displaying regular messages and replies differently
+2. Implement a message selection mechanism to allow users to choose which message to reply to
+3. Set up local state management to store messages and their relationships
+4. Handle edge cases like missing original messages or pagination of history
+5. Implement error handling for failed message sends or history queries
+6. Add visual indicators for when a message is a reply and which message it's replying to
+
+Here's a simple React component that demonstrates how to implement a chat interface with basic reply functionality using the `useMessages` hook:
+
+<Code fixed="true">
```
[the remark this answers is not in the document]
splindsay-92: I've now made it so all code segments have both :)

src/pages/docs/guides/chat/replies.mdx
```diff
@@ -0,0 +1,583 @@
...
+
+2. **Message availability**: If a quoted message is very old, it might not be available in history anymore, depending on your Ably account's message retention settings.
+
+3. **Message updates**: Because messages can be updated (including the metadata), it's possible to accidentally remove the quoted message reference. Be careful when updating messages that are part of a reply chain.
+
+4. **No server-side validation**: Since metadata is not validated by the server, you should ensure that your application logic correctly handles cases where the metadata might be malformed or missing.
+
+5. **Local state management**: Maintaining a local state of messages will improve performance by reducing the need to query history.
+
+6. **Tracking replies**: There isn't an easy way to know if a message has been replied to, other than to see its replies in history. This is why the historyBeforeSubscribe method is useful to ensure you have the full context of messages in your room.
```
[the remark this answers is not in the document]
splindsay-92: I've reworked this, the focus is now on history, but I've mentioned historyBeforeSubscribe as being useful for initial fetch.

src/pages/docs/guides/chat/replies.mdx
```diff
@@ -0,0 +1,583 @@
...
+
+1. Subscribe to messages in the room using `room.messages.subscribe()`.
+2. The subscription returns an object that includes the `historyBeforeSubscribe` method.
+3. Call this method with a limit parameter (e.g., `{ limit: 50 }`) to get recent history.
+4. Process the results to identify any reply messages without their original messages.
+5. If you find replies without originals, paginate back further using the `next()` method on the result.
+6. Continue this process until all replies have their original messages or no more history is available.
+
+Of course, if you have a lot of history and the reply came much later, it might be better to use the `history` function to load only the specific message you care about, as shown in the previous section.
+
```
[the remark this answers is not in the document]
splindsay-92: I've reworked this section a bit too, the default is to use `history` now, and that it might be more beneficial to load individual message on demand.

src/data/nav/chat.ts
m-hulbert: I'm not sure about removing it if there's only 1 language. I think we should remove the ability to click the dropdown, but I personally think it makes sense to keep it for consistency (though I'll check to get a UX perspective).

src/data/nav/chat.ts
```diff
@@ -129,6 +129,11 @@ export default {
           name: 'Livestream chat',
           link: '/docs/guides/chat/build-livestream',
         },
+        {
+          name: 'Replies',
+          link: '/docs/guides/chat/replies',
+          languages: ['react', 'javascript'],
```
m-hulbert, line 135: This isn't necessary if it's now relevant to both JS and React - the dropdown gets automatically populated and this is just an override.

src/data/nav/chat.ts
```diff
@@ -129,6 +129,11 @@ export default {
           name: 'Livestream chat',
           link: '/docs/guides/chat/build-livestream',
         },
+        {
+          name: 'Replies',
```
m-hulbert, line 133: I'd suggest referencing quotes here as well for completeness and scanning purposes.

src/pages/docs/guides/chat/replies.mdx
```diff
@@ -0,0 +1,803 @@
+---
+title: "Implementing Replies and Quotes with Chat"
+meta_description: "Learn how to implement message replies and quotes in your Ably Chat applications using TypeScript. This guide shows how to use metadata to create quote and reply functionality."
+meta_keywords: "ably chat, message replies, message quotes, chat replies, chat quotes, javascript chat replies, typescript chat replies, chat metadata, ably chat metadata, realtime chat replies"
+---
+
+This guide will show you how you could implement a simple solution for things like replies and quotes in your Ably Chat application.
```
m-hulbert, line 7: I'm going to make the same comment as I did for file uploads in that I think it's immaterial to an end-user that there isn't a dedicated API for handling this. I think this should just sit in the existing 'feature' docs and it isn't necessary to call it out as a guide.

src/pages/docs/guides/chat/replies.mdx
```diff
@@ -0,0 +1,803 @@
...
+While the two features have different applications, they share the same fundamental principle of referencing other messages in the chat.
+As such, we will cover only replies in this guide, but the same principles can be applied to quotes.
```
m-hulbert, line 9: Avoid first person "we" please 🙂

src/pages/docs/guides/chat/replies.mdx
```diff
@@ -0,0 +1,803 @@
...
+
+While cross-room replies are possible, this guide will focus on providing replies within the same room for simplicity.
+
+## Understanding the current limitations <a id="limitations"/>
```
m-hulbert, line 13: This is starting out on a negative upfront. I don't think we need to be apologetic or mention there isn't a dedicated API. It's possible  to do them and people just want to get to the _how_.

src/pages/docs/guides/chat/replies.mdx
```diff
@@ -0,0 +1,803 @@
...
+While Ably Chat doesn't have built-in support for replies or quotes, you can implement this functionality by using the metadata fields that can be sent along with chat messages.
+
+Every message in Ably Chat has:
+- A `serial` field (a UUID that uniquely identifies the message)
+- A `createdAt` field (when the message was created)
+- A `metadata` field (key-value store for custom data)
+
+By storing the serial and creation time of a message you want to quote in the metadata of a new message, you can create a reference between messages that your application can interpret and display accordingly.
+
+## Implementing replies using message metadata <a id="implementing-replies"/>
```
m-hulbert, line 24: I'd keep this active voice, so 'Implement replies using message metadata'.

src/pages/docs/guides/chat/replies.mdx
```diff
@@ -0,0 +1,803 @@
...
+  );
+};
+```
+</Code>
+
+Your message subscription logic should check for the reply metadata and handle it appropriately.
+
+When displaying a reply message, you might want to show a preview of the original message, the name of the original sender, or a visual indication that it's a reply.
+
+## Finding quoted messages that aren't in local state <a id="finding-quoted-messages"/>
```
m-hulbert, line 334: We state at the top that we're only covering replies, but then we start referencing quotes here.

src/pages/docs/guides/chat/replies.mdx
```diff
@@ -0,0 +1,803 @@
...
+
+## Implementing replies using message metadata <a id="implementing-replies"/>
+
+To implement a reply to a message, you'll need to:
+
+1. Store the original message's information when a user wants to reply to it.
+2. Include that information in the metadata when sending the reply.
+3. Process incoming messages to check for reply metadata.
+
+### Step 1: Store information about the message being replied to <a id="store-information"/>
```
m-hulbert, line 32: I think we should move away from making this step-based and just discuss the process as we normally would for another feature.

src/pages/docs/guides/chat/replies.mdx
```diff
@@ -0,0 +1,803 @@
...
+    </div>
+  );
+};
+</If>
+```
+</Code>
+
+Ably Chat doesn't provide built-in UI components or reply functionality **yet**, so you'll need to implement these aspects yourself using the metadata approach described in this guide. The steps outlined in the previous sections provide the foundation for building a complete reply system in your application.
+
+## Limitations and considerations <a id="limitations-and-considerations"/>
```
m-hulbert, line 773: This is the second section covering limitations 🙂

m-hulbert (commented):
I've just dropped some high-level comments for now on how I think we should position these types of features.

src/pages/docs/guides/chat/replies.mdx
splindsay-92: I could change this to `Finding parent messages that have been replied to...`?

src/data/nav/chat.ts
splindsay-92: Done.

src/pages/docs/guides/chat/replies.mdx
splindsay-92: I've reworded things a bit now as you recommended :)

src/pages/docs/guides/chat/replies.mdx
splindsay-92: Corrected that, thanks! :D

PR #2722, opened 2025-07-16
[EDU-1976] - Added Flutter getting started guide

GregHolmes:
Added the Flutter getting started guide for Pub/Sub

### DIFF
--- a/src/pages/docs/getting-started/flutter.mdx
+++ b/src/pages/docs/getting-started/flutter.mdx
@@ -17,64 +17,74 @@ It will take you through the following steps:
 * Retrieve the messages you sent in the guide from history.
 * Close a connection to Ably when it is no longer needed.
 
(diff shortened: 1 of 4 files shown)


src/pages/docs/getting-started/flutter.mdx
```diff
@@ -0,0 +1,758 @@
...
+<Code>
+```flutter
+// Add 10 second delay and close Ably connection
+Future.delayed(const Duration(seconds: 10), () async {
+  await AblyService().close();
+});
+```
+</Code>
+
+This ensures the connection is closed when your application terminates, freeing resources and removing listeners.
```
ttypic, line 744: This actually close connection after 10s, not when your application terminates, I would use another wording

ttypic (approved):
Looks good

PR #2738, opened 2025-07-24
[WEB-4478] fix: test for internal links when rewriting more comprehensively

jamiehenson:
When rewriting links, check that they're internal more effectively, checking for anchor links isn't enough.

Review link: https://ably-docs-web-4478-link-yfjmim.herokuapp.com/


m-hulbert (approved):
🚀

PR #2739, opened 2025-07-24
[WEB-4478] chore: add bin/dev script to set up environment and run dev server

jamiehenson:
The same script as from [this](https://github.com/ably/voltaire/pull/1126). Adds a `bin/dev` script to automate running a dev env in the same way as other repos. As with Voltaire, still on Yarn 1.


matt423 (approved):
Had thoughts about future improvements(for all our `./bin/dev` scripts really) about more config options for things such as ports but definitely one for future PRs

PR #2741, opened 2025-07-25
Update stats to reflect sent messages includes recevied

mattheworiordan:
The copy previously made it sounds like only messages sent were counted, which could be confusing for users who think this is messages sent to Ably, or sent from Ably, but not both.

See somewhat related [Slack message](https://ably-real-time.slack.com/archives/CF7GGBF2N/p1753425019299029) which surfaced this.

### DIFF
--- a/src/pages/docs/metadata-stats/stats.mdx
+++ b/src/pages/docs/metadata-stats/stats.mdx
@@ -373,10 +373,10 @@ All messages metrics include all messages types, such as those sent and received
 
 | Metric | Description |
 | --- | --- |
-| **messages.all.all.count** | Total number of messages that were successfully sent, summed over all message types and transports. |
-| **messages.all.all.billableCount** | Total number of billable messages that were successfully sent, summed over all message types and transports. |
-| **messages.all.all.data** | Total message size of all messages that were successfully sent, summed over all message types and transports. |
-| **messages.all.all.uncompressedData** | Total uncompressed message size, excluding delta compression. |
+| **messages.all.all.count** | Total number of messages that were successfully received and sent by Ably, summed over all message types and transports. |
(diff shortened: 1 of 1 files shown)


content/metadata-stats/stats.textile
```diff
@@ -375,10 +375,10 @@ h3(#messages). All messages
...
 
 |_. Metric |_. Description |
-| messages.all.all.count | Total number of messages that were successfully sent, summed over all message types and transports. |
-| messages.all.all.billableCount | Total number of billable messages that were successfully sent, summed over all message types and transports. |
-| messages.all.all.data | Total message size of all messages that were successfully sent, summed over all message types and transports. |
-| messages.all.all.uncompressedData | Total uncompressed message size, excluding delta compression. |
+| messages.all.all.count | Total number of messages that were successfully received or sent by Ably, summed over all message types and transports. |
+| messages.all.all.billableCount | Total number of billable messages that were successfully received or sent by Ably, summed over all message types and transports. |
+| messages.all.all.data | Total data in messages successfully received or sent by Ably, summed over all message types and transports. |
+| messages.all.all.uncompressedData | Total data in messages successfully received or sent by Ably, excluding savings provided delta compression. |
```
m-hulbert, line 381: Should this be 'received and sent' rather than 'or' in that case?

SimonWoolf:
Could we consider making these sort of wording improvements in the authoritative schema at https://github.com/ably/ably-common/blob/main/json-schemas/src/app-stats.json , and having a docs repo script that populates this table from that we can run every so often? So we're not making different wording improvements to the two independently.

m-hulbert:
Sorry, dropped the ball on this in the MDX conversion. 

I've rebased this to fix it for now (with my suggested change included). 

@SimonWoolf I think that's a really sensible idea. I'll add it to my list to look at standardising in the schema 👍

PR #2746, opened 2025-07-29
Update pubsub nav to match chat

zknill:
Pubsub uses "Getting started" -> "Guides", and Chat uses "Getting started" -> "Quickstarts".

image
image
image

### DIFF
--- a/src/data/nav/chat.ts
+++ b/src/data/nav/chat.ts
@@ -16,13 +16,8 @@ export default {
           link: '/docs/chat',
           index: true,
         },
-      ],
-    },
-    {
-      name: 'Getting started',
-      pages: [
         {
-          name: 'Quickstarts',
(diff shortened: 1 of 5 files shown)


GregHolmes:
Sorry for the late review on this @zknill , I've spoken among the DevEd team and we've decided to slightly change what you'd proposed in this PR. I believe Chat's nav / pages is going to be in review by Andy when he returns, but also, there's a fair bit of redesign of the nav for all products in the near future. So the changes committed are just for the interim.

What do you think?
I've also tagged @m-hulbert

m-hulbert (commented):
I think this is the most elegant solution until we fix this with the wider changes we have planned. 

Can we update the LO title to be 'Getting started' if we go this route though too, please?

PR #2753, opened 2025-08-06
Fix naming of outbound streaming overview

zknill:
The outbound streaming header has an "overview" where the page title is "Firehose overview", but we haven't mentioned firehose before.

Align the nav and page titles.


m-hulbert:
Hey @zknill 

Thanks for this. I actually merged a very large PR for integrations last week that incorporated this change as it was tidying up a lot of the terminology around integrations.

PR #2769, opened 2025-08-14
[WIP, ECO-5468] Add Swift documentation for LiveObjects

lawrence-forooghian:
### Checklist


sacOO7 (commented):
Need to resolve conflicts with `main`

src/pages/docs/liveobjects/counter.mdx
```diff
@@ -98,8 +122,34 @@ const { unsubscribe } = counter.subscribe(() => console.log(counter.value()));
...
+}
+// To remove the listener
+subscriptionResponse.unsubscribe()
+```
 </Code>
 
+<If lang="swift">
+To remove a counter update listener from _inside_ the listener function, you can call `unsubscribe()` on the subscription response that is passed as the second argument to the listener function:
+
+```swift
```
m-hulbert, line 143: This needs to be wrapped in `<Code>` tags.

src/pages/docs/liveobjects/map.mdx
```diff
@@ -94,8 +122,30 @@ const { unsubscribe } = map.subscribe(() => console.log('Map updated'));
...
+}
+// To remove the listener
+subscriptionResponse.unsubscribe()
+```
 </Code>
 
+<If lang="swift">
+To remove a map update listener from _inside_ the listener function, you can call `unsubscribe()` on the subscription response that is passed as the second argument to the listener function:
+
+```swift
```
m-hulbert, line 139: This needs to be wrapped in `<Code>` tags.

src/data/languages/languageData.ts
```diff
@@ -33,6 +33,7 @@ export default {
   },
   liveObjects: {
     javascript: '2.9',
+    swift: '0.1',
```
m-hulbert, line 36: It's a little odd that we're tying JS/Java to their Pub/Sub version but having Swift at 0.1 just because of how they're imported (just a comment).

src/pages/docs/liveobjects/counter.mdx
```diff
@@ -111,18 +161,23 @@ counter.subscribe(listener);
...
+counter.unsubscribeAll();
+```
 </Code>
 
 ## Update LiveCounter <a id="update"/>
 
-Update the counter value by calling `LiveCounter.increment()` or `LiveCounter.decrement()`. These operations are synchronized across all clients and trigger data subscription callbacks for the counter, including on the client making the request.
+Update the counter value by calling <If lang="javascript">`LiveCounter.increment()` or `LiveCounter.decrement()`</If><If lang="swift">`LiveCounter.increment(amount:)` or `LiveCounter.decrement(amount:)`</If>. These operations are synchronized across all clients and trigger data subscription callbacks for the counter, including on the client making the request.
```
m-hulbert, line 180: I'd personally keep all these occurrences as `LiveCounter.increment()` since the usage is immediately below to save ending up with 10+ theoretical if statements, unless you think it's that important to keep in.

src/data/languages/languageData.ts
lawrence-forooghian: The difference is that in JS and Java, the LiveObjects plugin is part of the core SDK and hence must have the same version. In Swift, we are forced to have a separate repository for the plugin, which has the happy side effect of allowing us to choose a semantic version number that reflects the fact that the LiveObjects API is experimental (JS and Java instead have to rely on code comments that state this).

src/pages/docs/liveobjects/map.mdx
lawrence-forooghian: Thanks, done

src/data/languages/languageData.ts
m-hulbert: Yeah, I totally understand the reasoning. Just from an external perspective it makes it look like JS and Java are far more mature than the Swift implementation.

src/pages/docs/liveobjects/counter.mdx
m-hulbert: With 10+ I was referring to if we were to implement LiveObjects in all SDKs and they all have slightly different signatures. 

I have no issues at all with keeping it, if it's the common pattern for Swift. We should extend this to other products though in that case (e.g. Chat definitely doesn't follow this format for Swift which is what made me question the need here).

sacOO7 (commented):
one more file conflict

src/pages/docs/liveobjects/counter.mdx
lawrence-forooghian: > We should extend this to other products though in that case (e.g. Chat definitely doesn't follow this format for Swift which is what made me question the need here).

Agreed — created https://github.com/ably/ably-cocoa/issues/2093 and https://github.com/ably/ably-chat-swift/issues/351

lawrence-forooghian:
@sacOO7 I've pushed again to incorporate https://github.com/ably/docs/pull/2775

umair-ably (approved):

src/data/languages/languageData.ts
m-hulbert: I think it's something we need to look at holistically that we should include on some guidance for releasing plugins. Not worth blocking this from releasing right now.

PR #2778, opened 2025-08-19
[WEB-4489] Relocate examples keys into config files hidden from sandpack

jamiehenson:
A followon to #2731. Sandpack runs what it shows, so we can't truncate the API key in the examples scripts as it will be interpreted as a malformed key. This PR moves the keys out into a separate file which is hidden from the Sandpack renderer, meaning that the apps will still work, but the keys will be hidden.


GregHolmes (approved):
This looks like a good solution to me.

PR #2811, opened 2025-09-04
[EDU-2025] - Update If component to take loggedIn prop

GregHolmes:
## Description

If Component Enhancement:
Added loggedIn prop for authentication-based content display
Shows/hides content based on user sign-in status 
Added coverage in MDXWrapper.test.tsx.


src/components/Layout/mdx/If.tsx
```diff
@@ -1,20 +1,36 @@
...
+  let shouldShow = true;
+
+  // Check language condition if lang prop is provided
+  if (lang !== undefined && language) {
+    const splitLang = lang.split(',');
+    shouldShow = shouldShow && splitLang.includes(language);
+  }
+
+  // Check logged in condition if loggedIn prop is provided
+  if (loggedIn !== undefined) {
```
jamiehenson, line 28: I would check for userContext.sessionState as well here, we may not always have that and don't want the access on the following line to error

jamiehenson (approved):
Looks good, just a small change to make first before merging

PR #2812, opened 2025-09-05
Add If loggedIn filter to API guest keys in getting started CLI section.

GregHolmes:
## Description

If the user is logged in they won't see the note that a guest key is used in the getting started guide. This is because if you're logged in the code blocks give your own API key. Where as if you're not, you'll see a demo API key, which you won't get from your dashboard when setting up the CLI.

It's difficult to replicate this locally as you need to have the docs and website set up together to test. If you want to test this and don't have it set up, please let me know and I can walk you through it.

Screenshot of CLI section of getting started guide when someone is logged in.
Screenshot 2025-09-05 at 10 08 22

Screenshot of CLI section of getting started guide when someone is **not** logged in.
Screenshot 2025-09-05 at 10 08 28


m-hulbert (commented):
Should this include Laravel, or are you updating that separately with all the new functionality?

PR #2815, opened 2025-09-08
Updated functionality on how examples use random name generator

GregHolmes:
I was unable to figure out why this isn't working. However with Cursor's assistance it seemed adamant that we're now meant to be using env variables to retrieve the channel name.

✅ Problem Solved
The issue was that the name URL parameter was undefined in the Sandpack environment because:
ExamplesRenderer.tsx: Was incorrectly passing { [languageKey]: getRandomChannelName() } instead of { CHANNEL_NAME: getRandomChannelName() } to the updateAblyConnectionKey function
chat-presence script: Was trying to read from URL parameters instead of environment variables
TypeScript errors: Missing null checks for DOM elements
✅ Changes Made
ExamplesRenderer.tsx: Fixed to pass CHANNEL_NAME as the environment variable key
config.ts: Added CHANNEL_NAME: import.meta.env.VITE_CHANNEL_NAME as string
script.ts:
Changed from urlParams.get('name') to config.CHANNEL_NAME
Added null check for onlineStatusDiv to fix TypeScript errors
Now the Sandpack environment properly receives the random channel name as an environment variable that gets replaced by the updateAblyConnectionKey function, and the chat-presence example should work correctly with the dynamically generated channel name!


GregHolmes:
> This makes sense, and makes good use of the hidden `config` files that were recently added.
> 
> I know you haven't ported everything yet, but there are a couple of examples that still use the query params approach (i.e. [this](https://github.com/ably/docs/blob/main/examples/pub-sub-history/react/src/Home.tsx#L37)), so I want to raise that in case this is significant.

There's no issue here, but thank you for bringing it to my attention. I think I would have missed that, and message encryption example.

PR #2830, opened 2025-09-15
[EDU-2044] - Remove SDK set up pages and redirect to getting started landing pages

GregHolmes:
## Description

This PR:

- Removes extra tile in Pub/Sub getting started landing page for Flutter. This was previously removed but seems to have been re-added.
- Removes both Pub/Sub and Chat SDK set up.
- Creates a Chat getting started landing page.
- Redirects the relevant SDK set up page to the getting started landing page.
- Updates SDKs page by removing the GitHub button, renames the Setup page to 'Get started' and links to the relevant getting started guide for that language.

### DIFF
--- a/src/pages/docs/chat/getting-started/index.mdx
+++ b/src/pages/docs/chat/getting-started/index.mdx
@@ -1,10 +1,10 @@
 ---
-title: "Get started with Chat"
-meta_description: "Get started with Ably Chat in your language or framework of choice. Learn how to send and receive messages, track online presence, fetch message history, implement typing indicators, among other features."
+title: "Getting started with Chat"
+meta_description: "Getting started with Ably Chat in your language or framework of choice. Learn how to send and receive messages, track online presence, fetch message history, implement typing indicators, among other features."
 meta_keywords: "Chat, Ably Chat SDKs, realtime messaging, send and receive messages, getting started guides, realtime communication, Ably tutorial, message history, presence and occupancy APIs, Ably CLI Chat"
 ---
 
-Get started with Ably Chat by choosing your language or framework.
+Getting started with Ably Chat by choosing your language or framework.
 
 You'll learn the essentials of building realtime chat applications, including how to create and manage chat rooms, send and edit messages, implement typing indicators, track user presence, retrieve message history, and send ephemeral reactions. You'll also discover how to use the Ably CLI for testing chat functionality and manage your chat applications through your Ably dashboard.
 
@@ -33,10 +33,16 @@ These are your first steps towards building a Chat application that can effortle
     link: '/docs/chat/getting-started/react-native',
   },
   {
-    title: 'Kotlin',
-    description: 'Start building Chat applications using Ably\'s Kotlin SDK.',
+    title: 'Kotlin (Android)',
+    description: 'Start building Chat applications for Android using Ably\'s Kotlin SDK.',
     image: 'icon-tech-kotlin',
-    link: '/docs/chat/getting-started/kotlin',
+    link: '/docs/chat/getting-started/android',
+  },
+  {
+    title: 'Kotlin (JVM)',
+    description: 'Start building Chat applications for JVM systems using Ably\'s Kotlin SDK.',
+    image: 'icon-tech-kotlin',
+    link: '/docs/chat/getting-started/jvm',
   },
   {
     title: 'Swift',
@@ -49,7 +55,7 @@ These are your first steps towards building a Chat application that can effortle
 
 ## React components
 
-Get started with pre-built React components that provide complete chat functionality out-of-the-box, allowing you to quickly integrate chat features into your applications.
+Getting started with pre-built React components that provide complete chat functionality out-of-the-box, allowing you to quickly integrate chat features into your applications.
 
 <Tiles>
 {[
--- a/src/pages/docs/getting-started/index.mdx
+++ b/src/pages/docs/getting-started/index.mdx
@@ -1,6 +1,6 @@
 ---
-title: "Get started with Pub/Sub"
-meta_description: "Get started with Ably Pub/Sub in your language or framework of choice. Learn how to publish, subscribe, track presence, fetch message history, and manage realtime connections."
+title: "Getting started with Pub/Sub"
+meta_description: "Getting started with Ably Pub/Sub in your language or framework of choice. Learn how to publish, subscribe, track presence, fetch message history, and manage realtime connections."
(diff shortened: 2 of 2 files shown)


src/components/SDKsPage/data.ts
```diff
@@ -211,29 +187,25 @@ export const data = {
...
-          setupLink: 'chat/setup?lang=javascript',
+          setupLink: 'chat/getting-started/javascript',
         },
         {
           title: 'React',
           text: 'Ably Chat React Hooks package.',
           image: { src: react, isWide: false },
-          githubRepoURL: 'https://github.com/ably/ably-chat-js',
-          setupLink: 'chat/setup?lang=react',
+          setupLink: 'chat/getting-started/react',
```
m-hulbert, line 196: Should we add a tile for React Native here since you wrote it?

src/pages/docs/chat/getting-started/index.mdx
```diff
@@ -0,0 +1,54 @@
...
+title: "Get started with Chat"
+meta_description: "Get started with Ably Chat in your language or framework of choice. Learn how to send and receive messages, track online presence, fetch message history, implement typing indicators, among other features."
+meta_keywords: "Chat, Ably Chat SDKs, realtime messaging, send and receive messages, getting started guides, realtime communication, Ably tutorial, message history, presence and occupancy APIs, Ably CLI Chat"
+redirect_from:
+  - /docs/chat/setup
+---
+
+Get started with Ably Chat by choosing your language or framework.
+
+You'll learn the basics, such as how to connect to Ably, send and receive messages, and manage the status of users with presence. You'll also be introduced to the Ably CLI and your Ably dashboard to interact with, and manage your apps.
```
m-hulbert, line 11: I think we should tailor this to the Chat version as you do more things, and those specific to Chat.

src/pages/docs/chat/getting-started/index.mdx
```diff
@@ -0,0 +1,54 @@
...
+    link: '/docs/chat/getting-started/kotlin',
+  },
+  {
+    title: 'Swift',
+    description: 'Start building Chat applications using Ably\'s Swift SDK.',
+    image: 'icon-tech-swift',
+    link: '/docs/chat/getting-started/swift',
+  },
+  {
+    title: 'React UI components',
```
m-hulbert, line 48: Is it worth pulling this out under a different heading with a different intro?

src/pages/docs/chat/setup.mdx
```diff
@@ -1,290 +0,0 @@
...
-
-<Aside data-type='important'>
-The examples use basic authentication to demonstrate features for convenience. In your own applications, basic authentication should never be used on the client-side as it exposes your Ably API key. Instead, use token authentication.
-</Aside>
-
-[Sign up](https://ably.com/sign-up) to Ably to create an API key in the dashboard or use the [Control API](/docs/platform/account/control-api) to create an API key programmatically.
-
-API keys and tokens have a set of [capabilities](/docs/auth/capabilities) assigned to them that specify which operations, such as subscribe or publish can be performed on which resources. To use the Chat SDK, the API key requires the following capabilities depending on which features are being used:
-
-| Feature | Capabilities |
```
m-hulbert, line 24: I think this info is lost if we don't move it elsewhere.

src/pages/docs/chat/setup.mdx
```diff
@@ -1,290 +0,0 @@
...
-        key = "{{API_KEY}}"
-        clientId = "<clientId>"
-    },
-)
-
-val chatClient = ChatClient(realtimeClient)
-```
-</Code>
-
-A [`ClientOptions`](/docs/api/realtime-sdk#client-options) object may be passed to the Pub/Sub SDK instance to further customize the connection, however at a minimum you must set an API key and provide a `clientId` to ensure that the client is [identified](/docs/auth/identified-clients).
```
m-hulbert, line 190: Similar comment here - we lose the info we have on logging for Chat, so we should move that elsewhere (maybe in connections?)

src/pages/docs/chat/connect.mdx
```diff
@@ -209,3 +209,72 @@ subscription.unsubscribe()
 <If lang="javascript,swift,kotlin">
 The discontinuity handler is accessible via the <If lang="javascript">[Room](https://sdk.ably.com/builds/ably/ably-chat-js/main/typedoc/interfaces/chat-js.Room.html#onDiscontinuity)</If><If lang="swift">[Room](https://sdk.ably.com/builds/ably/ably-chat-swift/main/AblyChat/documentation/ablychat/room)</If><If lang="kotlin">[Room](https://sdk.ably.com/builds/ably/ably-chat-kotlin/main/dokka/chat-android/com.ably.chat/-room/index.html)</If> object.
 </If>
+
+## Logging <a id="logging"/>
```
m-hulbert, line 213: I think this needs to be removed as we now duplicate it here and in the setup page.

franrob-projects:
> @franrob-projects If I understand correctly on your first point, you're recommending that we implement some sort of functionality that checks whether the lang param the user has when accessing sdk set up page should determine at the point of redirect which getting started page they go to? 

Yes that I was I was refering to.


> If you compare the two, Pub/Sub has (in exact order):
> 
> * JavaScript
> * Node.js
> * React
> * React Native
> * Kotlin
> * Swift
> * ... And the others.
> 
> Chat has (in order):
> 
> * JavaScript
> * React
> * React Native
> * Kotlin
> * Swift
> 
> I apologise if I misunderstood but was this the part you referring to in inconsistent tile ordering?

There is `[Getting started: Chat in Android]` in between `React native` and `Kotlin`.

GregHolmes:
> On the Pub/sub get started overview page, we have the title "get started", then on all these guides, we have "getting started".

Thanks for the comment on this. I've pushed a fixup to correct this in both landing pages. (Chat and Pub/Sub)

I'm still confused on what you mean:

> There is no overview page for Getting started with Chat; instead, the "overview page" for the Chat getting started guides is a list under [About Chat](https://ably.com/docs/chat).

There is a landing page for Chat getting started guides? It's the first commit in this branch/PR?
We've decided not to remove the sdk setup page for Chat at this moment as there's key authentication points in it that'll need to be brought into the ongoing authentication rewrite. So for the time being it's staying where it is, this is why there's not a redirect for Chat in this PR.

My other PR also starts working on improving the landing page (not getting started overview) for Chat product which changes the list to tiles and improves the benefits of why you should use Ably for your Chat etc.

PR #2837, opened 2025-09-16
Add client-reactions endpoint in chat http api reference

vladvelici:
https://ably.atlassian.net/browse/CHA-1099

### DIFF
--- a/static/open-specs/chat.yaml
+++ b/static/open-specs/chat.yaml
@@ -108,22 +108,26 @@ components:
         reactions:
           type: object
           description: Represents a summary of all reactions on a message
-          properties:
-            unique:
-              type: object
-              description: Represents a summary of all unique reactions on a message
-              additionalProperties:
-                $ref: '#/components/schemas/SummaryClientIdList'
-            distinct:
-              type: object
-              description: Represents a summary of all distinct reactions on a message
-              additionalProperties:
-                $ref: '#/components/schemas/SummaryClientIdList'
-            multiple:
-              type: object
-              description: Represents a summary of all multiple reactions on a message
-              additionalProperties:
-                $ref: '#/components/schemas/SummaryClientIdCounts'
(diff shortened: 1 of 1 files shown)


static/open-specs/chat.yaml
```diff
@@ -108,22 +108,26 @@ components:
...
+      description: Represents a summary of all reactions on a message
+      properties:
+        unique:
+          type: object
+          description: Represents a summary of all unique reactions on a message
+          additionalProperties:
+            $ref: '#/components/schemas/SummaryClientIdList'
+        distinct:
+          type: object
+          description: Represents a summary of all distinct reactions on a message
```
AndyTWF, line 123: This is perhaps a misleading comment and muddies the waters between this and unique (distinct reactions, unique reactions - sounds like the same thing).

Can we relate it more directly to the summary schemes / behaviour?

static/open-specs/chat.yaml
```diff
@@ -1073,3 +1077,98 @@ paths:
...
+          in: query
+          required: false
+          description: |
+            If requesting data for a client ID other than the authenticated one, provide the required client ID in this parameter.
+          schema:
+            type: string
+            example: "like"
+        - $ref: '#/components/parameters/ClientIdHeader'
+        - $ref: '#/components/parameters/AuthorizationHeader'
+        - $ref: '#/components/parameters/AcceptHeader'
```
AndyTWF, line 1124: See David's PR here: https://github.com/ably/docs/pull/2840/files

We should also add ClientIdParam

static/open-specs/chat.yaml
vladvelici: I've rephrased. I can change them to more closely match the descriptions in [this table](https://ably.com/docs/chat/rooms/message-reactions#types-of-reactions) but this is a reference so not sure how long to make it?

PR #2845, opened 2025-09-22
Add images to getting started guides

GregHolmes:
## Description

This PR adds an image of how the getting started guide looks at the beginning of the guide IF there is a UI to show in an image.


franrob-projects (approved):
LGTM. Two things to note.
- The blue text (where used) looks somewhat pixelated (see snap).
- When I click on an image to enlarge it, it opens up in a new window (which isn't the norm).

Image

src/images/content/screenshots/getting-started/pub-sub-flutter-getting-started-guide.png
m-hulbert, line 1: There's a horizontal line at the very top of this image that renders in page.

PR #2862, opened 2025-09-30
feat: generate markdown static files for LLM agent token optimization

mattheworiordan:
Generate both HTML and Markdown versions of each documentation page to optimize token usage for LLM crawlers and AI agents. Research shows that serving markdown instead of HTML can reduce token consumption by 60-80%, significantly improving efficiency and reducing costs for AI-powered tools accessing documentation.

Reference: https://x.com/cramforce/status/1972430376149913715

Internal Slack conversation: https://ably-real-time.slack.com/archives/C07C48W7K1A/p1759170942282069

## Implementation

- Added post-build hook to convert HTML pages to clean Markdown format
- Configured nginx content negotiation to serve markdown when requested
- Added validation script to ensure markdown generation completeness
- Integrated markdown generation into CI/CD pipeline
- Added UI button with markdown icon for user access (see below, common pattern in other sites)

MO screenshot 2025-09-30 at 21 33 06

Note there is a corresponding PR in the website which ensure the `Accept: text/markdown` header is used to route to the markdown file.

## Usage

**Via content negotiation (for agents/crawlers):**
```bash
curl -H "Accept: text/markdown" https://ably.com/docs/channels
```

**Direct file access:**
```bash
curl https://ably.com/docs/channels/index.md
```

**Via UI:**
Click the Markdown icon button in the "Open In" section on any page

## Technical Details

- Uses Turndown library for HTML to Markdown conversion
- Preserves code block language annotations
- Removes navigation, headers, footers and UI chrome
- Markdown files located at `/docs/{page-path}/index.md`
- Skips redirect pages (324 redirects detected)
- Successfully generates markdown for 209/210 content pages
- No frontmatter - clean markdown content only

### DIFF
--- a/bin/validate-markdown.ts
+++ b/bin/validate-markdown.ts
@@ -1,0 +1,113 @@
+#!/usr/bin/env node
+
+/**
+ * Validates that markdown files exist for all HTML pages in the public directory.
+ * This script ensures the markdown generation process completed successfully.
+ */
+
+import * as fs from 'fs';
+import * as path from 'path';
+import fastGlob from 'fast-glob';
+
+const publicDir = path.join(process.cwd(), 'public', 'docs');
+
+interface ValidationResult {
+  totalPages: number;
+  markdownFound: number;
+  markdownMissing: number;
+  redirectPages: number;
+  missingFiles: string[];
+}
+
+const validateMarkdownFiles = async (): Promise<ValidationResult> => {
+  // Find all index.html files in the docs directory
+  const htmlFiles = await fastGlob('**/index.html', {
+    cwd: publicDir,
+    absolute: false,
+  });
+
+  const result: ValidationResult = {
+    totalPages: htmlFiles.length,
+    markdownFound: 0,
+    markdownMissing: 0,
+    redirectPages: 0,
+    missingFiles: [],
+  };
+
+  for (const htmlFile of htmlFiles) {
+    // Get the directory of the HTML file
+    const dir = path.dirname(htmlFile);
+
+    // Check if this is a redirect page (skip validation for these)
+    const htmlPath = path.join(publicDir, htmlFile);
+    const htmlContent = fs.readFileSync(htmlPath, 'utf8');
+
+    if (htmlContent.length < 1000 && htmlContent.includes('window.location.href')) {
+      result.redirectPages++;
+      continue; // Skip redirect pages
+    }
+
+    // Check if corresponding markdown file exists
+    const markdownFile = path.join(publicDir, dir, 'index.md');
+
+    if (fs.existsSync(markdownFile)) {
+      result.markdownFound++;
+
+      // Verify the markdown file has content
+      const stats = fs.statSync(markdownFile);
+      if (stats.size === 0) {
+        console.warn(`⚠️  Warning: ${dir}/index.md is empty`);
+      }
+    } else {
+      result.markdownMissing++;
+      result.missingFiles.push(dir);
+    }
+  }
+
+  return result;
+};
+
+const main = async () => {
+  console.log('🔍 Validating markdown files...\n');
+
+  if (!fs.existsSync(publicDir)) {
+    console.error(`❌ Error: Public docs directory not found: ${publicDir}`);
+    console.error('   Make sure to run this script after the build process.');
+    process.exit(1);
+  }
+
+  try {
+    const result = await validateMarkdownFiles();
+
+    console.log(`📊 Validation Results:`);
+    console.log(`   Total HTML pages: ${result.totalPages}`);
+    console.log(`   🔀 Redirect pages (skipped): ${result.redirectPages}`);
+    console.log(`   📄 Content pages: ${result.totalPages - result.redirectPages}`);
+    console.log(`   ✅ Markdown files found: ${result.markdownFound}`);
+    console.log(`   ❌ Markdown files missing: ${result.markdownMissing}`);
+
+    if (result.markdownMissing > 0) {
+      console.log('\n⚠️  Missing markdown files:');
+      result.missingFiles.slice(0, 10).forEach((file) => {
+        console.log(`   - ${file}/index.md`);
+      });
+
+      if (result.missingFiles.length > 10) {
+        console.log(`   ... and ${result.missingFiles.length - 10} more`);
+      }
+
+      console.log('\n❌ Validation failed: Some markdown files are missing.');
+      console.log('   This may indicate an issue with the markdown generation process.');
+      process.exit(1);
+    }
+
+    // Calculate coverage percentage
+    const coverage = (result.markdownFound / result.totalPages) * 100;
+    console.log(`\n✅ Validation passed! Markdown coverage: ${coverage.toFixed(1)}%`);
+  } catch (error) {
+    console.error('❌ Error during validation:', error);
+    process.exit(1);
+  }
+};
+
+main();
--- a/data/onPostBuild/generateMarkdown.ts
+++ b/data/onPostBuild/generateMarkdown.ts
@@ -1,0 +1,214 @@
+import { GatsbyNode } from 'gatsby';
+import * as path from 'path';
+import * as fs from 'fs';
+import TurndownService from 'turndown';
+import cheerio from 'cheerio';
+import fastGlob from 'fast-glob';
+
+/**
+ * This script generates Markdown static files alongside HTML files for each page.
+ * This allows LLM crawlers and other clients to request markdown versions of pages
+ * which are significantly more token-efficient than HTML.
+ */
+
+const REPORTER_PREFIX = 'generateMarkdown:';
+
+// Configure Turndown for documentation-friendly markdown
+const createTurndownService = () => {
+  const turndownService = new TurndownService({
+    headingStyle: 'atx',
+    codeBlockStyle: 'fenced',
+    bulletListMarker: '-',
+    emDelimiter: '_',
+    strongDelimiter: '**',
+    linkStyle: 'inlined',
+    linkReferenceStyle: 'full',
+  });
+
+  // Preserve code block language annotations
+  turndownService.addRule('fencedCodeBlock', {
+    filter: (node) => {
+      return node.nodeName === 'PRE' && node.firstChild && node.firstChild.nodeName === 'CODE';
+    },
+    replacement: (content, node) => {
+      const codeNode = node.firstChild as HTMLElement;
+      const className = codeNode.getAttribute('class') || '';
+      const languageMatch = className.match(/language-(\w+)/);
+      const language = languageMatch ? languageMatch[1] : '';
+
+      // Get the actual code content
+      const code = codeNode.textContent || '';
+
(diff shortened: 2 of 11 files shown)


src/components/Layout/RightSidebar.tsx
```diff
@@ -321,6 +321,34 @@ const RightSidebar = () => {
...
+                content="View in Markdown"
+                triggerElement={
+                  <img
+                    src="/icons/markdown-mark.svg"
+                    alt="View in Markdown"
+                    className="w-5 h-5 transition-opacity opacity-60 group-hover/markdown-link:opacity-100"
+                  />
+                }
+              >
+                View in Markdown
```
jamiehenson, line 349: "View in Markdown" is inconsistent with the rest of the tooltips here as the surrounding context is "Open in ...", so "Open in View in Markdown" isn't right. Just "Markdown" would be better here

src/components/Layout/RightSidebar.tsx
```diff
@@ -321,6 +321,34 @@ const RightSidebar = () => {
...
               </a>
             ))}
+            <a
+              href={`${location.pathname.replace(/\/$/, '')}/index.md`}
+              className="flex h-5 ui-theme-dark group/markdown-link cursor-pointer"
+              onClick={(e) => {
+                // In development mode, markdown files aren't generated, so show alert immediately
+                if (process.env.NODE_ENV === 'development') {
+                  e.preventDefault();
+                  alert('Markdown files are only available in production builds. Run "yarn build" to generate them.');
```
jamiehenson, line 331: Alerts are pretty poor UX, but given this is just for development, it's not a view I'll push strongly here. I would clarify that `yarn serve` is needed in addition to `yarn build` to run a local "production" version

src/components/Layout/RightSidebar.tsx
```diff
@@ -321,6 +321,34 @@ const RightSidebar = () => {
                 </Tooltip>
               </a>
             ))}
+            <a
+              href={`${location.pathname.replace(/\/$/, '')}/index.md`}
+              className="flex h-5 ui-theme-dark group/markdown-link cursor-pointer"
+              onClick={(e) => {
+                // In development mode, markdown files aren't generated, so show alert immediately
```
jamiehenson, line 328: Let's get rid of this comment, the code is descriptive enough

src/components/Layout/RightSidebar.tsx
```diff
@@ -321,6 +321,34 @@ const RightSidebar = () => {
                 </Tooltip>
               </a>
             ))}
+            <a
+              href={`${location.pathname.replace(/\/$/, '')}/index.md`}
+              className="flex h-5 ui-theme-dark group/markdown-link cursor-pointer"
+              onClick={(e) => {
```
jamiehenson, line 327: Consider opening this in a new tab like the other links. It's not external, but it is a disruptive loss of context to replace the window you're on

bin/validate-markdown.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  for (const htmlFile of htmlFiles) {
+    // Get the directory of the HTML file
+    const dir = path.dirname(htmlFile);
+
+    // Check if this is a redirect page (skip validation for these)
+    const htmlPath = path.join(publicDir, htmlFile);
+    const htmlContent = fs.readFileSync(htmlPath, 'utf8');
+
+    if (htmlContent.length < REDIRECT_PAGE_MAX_SIZE && htmlContent.includes('window.location.href')) {
```
jamiehenson, line 48: This seems flimsy to me. Looking at the generated html, we have a pretty consistent case for this - something like 
```
<script>window.location.href="/docs/auth"</script>
```
We can look for this directly.

bin/validate-markdown.ts
```diff
@@ -0,0 +1,116 @@
...
+      result.redirectPages++;
+      continue; // Skip redirect pages
+    }
+
+    // Check if corresponding markdown file exists
+    const markdownFile = path.join(publicDir, dir, 'index.md');
```
jamiehenson, line 54: More relevant for the generation script, but, and this is grounded more in my own opinion than anything, `/docs/thing/index.md` is less ergonomic to me than `/docs/thing.md`. We get away with it with HTML, but having "index" in a URL is a dated practice.

If you agree, the upshot is that the only thing that would need to happen is to rename the generated MD file and move it a directory up

data/onPostBuild/generateMarkdown.ts
```diff
@@ -0,0 +1,210 @@
...
+/**
+ * This script generates Markdown static files alongside HTML files for each page.
+ * This allows LLM crawlers and other clients to request markdown versions of pages
+ * which are significantly more token-efficient than HTML.
+ */
+
+const REPORTER_PREFIX = 'generateMarkdown:';
+
+// Constants for content validation
+const REDIRECT_PAGE_MAX_SIZE = 1000; // Maximum size in bytes for redirect pages
```
jamiehenson, line 17: As referred to in the other comment, this seems flimsy

data/onPostBuild/generateMarkdown.ts
```diff
@@ -0,0 +1,210 @@
...
+const MIN_CONTENT_LENGTH = 100; // Minimum content length to consider meaningful
+
+// Configure Turndown for documentation-friendly markdown
+const createTurndownService = () => {
+  const turndownService = new TurndownService({
```
jamiehenson, line 22: No issue with the Turndown choice, but I don't know the basis behind this config - worth checking that it's a "standard" one to assist with LLM comprehension

data/onPostBuild/generateMarkdown.ts
```diff
@@ -0,0 +1,210 @@
...
+    codeBlockStyle: 'fenced',
+    bulletListMarker: '-',
+    emDelimiter: '_',
+    strongDelimiter: '**',
+    linkStyle: 'inlined',
+    linkReferenceStyle: 'full',
+  });
+
+  // Preserve code block language annotations
+  turndownService.addRule('fencedCodeBlock', {
```
jamiehenson, line 33: Not sure if the related bit, but we should scrub out code block lines in generated MD. We get nice blocks surrounded by triple back-ticks, but then a series of numbers 1-N before it, each on their own lines

data/onPostBuild/generateMarkdown.ts
```diff
@@ -0,0 +1,210 @@
...
+      const language = languageMatch ? languageMatch[1] : '';
+
+      // Get the actual code content
+      const code = codeNode.textContent || '';
+
+      return '\n\n```' + language + '\n' + code + '\n```\n\n';
+    },
+  });
+
+  // Remove navigation, headers, footers, and other UI elements
```
jamiehenson, line 50: I know this is how Claude operates, but we should remove very obvious comments, they're just redundant.

data/onPostBuild/generateMarkdown.ts
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    const html = fs.readFileSync(htmlPath, 'utf8');
+
+    // Check if this is a redirect page (very small file with window.location.href)
+    if (html.length < REDIRECT_PAGE_MAX_SIZE && html.includes('window.location.href')) {
+      return null; // Skip redirect pages
+    }
+
+    const $ = cheerio.load(html);
```
jamiehenson, line 70: I almost had to break glass in case of jQuery here 😄

data/onPostBuild/generateMarkdown.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    // Remove unwanted elements
+    $('nav, header, footer, script, style, noscript, .sidebar, .navigation').remove();
```
jamiehenson, line 73: I don't think sidebar and navigation exist as selectors to kill, not sure if there was another intention

data/onPostBuild/generateMarkdown.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    // If we can't find a main content area, fall back to body but remove header/footer
+    if (!mainContent) {
+      $('body > header, body > footer, body > nav').remove();
+      mainContent = $('body').html();
+    }
+
+    // Check if content is meaningful (more than just whitespace/empty tags)
+    if (mainContent && mainContent.trim().length < MIN_CONTENT_LENGTH) {
+      return null; // Skip pages with minimal content
```
jamiehenson, line 87: If we skip stuff for certain pages, how do we handle this in the frontend? The Open in Markdown button is available all the time, but the pages may not be

data/onPostBuild/generateMarkdown.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    // Determine the HTML file path
+    const htmlPath = path.join(publicDir, 'docs', slug, 'index.html');
+
+    // Extract main content from HTML
+    const htmlContent = extractMainContent(htmlPath);
+
+    if (!htmlContent) {
+      reporter.warn(`${REPORTER_PREFIX} Could not extract content for ${slug}`);
+      failCount++;
```
jamiehenson, line 187: Is this a failure? The asset is not generated indeed but there's a difference between "this failed" and "the content has been skipped"

jamiehenson (changes requested):
Few things to look into here. The actual MD generation looks decent, obvs it's not optimised for human reading (and is mostly to get token consumption down), and it could be cleaned up a bit more - but it's not super necessary.

src/components/Layout/RightSidebar.tsx
mattheworiordan: I don't agree, because you're not opening in Markdown. I didn't want to fundamentally change the UI, so felt this is in fact better.

src/components/Layout/RightSidebar.tsx
mattheworiordan: Fair

data/onPostBuild/generateMarkdown.ts
mattheworiordan: Fine

data/onPostBuild/generateMarkdown.ts
mattheworiordan: I think it is a failure because we're assuming all pages do have content (your point above). So I think this is valid, unless we have conditional logic for when we show the View Markdown option

data/onPostBuild/generateMarkdown.ts
kennethkalmer: Yeah this is the one we just need to nail down. My next move in #2657 was to solve this, and I was going to go down the path of generating files with double extensions, ie `thing.javascript.md` or `thing.swift.md`, rather than using `-`.

The other thing is that it is not just code blocks, Mark can use a component to restrict pieces of copy to only show up under certain languages... My (untested) idea was to load up the page with jsdom and manipulate the selected language so the _whole_ DOM is updated correctly before passing the markup to Turndown...

kennethkalmer: @mattheworiordan I'm not sure if you saw my earlier comment here

m-hulbert:
This has been superseded by https://github.com/ably/docs/pull/3000

PR #2886, opened 2025-10-15
Migrate Realtime SDK API spec channel-metadata page to MDX

GregHolmes:
## Description

This PR converts the Realtime SDK channel metadata API reference page from Textile format to MDX, as part of a full textile to MDX migration.


m-hulbert (approved):
Needs the tags in the codeblock but LGTM other than that.

PR #2942, opened 2025-11-03
[EDU-2112] - Migrate REST SDK API spec Push-Admin page to MDX

GregHolmes:
## Description

This PR converts the REST SDK Push-Admin API reference page from Textile format to MDX, as part of a full textile to MDX migration.


src/pages/docs/api/rest-sdk/push-admin.mdx
```diff
@@ -0,0 +1,1553 @@
...
+
+`subscribeDevice(): Promise<void>`
+
+</If>
+
+Subscribe your device to the channel's push notifications.
+
+<If lang="javascript,nodejs">
+
+##### Returns
```
franrob-projects: As with the original. There are lots of the same "returns" section. Even more so in this part of the page

Image

PR #2946, opened 2025-11-04
Add docs for message updates and deletes

SimonWoolf:
Main new page: /docs/messages/updates-deletes

(but the majority of the diff is in the api docs, which, sorry to complain, seem to be getting increasingly painful to make any changes to -- there now seem to be six near-duplicate copies of the Message fields table in various places which needed updating and I wouldn't be surprised if there were more I missed)


GregHolmes:
Branched and merged in: https://github.com/ably/docs/pull/2970

PR #2973, opened 2025-11-21
[WEB-4788] Fix DOMParser SSR build error in changelog component

kennethkalmer:
## Summary
Fixes the DOMParser build error that was preventing `yarn build` from completing successfully.

## Problem
The `stripHtml` function in the changelog component was using `DOMParser`, which is only available in the browser. This caused the build to fail during static HTML generation with:
```
Error parsing changelog content: ReferenceError: DOMParser is not defined
```

## Solution
- Added browser environment check to detect SSR vs browser context
- During SSR/build: Use cheerio for HTML parsing (loaded via `require()` to prevent browser bundle inclusion)
- In browser: Use native `DOMParser` for optimal performance
- Changelog descriptions now render correctly in both SSR and browser with no bundle size impact

## Test plan
- ✅ Build completes successfully (`yarn build`)
- ✅ All 240 pages generate without errors
- ✅ Cheerio excluded from browser bundle (uses conditional `require()`)
- ✅ Changelog entries display properly with descriptions

### DIFF
--- a/src/components/Homepage/Changelog.tsx
+++ b/src/components/Homepage/Changelog.tsx
@@ -84,6 +84,16 @@ const stripHtml = (html: string | null | undefined): string => {
     return '';
   }
 
+  // Check if we're in a browser environment
+  if (typeof window === 'undefined' || typeof DOMParser === 'undefined') {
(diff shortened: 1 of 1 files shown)


GregHolmes (approved):
Tested and the error was gone!

PR #3215, opened 2026-02-19
[DX-475] Updated docs section for handling self published message

sacOO7:
- Added guidence for avoiding message duplication after sending self message.
- Fixed https://ably.atlassian.net/browse/DX-475
- `Deduplicate with optimistic UI` section can be updated with `echoMessages=false` if https://ably.atlassian.net/browse/DX-456 is fixed.
 

## Summary by CodeRabbit

* **Documentation**
  * Added a new section on handling self-published messages in chat, describing two approaches—wait for the subscriber and optimistic UI with deduplication—and including sample patterns and snippets for JavaScript, React, Swift, Kotlin, and Android.
  * Minor Kotlin code sample correction: adjusted a message list type signature.

### DIFF
--- a/src/pages/docs/chat/rooms/messages.mdx
+++ b/src/pages/docs/chat/rooms/messages.mdx
@@ -228,6 +228,90 @@ fun MyComponent(room: Room) {
 ```
 </Code>
 
+### Handle self-published messages <a id="self-published-messages"/>
+
+<Aside data-type='important'>
+When you send a message using `send()`, the server echoes it back to all subscribers in the room, including the sender. 
+If your application adds the message to the UI immediately after calling `send()` and also appends it when received via `subscribe()`, the message will appear twice.
+To avoid this, we need to add safety check in the subscribe method that validates incoming message serial and version against existing messages.
+</Aside>
+
+<Code>
+```javascript
+const {unsubscribe} = room.messages.subscribe((event) => {
+  // Early return if a message with the same serial and version.serial already exists
+  const existingMessage = myMessageList.find(msg => msg.serial === event.message.serial);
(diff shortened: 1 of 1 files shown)


src/pages/docs/chat/rooms/messages.mdx
```diff
@@ -230,6 +228,98 @@ fun MyComponent(room: Room) {
 ```
 </Code>
 
+### Handle self-published messages <a id="self-published-messages"/>
+
+When you send a message using `send()`, the server echoes it back to all subscribers in the room, including the sender. If your application adds the message to the UI immediately before/after calling `send()` and also appends it when received via `subscribe()`, the message will appear twice. There are two approaches to handle this.
+
+#### Wait for the subscriber <a id="wait-for-subscriber"/>
+
+The recommended approach is to not add the message to the UI immediately before/after calling `send()`. Instead, only append messages to the UI inside the `subscribe()` listener. Since the server echoes every sent message back to the sender as a subscriber event, the message will still appear in the UI when it arrives through the subscription. This eliminates the duplication problem entirely and requires no deduplication logic in the subscriber.
```
splindsay-92: Users still have to contend with cross-region publishes, so they still need to handle ordering. They also need to apply updates/deletes, which means existence checks. At this point, the deduplication logic is essentially mostly implemented (and we recommend to use the `.with()` function to handle this in JS land). 

I agree with handling concurrent writes to a data structure being a concern, but I don't think this means we should recommend in all cases for non-optimistic updates.

splindsay-92 (commented):
Just a few of my comments still to address. The biggest one is around recommending optimistic updates on `message.creates`, generally I don't think we should recommend against it.

PR #3228, opened 2026-02-26
[DX-285] React Native push notifications guide

maratal:
## Description

React Native push notifications tutorial.

https://ably.atlassian.net/browse/DX-285

### Checklist


src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,610 @@
...
+```xml
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+    ...
+</manifest>
+```
+</Code>
+
+<Aside>
```
GregHolmes: This has no data-type attribute. All asides should. It could just be <Aside data-type='note'>

GregHolmes (changes requested):
Looks good! Just a couple comments.

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,610 @@
...
+At the time of writing React Native 0.84's Gradle plugin is not compatible with Gradle 9. If you see a `JvmVendorSpec.IBM_SEMERU` error during the Android build, update `android/gradle/wrapper/gradle-wrapper.properties` to use Gradle 8.13: `distributionUrl=https\://services.gradle.org/distributions/gradle-8.13-bin.zip`.
```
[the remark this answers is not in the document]
maratal: I just removed version number.

src/pages/docs/push/getting-started/react-native.mdx
maratal: Fixed, added `data-type='note'`.

splindsay-92 (commented):
Just an initial comment again on if we should be using ably-js react hooks here. Also, would it make this whole guide a bit simpler/shorted if we made use of the dashboards ability to send notifications to device/clientId/Channel? We could still have basic code segments to show how its done, but there'd be no need for the additional UI/handling (also this page is due for a refurb). 

I won't block the guide based on the current design though, but just a thought that might make things cleaner :)

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,570 @@
...
+At the time of writing React Native Gradle plugin is not compatible with Gradle 9. If you see a `JvmVendorSpec.IBM_SEMERU` error during the Android build, update `android/gradle/wrapper/gradle-wrapper.properties` to use Gradle 8.13: `distributionUrl=https\://services.gradle.org/distributions/gradle-8.13-bin.zip`.
+</Aside>
+
+#### Configure iOS project <a id="configure-ios"/>
+
+Open `ios/PushTutorial.xcworkspace` in Xcode and add the `Push Notifications` capability: select your target, go to **Signing & Capabilities**, and click **+ Capability**.
+
+Add `use_modular_headers!` to `ios/Podfile` after `prepare_react_native_project!`:
+
+```ruby
```
m-hulbert, line 106: I don't think this should be ruby?

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,570 @@
...
+```xml
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+    ...
+</manifest>
+```
+
+<Aside data-type='note'>
+At the time of writing React Native Gradle plugin is not compatible with Gradle 9. If you see a `JvmVendorSpec.IBM_SEMERU` error during the Android build, update `android/gradle/wrapper/gradle-wrapper.properties` to use Gradle 8.13: `distributionUrl=https\://services.gradle.org/distributions/gradle-8.13-bin.zip`.
```
m-hulbert: We should avoid things like 'at time of writing' etc. Can we just state we should use 8.13 instead if it's still not supported in 9?

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,570 @@
...
+
+`activatePush()` does the following:
+
+1. Requests notification permission from the user.
+2. Obtains the FCM registration token from Firebase.
+3. Registers the device with Ably's push notification service using the token.
+
+After successful activation, `deviceId` contains the unique device ID assigned to this installation.
+
+<Aside data-type='note'>
```
m-hulbert: Do we need all this exposition? Three notes in a row suggests this is way too complex. Do we need to explain which methods are used and why?

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,570 @@
...
+  FirebaseApp.configure()
+  // ... rest of existing setup
+}
+```
+
+Add all further code to `App.tsx`.
+
+## Step 1: Set up Ably <a id="step-1"/>
+
+Replace the contents of `App.tsx` with the following to initialize the Ably Realtime client, wrap the app in `AblyProvider` and `ChannelProvider`, and subscribe to a channel for incoming messages using the `useChannel` hook:
```
m-hulbert: Should we keep this to the Pub/Sub rather than a 'realtime client' so people know what it is?

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,570 @@
...
+    tokenRefreshUnsubscribeRef.current = null;
+    showStatus('Push notifications deactivated.');
+  } catch (error: any) {
+    showStatus(`Failed to deactivate push: ${error.message}`);
+  }
+}
+```
+</Code>
+
+The `transportType` is set to `fcm` on both platforms because `messaging().getToken()` always returns an FCM registration token, even on iOS. Firebase exchanges the APNs device token for an FCM token internally, so Ably communicates with Firebase rather than APNs directly (instead Firebase sends push notifications to iOS devices via APNs).
```
m-hulbert: Do we need to explain this? Would someone know this if they're React Native devs or familiar with Push? Maybe even a short code comment for iOS would be sufficient?

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,570 @@
...
+On Android, use `notifee.requestPermission()` instead of React Native's `PermissionsAndroid` because it handles the `POST_NOTIFICATIONS` permission consistently across Android API levels. On iOS, `messaging().requestPermission()` shows the system notification permission dialog; Firebase Messaging calls the native iOS permission APIs internally.
+</Aside>
+
+<Aside data-type='note'>
+In React Native `realtime.push.activate()` is not used because it's a JS SDK and it registers browser for push notifications, which is not what you want in a React Native app. Instead, use `realtime.push.admin.deviceRegistrations` API to register the device.
+</Aside>
+
+## Step 3: Subscribe and test push notifications <a id="step-3"/>
+
+The FCM SDK handles background push notifications automatically and displays them as system notifications. For foreground handling, use `@notifee/react-native` to display notifications while the app is open.
```
m-hulbert, line 329: Should we choose one approach or the other? Or explain how to use both.

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,570 @@
...
+    showStatus(`Unsubscribed from push on channel: ${CHANNEL_NAME}`);
+  } catch (error: any) {
+    showStatus(`Failed to unsubscribe: ${error.message}`);
+  }
+}
+```
+</Code>
+
+<Aside data-type='note'>
+Sending push notifications using `deviceId` or `clientId` requires the `push-admin` capability for your API key. Use this method for testing purposes. In production, send push notifications from your backend server by publishing messages with a `push` `extras` field to a channel.
```
m-hulbert: Could the capabilities be covered in the pre-reqs rather than mid-guide? Which method are we then referring to here?

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,570 @@
...
+```
+</Code>
+
+You can also open each platform project in their respective IDEs and run from there.
+
+Tap **Activate Push** and wait until the status message displays your device ID. Try sending a test push notification using the Ably CLI commands shown in Step 3.
+
+### Send push via channel <a id="step-4-send-channel"/>
+
+To test push notifications via channel, tap **Subscribe to Channel** in the app and then send a push notification to the channel using the Ably CLI:
```
m-hulbert, line 517: There's a lot of 'testing' here and do we need both options? I think I agree with Steven that we should either use the dashboard (or more likely the CLI) for pushing the notifications as you wouldn't tend to be sending via the app in reality.

m-hulbert (commented):
This is getting there Marat. I've just put a few comments in, but I'd say in general we need to be opinionated on a method rather than consistently offer multiple options. Also agree with @splindsay-92 about keeping this simpler where it makes sense / reflects real-world usage.

src/pages/docs/push/getting-started/react-native.mdx
maratal: Changed to `text`.

src/pages/docs/push/getting-started/react-native.mdx
maratal: I've meant that if a reader wants to display pushes inside their app (foreground), they should use notifee (code example already shows how). If app in the background though, it will be handled by Android itself and displayed in the Notification Center.

src/pages/docs/push/getting-started/react-native.mdx
maratal: I'm now thinking that it's an insignificant implementation detail which can be easily fixed by android developer familiar with android studio enough to dare implement pushes. I'm removing this.

src/pages/docs/push/getting-started/react-native.mdx
maratal: Moved two of them in a short code comment.

src/pages/docs/push/getting-started/react-native.mdx
maratal: fixed

src/pages/docs/push/getting-started/react-native.mdx
maratal: Moved to code comment (I'm not sure what you've meant by "for iOS" tho, it's cross-platform js code snippets).

src/pages/docs/push/getting-started/react-native.mdx
maratal: Moved to Prerequisites.

src/pages/docs/push/getting-started/react-native.mdx
maratal: Not sure I follow, do you mean to remove code snippet for js and only leave CLI? (js one was Steven suggestion.)

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,557 @@
...
+
+This guide will get you started with Ably Push Notifications in a new React Native application.
+
+You'll learn how to set up your application with Firebase Cloud Messaging (FCM), register devices with Ably, send push notifications, subscribe to channel-based push, and handle incoming notifications on both iOS and Android.
+
+## Prerequisites <a id="prerequisites"/>
+
+1. [Sign up](https://ably.com/signup) for an Ably account.
+2. Create a [new app](https://ably.com/accounts/any/apps/new), and create your first API key in the **API Keys** tab of the dashboard.
+3. Your API key needs the `publish`, `subscribe`. Also add `push-admin` capabilities to test sending push notifications (in production, use your backend server to send push notifications by publishing messages with an `extras.push` field to a channel).
```
m-hulbert:
```suggestion
  * Your API key needs the `publish`, `subscribe` capabilities. 
  * Also add the `push-admin` capability if you're using the same API key to send a push notification. In production this would more likely be a server using a different API key.
```

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,557 @@
...
+4. For channel-based push, add a rule for the channel with **Push notifications enabled** checked. In the dashboard left sidebar: **Configuration** → **Rules** → **Add** or **Edit** a rule, then enable the Push notifications option. See [channel rules](/docs/channels#rules) for details.
```
m-hulbert:
```suggestion
3. Add a rule to a channel so you can test sending push notification via a channel. Select [**Rules**](https://ably.com/accounts/any/apps/any/app_namespaces) in the Ably dashboard, add a new rule and enable the **Push notifications** option.
```

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,557 @@
...
+cd PushTutorial
+npm install ably @react-native-firebase/app @react-native-firebase/messaging @notifee/react-native
+```
+</Code>
+
+#### Configure Android project <a id="configure-android"/>
+
+Apply the Google Services plugin in `android/build.gradle`:
+
+```kotlin
```
m-hulbert: Missing `<Code>` tags on a view code blocks here.

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,557 @@
...
+  TouchableOpacity,
+  View,
+} from 'react-native';
+import {SafeAreaView} from 'react-native-safe-area-context';
+import * as Ably from 'ably';
+import {AblyProvider, ChannelProvider, useChannel, useAbly} from 'ably/react';
+import messaging from '@react-native-firebase/messaging';
+import notifee, {AuthorizationStatus} from '@notifee/react-native';
+
+const CHANNEL_NAME = 'exampleChannel1';
```
m-hulbert: Can we set this to something like `my-first-push-channel` everywhere for consistency with the other getting started guides?

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,557 @@
...
+# iOS
+npx react-native run-ios --device
+```
+</Code>
+
+You can also open each platform project in their respective IDEs and run from there.
+
+Tap **Activate Push** and wait until the status message displays your device ID. Try sending a test push notification using the Ably CLI commands shown in Step 3.
+
+### Send push via channel <a id="step-5-send-channel"/>
```
m-hulbert: Should this be a step 6 in its own right rather than a sub section of building your UI? 

Then we could have sub-sections for publishing via a channel and publishing directly to a device (or client). This might make more sense than having the publish directly to a client part under "test admin notifications" - which we should definitely rename as well.

m-hulbert:
```suggestion
## Step 6: Publish a push notification via a channel <a id="step-5-send-channel"/>
```

We should probably use 'publish' throughout since that's what the existing docs use.

src/pages/docs/push/getting-started/react-native.mdx
m-hulbert: I was referring to showing how to publish via code and also using the CLI. Since this is a simple getting started guide I think we should show 1 or the other, rather than all the options. I'd suggest using the CLI for simplicity (and then you could note or link to the code equivalent in the existing docs).

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,557 @@
...
+```
+
+Add all further code to `App.tsx`.
+
+## Step 1: Set up Ably <a id="step-1"/>
+
+Replace the contents of `App.tsx` with the following to initialize the Ably Pub/Sub client, wrap the app in `AblyProvider` and `ChannelProvider`, and subscribe to a channel for incoming messages using the `useChannel` hook:
+
+<Code>
+```javascript
```
splindsay-92: shouldn't all these code blocks be react, since this is a react native guide?

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,557 @@
...
+
+    await messaging().registerDeviceForRemoteMessages(); // Required to receive push notifications on iOS, no-op on Android
+    const fcmToken = await messaging().getToken();
+    await saveDeviceRegistration(fcmToken);
+
+    tokenRefreshUnsubscribeRef.current?.();
+    tokenRefreshUnsubscribeRef.current = messaging().onTokenRefresh(async newToken => {
+      try {
+        await saveDeviceRegistration(newToken);
+      } catch (error: any) {
```
splindsay-92: I believe the call here should return an errorInfo, perhaps worth making that clear in a cast?

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,557 @@
...
+  };
+}, []);
+```
+</Code>
+
+Use the Ably CLI to send a test push notification to your client ID:
+
+<Code>
+```shell
+ably push publish --client-id push-tutorial-client \
```
splindsay-92: What version of the cli are you using here? 👀

splindsay-92 (commented):
Just a few small comments

src/pages/docs/push/getting-started/react-native.mdx
maratal: The point of "Step 3: Test admin push notifications" was so user could see pushes at work rather early in the guide. WDYT @m-hulbert Also what would be the alternative name? (because user should understand that this is admin push, not real app use case)

src/pages/docs/push/getting-started/react-native.mdx
maratal: I'm using `client-id` flag, which is described here - https://ably.com/docs/cli/push/publish#client-id
Or what do you mean?

src/pages/docs/push/getting-started/react-native.mdx
maratal: done

src/pages/docs/push/getting-started/react-native.mdx
maratal: done

src/pages/docs/push/getting-started/react-native.mdx
maratal: fixed

src/pages/docs/push/getting-started/react-native.mdx
maratal: done

src/pages/docs/push/getting-started/react-native.mdx
maratal: done

src/pages/docs/push/getting-started/react-native.mdx
maratal: done

src/pages/docs/push/getting-started/react-native.mdx
maratal: fixed

src/pages/docs/push/getting-started/react-native.mdx
maratal: done

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,520 @@
...
+```
+</Code>
+
+## Step 3: Subscribe and test push notifications <a id="step-3"/>
+
+### Subscribe to channel push notifications <a id="step-3-subscribe"/>
+
+To subscribe your device to a channel so it can receive channel-based push notifications, add the following functions inside `PushScreen`, after the functions from Step 2:
+
+<Code fixed="true">
```
m-hulbert: You only need `fixed=true` for things like `text` - not for every code block.

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,520 @@
...
+  return () => {
+    unsubscribe();
+  };
+}, []);
+```
+</Code>
+
+## Step 3: Subscribe and test push notifications <a id="step-3"/>
+
+### Subscribe to channel push notifications <a id="step-3-subscribe"/>
```
m-hulbert:
```suggestion
## Subscribe to channel push notifications <a id="step-3"/>
```

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,520 @@
...
+  } catch (error) {
+    showStatus(`Failed to unsubscribe: ${(error as Ably.ErrorInfo).message}`);
+  }
+}
+```
+</Code>
+
+### Build the UI <a id="step-3-ui"/>
+
+Update the `return` statement in `PushScreen` to add buttons that call all the push functions:
```
m-hulbert:
```suggestion
### Build the UI <a id="step-3-ui"/>

Build a UI in your app to add buttons that call all push functions. Update the `return` statement in `PushScreen`:
```

Do you think this should just be a step in its own right? That would make this step 4 and then publish step 5?

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,520 @@
...
+npx react-native run-android
+
+# iOS
+npx react-native run-ios --device
+```
+</Code>
+
+You can also open each platform project in their respective IDEs and run from there.
+
+### Publish a push notification <a id="step-3-publish"/>
```
m-hulbert:
```suggestion
## Step 4: Publish a push notification <a id="step-4"/>
```

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,520 @@
...
+
+In the app tap **Activate Push** and wait until the status message displays your device ID.
+
+#### Direct push notification <a id="step-3-direct"/>
```
m-hulbert:
```suggestion
### Publish directly to your device <a id="step-4-direct"/>
```

src/pages/docs/push/getting-started/react-native.mdx
```diff
@@ -0,0 +1,520 @@
...
+<Code fixed="true">
+```shell
+ably push publish --client-id push-tutorial-client \
+  --title "Hello" \
+  --body "World!" \
+  --data '{"foo":"bar"}'
+```
+</Code>
+
+#### Channel push notification <a id="step-3-channel"/>
```
m-hulbert:
```suggestion
### Publish via a channel <a id="step-4-channel"/>
```

m-hulbert (commented):
A few really minor fixes and then LGTM @maratal !

src/pages/docs/push/getting-started/react-native.mdx
maratal: removed for `react` code blocks

src/pages/docs/push/getting-started/react-native.mdx
maratal: applied

src/pages/docs/push/getting-started/react-native.mdx
maratal: done (back to 5 steps)

src/pages/docs/push/getting-started/react-native.mdx
maratal: done

src/pages/docs/push/getting-started/react-native.mdx
maratal: done

src/pages/docs/push/getting-started/react-native.mdx
maratal: done

m-hulbert (approved):

PR #3292, opened 2026-03-19
[DX-531] Push getting started guide for NextJS

maratal:
## Description

Getting started guide for NextJS and React using Pub/Sub.

https://ably.atlassian.net/browse/DX-531

### Checklist


splindsay-92 (commented):
Before I start the review, I checked the ticket which states this is a _**Getting started guide for NextJS and React using Pub/Sub**_. Is this guide supposed to just be a NextJS with plain ably-js, is there a NextJS + React coming too? :)

maratal:
## React hooks rewrite

Updated the guide to use the new React hooks from [ably/ably-js#2187](https://github.com/ably/ably-js/pull/2187) (`usePushActivation`, `usePush`, `useAbly`, `useChannel`, `AblyProvider`, `ChannelProvider`).

### Structure change

The monolithic `PushPage` component is replaced by four focused files:

| File | Responsibility |
|---|---|
| `page.tsx` | Entry point — `dynamic(() => import('./PushApp'), { ssr: false })` |
| `PushApp.tsx` | `AblyProvider` + `ChannelProvider` setup, shared log state |
| `PushActivationBanner.tsx` | `usePushActivation` + `useAbly` |
| `ChannelSubscription.tsx` | `usePush` + `useChannel` + service worker listener |
| `NotificationLog.tsx` | Pure presentational, no Ably dependency |

### Key changes

**SSR handling**: The previous guide created the Ably client inside `useEffect` to avoid SSR. The new guide uses `dynamic` with `ssr: false` on the page entry point, which means `PushApp.tsx` is only ever loaded client-side and the client can be safely instantiated at module scope — cleaner than managing a ref.

**Activation** (`PushActivationBanner`): `usePushActivation()` replaces manual calls to `realtime.push.activate()`. Returns `localDevice` (reactive, persisted across page loads via `localStorage`) — used to guard the send-to-device/client buttons so `client.device()` is never called before activation.

**Channel subscriptions** (`ChannelSubscription`): `usePush(channelName)` replaces `channelRef.current.push.subscribeDevice()`. The `isActivated` flag it returns is reactive — when `PushActivationBanner` calls `activate()`, the subscribe buttons in `ChannelSubscription` enable automatically with no extra wiring, via a shared module-level store in `PushActivationState.ts`. The `channel` returned by `usePush` is used directly for publishing, removing the need for a separate `channelRef`.

**Realtime messages**: `useChannel` replaces the manual `channel.subscribe()` call inside `useEffect`.

**Error handling**: `connectionError` and `channelError` come directly from `usePush` rather than being managed as separate state.

**Service worker**: unchanged.

src/pages/docs/push/getting-started/nextjs.mdx
```diff
@@ -0,0 +1,468 @@
...
+## Step 2: Set up push notifications <a id="step-2"/>
+
+Create `src/app/push/PushActivationBanner.tsx`. This component uses the `usePushActivation` hook to activate and deactivate the device.
+
+<Code>
+```react
+'use client';
+
+import { useEffect } from 'react';
+import { usePushActivation } from 'ably/react';
```
splindsay-92, line 176: I don't think we have this hook do we? I think I'd suggested it, but I've not implemented it yet.. If we want to use it in this doc, we'll have to agree to push in forward and get it into ably-js :)

splindsay-92 (changes requested):
Lets discuss on Monday and see what we want to do with these hooks

src/pages/docs/push/getting-started/nextjs.mdx
maratal: Hey @splindsay-92, was this decision ever made in js team?

PR #3303, opened 2026-03-27
Update docs navigation to horizontal

m-hulbert:
## Description

This PR updates the docs navigation;
* Adds a horizontal layout split between Platform and Products.
* Adds a sub-menu for listing of Products.
* Expands 2nd level navigation items by default to increase discoverability. 
* Minor navigation updates to all products to remove 'orphaned' single pages (such as About X product). 
* Adds the search bar back into the header. 
* Strips the utility bar functionality; the language selector goes back to the right-hand nav, and the 'copy as' and 'view as' functionality gets rolled into a single split button.


src/components/Layout/hooks/useShowLanguageSelector.ts
```diff
@@ -0,0 +1,18 @@
+import { useMemo } from 'react';
```
jamiehenson: I don't think this needs to be its own hook, it is only used in one place - best to localise this to where it's used.

src/components/Layout/Header.tsx
```diff
@@ -151,51 +149,114 @@ const Header: React.FC = () => {
...
+                    K
+                  </kbd>
+                </span>
+              </button>
+            </div>
+          )}
+        </div>
+        <Tooltip.Provider delayDuration={0} disableHoverableContent>
+          <div className="hidden md:flex gap-2 pt-3 md:py-0 px-4 md:px-0 shrink-0">
             <button
```
jamiehenson: From what I can see this Ask AI button is no longer gated by `inkeepChatEnabled` - sounds like a problem, unless it's now on everywhere all the time without the switch

src/components/Layout/ProductBar.tsx
```diff
@@ -0,0 +1,131 @@
+import { useMemo } from 'react';
```
jamiehenson, line 1: No tests for this file

src/components/Layout/CopyForLLM.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+      track('markdown_copy_link_clicked', {
+        location: location.pathname,
+      });
+    } catch (error) {
+      console.error('Failed to copy markdown:', error);
+      setCopyFeedback('Error!');
+    }
+
+    setTimeout(() => setCopyFeedback(null), 2000);
```
jamiehenson: If the user navigates within 2s of clicking copy, this fires setState on an unmounted component. Minor; the abortController pattern already used for the fetch is the right model to extend here (or keep a ref and clearTimeout on unmount).

data/onCreatePage.ts
```diff
@@ -12,7 +12,7 @@ export type LayoutOptions = {
 const mdxWrapper = path.resolve('src/components/Layout/MDXWrapper.tsx');
 
 const pageLayoutOptions: Record<string, LayoutOptions> = {
-  '/docs': { leftSidebar: true, rightSidebar: false, template: 'index', mdx: false },
+  '/docs': { leftSidebar: false, rightSidebar: false, template: 'index', mdx: false },
```
jamiehenson, line 15: No left bar on the homepage?

src/components/Examples/ExamplesFilter.tsx
```diff
@@ -111,18 +112,22 @@ const ExamplesFilter = ({
 
   return (
     <>
-      <div className="h-[2.125rem] sm:h-[1.875rem] w-5 absolute left-2 top-1 flex items-center justify-center select-none cursor-default">
-        <Icon name={'icon-gui-magnifying-glass-outline'} size="1rem" />
+      <div className="relative w-full">
```
jamiehenson, line 115: Cool but is this in scope with the rest of the PR feature?

src/components/Layout/CopyForLLM.tsx
```diff
@@ -0,0 +1,203 @@
...
+    const docUrl = `https://ably.com${page.link}.md`;
+    const prompt = `Fetch the documentation from ${docUrl} and tell me more about ${product ? productData[product]?.nav.name : 'Ably'}'s '${page.name}' feature${language ? ` for ${languageInfo[language]?.label}` : ''}`;
+    const gptPath = `https://chatgpt.com/?q=${encodeURIComponent(prompt)}`;
+    const claudePath = `https://claude.ai/new?q=${encodeURIComponent(prompt)}`;
+    const perplexityPath = `https://www.perplexity.ai/?q=${encodeURIComponent(prompt)}`;
+
+    return [
+      { model: 'gpt', label: 'Open in ChatGPT', icon: 'icon-tech-openai', link: gptPath },
+      { model: 'claude', label: 'Open in Claude', icon: 'icon-tech-claude-mono', link: claudePath },
+      { model: 'perplexity', label: 'Open in Perplexity', icon: 'icon-tech-perplexity', link: perplexityPath },
```
jamiehenson: Not seeing a logo for this on the review app - package bump? I know you were asking about this before

jamiehenson (changes requested):
I like it, it's close! Sat with it a little more since demo earlier. Code is decent. I like the simplifications, I like the break up of the utility bar and the shuffling round of various things. Product bar works well. Pretty minor stuff from here tbh and some of it is subjective.

Small observations:
- I understand why the left nav contents was padded in (to align with the ably logo and the product row), but it looks weird and overly big to me, particularly when the padding _between_ elements is so small. Honestly, I'd lower the side and top padding 

Image

-----

- Product row on mobile looks a bit padded in as well, weird when scrolling

Image

src/components/Layout/hooks/useShowLanguageSelector.ts
m-hulbert: Good shout, have moved it: bf4e1c79f

src/components/Layout/Header.tsx
m-hulbert: Yes, I forgot to switch it back after testing. 

Fixed in: 361b609c2 and the tests in: 99572c4f7

src/components/Layout/ProductBar.tsx
m-hulbert: Added in c0d6e3d62

src/components/Layout/CopyForLLM.tsx
m-hulbert: 51460675d

data/onCreatePage.ts
m-hulbert: Yes, this is by design as we'd have to pick one of the products to show, rather than all of them now.

src/components/Examples/ExamplesFilter.tsx
m-hulbert: This was just to give it parity with the updated search bar in the header - can pull out if you prefer?

src/components/Layout/CopyForLLM.tsx
m-hulbert: I've commented this out for now - v18 of ably-ui had a few breaking changes that I don't want to address in this PR. Aware I could just copy the image separately, but feel like we should do an ably-ui removal sweep in 1 go.

src/components/Examples/ExamplesFilter.tsx
jamiehenson: It's fine to leave in

jamiehenson (approved):
Good work! A nice evolution of the current UX

PR #3307, opened 2026-04-01
[DX-329] Implemented reference documentation for ably-cli tool

sacOO7:
- Fixed https://ably.atlassian.net/browse/DX-329
- Added reference documentation for [ably-cli](https://github.com/ably/ably-cli) 
- Updated quickstart documentation under https://ably.com/docs/platform/tools/cli
- Commands format is similar to [GitHub CLI Reference Doc](https://cli.github.com/manual/gh_codespace_create)


src/pages/docs/platform/tools/cli.mdx
```diff
@@ -51,22 +43,74 @@ To subscribe:
 
 <Code>
 ```shell
+# Attach to channel with name channel-1
```
AndyTWF: Not just attach though, start receiving messages

src/pages/docs/platform/tools/cli.mdx
```diff
@@ -51,22 +43,74 @@ To subscribe:
...
+<Code>
+```shell
 # To publish 5 messages:
 ably channels publish --count 5 channel-1 "Message number {{.Count}}"
 ```
 </Code>
 
+<Code>
+```shell
+# Output
```
AndyTWF, line 96: In some examples we put command + output in one, others we split, we should be consistent.

src/pages/docs/platform/tools/cli.mdx
```diff
@@ -82,40 +126,75 @@ You can use the Ably CLI for undertaking operations such as:
...
+<Code>
+```shell
+ably-interactive
+```
+</Code>
+
+Interactive mode provides the following features:
+
+* **Command history**: Previous commands are saved and can be accessed with up/down arrows.
+* **Tab completion**: Full support for command and flag completion.
```
AndyTWF, line 194: Tab completion is a feature of the base CLI, no?

src/pages/docs/cli/index.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+## Overview
+
+For an overview of the Ably CLI, including [installation instructions](/docs/platform/tools/cli#installation) and [feature highlights](/docs/platform/tools/cli#feature-set), see the [Ably CLI](/docs/platform/tools/cli) page.
+
+## Global flags <a id="global-flags"/>
+
+The following global flags are available on most CLI commands:
+
+### `--json` <a id="json"/>
```
AndyTWF, line 13: Worth calling out explicitly that this is NDJSON and can be used for programmatic usage?

src/data/nav/platform.ts
```diff
@@ -386,6 +386,363 @@ export default {
...
+                { name: 'delete', link: '/docs/cli/integrations/delete' },
+                { name: 'get', link: '/docs/cli/integrations/get' },
+                { name: 'list', link: '/docs/cli/integrations/list' },
+                { name: 'update', link: '/docs/cli/integrations/update' },
+              ],
+            },
+            {
+              name: 'Logs',
+              pages: [
+                { name: 'overview', link: '/docs/cli/logs', index: true },
```
AndyTWF: For this and other commands, not sure how I feel about `overview` - makes it look like a command, when it isnt

src/pages/docs/cli/accounts/list.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+## Synopsis
+
+<Code>
+```shell
+ably accounts list [flags]
+```
+</Code>
+
+## Global flags
```
AndyTWF, line 17: Having this repeated over every single page is going to get very repetitive. Is it better to have a "global flags" page and that's it? Otherwise if we add another global flag then it mean lots of updates.

Alternatively a partial, @m-hulbert will probably have the best advice on this

src/pages/docs/cli/login.mdx
```diff
@@ -0,0 +1,68 @@
...
+<Code>
+```shell
+ably login [token] [flags]
+```
+</Code>
+
+Use the `ably login` command to authenticate with your Ably account.
+
+<Aside data-type="note">
+This command is an alias for [`ably accounts login`](/docs/cli/accounts/login).
```
AndyTWF, line 18: If this command is just an alias, should we just send users to the docs for that command rather than repeating ourselves?

src/pages/docs/cli/login.mdx
```diff
@@ -0,0 +1,68 @@
...
+</Aside>
+
+## Arguments
+
+Pass the following arguments to the `ably login` command:
+
+### `[token]` <a id="token"/>
```
AndyTWF: I dont think we should wrap the param names in []

src/pages/docs/cli/login.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+The access token for your Ably account.
+
+If not provided, the CLI will prompt you to enter an access token, or open a browser window for you to obtain one from your Ably account.
+
+## Flags
+
+You can optionally pass the following flags to the `ably login` command:
```
AndyTWF: Personally I don't think this sentence is required at the start of every `Flags` block as it's a given

src/pages/docs/cli/apps/index.mdx
```diff
@@ -0,0 +1,32 @@
...
+| Command | Description |
+| ------- | ----------- |
+| [`ably apps create`](/docs/cli/apps/create) | Create a new Ably application. |
+| [`ably apps current`](/docs/cli/apps/current) | Show the currently selected app. |
+| [`ably apps delete`](/docs/cli/apps/delete) | Permanently delete an Ably application. |
+| [`ably apps list`](/docs/cli/apps/list) | List all apps in the current account. |
+| [`ably apps switch`](/docs/cli/apps/switch) | Switch to a different Ably app. |
+| [`ably apps update`](/docs/cli/apps/update) | Update an app's configuration. |
+
+## Rule management commands
```
AndyTWF: I'm thinking we shouldn't list nested commands here, or if we do, just a link pointing down a level

src/pages/docs/cli/apps/rules/create.mdx
```diff
@@ -0,0 +1,111 @@
...
+meta_keywords: "ably cli, cli, apps, rules, create, channel rule, namespace, ably apps rules create"
+---
+
+<Code>
+```shell
+ably apps rules create --name <name> [flags]
+```
+</Code>
+
+Use the `ably apps rules create` command to create a new [channel rule](/docs/channels/channel-rules-namespaces) for an Ably application. Channel rules define behavior for channels matching a specific namespace.
```
AndyTWF: This link is a 404 - please check any links are valid

src/pages/docs/cli/channels/append.mdx
```diff
@@ -0,0 +1,87 @@
+---
+title: "Channels append"
+meta_description: "Append data to a message on an Ably Pub/Sub channel using the CLI."
+meta_keywords: "ably cli, cli, channels, append, message, pub/sub, ably channels append"
+---
+
+<Code>
+```shell
+ably channels append <channel> <serial> <message> [flags]
```
AndyTWF, line 13: There are three different ways (that I've counted) that we name arguments.

1. `name`
2. `<name>`
3. `[name]`

We should pick one and be consistent

AndyTWF (changes requested):
There's a few issues to fix here so I'll pause my review for now

src/pages/docs/platform/tools/cli.mdx
sacOO7: No it's not supposed by default. It needs shell specific configuration using [ably autocomplete](https://github.com/ably/ably-cli/blob/main/docs/Auto-completion.md)

src/data/nav/platform.ts
sacOO7: Hmm, maybe we should rename it to `index`, I will check 👍 
Also, all sub-commands can be made Pascal-Case similar to top level commands

src/pages/docs/cli/channels/append.mdx
sacOO7: This follows industry convention for naming required/optional arguments
As per http://docopt.org/

  1.  `<name>` / `[name]` in Synopsis = shows command syntax with required/optional markers respectively
  2.  `name` in argument headings = identifies the argument being documented

 Sources that codify the same convention:
  - docopt (http://docopt.org/) — formalizes these decades-old conventions into a parseable spec
  - man-pages(7) (https://man7.org/linux/man-pages/man7/man-pages.7.html) — Linux man page standard
  - Microsoft Syntax Key
  (https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/command-line-syntax-key) — same convention independently
  - [Utility Argument Syntax](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) - IEEE Std 1003.1-2017 (POSIX.1-2017), Section 12.1

src/pages/docs/cli/apps/rules/create.mdx
sacOO7: Fixed -> https://github.com/ably/docs/pull/3307/commits/3c9c40cfb8c7188648e91d12ee82c07d3bc3dea1

src/pages/docs/cli/login.mdx
sacOO7: Fixed -> https://github.com/ably/docs/pull/3307/commits/3c9c40cfb8c7188648e91d12ee82c07d3bc3dea1

src/pages/docs/cli/login.mdx
sacOO7: Fixed -> https://github.com/ably/docs/pull/3307/commits/3c9c40cfb8c7188648e91d12ee82c07d3bc3dea1

src/pages/docs/platform/tools/cli.mdx
sacOO7: Fixed -> https://github.com/ably/docs/pull/3307/commits/3c9c40cfb8c7188648e91d12ee82c07d3bc3dea1

src/pages/docs/cli/channels/append.mdx
sacOO7: Also, all commands with arg/flags are kept **simple** and **readable** similar to [GitHub CLI Reference Doc](https://cli.github.com/manual/gh_codespace_create)

src/pages/docs/cli/index.mdx
sacOO7: Yeah, will update accordingly 👍

src/pages/docs/cli/login.mdx
sacOO7: Actually, `Login` is a standalone top-level command and the first one users run right after installing the CLI, so it makes sense to document it explicitly. Also, since the current page already references it via an `alias`, it might be better to give it a dedicated page. 
We could shorten the existing content to avoid duplication, but since the overall content wouldn’t really change, keeping it as is also works. I lean toward the second approach.
wdyt @m-hulbert

src/data/nav/platform.ts
sacOO7: Better to create two subsections `Overview` and `Commands`. `Commands` will have all `sub-commands` for given `top-level` command.

src/pages/docs/cli/apps/index.mdx
sacOO7: Fixed -> https://github.com/ably/docs/pull/3307/commits/a00f71e176bd0996713d846d09c9abd620fad332

src/data/nav/platform.ts
sacOO7: Fixed -> https://github.com/ably/docs/pull/3307/commits/a00f71e176bd0996713d846d09c9abd620fad332

src/pages/docs/platform/tools/cli.mdx
sacOO7: I tried putting command and output in 1 command, it looks super weird, so currently all examples in the file have separate `command` and `output`

src/pages/docs/cli/index.mdx
sacOO7: Fixed -> https://github.com/ably/docs/pull/3307/changes/97cd99c0234f3ec3eee5b551b46561ffe5ca9498

src/pages/docs/cli/accounts/list.mdx
sacOO7: Maybe we can get rid of the page specific `Global flags` section and add a note pointing towards `Global Flags` section on **ROOT** `Ably CLI Reference/Overview` page. 
Currently, we have included examples that use either `--json` or `--pretty-json`.

sacOO7: Some commands like 
1. [Accounts Current](https://ably-docs-feature-ably--sze1pz.herokuapp.com/docs/cli/accounts/current) : `ably accounts current [flags]`
2. [Accounts List](https://ably-docs-feature-ably--sze1pz.herokuapp.com/docs/cli/accounts/list): `ably accounts list [flags]`

Only supports `Global Flags`.

src/pages/docs/platform/tools/cli.mdx
AndyTWF: Sure, but it's not a feature of _just_ the interactive mode right?

src/pages/docs/cli/channels/append.mdx
AndyTWF: That's fine if we want to follow docopt as a standard, but:

- `[flags]` would need to become `[options]` - `[flags]` would be interpreted as a positional argument

src/pages/docs/cli/integrations/delete.mdx
```diff
@@ -0,0 +1,75 @@
...
+meta_keywords: "ably cli, cli, integrations, delete, rule, ably integrations delete"
+---
+
+Use the `ably integrations delete` command to delete an integration rule.
+
+## Synopsis
+
+<Code>
+```shell
+ably integrations delete <integration-id> [flags]
```
AndyTWF: For some of these we use `rule-id`, others `integration-id` - can we pick one?

src/pages/docs/cli/rooms/messages/update.mdx
```diff
@@ -0,0 +1,99 @@
...
+```shell
+ably rooms messages update my-room "serial-001" "Updated text" --headers '{"source":"cli"}'
+```
+</Code>
+
+Update a room message and output the result in JSON format:
+
+<Code>
+```shell
+ably rooms messages update my-room "serial-001" "Updated text" --json
```
AndyTWF, line 92: I would suggest for completeness, we make a realistic looking serial here (and in other commands where the serial is an arg) - that way its clear for a consumer what they're putting here

AndyTWF (commented):
Couple of small things

src/pages/docs/cli/integrations/delete.mdx
sacOO7: Actually, I cross checked, that's what we use in `ably-cli` tool too

sacOO7: Maybe, we need to update the same at both places 👍

src/pages/docs/cli/rooms/messages/update.mdx
sacOO7: The main issue is that serials are quite long, which makes the command harder to read.
The above command with actual serial will look like 
```
ably rooms messages update my-room "01775550200065-000@108sLXM8gC109L72259911:000 "Updated text" --json
```
Maybe we can put shorthand serial
```
ably rooms messages update my-room "0177555@serial:000 "Updated text" --json
```
wdyt 

cc @m-hulbert

maratal (commented):
you should update push  command, it now supports flag `--message`  in conjunction with `--channel`  which sets realtime message as well, see https://github.com/ably/ably-cli/pull/310

something like this
```
The target channel name. Publishes a push notification via the channel using extras.push. Ignored if --device-id, --client-id, or --recipient is also provided. Use together with --message to provide realtime message to push recipients.
```

src/pages/docs/cli/integrations/delete.mdx
sacOO7: Fixed https://github.com/ably/docs/pull/3307/commits/55693f9d691990fd964008399af4d66cf0c3fde2

src/pages/docs/platform/tools/cli.mdx
sacOO7: True but as per command name `ably autocomplete`, it's the main feature of the command

sacOO7: We have ported documentation as is from `ably-cli/README.md`

umair-ably:
Couple of comments from the demo session:

- "Required" should be a badge of sorts and not look like the same code tag it is next to
- Output examples are redundant - we don't think a user would or should care, they care about what the CLI can do and how, not what the output looks like. Also, it's one more thing to maintain if we update our formatting again
- I also prefer options instead of flags (options is more behavioural, flags is technical/mechanical - we should obfuscate terminology where appropriate) 
- Each "option" should exhaust it's potential values and not only show a subset of examples (otherwise a user has no way of knowing all the possible values they could set for an option)

sacOO7:
TODO

- [x] 1. Mention json flags programmatic usage in [tools/cli page](https://ably.com/docs/platform/tools/cli).
- [x] 2. Remove output examples snippets from the index page.
- [x] 3. Make sure all annotation types are well covered.
- [x] 4. Fix a bug where weird symbol is shown at the bottom See also section of markdown page -> This automatically gets fixed when markdown is served as `text/markdown` or `text/plain`
- [x] 5. Re-arrange ably-cli commands with correct order in nav bar. 
- [x] 6. Update commands/examples as per latest merged PRs
- [x] 7. Update messageSerial with actual serial or exampleSerial placeholder, same should be applicable for other arguments.
- [x] 8. Mark required as `badge` component from documentation.
- [x] 9. Replace flags with `Options` -> https://github.com/ably/docs/pull/3307#issuecomment-4268356340 and https://github.com/ably/docs/pull/3307#discussion_r3087820720

maratal:
Update for push commands - https://github.com/ably/docs/pull/3353

AndyTWF (approved):
Looks good, but noting that this requires a CLI release prior to merging

src/pages/docs/cli/autocomplete.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+The shell type to set up autocomplete for. Supported values: `zsh`, `bash`, `powershell`.
+
+If no shell is specified, the default shell is used.
+
+## Options
+
+### `--refresh-cache | -r` <a id="refresh-cache"/>
+
+Refresh the autocomplete cache, ignoring displaying instructions.
```
jamiehenson: nit: Might want to re-word this, doesn't make grammatical sense to me. "instead of displaying instructions" ?

src/pages/docs/cli/index.mdx
```diff
@@ -0,0 +1,49 @@
...
+| [`ably push`](/docs/cli/push) | Manage push notifications |
+| [`ably logs`](/docs/cli/logs) | Streaming and retrieving logs from Ably |
+| [`ably stats`](/docs/cli/stats) | View statistics for your Ably account or apps |
+| [`ably config`](/docs/cli/config) | Manage Ably CLI configuration |
+| [`ably bench`](/docs/cli/bench) | Commands for running benchmark tests |
+| [`ably connections`](/docs/cli/connections) | Interact with Ably Pub/Sub connections |
+| [`ably support`](/docs/cli/support) | Get support and help from Ably |
+| [`ably autocomplete`](/docs/cli/autocomplete) | Display autocomplete installation instructions |
+| [`ably status`](/docs/cli/status) | Check the status of the Ably service |
+| [`ably help`](/docs/cli/help) | Display help for ably |
```
jamiehenson, line 51: nit: `interactive` and `version` are docs in the PR but aren't in this table. `interactive` also isn't in the sidebar.

src/pages/docs/cli/accounts/list.mdx
```diff
@@ -0,0 +1,72 @@
...
+
+<Code>
+```shell
+ably accounts list --pretty-json
+```
+</Code>
+
+## Output
+
+The `ably accounts list` command displays list of accounts in format:
```
jamiehenson: nit: another grammar tidyup

jamiehenson (approved):
Given the depth of docs added here I've naturally not read them all, but the composition of the MDX docs on the review app looks good to me, and I only have a handful of minor non-blocking nits to add. The orphaned pages nit is the biggest one. Nice job though.

Obvs need to verify that the cli has been released before this as @AndyTWF mentioned, and my agent picked up a couple of unresolved comments in the thread.

Not a consideration for this PR, but I do wonder what our best strategy will be for keeping these docs in sync with the rapid development of the CLI.

sacOO7:
TODO's

- [x] 1. Resolve pending review comments
- [x] 2. Fix `On this page` section, shows `Required` as a unnecessary text.
- [x] 3. Replace `env. variables usage github documentation` with `env. variables usage documentation`
- [x] 4. Fix related topics incorrectly displayed -> https://ably-docs-feature-ably--sze1pz.herokuapp.com/docs/cli/channels/publish.md

src/pages/docs/cli/autocomplete.mdx
sacOO7: Fixed — updated to "instead of displaying instructions".

src/pages/docs/cli/index.mdx
sacOO7: Added `version` to the commands table. For the sidebar, `version` is already there. Skipping `interactive` in the sidebar since it's an alpha feature with no command-specific flags, and it's already covered in the [Interactive mode](/docs/platform/tools/cli#interactive-mode) section of the CLI overview page.

src/pages/docs/cli/accounts/list.mdx
sacOO7: Fixed — updated to "displays a list of accounts in the following format:".

sacOO7: https://github.com/ably/docs/pull/3307/commits/b1227f675182527f9be229c895d485eefd140c21

src/pages/docs/cli/autocomplete.mdx
sacOO7: https://github.com/ably/docs/pull/3307/commits/b1227f675182527f9be229c895d485eefd140c21

src/pages/docs/cli/index.mdx
sacOO7: https://github.com/ably/docs/pull/3307/commits/b1227f675182527f9be229c895d485eefd140c21

src/pages/docs/platform/tools/cli.mdx
```diff
@@ -75,47 +77,128 @@ ably help
...
+</Code>
+
+Interactive mode provides the following features:
+
+* **Command history**: Previous commands are saved and can be accessed with up/down arrows.
+* **Tab completion**: Full support for command and option completion.
+* **Ctrl+C handling**: A single Ctrl+C interrupts the current command and returns to the prompt. A double Ctrl+C (within 500ms) force quits the shell.
+* **No "ably" prefix needed**: Commands can be typed directly, for example `channels list` instead of `ably channels list`.
+
+## Environment variables <a id="environment-variables"/>
```
umair-ably: As discussed in standup and CLI PR... this should be it's own page (I think it can live under `pages/docs/cli`)... 

Any links to the cli markdown should be removed. There are a couple approaches we can take but with some of the agent/LLM eval based work I have planned for this quarter in mind, I would urge you to remember that not all agents will read the docs site - we should always ask if an agent (and human) can self-serve the CLI?

My recommendation for this would be:
1. This is now it's own page under `pages/docs/cli`
2. We keep the page very brief e.g. only the 3 auth type env vars mentioned, again highlighting it as an alternative to `ably login` - similar to what you have here, but having removed all links to the external markdown. 
3. We should update the CLI to add a custom help topic specifically for env vars (oclif supports this afaik). This section can have the exhaustive list of env vars that are currently in your general usage markdown doc (so not the internal env vars)
4. We update this new page to mention users can call `--help` to see the full list 

This allows the env vars to be self-documenting to a certain degree (same as all other CLI commands are). You can make a call on if you want to remove the general env var documentation from the repo - we likely won't need it after these changes as this new page tied in with an actual help section should make that doc redundant

umair-ably: I'd also add `ABLY_ENDPOINT` to this page - host overrides would be important for a human to see when passing through the docs. All others are more config/cool tricks and don't need mentioning on this page imo

sacOO7: Sure 👍

sacOO7: Updated documentation

umair-ably: nice, looks much better

umair-ably (approved):

PR #1937, opened 2023-06-26
[EDU-1239] Update the docs information architecture

m-hulbert:
## Description

This PR updates content throughout the docs site to refresh the information architecture. At a high level it:

* Restructures the side navigation to group content by feature.
  * Combines the REST and realtime SDK content.
  * Adds a component to code blocks to have tabs at the interface level.
* Updates the homepage `/docs` to reflect new format.
* Updates in-page navigation for the majority of pages to enforce consistency. 
* Languages are now displayed by their latest SDK version.
* Introduces a more in-depth notes component.
* Replicates the content from `/download` and `/download/sdk-feature-support-matrix` in the docs.
* Removes content previously versioned by protocol version.
* Redirects deleted content to the most relevant new page. 

See [JIRA](https://ably.atlassian.net/browse/EDU-1239) for full details.

## Review

See review app: https://ably-docs-ia-update-dthvttqu6l.herokuapp.com/docs

PR #3334, opened 2026-04-13
 Rewind param: stop making the non-ably-sdk fallback method the first thing we tell people about

SimonWoolf:
It doesn't make any sense to start out by saying "rewind can **also** be set by..." and then the MQTT/SSE fallback method when we haven't told anyone the actual, main method that they should use with ably SDKs yet. Nor to have that as a top-level paragraph followed by an inset note explaining when you should use that method (for mqtt and sse), it's not at all obvious that the inset note is meant to tell you when to use the previous _not_-inset paragraph

I think this was a regression from https://github.com/ably/docs/pull/1937 

also fix a bunch of places on that page where we call it a channel option not a channel param

### DIFF
--- a/src/pages/docs/channels/options/rewind.mdx
+++ b/src/pages/docs/channels/options/rewind.mdx
@@ -128,7 +128,7 @@ channel.subscribe().listen((message) {
 </Code>
 
 <Aside data-type="note">
-If using [MQTT](/docs/protocols/mqtt) or [SSE](/docs/protocols/sse), see [Channel options without Ably SDK support](/docs/channels/options#other-sdks) for how to specify a rewind param.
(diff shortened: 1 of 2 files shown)


src/pages/docs/channels/options/rewind.mdx
```diff
@@ -135,6 +127,10 @@ channel.subscribe().listen((message) {
 ```
 </Code>
 
+<Aside data-type="note">
+If using [MQTT](/docs/protocols/mqtt) or [SSE](/docs/protocols/sse), see [Channel options without Ably SDK support](/docs/channels/options#other-sdks) for how to specify a rewind param.
```
m-hulbert:
```suggestion
If using [MQTT](/docs/protocols/mqtt) or [SSE](/docs/protocols/sse), you can [specify channel params](/docs/channels/options#other-sdks) as a query string.
```

m-hulbert (approved):
Small suggestion to make the link inline, but LGTM otherwise.

PR #3343, opened 2026-04-16
Restructure AI Transport docs around concepts

zknill:
Replace the "How it works" section with a dedicated "Concepts" section containing five pages: Sessions, Messages and conversation tree, Turns, Transport, and Authentication. These pages explain AI Transport's conceptual model (sessions, the conversation tree, turn lifecycle, client/agent transports, and the codec) as logical entities rather than code walkthroughs.

Merge the "Framework guides" section into "Getting started" so all Vercel AI SDK content lives on a single page. Add a new "Core SDK" getting started page for using AI Transport's generic React hooks without a framework wrapper.

Strengthen the "Why" page with clearer durable session properties (resilient delivery, continuity across surfaces, live control) and update the overview page with a Core SDK tile.

Update cross-references across all 16 feature pages to point to the new concepts paths. Add redirects from all old URLs.


zknill:
@rainbowFi I've made extra changes: 
1. Re-added the framework guide for vercel, but made it actually good
2. Fixed _all_ the code api references and code examples (which were basically all entirely wrong). 

This is ofc not the 'final destination' for these docs, but there is a _LOT_ on the website that is currently just-plain-wrong. So I think this is a "if this PR is broadly fine, lets merge and iterate" kind of situation rather than expecting to 1-shot perfection (not that that's what you've been arguing for 🙇 ). 🙏

src/pages/docs/ai-transport/getting-started/core-sdk.mdx
```diff
@@ -0,0 +1,251 @@
+---
+title: "Get started with the Core SDK"
```
rainbowFi, line 2: Based on the code, I think this is about using Vercel AI SDK Core to interact with the LLM but retaining complete control of the interface between agent and client. Can you please make that clear throughout the text, so people don't think Ably is providing a Core SDK? Also needs a new title that includes Vercel, without clashing with the other Vercel article in the nav....

rainbowFi: Zak and I chatted, he's going to work on naming this to avoid the confusion I fell into :D

rainbowFi (approved):

PR #3386, opened 2026-05-27
docs: add error code 40025 for unsupported SDK API

umair-ably:
## Summary

- Documents the new `40025` error code introduced in [ably/ably-js#2226](https://github.com/ably/ably-js/pull/2226), thrown when an SDK method is called using an API shape that is no longer supported.
- Generic framing (the code is reusable across SDKs/APIs), with the ably-js v1 → v2 callback case given as the concrete example and a link to the v2 migration guide.
- Placed in numeric order between `40023` and `40030` in `src/pages/docs/platform/errors/codes.mdx`.

## Test plan

- [ ] Visit `/docs/platform/errors/codes#40025` in a preview build and confirm the heading anchor renders correctly
- [ ] Confirm the entry sits between 40023 and 40030 in the rendered page
- [ ] Confirm the migration guide link resolves

### DIFF
--- a/src/pages/docs/platform/errors/codes.mdx
+++ b/src/pages/docs/platform/errors/codes.mdx
@@ -154,6 +154,14 @@ Examples of this error include attempting to use [channel objects](/docs/liveobj
 Resolution: The following steps can help resolve this issue:
 * Upgrade the Ably SDK to a version that supports the requested operation.
 
+## 40025: Unsupported SDK API <a id="40025"/>
+
+This error occurs when an SDK method is called using an API shape that is no longer supported.
+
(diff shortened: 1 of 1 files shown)


src/pages/docs/platform/errors/codes.mdx
```diff
@@ -154,6 +154,14 @@ Examples of this error include attempting to use [channel objects](/docs/liveobj
 Resolution: The following steps can help resolve this issue:
 * Upgrade the Ably SDK to a version that supports the requested operation.
 
+## 40025: Unsupported SDK API <a id="40025"/>
+
+This error occurs when an SDK method is called using an API shape that is no longer supported.
```
AndyTWF, line 159: Specifically API shape, or just "a version of the API"?

umair-ably: I'm leaning towards shape because "version of the API" implies we're talking about deprecated versions and not individual deprecated APIs (which this error code can also apply to)

m-hulbert: If it's also thrown for when an API is deprecated, wouldn't that mean the error message doesn't read correctly?

PR #3394, opened 2026-06-03
Add Pub/Sub "Live voting" example (message annotations)

SimonWoolf:
## What

Adds an interactive **Live voting** example at `/examples/pub-sub-live-voting`, focused on **message annotations**: a poll is a message, each vote is a `vote:unique.v1` annotation, and Ably aggregates votes into a summary — no database, no client-side tallying.

The live widget shows **voter + presenter** panes side by side. On load a sandbox-only bootstrap auto-starts the d-pad poll and seeds a few votes so the presenter (heatmap + leader badge + vote bubbles) is alive immediately; the voter pane then votes live on top.

live-voting-example-readme

## Structure

Everything for the full app lives under `examples/pub-sub-live-voting/javascript/` so the in-page **Open in CodeSandbox** export is complete:

- client at the `javascript/` root (voter/presenter/admin views, branched by `?role=`)
- `server/` — Express: role-scoped JWT `/auth`, shows API, static `SHOWS_FILE` **and** Postgres stores
- `data/` (static poll data) and `database/` (schema + migrations)

Sandbox vs. clone: the live widget has no backend, so it falls back to a raw key (loudly commented as demo-only) and a per-role channel mode split; a real clone uses the server's token auth. The canonical runnable source lives in the standalone **[ably-demos/live-voting-with-annotations](https://github.com/ably-demos/live-voting-with-annotations)**, which the README and the new "View on GitHub" override point to.

## Shared-infra touches (kept minimal / backward-compatible)

- `ExamplesRenderer`: per-id start routes (`?role=voter` / `?role=presenter&demo=1`) + Voter/Presenter pane labels; add `qrcode` dependency.
- `examples.tsx` + `Example` type: optional `githubUrl` to override the "View on GitHub" target (defaults unchanged for every other example).
- `gatsby-config`: `ignore` `node_modules`/`dist` on the examples source — required now that a nested `server/` can have its own installed deps under a sourced language dir.

## Verification

`yarn lint` clean (0 errors); standalone server `npm install` + `tsc` clean; client `vite build` clean. Live behaviour (two panes, voting → heatmap/bubbles) should be eyeballed on the review app — it needs the *Message annotations* channel rule on the `voting:*` namespace of the demo app.


SimonWoolf:
@m-hulbert any chance of a review of this?

JoaoDiasAbly (approved):
Overall LGTM

PR #3396, opened 2026-06-04
[DX-1371] Update search and Ask AI behaviour

m-hulbert:
## Description

This PR updates the Search bar and Ask AI button to match styles between local and production. It also changes the rendering so that they're always visible in a local setup, however the triggers are disabled.

### Checklist

### DIFF
--- a/src/components/Layout/Header.test.tsx
+++ b/src/components/Layout/Header.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { render, screen, fireEvent } from '@testing-library/react';
 import { useStaticQuery } from 'gatsby';
+import { track } from '@ably/ui/core/insights';
 import Header from './Header';
 import UserContext from 'src/contexts/user-context';
 
@@ -15,6 +16,10 @@ jest.mock('src/contexts/layout-context', () => ({
       template: null,
     },
   }),
+}));
+
+jest.mock('@ably/ui/core/insights', () => ({
+  track: jest.fn(),
 }));
 
 jest.mock('src/components/Icon', () => {
@@ -143,7 +148,7 @@ describe('Header', () => {
     expect(searchBar).not.toBeInTheDocument();
   });
 
-  it('does not render Ask AI button and chat bar when inkeepChatEnabled is false', () => {
+  it('always renders the Ask AI button, but not the chat instance, when inkeepChatEnabled is false', () => {
     (useStaticQuery as jest.Mock).mockReturnValue({
       site: {
         siteMetadata: {
@@ -157,12 +162,13 @@ describe('Header', () => {
 
     render(<Header />);
 
-    expect(screen.queryByText('Ask AI')).not.toBeInTheDocument();
+    // The Ask AI button always renders; only the Inkeep chat instance is flag-gated.
+    expect(screen.getByText('Ask AI')).toBeInTheDocument();
     const chatBar = document.getElementById('inkeep-ai-chat');
     expect(chatBar).not.toBeInTheDocument();
   });
 
-  it('does not render search bar or Ask AI button when both flags are false', () => {
+  it('always renders the search trigger and Ask AI button, but not the Inkeep instances, when both flags are false', () => {
     (useStaticQuery as jest.Mock).mockReturnValue({
       site: {
         siteMetadata: {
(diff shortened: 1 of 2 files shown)


src/components/Layout/Header.tsx
```diff
@@ -222,59 +264,26 @@ const Header: React.FC = () => {
...
-                  }
-                }}
-              >
-                <Icon name="icon-gui-sparkles-outline" size="20px" />
-                <span>Ask AI</span>
-              </button>
-            )}
+                if (chatButton) {
+                  chatButton.click();
+                }
```
jamiehenson: This Ask AI onClick is now a near-twin of `openInkeepSearch` up top — query `#X > div` → shadow `button` → track → click. Worth pulling both into one `openInkeepWidget(hostSelector, trackEvent)` so the shadow-DOM reach-in lives in a single place.

~ 𝒞𝓁𝒶𝓊𝒹𝑒

src/components/Layout/Header.test.tsx
```diff
@@ -176,7 +177,10 @@ describe('Header', () => {
 
     render(<Header />);
 
-    expect(screen.queryByText('Ask AI')).not.toBeInTheDocument();
+    // Our own search trigger and the Ask AI button are always present; the Inkeep
+    // search/chat instances only mount when their flags are enabled.
+    expect(screen.getByText('Search')).toBeInTheDocument();
+    expect(screen.getByText('Ask AI')).toBeInTheDocument();
```
jamiehenson, line 188: These only assert the buttons render, not that clicking them opens anything — so if the shadow-DOM open path breaks, the tests stay green. A click-wiring test (mock `querySelector`, assert the trigger fires) would catch that. The real shadow DOM isn't jsdom-friendly so it's partial at best, but the click path is the risky bit worth pinning down.

~ 𝒞𝓁𝒶𝓊𝒹𝑒

jamiehenson (commented):
Mostly a heads-up plus a couple of nits — not blocking.

The bit I want to be sure of: the visible Search button now opens the modal by clicking into `#inkeep-search`'s shadow DOM while that instance sits in a `display:none` holder — and that reach-in has only ever been done against the chat widget before. If Inkeep's search trigger isn't a bare `<button>` (or is nested differently) it silently no-ops and the button does nothing, with no error, and the jsdom tests only check it renders.

So — are we confident this works in both environments? i.e. unconfigured Inkeep (local / flags off, button inert by design) *and* configured Inkeep (prod / flags on, button actually opens the modal). I've kicked off a review app to poke at it.

Two nits inline. Also tiny: title typo, "behavor" → "behaviour".

~ 𝒞𝓁𝒶𝓊𝒹𝑒

src/components/Layout/Header.tsx
m-hulbert: I've addressed this in 91bad229fe13cd39ff34863c2d85c7beb99bffb6

src/components/Layout/Header.test.tsx
m-hulbert: This one too: 91bad229fe13cd39ff34863c2d85c7beb99bffb6

jamiehenson (commented):
Looks good in the review app — header search bar + Ask AI render identically local vs prod, and the inert local triggers behave. Good to merge after a couple of small things.

**Mandatory**

- `openInkeepWidget` calls `track(eventName)` before checking the trigger resolves, so inert clicks (local / no widget) still fire `docs_search_button_clicked` / `docs_ask_ai_button_clicked` analytics with nothing actually opening. Reorder so we only track a real open:
  ```ts
  const openInkeepWidget = (hostSelector: string, eventName: string) => {
    const trigger = document.querySelector(`${hostSelector} > div`)?.shadowRoot?.querySelector('button');
    if (!trigger) return;
    track(eventName);
    trigger.click();
  };
  ```

**Nits**

- `SearchTrigger` is a plain arrow; rest of the file types components as `React.FC`. Tiny consistency thing.
- The new block comments run a bit longer than the file's one-liner style — happy to leave them though, they earn it explaining the shadow-DOM reach-in.
- Worth squashing the `fixup!` commit before merge.

~ 𝒞𝓁𝒶𝓊𝒹𝑒

jamiehenson (approved):
Re-reviewed — all points addressed: `track()` now fires only on a real open, `SearchTrigger` typed as `React.FC`, and the branch is squashed + rebased on latest main (clean reconcile with the DX-1128 Icon migration). CI green. Approving.

~ 𝒞𝓁𝒶𝓊𝒹𝑒

PR #3397, opened 2026-06-04
docs(auth): standardise auth endpoint examples on /api/auth/token

umair-ably:
## Summary

Brings the core auth, Chat, Spaces and Pub/Sub docs onto a single token-endpoint convention so that, together with the AI Transport docs, the whole site teaches one consistent auth setup: a short-lived JWT fetched from your server via `authCallback` or `authUrl`, from one example endpoint path.

This PR originally also covered the AI Transport pages (executing `ait-auth-consistency-plan.md`). **That half has since been captured and merged in #3399** ("AI Transport docs: auth endpoint URL and example channel namespaces"), which moved the AI Transport docs onto `/api/auth/token`, added a dedicated *Set up authentication* getting-started page, and reframed the authentication concept page. After rebasing onto `main`, those commits became no-ops and dropped out — so what remains here is only the non-AI-Transport work that #3399 did not touch.

### What changed

- **Centralise `authUrl` vs `authCallback` guidance** in the token overview (`auth/token/index.mdx`): a comparison table and per-mechanism guidance, framed as equivalent and current, linked from the auth index. The choice is now documented in one place instead of being re-explained per page.
- **Standardise the example auth endpoint on `/api/auth/token`** across `auth/token/*`, `chat/*`, `spaces/*` and `pub-sub/*`. Client `fetch`/`authUrl` calls and server route examples that previously used `/auth`, `/api/ably-token` or `/api/ably-jwt` now all use `/api/auth/token`, matching the convention #3399 established in the AI Transport docs (including the canonical `auth/token/jwt.mdx` examples).

### Why `/api/auth/token`

The path is type-neutral by design: the endpoint returns a JWT, so a path containing `ably-token` would mislead readers into thinking it returns Ably's native Token type. Aligning on `/api/auth/token` matches the path already used throughout the AI Transport docs after #3399.

### Out of scope

A few pages outside the core auth/Chat/Spaces setup flow still use other illustrative endpoint names (`connect/states` → `/obtainToken`, `getting-started/php` → `/get-auth-jwt`, `platform/account/enterprise-customization` → `/auth`, and the Spaces feature pages' `<authEndpoint>` placeholder). These were left untouched to keep this PR's scope tight; a follow-up sweep can bring them onto `/api/auth/token` if wanted.

## Test plan

- [ ] CI green (build + lint)
- [ ] Token overview renders the new `authUrl` vs `authCallback` table and guidance
- [ ] No remaining `/auth`, `/api/ably-token` or `/api/ably-jwt` endpoint examples in `auth/`, `chat/`, `spaces/`, `pub-sub/`
- [ ] Cross-page navigation (auth index → token overview → JWT/Ably Tokens client setup) resolves and reads consistently

### DIFF
--- a/src/pages/docs/auth/token/ably-tokens.mdx
+++ b/src/pages/docs/auth/token/ably-tokens.mdx
@@ -125,16 +125,16 @@ final tokenRequest = rest.auth.createTokenRequest(tokenParams: tokenParams);
 ```
 </Code>
 
-### Client usage
+### Client usage <a id="client-usage"/>
 
-The client SDK automatically handles TokenRequests returned from your auth endpoint:
+The client SDK automatically handles TokenRequests returned from your auth endpoint. These examples use `authCallback`, but `authUrl` is also valid; see [choosing how the client fetches tokens](/docs/auth/token#auth-mechanism) for guidance on which to use.
 
 <Code>
 ```realtime_javascript
 const realtime = new Ably.Realtime({
   authCallback: async (tokenParams, callback) => {
     try {
-      const response = await fetch('/api/ably-token');
+      const response = await fetch('/api/auth/token');
       const tokenRequest = await response.json();
       callback(null, tokenRequest);
     } catch (error) {
@@ -148,7 +148,7 @@ const realtime = new Ably.Realtime({
 const realtime = new Ably.Realtime({
   authCallback: async (tokenParams, callback) => {
     try {
-      const response = await fetch('/api/ably-token');
+      const response = await fetch('/api/auth/token');
       const tokenRequest = await response.json();
       callback(null, tokenRequest);
     } catch (error) {
@@ -163,7 +163,7 @@ import aiohttp
 
 async def get_token_request(*args, **kwargs):
     async with aiohttp.ClientSession() as session:
-        async with session.get('/api/ably-token') as response:
+        async with session.get('/api/auth/token') as response:
             if response.status != 200:
                 raise Exception(f"Auth failed: {response.status}")
             return await response.json()
@@ -340,7 +340,7 @@ The client code for Ably Token direct is the same as for TokenRequest — the SD
 const realtime = new Ably.Realtime({
   authCallback: async (tokenParams, callback) => {
     try {
-      const response = await fetch('/api/ably-token');
+      const response = await fetch('/api/auth/token');
       const tokenDetails = await response.json();
       callback(null, tokenDetails);
     } catch (error) {
@@ -354,7 +354,7 @@ const realtime = new Ably.Realtime({
 const realtime = new Ably.Realtime({
   authCallback: async (tokenParams, callback) => {
     try {
-      const response = await fetch('/api/ably-token');
+      const response = await fetch('/api/auth/token');
       const tokenDetails = await response.json();
       callback(null, tokenDetails);
     } catch (error) {
@@ -369,7 +369,7 @@ import aiohttp
 
 async def get_ably_token(*args, **kwargs):
     async with aiohttp.ClientSession() as session:
-        async with session.get('/api/ably-token') as response:
+        async with session.get('/api/auth/token') as response:
             if response.status != 200:
                 raise Exception(f"Auth failed: {response.status}")
             return await response.json()
--- a/src/pages/docs/auth/token/ably-tokens.mdx
+++ b/src/pages/docs/auth/token/ably-tokens.mdx
@@ -127,7 +127,7 @@ final tokenRequest = rest.auth.createTokenRequest(tokenParams: tokenParams);
 
 ### Client usage <a id="client-usage"/>
 
-The client SDK automatically handles TokenRequests returned from your auth endpoint. These examples use `authCallback`, but `authUrl` is also valid; see [choosing how the client fetches tokens](/docs/auth/token#auth-mechanism) for guidance on which to use.
+The client SDK automatically handles the `TokenRequest` returned from your auth endpoint. These examples use `authCallback`, but `authUrl` is also valid; see [choosing how the client fetches tokens](/docs/auth/token#auth-mechanism) for guidance on which to use.
 
 <Code>
 ```realtime_javascript
--- a/src/pages/docs/auth/token/index.mdx
+++ b/src/pages/docs/auth/token/index.mdx
@@ -16,7 +16,7 @@ Token authentication is the recommended authentication method to use client-side
 SDKs for [Chat](/docs/chat) and [Spaces](/docs/spaces) products rely on the underlying Pub/Sub library for their connection and authentication, so there is no need to configure auth separately in those SDKs.
 </Aside>
 
-## Choosing a token mechanism <a id="choosing"/>
+## Choosing a token format <a id="choosing"/>
 
 Ably supports two token formats: JWT, the primary and recommended format, and Ably tokens, a legacy format. Each format is associated with a different flow for issuing tokens to clients.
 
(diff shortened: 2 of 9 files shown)


src/pages/docs/ai-transport/getting-started/core-sdk.mdx
```diff
@@ -164,7 +166,7 @@ function Providers({ children }) {
   const [client, setClient] = useState(null);
 
   useEffect(() => {
-    const ably = new Ably.Realtime({ authUrl: '/auth', clientId: 'user' });
```
rainbowFi: AIT requires clientId to be set so that it can attribute turns etc. If we don't specify it here, will the SDK extract it from the JWT?

umair-ably: yes, exactly that, and we should look to make that clearer. 

As it was, using JWT auth and then also providing a clientId could cause issues that we're not being upfront about e.g. if the JWT is scoped to a very specific clientID ("Bob"), and the user passes in a different clientId as a param in the constructor ("Alice"), they'd be hit with a client mismatch error (which is increasingly likely if a user just lifts and shifts this code, or lets an LLM copy it as is)

rainbowFi:
Text changes look good. The urls all seem to be `/auth`, I thought we were going with `/api/auth/token`?

src/pages/docs/auth/token/ably-tokens.mdx
```diff
@@ -125,16 +125,16 @@ final tokenRequest = rest.auth.createTokenRequest(tokenParams: tokenParams);
 ```
 </Code>
 
-### Client usage
+### Client usage <a id="client-usage"/>
 
-The client SDK automatically handles TokenRequests returned from your auth endpoint:
+The client SDK automatically handles TokenRequests returned from your auth endpoint. These examples use `authCallback`, but `authUrl` is also valid; see [choosing how the client fetches tokens](/docs/auth/token#auth-mechanism) for guidance on which to use.
```
paddybyers:
```suggestion
The client SDK automatically handles the `TokenRequest` returned from your auth endpoint. These examples use `authCallback`, but `authUrl` is also valid; see [choosing how the client fetches tokens](/docs/auth/token#auth-mechanism) for guidance on which to use.
```

src/pages/docs/auth/token/index.mdx
```diff
@@ -68,6 +68,21 @@ Ably enforces maximum TTL (time-to-live) limits:
 Attempting to create a token with a TTL that exceeds these limits will result in an error code [40003](/docs/platform/errors/codes#40003).
 </Aside>
 
+## Choosing how the client fetches tokens <a id="auth-mechanism"/>
+
+`authUrl` and `authCallback` are the two ways a client fetches a token from your server. Both apply to any token format, both are current, and both refresh tokens automatically before expiry. Choose based on the client environment and how much control you need over the request.
```
paddybyers:
```suggestion
`authUrl` and `authCallback` are alternative ways to enable a client to fetch tokens from your server when required, and they apply to both supported token formats. Choose based on the client environment and how much control you need over the request.
```

src/pages/docs/auth/token/index.mdx
```diff
@@ -68,6 +68,21 @@ Ably enforces maximum TTL (time-to-live) limits:
...
+
+| | `authUrl` | `authCallback` |
+| --- | --- | --- |
+| How it works | The SDK makes an HTTP request to a URL you specify and uses the response as the token. | The SDK calls a function you provide, which returns the token. |
+| Best for | Web clients that can pass cookies automatically to your auth endpoint. | Native and non-web clients, or any case where you need more control. |
```
paddybyers:
```suggestion
| Best for | Web clients that can pass cookies or other session context automatically to your auth endpoint. | Native and non-web clients, or any case where you need more control. |
```

src/pages/docs/auth/token/index.mdx
```diff
@@ -68,6 +68,21 @@ Ably enforces maximum TTL (time-to-live) limits:
...
+| Control over the request | Limited to `authMethod`, `authHeaders`, and `authParams` [auth options](/docs/api/realtime-sdk/types#auth-options). | Full: run any logic to obtain the token, set custom headers, and handle errors. |
+| Setup | A single client option pointing at your endpoint. | A function in your client code. |
+
+Use `authUrl` when a simple HTTP request to your server returns the token, particularly for web clients where the browser sends session cookies automatically. Reach for `authCallback` when the token fetch is non-trivial, for example retrieving the token from native storage, adding custom authorization headers, or handling errors yourself. `authCallback` is also the option for non-web clients, though native SDKs support `authUrl` as well.
```
paddybyers:
```suggestion
Use `authUrl` when a simple HTTP request to your server returns the token, particularly for web clients where the browser sends session cookies automatically. Use `authCallback` when the token fetch is non-trivial, for example retrieving the token from native storage, adding custom authorization headers, or handling errors yourself. `authCallback` is also the option for non-web clients, though native SDKs support `authUrl` as well.
```

paddybyers (changes requested):
I made a few suggestions.

I realise it's unrelated to this PR, but it's odd that there's a diagram for the non-recommended method, but no diagram for JWTs.

umair-ably:
thanks for the review @paddybyers, I've applied all your suggestions in a single commit - https://github.com/ably/docs/pull/3397/commits/0932bd408e5b4071c73a608b03db210dd2584a94

umair-ably:
>  realise it's unrelated to this PR, but it's odd that there's a diagram for the non-recommended method, but no diagram for JWTs.

fair point! I'm not sure what the original motivations were for leaving JWTs out but I would assume it's because it's a pretty standard auth mechanism. 

We also have a diagram for Embedded Token JWTs (which naturally builds on JWTs) but I'm not against an explicit one on the overview page - it'd make the page nicer to scan. Will track independently as a follow up for this

src/pages/docs/auth/token/ably-tokens.mdx
AndyTWF, line 9: Not this PR, but here `TokenRequest` has a space which I dont think it should

src/pages/docs/auth/token/ably-tokens.mdx
```diff
@@ -125,16 +125,16 @@ final tokenRequest = rest.auth.createTokenRequest(tokenParams: tokenParams);
...
-The client SDK automatically handles TokenRequests returned from your auth endpoint:
+The client SDK automatically handles the `TokenRequest` returned from your auth endpoint. These examples use `authCallback`, but `authUrl` is also valid; see [choosing how the client fetches tokens](/docs/auth/token#auth-mechanism) for guidance on which to use.
 
 <Code>
 ```realtime_javascript
 const realtime = new Ably.Realtime({
   authCallback: async (tokenParams, callback) => {
     try {
-      const response = await fetch('/api/ably-token');
+      const response = await fetch('/api/auth/token');
```
AndyTWF, line 137: `api/auth/token` is generic enough that readers/LLMs might confuse it with an Ably endpoint (which doesn't exist). Do you have any concerns here?

src/pages/docs/auth/token/index.mdx
```diff
@@ -68,6 +68,21 @@ Ably enforces maximum TTL (time-to-live) limits:
 Attempting to create a token with a TTL that exceeds these limits will result in an error code [40003](/docs/platform/errors/codes#40003).
 </Aside>
 
+## Choosing how the client fetches tokens <a id="auth-mechanism"/>
+
+`authUrl` and `authCallback` are alternative ways to enable a client to fetch tokens from your server when required, and they apply to both supported token formats. Choose based on the client environment and how much control you need over the request.
+
+| | `authUrl` | `authCallback` |
```
AndyTWF, line 75: There's also the `token` parameter which whilst not widely used, is still used with valid usecases. We should probably document it even if it's a "dont use this unless you know what you're doing"

src/pages/docs/auth/token/index.mdx
```diff
@@ -68,6 +68,21 @@ Ably enforces maximum TTL (time-to-live) limits:
...
+| --- | --- | --- |
+| How it works | The SDK makes an HTTP request to a URL you specify and uses the response as the token. | The SDK calls a function you provide, which returns the token. |
+| Best for | Web clients that can pass cookies or other session context automatically to your auth endpoint. | Native and non-web clients, or any case where you need more control. |
+| Control over the request | Limited to `authMethod`, `authHeaders`, and `authParams` [auth options](/docs/api/realtime-sdk/types#auth-options). | Full: run any logic to obtain the token, set custom headers, and handle errors. |
+| Setup | A single client option pointing at your endpoint. | A function in your client code. |
+
+Use `authUrl` when a simple HTTP request to your server returns the token, particularly for web clients where the browser sends session cookies automatically. Use `authCallback` when the token fetch is non-trivial, for example retrieving the token from native storage, adding custom authorization headers, or handling errors yourself. `authCallback` is also the option for non-web clients, though native SDKs support `authUrl` as well.
```
AndyTWF, line 82: You can set custom auth headers with `authUrl`?

AndyTWF: This paragraph in general just seems like a re-statement of the table, is it needed?

PR #3400, opened 2026-06-09
[DX-631] Pub/Sub JavaScript SDK API references (realtime interface)

m-hulbert:
## Description

This PR proposes a new format for the Pub/Sub JavaScript SDK (realtime interface) API references. 

It uses the new components and design for API references introduced for the Chat API refs, updates the content to the latest SDK version (2.22.1) and tidies up anything that was legacy. 

It also introduces a slightly new IA to better group like APIs, and ensures a better blend between the actual interface names used by the SDK (e.g. `RealtimeChannel`) and the concepts themselves (e.g. 'Channels').

For now, this doesn't update any links or remove the existing API references. This will follow once REST is complete for JavaScript, as we'll need to update the query param behaviour for API reference links. 1

### Checklist


src/pages/docs/pub-sub/api/javascript/realtime/auth.mdx
```diff
@@ -0,0 +1,278 @@
+---
+title: Auth
```
umair-ably, line 2: Just read through the auth page, it's infinitely better than what we originally had! The default values and even some side effects are way easier to understand.

Just a couple of notes that are probably only relevant for this page... "Auth" is both a concept and an interface, so I think we need to be stronger in the intro when framing what is and isn't on this page i.e. I think we should lead with JWTs more strongly and then highlight this page is largely only relevant if you're using the other auth types

src/pages/docs/pub-sub/api/javascript/realtime/auth.mdx
```diff
@@ -0,0 +1,278 @@
...
+meta_description: "API reference for the Authentication (Auth) interface in the Ably Pub/Sub JavaScript Realtime SDK."
+meta_keywords: "Ably Pub/Sub SDK, JavaScript, Auth API, authorize, createTokenRequest, requestToken, revokeTokens, AuthOptions, TokenParams, TokenDetails, TokenRequest"
+---
+
+The `Auth` interface is responsible for managing tokens and [authentication](/docs/auth) for an Ably client. Access it via the `auth` property of a [`Realtime`](/docs/pub-sub/api/javascript/realtime/realtime-client) client instance.
+
+The principal use of `Auth` is to create Ably `TokenRequest`s with `createTokenRequest()`, or to obtain Ably Tokens from Ably with `requestToken()`, and then issue these short-lived tokens to less-trusted clients. Typically, your servers should be the only devices to hold the [private API key](/docs/auth#api-keys), which is used to sign Ably `TokenRequest`s or request Ably Tokens. If you adopt this model, your private API key is never shared with clients directly.
```
umair-ably: `TokenRequest`s formats weirdly... the old prose of "`TokenRequest` objects" reads nicer

umair-ably: I think the aside is too low down and also doesn't necessarily highlight that this page is largely irrelevant if you do use JWTs... 

I think the intro itself could be reframed to something like:

> The principal use of Auth is to issue short-lived credentials to less-trusted clients without sharing your private API key. The recommended approach for most applications is Ably JWTs, which let your server mint tokens using standard JWT libraries - examples can be seen here. 

> Ably also supports signed TokenRequests... blah blah blah... details below

src/pages/docs/pub-sub/api/javascript/realtime/realtime-channel.mdx
```diff
@@ -0,0 +1,760 @@
+---
+title: RealtimeChannel
```
umair-ably, line 2: The ordering on this page is a little strange imo... I'd expect to see all the channel-centric operations first e.g. attaching, listeners, etc.. first, followed by all the message operations e.g. history, append, etc...

ttypic (approved):
LGTM

umair-ably (approved):

PR #3404, opened 2026-06-12
ait: improvements to ai transport channel rule page

owenpearson:
fix a couple of issues with this page:

- The page talks about what failure you will see when you don't have the rule configured, and the side effects of configuring the rule before it gets to what a namepsace is. A new user trying to get started probably won't know what a namespace is and won't care about the error behaviour, so namespaces ought to be introduced earlier.
- The page is just called "Enable channel rules" but really it's a section of the getting started guide for AI transport detailing how to configure a specific rule required for AI Transport, the fact that you're in the getting started section of the AI Transport section is not that prominent when coming here from a link so we shouldn't leave that implicit in the page title.

https://ably-real-time.slack.com/archives/C09SY1AQGK0/p1781260131449549?thread_ts=1781090478.352169&cid=C09SY1AQGK0

### DIFF
--- a/src/data/nav/aitransport.ts
+++ b/src/data/nav/aitransport.ts
@@ -34,7 +34,7 @@ export default {
           link: '/docs/ai-transport/getting-started/authentication',
         },
         {
-          name: 'Enable channel rules',
+          name: 'Configure the channel rule',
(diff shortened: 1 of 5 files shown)


rainbowFi (approved):
LGTM, thanks for dealing with this!

PR #3410, opened 2026-06-17
DX-1128: docs local cn + heights utils (drop @ably/ui/core/utils)

jamiehenson:
Part of **DX-1128**. Drops `@ably/ui/core/utils` by vendoring the two utilities docs uses.

## What
- `src/utilities/cn.ts` — `twMerge(clsx(...))`, local replacement for `@ably/ui/core/utils/cn`.
- `src/utilities/heights.ts` — layout height helpers, replacing `@ably/ui/core/utils/heights`.
- `clsx` + `tailwind-merge` promoted to direct deps (were transitive via `@ably/ui`).
- Codemodded ~34 import sites across components/pages.

## Behaviour
Pure import-source swap — same functions, same output. No visual change.

## Stack (DX-1128)

|  | PR | Role |
|--|----|------|
|  | #3409 | Design foundation — vendor `@ably/ui` tokens + reset/core CSS _(merged)_ |
| ▶ | #3410 | Local `cn` + `heights` utils |
|  | #3411 | Docs-native `Icon` component (Ably glyphs only; Heroicons imported directly) |
|  | #3412 | Presentational components (Badge, Button, LinkButton, FeaturedLink, Status) |
|  | #3413 | Interactive components (SegmentedControl, TabMenu) |
|  | #3414 | ProductTile + product data (leaf) |
|  | #3415 | syntax-highlighter util + registry |
|  | #3416 | Code + CodeSnippet — completes the lift; prunes dead CSS + drops `@ably/ui` Tailwind scan |

_#3410–#3416 are a linear stack, each building on the previous; #3409 has merged to `main`. ▶ = this PR._

## Testing

- Load the review app.
- Check a sufficient sample of pages load with no visual regressions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3409, opened 2026-06-17
DX-1128: docs local design foundation (vendor @ably/ui tokens + reset CSS)

jamiehenson:
Part of **DX-1128 / DXDR-005** (retire the shared `@ably/ui` package). Foundation PR — vendors `@ably/ui`'s Tailwind theme + CSS into docs so the package is no longer the styling source.

## What
- **Flattens** `@ably/ui`'s base Tailwind config into `tailwind.config.js` — the theme (colours, `ui-text-*` typography, screens, keyframes, plugins) is now inlined here instead of `require('@ably/ui/tailwind.extend.js')`.
- **Vendors** the reset + core + component CSS under `src/styles/ui/` (`properties.css` token vars, `text`/`buttons`/`forms`/`shadows`/`utils`, reset/normalize, Code/Slider/Flash component CSS, syntax-highlighter). `global.css` `@import`s these locally.
- Vendored tree added to `.prettierignore`.

## Self-sufficiency
The `node_modules/@ably/ui` Tailwind content glob and the full vendored CSS are **retained** here, so this PR (and every PR up to #3415) is independently deployable — the still-present `@ably/ui` components in the early stack keep their utility classes. The dead-CSS prune + glob removal land in #3416 once everything is local.

## Behaviour
No visual change intended. Theme verified equal to `@ably/ui`'s `extendConfig` output; `global.css` compiles to ~8.4k lines with `--color-*`, `ui-text-*`, `hljs` all present.

## Stack (DX-1128)

|  | PR | Role |
|--|----|------|
| ▶ | #3409 | Design foundation — vendor `@ably/ui` tokens + reset/core CSS |
|  | #3410 | Local `cn` + `heights` utils |
|  | #3411 | Docs-native `Icon` component |
|  | #3412 | Presentational components (Badge, Button, LinkButton, FeaturedLink, Status) |
|  | #3413 | Interactive components (SegmentedControl, TabMenu) |
|  | #3414 | ProductTile + product data (leaf) |
|  | #3415 | syntax-highlighter util + registry |
|  | #3416 | Code + CodeSnippet — completes the lift; prunes dead CSS + drops `@ably/ui` Tailwind scan |
|  | #3417 | Combined view of #3409–#3416 — review/demo only (based on `main`) |

_#3409–#3416 are a linear stack (each builds on the previous); #3417 is the combined view. ▶ = this PR._

## Testing

- Load the review app.
- Check a sufficient sample of pages load with no visual regressions.

[...]

PR #3411, opened 2026-06-17
DX-1128: docs-native Icon component and direct HeroIcon usage (drop @ably/ui/core/Icon)

jamiehenson:
Part of **DX-1128**. Replaces `@ably/ui/core/Icon` with a docs-native `Icon` (`src/components/Icon/`).

## What
- `index.tsx` — the docs-native `Icon`, which now renders **Ably's own glyphs only** (product / social / tech / `gui-prod` / badge); `set-unique-ids.ts` (useId-keyed) rewrites gradient/mask ids per instance; null fallback for unknown names.
- `glyphs/` — those glyphs as typed `.tsx` forwardRef components + a generated index, produced by running `@ably/ui`'s own SVGR pipeline (`generate-icons.ts` config) over its source SVGs. Byte-equivalent SVG output to the compiled originals — no visual change.
- **Heroicons are no longer resolved by name through `Icon`.** Every former `<Icon name="icon-gui-*-{outline,solid,mini,micro}">` is now a direct `@heroicons/react` component import at the call site, so they tree-shake. `@heroicons/react` promoted to a direct dep (outline/solid/mini/micro → 24/24/20/16).
- Codemodded the direct heroicon call sites; `Header`/`LanguageSelector` tests updated to query by role/label (and the burger button gained an `aria-label`).

## Behaviour
No visual change intended. Glyph coverage verified (no misses); Heroicons render identically as direct components.

## Stack (DX-1128)

|  | PR | Role |
|--|----|------|
|  | #3409 | Design foundation — vendor `@ably/ui` tokens + reset/core CSS _(merged)_ |
|  | #3410 | Local `cn` + `heights` utils |
| ▶ | #3411 | Docs-native `Icon` component (Ably glyphs only; Heroicons imported directly) |
|  | #3412 | Presentational components (Badge, Button, LinkButton, FeaturedLink, Status) |
|  | #3413 | Interactive components (SegmentedControl, TabMenu) |
|  | #3414 | ProductTile + product data (leaf) |
|  | #3415 | syntax-highlighter util + registry |
|  | #3416 | Code + CodeSnippet — completes the lift; prunes dead CSS + drops `@ably/ui` Tailwind scan |

_#3410–#3416 are a linear stack, each building on the previous; #3409 has merged to `main`. ▶ = this PR._

## Testing

- Load the review app.
- Check a sufficient sample of pages load with no visual regressions.


kennethkalmer (commented):
Quick browse around the site showed nothing _obviously_ broken, but botticus picked up some things to investigate when I asked it to review and cross-reference with the original code in `@ably/ui`

src/components/Layout/Footer.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useEffect, useMemo, useState } from 'react';
 import { useLocation } from '@reach/router';
-import Icon from '@ably/ui/core/Icon';
-import { IconName } from '@ably/ui/core/Icon/types';
+import Icon from 'src/components/Icon';
```
kennethkalmer, line 3: **Live regression — the "Was this page helpful?" feedback icons render nothing.**

This now imports the docs-native `Icon`, which only resolves vendored Ably glyphs. But `feedbackButtons` (lines 65-83) stores Heroicon names — `icon-gui-hand-thumb-up-outline/solid`, `icon-gui-hand-thumb-down-*`, `icon-gui-hand-raised-*` — and they're rendered via `<Icon name={button.colorIcon}>` / `name={button.monoIcon}` at lines 201-214. None are in `glyphs/`, so `glyphs[name]` is `undefined` and `Icon` returns `null`. The thumbs/hand icons silently disappear; only the text labels remain.

The Heroicon codemod converted JSX call sites but missed these because the names live in a `Record` data structure as string literals, not as `<Icon>` JSX. Contradicts the PR's "Glyph coverage verified (no misses)".

Fix: store the Heroicon **component** refs in `feedbackButtons` (e.g. `HandThumbUpIcon` from `@heroicons/react/24/outline` + `/24/solid`) and render them directly, matching the rest of the codemod.

(Note: `monoIcon`/`colorIcon` are typed `IconName` = `` `icon-${string}` ``, so TypeScript can't catch this — see comment on `types.ts`.)

src/components/Icon/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  style,
+  ...rest
+}: IconProps): React.ReactElement | null => {
+  const Component = glyphs[name] as SvgComponent | undefined;
+
+  // Per-instance id rewriting so a gradient/mask glyph rendered more than once on a
+  // page doesn't collide on duplicate DOM ids.
+  const uniqueId = useId();
+  const ref = useCallback((el: SVGSVGElement | null) => setUniqueIds(el, uniqueId), [uniqueId]);
+
```
kennethkalmer, line 39: Unknown name → `null` with no signal. Combined with the permissive `IconName` type, any typo or un-vendored icon disappears silently — no compile error, no runtime warning, and tests mock `Icon` so they won't catch it. This is exactly how the Footer feedback icons (see `Footer.tsx`) break with zero feedback.

The old `@ably/ui` Icon resolved the full website set (~hundreds) plus a Heroicon name fallback; this resolves ~60 vendored glyphs. Suggest a dev-only guard so misses surface during development:

```ts
if (!Component) {
  if (process.env.NODE_ENV !== 'production') {
    console.warn(`[Icon] no glyph for "${name}" — render skipped`);
  }
  return null;
}
```

src/components/Icon/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+import type { IconName } from './types';
+
+export type { IconName };
+
+export interface IconProps extends Omit<React.SVGProps<SVGSVGElement>, 'color'> {
+  name: IconName;
+  /** CSS width & height, e.g. "1rem" or "20px". */
+  size?: string;
+  /** Tailwind text-colour class — applied via `currentColor`. */
+  color?: string;
```
kennethkalmer, line 16: Comment nit: says "applied via `currentColor`", but `color` is just appended as a class via `cn(color, additionalCSS, className)` (line 44). No `currentColor` involved. Reword to "applied as a class".

src/components/Icon/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const uniqueId = useId();
+  const ref = useCallback((el: SVGSVGElement | null) => setUniqueIds(el, uniqueId), [uniqueId]);
+
+  if (!Component) {
+    return null;
+  }
+
+  return (
+    <Component
+      ref={ref}
```
kennethkalmer, line 51: `aria-hidden` is now forced on every icon. Check the icon-only social links in `Footer.tsx` (~line 304) and anywhere an icon is the sole content of an interactive/`<a>` element — with the icon hidden and no text, those lose their accessible name. The Header burger + LanguageSelector got `aria-label`s; confirm the social links are covered too.

src/components/Icon/types.ts
```diff
@@ -0,0 +1,3 @@
+// Permissive on purpose: covers the static glyph names and the dynamic
+// `icon-tech-${language}` usages without enumerating every glyph.
+export type IconName = `icon-${string}`;
```
kennethkalmer: `` `icon-${string}` `` accepts any matching string, so a name with no vendored glyph (e.g. the Footer Heroicon names) typechecks fine and then silently renders `null`. Pragmatic given the dynamic `icon-tech-${lang}` usage, but it removes the safety net. At minimum pair it with the dev-warn suggested in `index.tsx`; ideally type as a union of vendored glyph keys ∪ `` `icon-tech-${string}` `` so static names are checked while dynamic tech icons stay open.

kennethkalmer (commented):
Review focused on icon-name resolution after the switch to the docs-native `Icon`. One live regression (Footer feedback icons), plus a few sharp edges around silent null-rendering. Details inline.

src/components/Layout/Footer.tsx
jamiehenson: Good catch. Fixed in `6c8343686` — `feedbackButtons` now hold the Heroicon component refs (`HandThumbUpIcon`/`HandThumbDownIcon`/`HandRaisedIcon`, outline + solid) and render directly, matching the rest of the codemod. It was gated behind `ENABLE_FEEDBACK = false` so dormant rather than user-visible, but no reason to leave it broken. Did the fix here in the Icon PR since that's where the breakage originates.

~ 𝒞𝓁𝒶𝓊𝒹𝑒

src/components/Icon/index.tsx
jamiehenson: Checked — the icon-only social links already carry `aria-label={`Visit Ably on ${link.key}`}` (Footer ~304), and the burger + LanguageSelector got labels too, so nothing loses its accessible name. Leaving `aria-hidden` forced.

~ 𝒞𝓁𝒶𝓊𝒹𝑒

src/components/Icon/index.tsx
jamiehenson: Added in `6c8343686` — dev-only `console.warn` when a name has no glyph, as suggested. Pairs with the narrower `IconName` (see the types.ts thread) so misses surface at build for static names and at runtime-in-dev for the dynamic tail.

~ 𝒞𝓁𝒶𝓊𝒹𝑒

src/components/Icon/types.ts
jamiehenson: Narrowed in `6c8343686`: `IconName = keyof typeof glyphs | ` `` `icon-tech-${string}` `` `, and the `glyphs` map now uses `satisfies` so its keys stay a literal union. Static names are checked against the registry; the dynamic `icon-tech-*` stays open. Plus the dev-warn above for the runtime tail.

~ 𝒞𝓁𝒶𝓊𝒹𝑒

src/components/Icon/index.tsx
jamiehenson: Fixed → "Tailwind text-colour class, appended to the icon's className." (`6c8343686`)

~ 𝒞𝓁𝒶𝓊𝒹𝑒

jamiehenson:
Thanks @kennethkalmer - the bot has done its responses but to summarise here, I've tightened up the type definition for the icons (tbh, I would like this to be as minimal as possible in future by leaning on HeroIcons more, but I'm already satisfied with the direction this PR goes in in terms of dethroning the overcomplicated Icon component) and pulled in something that was further down the stack but opened up a regression here.

kennethkalmer (approved):
All five review threads addressed and verified against the pushed code:

- Footer feedback icons now hold direct Heroicon component refs (outline + solid), rendered directly — matches the codemod. (Confirmed dormant behind ENABLE_FEEDBACK, but right to fix.)
- IconName narrowed to `keyof typeof glyphs | `icon-tech-${string}`` with the glyphs map closed via `satisfies` — static names now typecheck against the registry.
- Dev-only `console.warn` on unresolved names covers the dynamic tail.
- Social links carry aria-labels, so forced `aria-hidden` keeps accessible names intact.
- Comment nit fixed.

LGTM.

PR #3419, opened 2026-06-17
feat(docs): support per-page JSON-LD via json_ld frontmatter [DX-1416]

AndyTWF:
## Why

This is the engineering dependency for the docs schema rollout in [DX-1416](https://ably.atlassian.net/browse/DX-1416) / MDR-076. The `docs-page-schema-jsonld` skill can now generate validated, per-page JSON-LD, but docs pages had no way to emit it — `MDXWrapper` could only render the auto-generated language schema (`TechArticle` + `SoftwareSourceCode`). This patch unblocks page-level schema so the AEO work can actually land on docs pages.

## What changed

- **`Layout.tsx`** — add `json_ld?: Record<string, unknown>` to the `Frontmatter` type.
- **`MDXWrapper.tsx`** — the `structuredData` memo now prefers `frontmatter.json_ld`:
  - present and an object → used verbatim, overriding the auto-generated language schema;
  - absent → existing behaviour unchanged;
  - present but not an object → `console.warn` and fall back to the generated schema.
- **`MDXWrapper.test.tsx`** — three new cases: override, `@graph` with no top-level `@type`, and malformed-value fallback + warning.

## Notes on approach

- The skill emits `@graph`-style documents that have `@context` but no top-level `@type`. The validated object is passed straight through to `<Head>`, so these render unchanged. I scoped the change to the two files named in the ticket and did **not** loosen `Head.tsx`'s `StructuredData` type — a local cast covers the `@graph` shape without widening the contract elsewhere.
- The ticket's illustrative `json_ld ?? schema` snippet had no validation; I implemented the fuller "Expected behaviour" spec (object check + warning), which the `??` shorthand wouldn't satisfy.
- The skill stores schema inline as JSON. Because YAML is a superset of JSON, an inline `{...}` parses to a JS object at frontmatter-load time (matching `Record<string, unknown>`), so no runtime `JSON.parse` is needed. A genuinely quoted string would hit the malformed-value branch (warn + fall back).

## Testing

- `jest src/components/Layout/MDXWrapper.test.tsx` — 15/15 pass.
- `eslint` — no errors on changed files.
- To test manually, go to `/docs/ai-transport/getting-started/vercel-ai-sdk` and check the `head` tags

[DX-1416]: https://ably.atlassian.net/browse/DX-1416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jamiehenson (commented):
GTG on the whole, obvs with the temp commit removed. 

One thing to raise before I approve here though - an overridden schema here loses the per-language schema granularity that the base schema offers (the `hasPart` field that gives `SoftwareSourceCode` entries per-language, which is relevant to docs given the multi-language nature of articles).

Given that this may be difficult to represent via per-page JSON-LD, this is a trade-off that I'll leave you to comment on. As you said, how truly impactful is this stuff anyway - it could be that the per-language fields are unnecessary when weighing up the value of this change.

AndyTWF:
@jamiehenson Currently an override is a full replacement - frontmatter json_ld ships verbatim and the generated TechArticle (incl. the per-language hasPart) is skipped.

I considered deep-merging the override onto the base so hasPart survives automatically, but I think full override is the right call, for one main reason: the base schema only ever emits a TechArticle, and the whole point of reaching for an override is usually to emit a different shape (FAQPage, HowTo, richer Article). The moment the type differs, merging hasPart of SoftwareSourceCode entries onto, say, a FAQPage produces incoherent schema - so merge only helps in the narrow "I want a TechArticle but tweak one field" case, which is the least likely reason to override in the first place. It also makes the emitted JSON-LD non-obvious (you can no longer see what ships by reading the frontmatter) and adds merge-semantics edge cases (array replace vs concat, type conflicts) for not much gain.

On the per-language fields specifically - agree with your instinct that they're low signal. SoftwareSourceCode pointing at ?lang= URLs isn't a documented rich-result enhancer; it's speculative. So I'd rather not build merge machinery to preserve it.

Proposal: keep full override (WYSIWYG, authoritative), and document in the skill that an override owns the whole schema, with a note that authors can include hasPart themselves if they want the per-language granularity on a multi-language page. Temp verification commit will be dropped before merge.

Happy to revisit if you feel the per-language data is worth more than we're giving it credit for.

jamiehenson (approved):
Nah, it's not a strongly held view. From what I remember, I put this initial schema work in Q4 last year as basically a punt itself with some finger-in-the-air decisions on fields to include based on some other examples in the wild.

A direct replacement via a prop with no fancy merging is cleaner and I can't justify the merge mess to preserve per language support otherwise

PR #3421, opened 2026-06-18
Support: Enterprise - Document legacy client options for custom endpoints

mclark-ably:
## Description

The enterprise customization page only shows the `endpoint` client option in its code examples. This option was introduced in certain SDKs at certain versions and is not universally available.

Customers on older SDK versions who configure a custom endpoint have no reference for the legacy approach, which can result in the option being silently ignored and traffic being routed incorrectly.

This PR:
- Adds a note to the "Modify your code" section pointing to the legacy options for customers whose SDK does not support `endpoint`
- Adds a new "Use legacy client options" section (between "Modify your code" and "Roll out your changes") with JavaScript examples for both the Ably subdomain and custom domain cases, using `environment` and `restHost`/`realtimeHost`/`fallbackHosts` respectively

### Checklist

### DIFF
--- a/src/pages/docs/platform/account/enterprise-customization.mdx
+++ b/src/pages/docs/platform/account/enterprise-customization.mdx
@@ -123,6 +123,10 @@ Repeat this step for all fallback hosts in the `fallbackHosts` array, if provide
 
 Now that you have tested your custom endpoints, you must modify your code to instantiate the Ably client with the provided `ClientOptions`.
 
+<Aside data-type='note'>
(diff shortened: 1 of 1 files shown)


mclark-ably:
@m-hulbert 

We discussed this a while ago and this came up on another support ticket recently so I have added this PR to see if the changes here make sense / look ok to highlight the older environment client option. I choose to stick with just an ably-js example for now as the current page only has these examples but it would probably good to get additional ones added in the future

m-hulbert (approved):

PR #3448, opened 2026-06-26
docs(ai-transport): add token streaming example, reframe append example

rainbowFi:
Refreshes the out-of-date AI Transport streaming examples (AIT-1030): adds a runnable token streaming example built on the AI Transport SDK, reframes the old append example under Pub/Sub, and fixes two example-gallery gaps found along the way.

### How to review

The branch is five single-purpose commits, best read in order:

1. **`fix(examples)`: product badge fix.** Example product keys (`ai_transport`, `liveobjects`) did not match `@ably/ui`'s ProductTile keys, so those gallery cards rendered without a badge. `ExamplesGrid` now maps them.
2. **`feat(examples)`: redirects for renamed example pages.** Gallery pages are generated at `/examples/<id>` from `src/data/examples`, and redirects were previously only sourced from MDX frontmatter, so renaming an example left a 404. Adds an optional `redirectFrom` field to `Example`, mirroring the MDX flow.
3. **`docs(pub-sub)`: reframe `ai-transport-message-per-response` as `pub-sub-message-append`.** The example always used raw Pub/Sub `appendMessage` + `rewind` rather than the SDK, so the READMEs, UI copy, and gallery entry now describe it accurately and point to AI Transport as the recommended approach for AI streaming. The code now models a generic text stream rather than an LLM response: `MockLLM` becomes `MockTextStream`, `Agent` becomes `Publisher`, and prompt-and-response naming becomes stream naming throughout. The channel stays in the `ai:` namespace because that namespace has mutable messages enabled, which `appendMessage()` requires; a code comment records that constraint. Also includes review fixes: a working `appendMessage()` doc link (`/docs/messages#append`; the API-reference anchor targets an H4 that gets no anchor generated), dead markup removed, API key placeholder aligned. The old URL redirects to the new one.
4. **`docs(ai-transport)`: the token streaming example (React + JavaScript)**, on `@ably/ai-transport` 0.4. Mirrors the [Core SDK getting started](https://ably.com/docs/ai-transport/getting-started/core-sdk): the agent flow (`createAgentSession` → `createRun` → `start` → drain `run.view` → `pipe` → `end`) and the client hooks (`ClientSessionProvider`, `useView`, `useClientSession`). The model is mocked so it runs without an API key, but the fake LLM emits the exact `UIMessageChunk` stream that `streamText(...).toUIMessageStream()` produces, so the agent code matches a production integration; the two necessary deviations (agent runs in-browser, token source mocked) are commented in the code. Both variants re-render on run lifecycle events (`session.tree.on('run')`) and gate streaming on `status === 'active'`, because `useView` alone would leave the UI stuck in the streaming state when a run reaches a terminal status (related SDK gap: AIT-1128). A Refresh button demonstrates recovery from channel history. Includes Sandpack deps, the root workspace dependency (`@ably/ai-transport ~0.4.0`, `ai ^6`, `ably ~2.23.0`) with the regenerated `yarn.lock`, gallery entry and image, and docs cross-links.
5. **`docs(ai-transport)`: remove `ai-transport-message-per-token`**, superseded by the token streaming example. Pure deletion plus a redirect for its URL.

### Verification

- Every `@ably/ai-transport` 0.4 call and every mocked `UIMessageChunk` shape was checked against the package's type definitions; both variants mirror the Core SDK getting started guide.
- All four changed example workspaces typecheck against the pinned 0.4.0 and build with vite; `yarn install` leaves the committed `yarn.lock` unchanged.
- Examples component and data tests pass (12), ETL tests pass (302), eslint clean on changed files.

### Known simplifications (intentional)

- There is a brief window after send, before the run is observed, where the prompt button re-enables; guarding it needs extra state that adds noise to a teaching example, and the in-browser agent responds within milliseconds.
- The React variant reloads history with a single `useView({ limit: 30 })` page while the JavaScript variant drains history via `hasOlder()`/`loadOlder()`; the JavaScript variant deliberately demonstrates the paging API.
- The append example keeps its pre-existing StrictMode quirk (a state update nested in an updater); its code is intentionally unchanged in this PR beyond the reframing copy.


src/components/Examples/ExamplesGrid.tsx
```diff
@@ -8,6 +8,12 @@ import { Image, ImageProps } from '../Image';
 import { DEFAULT_EXAMPLE_LANGUAGES } from '../../data/examples/';
 import { Example } from '../../data/examples/types';
 
+// Example product keys don't all match @ably/ui's ProductTile keys.
+const PRODUCT_KEY_MAP: Record<string, ProductName> = {
```
VeskeR: don't know about the @ably/ui, but shouldn't we fix example product keys instead? so they use correct ones from 'src/components/ui/ProductTile/data'?

data/createPages/index.ts
```diff
@@ -120,6 +120,18 @@ export const createPages: GatsbyNode['createPages'] = async ({
         layout: { sidebar: false, template: 'examples' },
       },
     });
+
+    exampleDatum.redirectFrom?.forEach((redirectFrom) => {
```
VeskeR: can't comment on this, worth checking with someone from docs team

examples/package.json
```diff
@@ -6,10 +6,10 @@
...
   },
   "workspaces": [
-    "ai-transport-message-per-response/javascript",
-    "ai-transport-message-per-response/react",
-    "ai-transport-message-per-token/javascript",
-    "ai-transport-message-per-token/react",
+    "ai-transport-token-streaming/javascript",
+    "ai-transport-token-streaming/react",
+    "pub-sub-message-append/javascript",
+    "pub-sub-message-append/react",
```
VeskeR: nit: it looks like "workspaces" and "scripts" below should be alphabetically sorted

examples/pub-sub-message-append/javascript/src/script.ts
```diff
@@ -1,31 +1,32 @@
 import * as Ably from 'ably';
-import { Agent } from './agent';
+import { Publisher } from './publisher';
 import { config } from './config';
 
-// Generate unique channel name for this session
-const CHANNEL_NAME = `ai:response-${crypto.randomUUID()}`;
+// Generate a unique channel name for this session. Channels in the ai
+// namespace have mutable messages enabled, which appendMessage() requires.
```
VeskeR: > Channels in the ai namespace have mutable messages enabled, which appendMessage() requires.

This is true for the example app ran on ably.com but not as a general rule and won't work if someones copies the repo. Maybe rephrase to clarify and point to /docs/ai-transport/getting-started/channel-rules?

examples/pub-sub-message-append/react/src/index.tsx
```diff
@@ -1,8 +1,8 @@
 import { StrictMode } from 'react';
 import { createRoot } from 'react-dom/client';
-import App from './App.tsx';
+import App from './App';
 
-createRoot(document.getElementById('root')!).render(
+createRoot(document.getElementById('root')).render(
```
VeskeR:
```suggestion
createRoot(document.getElementById('root')!).render(
```

VeskeR: also apply the same to `examples/ai-transport-token-streaming/react/src/index.tsx` file

VeskeR (changes requested):
LGTM, couple of minor edits:

1. We also need to update `ai-transport-token-streaming` example to 0.5.0 ait sdk. The only breaking change is that `AgentSession.close()` is gone, and in non-durable execution examples we should call `AgentSession.end()`. I'm okay with this being changed in the follow-up PR, or here.
2. my Claude flagged a bunch of pre-existing broken anchor links, like `/docs/channels#create → /docs/channels#use`. Could you please double check in the affected files?

VeskeR (approved):
LGTM content wise.

Two outstanding questions above to the docs/devex team to resolve

rainbowFi:
Agree with Mark that no redirects are required. Moved the badging fix out into https://github.com/ably/docs/pull/3483 for wider discussion/work.

PR #3466, opened 2026-07-09
[DX-1518] feat: site-wide dark mode + theme toggle

jamiehenson:
## Site-wide dark mode + theme toggle

Adds a site-wide dark mode to the docs, with a light / dark / system toggle that defaults to the reader's OS preference. Jira: **DX-1518**.

The `@ably/ui` cut-off (DX-1128) is what made this tractable — the localised components already carried their `dark:` variants and the vendored highlight.js theme already had `.ui-theme-dark` rules; they were just dormant because nothing ever set the class. This PR activates that and closes the gaps.

### How it works
- **No-flash init** — an inline script in `gatsby-ssr` sets `ui-theme-light` / `ui-theme-dark` on `<html>` from `localStorage` (or the OS preference) before first paint, so the static build never flashes the wrong theme.
- **Theme context** — `ThemeProvider` / `useTheme` (light / dark / system), persisted to `localStorage`, reflected as the `ui-theme-*` class which drives Tailwind's `dark:` variants (`darkMode: ['selector', '.ui-theme-dark']`). Storage reads are guarded and values normalised, so blocked storage or a corrupt value falls back to the default instead of crashing the root provider.
- **Toggle** — a three-cell segmented control (System / Dark / Light) in the footer, under the status unit.
- **Systemic typography** — the vendored `ui-text-*` classes hardcoded dark colours (only labels had a dark override); added dark overrides for headings/body + fixed an inline-code rule that targeted `.dark` instead of `.ui-theme-dark`.
- **Page canvas** — `body` had no background, so dark mode showed white behind themed sections; set an explicit body background per theme.
- **JS-driven renderers** wired to the resolved theme: the Sandpack examples editor and the Redoc API reference (which gets a dedicated dark theme object, re-initialised safely on toggle — script load memoised, listeners/observer torn down between inits so nothing accumulates).
- **Odds and ends** — dark logo variant, admonitions, examples grid, decorative graphics hidden in dark, remaining light-only components.

### Design-review polish
A round of designer feedback, addressed on top of the base implementation:
- **Tech-logo glyphs** — Objective-C, PHP, Vercel and the generic web/code glyph hardcoded near-black fills (invisible on dark); switched to `currentColor` so they inherit the theme-aware text colour. (Next.js and Postgres are two-tone marks that need a separate dark asset — deferred, see below.)
- **Homepage platform card** — added dark variants of the three stack-layer SVGs, CSS-toggled against the light set (`dark:hidden` / `hidden dark:block`).
- **Links** — `.ui-link` plus the hardcoded blue call sites (FeaturedLink, API-reference breadcrumbs) resolve to the GUI blue (`#4DA6FF`) in dark.
- **Admonitions** — designer palette applied: coloured fills at `100 / 900` (light / dark), strokes standardised to `500 / 500`; `neutral` unchanged (`100 / 1200`, `500 / 800`); `usp` kept deliberately borderless/fill-less with an `orange-700 / 500` stroke. usp headline made visible in dark.
- **Tiles** — headings flip to `neutral-000`.
- **Navigation selection** — header, product-bar and left-nav selected states softened from the bold orange to a subtle `orange-1100` tint (keeping the `orange-600` accent border); also fixed a left-nav light-on-light legibility bug where the selected item kept its light `orange-100` fill in dark.
- **Native controls** — `color-scheme` now follows the theme, so scrollbars and form controls render dark.
- **Borders** — the dark-mode inversion of `neutral-300` borders moved from `neutral-1000` to `neutral-1100` site-wide (cards, panels, API tables, dropdowns, code-block chrome, dividers) for subtler, lower-contrast edges; two off-standard borders (`neutral-200`, legacy `light-grey`) normalised to the standard `neutral-300` / `neutral-1100` pair.
- **Typography safety net** — the `ui-text-*` dark overrides (`title/h1–h5/quote`, `p1–p4`, `sub-header`, `label1–4`, `code-inline`, links) auto-flip role-classed text without a per-call-site `dark:`. All are wrapped in `:where()` (zero specificity) so they beat the hardcoded base by source order yet defer to any explicit call-site `dark:text-*` — which is what preserves e.g. the breadcrumb's dimmed current-page crumb and inactive nav/tab states (`text.css` loads after Tailwind's utilities, so a plain override would wrongly win). A sweep confirmed residual dark-on-dark risk is effectively nil apart from the two deferred two-tone logos.

### Verified in-browser (dark + a light/toggle pass)
Homepage, MDX doc pages (prose, tables, admonitions, code blocks, inline code), the Redoc API reference, the examples listing, and header/footer chrome. The design-review round above was verified the same way — homepage platform card, `<Tiles>` grid (incl. the Objective-C / PHP / web tech-logo tiles), the header / product-bar / left-nav selection states, and each admonition variant. The Redoc re-init leak fix was measured directly (scroll-listener add/remove balanced across a toggle; single Redoc render). `yarn lint` clean; **190 unit tests pass**.

### Out of scope (follow-up)
**Content imagery (Phase 4)** — inline MDX diagrams/screenshots with baked-in light backgrounds (e.g. the platform architecture diagram, example preview shots) need dark asset variants; deferred to a separate effort. The default theme is `system` (it follows the reader's OS), so dark-preferring readers may see these light-background images on dark pages until the dark variants land.

**Two-tone tech-logo glyphs** — Next.js and Postgres are multi-tone marks (a filled shape with a contrasting mark knocked over it); a single `currentColor` can't flip both tones, so correct dark rendering needs a separate dark asset with a CSS toggle (like the platform-card layers). Deferred.

Not yet eyeballed: the live Sandpack editor (gated behind a logged-in API key), a mobile/responsive pass, and the nested API reference tables — the systemic fixes should cover them but they warrant a look on the review app.

### Testing
Add the `review-app` label for a deployment, then use the footer toggle (System / Dark / Light) and walk the page types above.

/cc @team-deved for review.


jamiehenson:
@m-hulbert @AndyTWF teed this up as a code review as from this point there's only gonna be incremental styling updates that won't change the body of the work. I appreciate you're both busy men, so will rope in other reviewers if this is too far down your list.

Before merge, this is pending signoff from design, but kicking off now as I believe we're close after two rounds.

src/components/Layout/ThemeToggle.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useTheme } from 'src/contexts/theme-context';
+import { DEFAULT_THEME, type Theme } from 'src/utilities/theme';
+
+const OPTIONS: { value: Theme; label: string; icon: typeof SunIcon }[] = [
+  { value: 'system', label: 'System theme', icon: ComputerDesktopIcon },
+  { value: 'dark', label: 'Dark theme', icon: MoonIcon },
+  { value: 'light', label: 'Light theme', icon: SunIcon },
+];
+
+const ThemeToggle: React.FC = () => {
```
AndyTWF, line 14: Feels like there should be a basic test here?

src/pages/docs/api/control-api.tsx
```diff
@@ -16,7 +16,7 @@ const ControlApi = () => {
     <>
       <Head title={meta_title} metaTitle={meta_title} canonical={canonical} description={meta_description} />
       <div className="ml-2 mt-8 mb-6">
-        <div className="text-gui-default hover:text-gui-hover focus:text-gui-focus focus:outline-gui-focus group ui-text-p2">
+        <div className="text-gui-default hover:text-gui-hover dark:text-gui-default-dark dark:hover:text-gui-hover-dark focus:text-gui-focus focus:outline-gui-focus group ui-text-p2">
```
AndyTWF, line 19: Very small nit, we use this in many places, is it worth a shared constant to prevent future divergence?

AndyTWF (commented):
Looking good, my retinas thank you. I would move the theme switcher to the top of the page if possible, the vast majority of people will probably never spot it down the bottom.

src/pages/docs/api/control-api.tsx
jamiehenson: This PR is supposed to be a quick win, I don't really want to get into the business of DRYing and restructuring.

jamiehenson:
@AndyTWF thanks. 

> I would move the theme switcher to the top of the page if possible, the vast majority of people will probably never spot it down the bottom.

I put it in the footer because it was there in one of the most recent designs I saw around this. I anticipated pushback from Design so put it in the footer - I can agree with this, it's a utility, not a feature. The header is also tightly contested for space which could have responsive implications. Design haven't said anything contrary

AndyTWF (approved):
Thanks for the feedback @jamiehenson - if design's happy with things, I'm happy :)

PR #3470, opened 2026-07-13
[DX-1525]: Collapse long snippets behind a "Show more" toggle

m-hulbert:
## Description

This PR collapses codeblocks by default if they are above 15 lines long, adding a 'show more' and 'show less' toggle so that it's easy to switch between them.

The intention is to improve the readability of pages with multiple, long codeblocks such as getting started guides.

### Checklist


umair-ably (approved):
1 nit... we can only calculate the height of the code blocks after the page has rendered, which in theory means all pages will load with the code blocks fully expanded, and then instantly redraw them to be collapsed if needed (which could cause a subtle flash or layout shift). 

However, I've tested this across several pages in the preview app and it looks perfect - nothing funky spotted

PR #3468, opened 2026-07-10
AI Transport docs: sort Features nav section alphabetically

VeskeR:
## Summary

The AI Transport **Features** navigation section had grown to 18 entries in an ad-hoc order, which made the sidebar hard to scan and made it difficult to locate a specific feature. This PR sorts that list alphabetically by display name so the section has a predictable, scannable order.

## Scope

Only the `Features` list in `src/data/nav/aitransport.ts` is reordered. Sections whose order carries meaning are intentionally left unchanged:

- **Getting started** and **Frameworks** follow the intended SDK progression (Core SDK → Vercel AI SDK → Vercel WDK → Temporal).
- **Concepts** follows the object-model / learning progression (Sessions → Connections → Runs → Steps → Invocations → …), so alphabetizing would scatter the model hierarchy.
- **Production**, **API reference**, and **Internals** are short and already grouped logically (for example, API reference keeps the Core → Vercel → Temporal SDK grouping).

No pages are added, removed, or relinked — the diff is a pure reordering (34 insertions / 34 deletions) and all 18 feature links are intact.

## New Features order

Agent presence · Branching, edit, and regenerate · Cancellation · Chain of thought · Concurrent turns · Database hydration · Double texting · Durable execution · History and replay · Human-in-the-loop · Interruption · LiveObjects State · Multi-device sessions · Optimistic updates · Push notifications · Reconnection and recovery · Token streaming · Tool calling

## Notes

- Based on `main` (not the in-flight WDK docs branch) so it is a clean, standalone change. The WDK nav entries live in Getting started and Frameworks, which are untouched here, so there is no overlap.
- Opened as a **draft** to confirm the ordering decision and scope before finalizing. If you also want the API reference method lists or the Internals section alphabetized, that is a quick follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3471, opened 2026-07-14
docs(ai-transport): document interruption and steering

zknill:
Documents the AI Transport interruption and steering feature, and verifies the affected agent-side docs against the SDK source (the `ait-1017-steering-steps` branch, the release target).

- Replaces the old `features/interruption.mdx` with `features/interruption-and-steering.mdx`, covering the three change-direction patterns: steer the active Run with a follow-up prompt, cancel and re-prompt, or send alongside as a concurrent Run.
- Adds a "Steer a Run" section to `concepts/runs.mdx` documenting the client-side `steer()` API (`{ published, outcome }`, `consumed`/`runTerminalReason`).
- Repoints the nav entry ("Interruption" → "Interruption and steering") and cross-links in `cancellation.mdx`, `concurrent-turns.mdx`, `double-texting.mdx`, and `why/index.mdx`.
- Fixes the agent loop in the steering page to match the SDK: `createRun` is synchronous (no `await`), drain `run.view` with `loadOlder()` before `run.start()`, feed `run.view.getMessages()` to the model (not `run.messages`, which is turn-only), and tear down with `session.end()`. Corrects a stray `ActiveRun` reference to `ClientRun`.
- Fixes the cancellation page's server-side abort-signal snippet, which fed `run.messages` (this Run's own turn) straight to the model, losing all prior conversation context. It now drains `run.view` and feeds the full conversation.


src/data/nav/aitransport.ts
```diff
@@ -132,76 +132,76 @@ export default {
       name: 'Features',
       pages: [
         {
-          name: 'Token streaming',
-          link: '/docs/ai-transport/features/token-streaming',
+          name: 'Agent presence',
```
VeskeR, line 135: bringing in @m-hulbert comment from https://github.com/ably/docs/pull/3468 regarding alphabetical ordering here:
> I agree that the feature list is rapidly expanding, and it would help to make that easier to navigate; however I'm not convinced that alphabetising the list is the right way to do this. In terms of releative 'importance' - is 'Agent presence' the first thing we believe should be read to understand the value proposition and how AI Transport works?

IMO, the core value proposition should be immediately stated and clear in the ["About AI Transport"](https://ably.com/docs/ai-transport). "How AI Transport works" is covered by ["Concepts"](https://ably.com/docs/ai-transport/concepts).
And "Features" then shows you what you can build with it. If you're just skimming through the docs to see the overall feature set, I don't think there's a material difference whether "Features" is sorted alphabetically or based on our assumptions of what we deem important - people will still be able to find what they want. But when you're returning to the documentation, having the list sorted makes it way easier to navigate.

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
```diff
@@ -0,0 +1,195 @@
...
+}
+
+await run.end({ reason: 'complete' });
+await session.end();
+```
+</Code>
+
+Each `pipe()` records which steering messages the agent had seen before producing that response. The client uses those records to resolve each message's outcome when the Run ends.
+
+## Implement cancel-then-send <a id="cancel-then-send"/>
```
VeskeR: should be "Cancel and re-prompt" to match the table above

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
```diff
@@ -0,0 +1,195 @@
...
+  ? <StopButton onClick={() => cancelAll()} />
+  : <SendButton onClick={handleSend} />;
+```
+</Code>
+
+## Edge cases and unhappy paths <a id="edge-cases"/>
+
+- If a steering message arrives just as the agent finishes its last `pipe()`, `outcome` resolves to `consumed: false`. This is normal under load. Show the user that the message didn't reach the agent rather than dropping it silently.
+- Once a Run has ended, `activeRun.steer()` rejects both promises immediately and publishes nothing. Only allow steering while `RunInfo.status === 'active'`.
+- If a Run is `'suspended'` (waiting on a tool approval, for example), `outcome` stays pending until the Run ends. Don't block UI updates on it.
```
VeskeR: A line higher up in "Steer the active Run" section:
> Absent if `outcome` resolves on a `run-suspend` instead of `run-end`

Suggesting `outcome` can resolve on suspend. But this lines says "`outcome` stays pending until the Run ends".
Looking at the implementation, it is an **already-consumed** steer resolves, and a **non-yet-resumed** stays pending.
Should clarify these two entries

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
```diff
@@ -0,0 +1,195 @@
...
+  - `consumed: true`: the agent saw the steering message before it sent its response.
+  - `consumed: false`: the Run ended before the agent saw the message.
+  - `runTerminalReason`: how the Run ended (`'complete'`, `'cancelled'`, or `'error'`). Absent if `outcome` resolves on a `run-suspend` instead of `run-end`.
+
+On the agent side, `Run.hasInput()` controls the loop. It returns `true` while there's input the agent hasn't responded to (the original prompt, or a steering message from the previous iteration), and `false` once the agent has caught up. The optional `onSteer` hook fires when a steering message arrives; use it to cancel an in-flight model call:
+
+<Code>
+```javascript
+const run = session.createRun(invocation, {
+  onSteer: () => {
```
VeskeR: I think it's better for `onSteer` example to have its own subsection under "Steer the active Run", showcasing the eager-cancel behavior and how you'd wire it up

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
```diff
@@ -0,0 +1,195 @@
...
+- Cancel is asynchronous. A small tail of tokens arrives after `session.cancel(runId)` returns and before the agent's `abortSignal` fires. The view emits them on the cancelled Run; treat them as belonging to that Run, not the new one.
+- A Run cancelled while awaiting a tool result ends with `'cancelled'`. Tool calls triggered before the cancel may still run on your server unless your handler honours the same `AbortSignal`.
+- A network drop on the client cancels nothing. The agent keeps streaming into the session. When the client reconnects, the response is still there. Check `session.view.runs()` on reconnect to decide whether to show Stop.
+- Send-alongside is rate-limited by your channel and any server-side concurrency you enforce. See [concurrent turns](/docs/ai-transport/features/concurrent-turns).
+- `'cancelled'` is reported on the Run's terminal `RunEndReason`. Don't infer cancellation from the absence of further tokens; subscribe to `view.on('run', ...)` and read the lifecycle event.
+
+## FAQ <a id="faq"/>
```
VeskeR, line 239: FAQ had dropped two items worth keeping from the previous version:
- Does cancel work from a different device?
- What if the agent endpoint is down when I send the follow-up?

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+### Can the agent ignore a steering message?
+
+Yes. `consumed: true` only means the agent saw the message, not that the LLM acted on it. Whether the model heeds new mid-flight input depends on the model, the system prompt, and what the agent passed in.
+
+### Can I cancel one of several concurrent Runs without touching the others?
+
+Yes. `session.cancel(runId)` only cancels the matching Run. Pull `runId`s from `session.view.runs()` to target the right one.
+
+### How does steering differ from concurrent turns?
```
VeskeR: We have three patterns for interruption and steering, so either compare them all here, or just drop this FAQ - we already described the diffference in the table above in "How it works"

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
```diff
@@ -0,0 +1,195 @@
...
+});
+
+// Drain run.view for the full conversation before starting. run.messages is
+// only this Run's own turn, so feed run.view to the model instead.
+while (run.view.hasOlder()) await run.view.loadOlder();
+await run.start();
+
+while (run.hasInput()) {
+  const conversation = run.view.getMessages().map(({ message }) => message);
+  const stream = await model.stream({ messages: conversation });
```
VeskeR: The rest of the page uses the Vercel codec in examples, so the agent loop should also use `streamText` + `toUIMessageStream()` here too rather than abstract `model.stream()`.

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
```diff
@@ -0,0 +1,195 @@
+---
+title: "Interruption and steering"
+meta_description: "Let users change direction mid-response in Ably AI Transport. Three patterns: steer the active Run with a follow-up prompt, cancel and re-prompt, or send alongside as a concurrent Run."
+meta_keywords: "AI interruption, interrupt, steering, steer, activeRun.steer, follow-up prompt, cancel and send, concurrent turns, AI Transport, Ably"
+intro: "Your users can change direction mid-response. AI Transport gives you three patterns: steer the active Run with a follow-up prompt, cancel and re-prompt, or send alongside as a concurrent Run."
+redirect_from:
+  - /docs/ai-transport/features/interruption
```
VeskeR, line 7: Recheck for stale `/interruption` links - found one on the [Roadmap](https://ably-docs-zak-docs-stee-dhtkzh.herokuapp.com/docs/ai-transport/roadmap) page

src/data/nav/aitransport.ts
zknill: Ok, well @m-hulbert if you have a concrete proposal for an order then please share that here. In the mean time, this re-ordering makes the features findable when you want to return to a page you already know exists. 

(I'm highly allergic to comments along the lines of "not this thing", without "this instead")

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
zknill: done

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
zknill: done

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
zknill: fixed, and added a few others

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
zknill: done

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
zknill: done

src/pages/docs/ai-transport/features/interruption-and-steering.mdx
zknill: dropped

.claude/skills/write-docs/SKILL.md
```diff
@@ -481,7 +497,7 @@ When reviewing someone else's docs PR (or your own at PR time), walk this list.
 **Writing standards**
 
 9. Page complies with the rules in [`writing-style-guide.md`](../../../writing-style-guide.md). Use the verification greps above to spot the common violations (em dashes, bold-prefix bullets, Latin abbreviations, subjective phrases, vague modals, "we" as the subject, AI-fingerprint patterns).
-10. Author ran `/deslop` (or equivalent AI-pattern check).
```
VeskeR, line 484: I think we should keep this as item 11

VeskeR (approved):
LGTM, one note about the skill change

.claude/skills/write-docs/SKILL.md
zknill: it's not a shared skill, so I think this is a meaningless line

PR #3472, opened 2026-07-15
control-v1: add missing type to conflationEnabled

surminus:
I generate the Terraform provider schemas from this spec, and tfplugingen-openapi silently skips any property that has no type. `conflationEnabled` in the three namespace schemas (`namespace_post`, `namespace_patch`, `namespace_response`) has a default, description and example but no `type`, so the attribute vanishes from the generated schema without any error. I scanned the rest of the spec and these three are the only untyped properties in the file.

The Terraform provider currently carries this fix as a local patch on its vendored copy of the spec (`codegen/spec-fixes.patch`), which I can drop once this lands.


m-hulbert (approved):
LGTM

PR #3473, opened 2026-07-17
[DX-1518] feat: dark mode for /examples (Sandpack previews)

jamiehenson:
## Dark mode for `/examples` (Sandpack previews)

Extends the site-wide dark mode (#3466, now merged) to the `/examples` pages. The runnable examples are small Vite subprojects surfaced through Sandpack; until now they were light-only regardless of the docs theme. This makes each **preview follow the docs theme** and gives every example **underlying dark-mode styling**.

### How it works

**Preview theming — bake the theme in, reload on change** (`src/components/Examples/sandpackThemeSync.ts` + `ExamplesRenderer`)

The examples run in a cross-origin Sandpack iframe, so the docs page can't restyle them directly. Rather than live cross-iframe messaging (an earlier attempt that proved timing-sensitive and flaky), the theme is **baked into each example's files**:

- A hidden module injected into every example adds `.dark` and sets Tailwind `darkMode:'class'`, and the `<head>` paints the dark canvas from the first frame (`background` + `color-scheme`) so there's no white flash before the bundle runs.
- The files are regenerated per theme and the `SandpackProvider` is **keyed on the resolved theme**, so a theme change fully reloads the preview on a fresh iframe. A full reload (not an in-place hot reload) tears down the example's side effects — e.g. its Ably connection — rather than stacking a second instance on each toggle. Trade-off: a short reload when the theme changes.
- The Sandpack preview container gets a dark background in dark mode (no white frame around the example).
- One shared fix: the examples' `--primary` token is identical in light and dark, so `uk-text-primary` (body text) would stay low-contrast on dark — remapped to `--foreground` in dark via a single injected rule.

**Example styling** — all 49 variants (React + JavaScript)

- Hardcoded light Tailwind utilities gain `dark:` counterparts (default grays); custom CSS gets `.dark`-scoped overrides via the shadcn CSS-variable blocks the examples already ship.
- Covers both static markup (`.tsx`/`.html`/`.css`) and the dynamically-rendered markup in `.ts` files (e.g. the annotations components, presence/spaces scripts).
- Semantic and per-user colours are preserved — status dots, toasts, avatar/cursor colours, annotation-type badge hues. `pub-sub-live-voting` (its own token set, not shadcn) got a dedicated `.dark` token block.

**Docs-side** — the example-page README renderer table chrome (`examples.tsx`) gained `dark:` variants.

### Verification (review app)

The Sandpack preview only renders behind a logged-in API key, so it can't be verified locally — it needs a deployed, signed-in build (the `review-app` label is applied). Confirmed live on the review app:

- Previews render in the docs theme with the dark canvas (no white container).
- Toggling System / Dark / Light reloads the previews into the matching theme, both directions, no manual refresh.
- A full reload on toggle (rather than a hot reload) — so no duplicate example instances / stacked presence members.

The latest first-frame-dark and full-reload fixes are deployed for a final pass. `yarn lint` clean; **196 unit tests pass**.

### Notes / follow-ups

- React previews may still show a brief white frame during the raw bundle fetch on a language switch (Sandpack serves its own template HTML there, which the injected `<head>` can't reach); JS previews paint dark from the first frame.
- A few card/row surfaces use `dark:bg-gray-900` on a `dark:bg-gray-900` parent and can blend; easy to bump to `gray-800` where noticed.
- The injected theme module import is visible at the top of the shown entry file, and the `darkMode` config in `index.html` — minor infra visible in the example code readers copy.
- Pre-existing `text-grey-500` (British spelling — not a real Tailwind class, no-op) left as-is in a couple of chat examples.

### Testing

Add/keep the `review-app` label for a deployment, then sign in, open a few example pages across product families, and toggle System / Dark / Light in the footer.

### DIFF
--- a/examples/ai-transport-message-per-response/javascript/index.html
+++ b/examples/ai-transport-message-per-response/javascript/index.html
@@ -7,15 +7,15 @@
     <title>AI Transport Message Per Response - JavaScript</title>
   </head>
 
-  <body class="bg-gray-100">
+  <body class="bg-gray-100 dark:bg-gray-800">
     <div class="max-w-6xl mx-auto p-5">
       <!-- Response section with always visible status -->
       <div class="mb-4">
         <div class="flex-1">
-          <div class="text-sm text-gray-600 mt-4 mb-2 flex justify-between">
+          <div class="text-sm text-gray-600 dark:text-gray-300 mt-4 mb-2 flex justify-between">
             <span id="prompt-display"></span>
             <div class="flex items-center gap-2">
-              <span class="text-xs bg-gray-200 px-2 py-1 rounded flex items-center gap-1">
+              <span class="text-xs bg-gray-200 dark:bg-gray-700 px-2 py-1 rounded flex items-center gap-1">
                 <span id="processing-status">Ready</span>
               </span>
               <!-- Disconnect/Reconnect button -->
@@ -24,7 +24,7 @@
               </button>
             </div>
           </div>
-          <div class="p-4 border border-gray-300 rounded-lg bg-gray-50 h-48 overflow-y-auto text-base leading-relaxed">
+          <div class="p-4 border border-gray-300 dark:border-gray-700 rounded-lg bg-gray-50 dark:bg-gray-900 h-48 overflow-y-auto text-base leading-relaxed">
             <span id="response-text">Select a prompt below to get started</span>
             <span id="cursor" class="text-blue-600"></span>
           </div>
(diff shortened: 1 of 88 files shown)


examples/ai-transport-message-per-response/javascript/index.html
```diff
@@ -36,7 +36,7 @@
         <div class="flex flex-wrap gap-2" id="prompt-buttons">
           <button
             id="prompt-button"
-            class="px-3 py-2 text-sm border rounded-md transition-colors bg-white hover:bg-blue-50 border-gray-300 hover:border-blue-300 cursor-pointer"
+            class="px-3 py-2 text-sm border rounded-md transition-colors bg-white dark:bg-gray-900 hover:bg-blue-50 dark:hover:bg-blue-950 border-gray-300 dark:border-gray-700 hover:border-blue-300 dark:hover:border-blue-700 cursor-pointer"
```
AndyTWF, line 39: This example looks... wrong on the demo site compared to prod. I can't see any buttons and it appears to have lost its structure?

AndyTWF (commented):
The colours seem fine, but looks like a few examples have lost their overall structure

examples/ai-transport-message-per-response/javascript/index.html
jamiehenson: Believable, I haven't scrutinised this. Will tell the bot to do better.

PR #3483, opened 2026-07-22
fix(examples): show product badge for AI Transport and LiveObjects cards

rainbowFi:
## What

AI Transport and LiveObjects example-gallery cards render with no product badge. The example product keys (`ai_transport`, `liveobjects`) don't match the existing set of keys. 

This PR correctly adds them, and also renames 'auth' to 'platform' for consistency.


splindsay-92 (approved):
lgtm

PR #3498, opened 2026-07-28
Zak/ait 1214 docs impovements

zknill:
## Summary

Two related pieces of AI Transport docs work.

**Correctness fixes.** Three things were wrong rather than just untidy:

- Statically-declared tools encode as `tool-${name}` parts, not `dynamic-tool`. Client-side lookups that matched only `dynamic-tool` silently missed them. Now matched via an `isToolPart` guard and the AI SDK's `getToolName`.
- The human-in-the-loop server example used a tool with no `execute`, which can only ever reach `input-available` — never the `approval-requested` state the client example checked for. Replaced with the demo-grounded pattern: one tool with both `execute` and a per-call `needsApproval` gate.
- Four core API pages said `view.send()` fires a POST to wake the agent. Only the Vercel `ChatTransport` does that; the core session is HTTP-free and the application must POST the invocation itself. `client-session.mdx` contradicted itself on the same page.

**Concepts consolidation.** Six concept pages folded into the three that carry the model: steps and invocations into Runs, connections into Sessions, authentication into the getting-started guide, codecs into codec architecture, infrastructure into Why AI Transport. Every old URL keeps a redirect and the nav drops the removed entries.

**House-style pass** across the section: "session" as the default term with "channel" reserved for channel-level detail (channel rules, capability scoping, wire headers, rate limits); primitives lowercased in prose and headings; links inline on the noun instead of trailing "see X for detail"; contrast constructions folded into plain sentences; "capability" reserved for the auth model; realtime as one word; code blocks labelled client or agent.

The rules are recorded in the in-repo `write-docs` skill so a later pass reapplies them instead of undoing them.


ttypic (approved):
LGTM

PR #3499, opened 2026-07-28
Preserve URL fragments when persisting the language param on links

m-hulbert:
## What

The MDX `Anchor` component carries the reader's `?lang=` (and `client_lang`/`agent_lang`) query params onto every internal link, but rebuilds the href as `url.pathname + url.search` — silently discarding `url.hash`.

## Impact

Whenever the current page URL has a language param (i.e. any time a reader has touched the language selector, or followed a `?lang=` link), **every anchored internal link on the site loses its `#fragment`** and lands at the top of the target page instead of the linked section.

## Fix

Include `url.hash` in the rebuilt href, plus a regression test pinning `/docs/channels#publish` → `/docs/channels?lang=javascript#publish`.

### DIFF
--- a/src/components/Markdown/MarkdownProvider.test.tsx
+++ b/src/components/Markdown/MarkdownProvider.test.tsx
@@ -96,5 +96,12 @@ describe('<Anchor/>', () => {
 
       expect(a).toHaveAttribute('href', '#');
     });
+
+    it('preserves the fragment on internal links with anchors', () => {
+      const { getByTestId } = render(<Anchor href="/docs/channels#publish">Example docs</Anchor>);
+      const a = getByTestId('link-internal');
+
+      expect(a).toHaveAttribute('href', '/docs/channels?lang=javascript#publish');
+    });
   });
 });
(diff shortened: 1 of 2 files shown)


AndyTWF (approved):
Approving — small, correct, and properly tested.

Verified locally by applying the diff: 6/6 pass in `MarkdownProvider.test.tsx`, and with only the `src` change reverted the new test fails (`/docs/channels?lang=javascript` instead of `/docs/channels?lang=javascript#publish`). So the test genuinely pins the bug rather than just restating current behaviour.

`pathname + search + hash` is the only valid ordering here, and existing behaviour is untouched — bare `#`/`#fragment` hrefs still short-circuit via `checkLinkIsInternal`, and hash-less links get an empty-string no-op. `GatsbyLink` handles `pathname?query#hash` and scroll restoration fine.

Good catch on the impact: since `Anchor` maps to every markdown `a`, this was breaking anchored links site-wide for anyone who had touched the language selector.

One non-blocking follow-up, out of scope here: `buildLinkWithParams` in `src/components/Layout/LeftSidebar.tsx` has the same latent problem — it does `` `${targetLink}?${paramString}` ``, so a nav entry with a fragment would produce the invalid `/docs/page#section?lang=javascript`, and one with its own query string would produce `?foo=1?lang=javascript`. Nothing is broken today (no `link:` value in `src/data/nav/` contains `#` or `?`), but reusing the `URL`-based approach there would stop that constraint being implicit.

PR #3504, opened 2026-07-29
[AIT-1225] Add an AI Transport pricing page

VeskeR:
Adds `/docs/ai-transport/pricing`, modelled on the existing Pub/Sub, Chat, LiveObjects, Spaces, and LiveSync pricing pages ([AIT-1225](https://ably.atlassian.net/browse/AIT-1225)).

AI Transport bills as ordinary Pub/Sub messages, so the page keeps the sibling shape (operations table → sessions and channels → connections → cost optimization) and adds three sections specific to how AI Transport generates messages:

- a per-run message breakdown
- append rollup, the main lever on a streamed response's message count
- message persistence, which the channel rule AI Transport requires forces across the namespace

Also included:

- nav entry in `src/data/nav/aitransport.ts`
- an `AI Transport operations` section in the platform-wide message-counting page, which every other product already has
- the `going-to-production` pricing section now links to the new page

Message counts are verified against `ably-ai-transport-js` 0.6.0 (publish and append call sites in `run-manager.ts`, `codec/encoder.ts`, `vercel/codec/outputs.ts`) rather than against prior docs. No rates on the page; it links to `/docs/platform/pricing` and sales for those.

## Open question for review

AITDR-002 states that append actions are not stored in persisted history, while the published [AI chatbot pricing example](https://ably.com/docs/platform/pricing/examples/ai-chatbot) bills persistence on every append. This page states only the platform rule (one additional message per stored message) and deliberately does not assert a per-append persistence multiplier. Worth confirming with product before merge.

Two follow-ups spotted but left out to keep the diff focused:

- `features/token-streaming.mdx` says to set `appendRollupWindow` "on the client"; it caps the publishing (agent) connection's rate, so the two pages should be reconciled.
- `/docs/integrations/kafka` in `going-to-production.mdx` is a pre-existing broken link.

[AIT-1225]: https://ably.atlassian.net/browse/AIT-1225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/docs/ai-transport/pricing.mdx
+++ b/src/pages/docs/ai-transport/pricing.mdx
@@ -5,7 +5,7 @@ meta_keywords: "AI Transport pricing, message counting, token streaming, append 
 intro: "How AI Transport operations contribute to your message count and strategies to optimize costs."
 ---
 
-The [AI Transport SDK](/docs/ai-transport) is built on top of [Pub/Sub](/docs/pub-sub). All AI Transport operations generate Pub/Sub messages that follow the same [counting rules](/docs/platform/pricing/message-counting), and Ably charges per published message rather than per token.
+The [AI Transport SDK](/docs/ai-transport) is built on top of [Pub/Sub](/docs/pub-sub). All AI Transport operations generate Pub/Sub messages that follow the same [counting rules](/docs/platform/pricing/message-counting): each message counts once inbound when it is published, once outbound for every subscriber that receives it, and once more when it is stored. Ably counts messages rather than tokens, and the [append rollup](#rollup) decides how many messages a streamed response becomes.
 
 ## AI Transport operations <a id="operations"/>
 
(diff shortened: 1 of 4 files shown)


src/pages/docs/ai-transport/pricing.mdx
```diff
@@ -0,0 +1,121 @@
...
+| [LiveObjects state](/docs/ai-transport/features/liveobjects) ||
+| Object operation | 1 inbound message per operation |
+| ObjectMessage delivery | 1 outbound message per connected client |
+| [Push notifications](/docs/ai-transport/features/push-notifications) ||
+| Push notification delivery | 1 outbound message per delivered notification |
+
+Each streamed part is its own channel message. A response that emits reasoning and then text opens and closes two messages, and every tool call opens another. The [wire protocol](/docs/ai-transport/internals/wire-protocol#event-names) lists the exact channel message each operation publishes.
+
+Ably bills messages in [5KiB chunks](/docs/platform/pricing/message-counting#message-size), and this matters more for streamed responses than for anything else. Individual appends are small and count as one each, but Ably stores and serves a streamed response as one aggregated message. A 20KiB response therefore counts as four messages everywhere it is counted whole: when it is stored, when it is retrieved from history, and when a reconnecting client receives the accumulated content.
```
paddybyers: > A 20KiB response therefore counts as four messages

This is incorrect. v6 pricing has bandwidth as a separate dimension; data doesn't just get translated into more messages, as was the case with earlier pricing dimensions.

paddybyers (changes requested):
I've not done a complete review, but I think there needs to be a pass that checks for accuracy against what v6 does, and then I will re-review.

src/pages/docs/ai-transport/pricing.mdx
VeskeR: I removed this from AIT pricing docs.
This was based on what seems to be the incorrect information at https://github.com/ably/docs/blob/003bb8823381bb67e827cdb7ab3e852c9b2d653b/src/pages/docs/platform/pricing/message-counting.mdx#L31, or see https://ably.com/docs/platform/pricing/message-counting#:~:text=Ably%20bills%20messages%20in%205KiB%20chunks.%20A%20message%20up%20to%205KiB%20counts%20as%201%20message.%20A%2050KiB%20message%20counts%20as%2010%20messages%2C%20and%20a%2016KiB%20message%20counts%20as%204%20messages.

paddybyers: I've raised https://github.com/ably/docs/issues/3506

